PR #2, opened 2016-10-27
Fix function signature for RPC readme

sectioneight:
This has been updated in recent versions of YARPC


sectioneight:
@uber-go/service-framework

PR #3, opened 2016-10-27
Register dynamic provider

anuptalwalkar:
GFM-109

### DIFF
--- a/core/config/config.go
+++ b/core/config/config.go
@@ -48,9 +48,10 @@ var (
 	_setupMux    sync.Mutex
 	_initialized bool
 
-	_envPrefix           = "APP"
-	_configProviderFuncs = []ProviderFunc{YamlProvider(), EnvProvider()}
-	_cpMux               sync.Mutex
+	_envPrefix            = "APP"
+	_staticProviderFuncs  = []ProviderFunc{YamlProvider(), EnvProvider()}
+	_dynamicProviderFuncs []DynamicProviderFunc
+	_cpMux                sync.Mutex
 )
 
 // Global returns the singleton configuration provider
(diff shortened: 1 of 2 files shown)


core/config/config.go
```diff
@@ -48,9 +48,10 @@ var (
 	_setupMux    sync.Mutex
 	_initialized bool
 
-	_envPrefix           = "APP"
-	_configProviderFuncs = []ProviderFunc{YamlProvider(), EnvProvider()}
-	_cpMux               sync.Mutex
+	_envPrefix            = "APP"
+	_staticProviderFuncs  = []ProviderFunc{YamlProvider(), EnvProvider()}
+	_dynamicProviderFuncs []DynamicProviderFunc
+	_cpMux                sync.Mutex
```
glibsm: random question: what's cp stand for?

core/config/config.go
```diff
@@ -144,41 +145,61 @@ func GetEnvironmentPrefix() string {
 // ProviderFunc is used to create config providers on configuration initialization
 type ProviderFunc func() (ConfigurationProvider, error)
 
+// DynamicProviderFunc is used to create config providers on configuration initialization
```
glibsm: We should probably mention why this provider takes a provider as an argument :)

Something along the lines of: "dynamic provider configuration settings (ports, keys, etc) can be supplied via env or yaml"

core/config/config.go
```diff
@@ -144,41 +145,61 @@ func GetEnvironmentPrefix() string {
...
 	_cpMux.Lock()
 	defer _cpMux.Unlock()
-	_configProviderFuncs = nil
+	_staticProviderFuncs = nil
 }
 
 // InitializeGlobalConfig initializes the ConfigurationProvider for use in a service
 func InitializeGlobalConfig() {
 	_setupMux.Lock()
 	defer _setupMux.Unlock()
```
glibsm: shouldn't this be the same mutex as config provider registration? i think you might have race conditions otherwise

core/config/config.go
anuptalwalkar: config provider* (TM)

core/config/config.go
anuptalwalkar: Ah, you are right. I will fix this.

PR #4, opened 2016-10-27
Fix one more YARPC discrepancy in readme

sectioneight:
[noci]


sectioneight:
@uber-go/service-framework

glibsm:
`[no ci]` with a space, i believe

PR #5, opened 2016-10-27
Update to latest yarpc

sectioneight:
transport.Options was removed


sectioneight:
@uber-go/service-framework

PR #6, opened 2016-10-27
Remove config.Global

sectioneight:
This was causing lots of headaches and leaked state between tests.

GFM-105

### DIFF
--- a/core/config/config.go
+++ b/core/config/config.go
@@ -140,7 +140,7 @@ func UnregisterProviders() {
 	_staticProviderFuncs = nil
 }
 
-// Initialize initializes the ConfigurationProvider for use in a service
+// Initialize creates a ConfigurationProvider for use in a service
 func Initialize() ConfigurationProvider {
 	var static []ConfigurationProvider
 	for _, providerFunc := range _staticProviderFuncs {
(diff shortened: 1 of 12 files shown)


sectioneight:
@uber-go/service-framework

core/config/config.go
```diff
@@ -175,15 +140,8 @@ func UnregisterProviders() {
...
-func InitializeGlobalConfig() {
-	_setupMux.Lock()
-	defer _setupMux.Unlock()
-	if _initialized {
-		return
-	}
-	_initialized = true
-
+// InitializeConfig initializes the ConfigurationProvider for use in a service
+func InitializeConfig() ConfigurationProvider {
```
anuptalwalkar: Lets just call it Initialize(), so that the call will be config.Initialize()

sectioneight: Great idea. Updating.

core/config/config.go
```diff
@@ -175,15 +140,8 @@ func UnregisterProviders() {
...
-func InitializeGlobalConfig() {
-	_setupMux.Lock()
-	defer _setupMux.Unlock()
-	if _initialized {
-		return
-	}
-	_initialized = true
-
+// Initialize creates a ConfigurationProvider for use in a service
+func Initialize() ConfigurationProvider {
```
sectioneight: sure. suggestions?

PR #7, opened 2016-10-27
Change metrics interface to provide Scope

glibsm:
Allows for more flexibility of creating a root scope and a reporter in one place.

### DIFF
--- a/core/metrics/metrics.go
+++ b/core/metrics/metrics.go
@@ -44,6 +44,7 @@ var (
 )
 
 // ScopeInit interface provides necessary data to properly initialize root metrics scope
+// service.Host conforms to this, but can't be used directly since it causes an import cycle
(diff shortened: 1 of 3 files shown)


glibsm:
@uber-go/service-framework

core/metrics/metrics.go
```diff
@@ -43,8 +43,14 @@ var (
 	DefaultReporterInterval = time.Second
 )
 
-// ReporterFunc is used during service init time to register the reporter
-type ReporterFunc func(config.ConfigurationProvider) (tally.StatsReporter, error)
+// ScopeInit interface provides necessary data to properly initialize root metrics scope
```
sectioneight: maybe mention that service.Host conforms to this interface and it's here to avoid an import cycle

PR #10, opened 2016-10-28
Improve make output

sectioneight:
Bring in some colors and labels. Respect verbosity setting for
generating example thrift code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@ SHELL := /bin/bash
 PROJECT_ROOT := go.uber.org/fx
 
 SUPPORT_FILES := .build
+include $(SUPPORT_FILES)/colors.mk
 include $(SUPPORT_FILES)/deps.mk
 include $(SUPPORT_FILES)/flags.mk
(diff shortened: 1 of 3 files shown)


sectioneight:
@uber-go/service-framework

Makefile
```diff
@@ -38,8 +39,12 @@ endif
 COVER_OUT := profile.coverprofile
 
 $(COV_REPORT): $(PKG_FILES) $(ALL_SRC)
-		$(ECHO_V)$(MAKE) -C examples/keyvalue/ kv/types.go
-		$(ECHO_V)$(OVERALLS) -project=$(PROJECT_ROOT) \
+	$(ECHO_V)echo "$(LABEL_STYLE)Generating example RPC bindings$(COLOR_RESET)"
+	$(ECHO_V)echo
```
glibsm: not sure we should literally echo the echo if `V=1` :)

sectioneight: Fair point: https://github.com/uber-go/uberfx/pull/14

PR #12, opened 2016-10-28
Kafka consumer

sectioneight:
OSS version of the kafka consumer.

A lot of the code was great as-is, just needed some tweaking to remove Uber specific stuff, and add errchecks.

GFM-65

### DIFF
--- a/modules/kafka/consumer/config_test.go
+++ b/modules/kafka/consumer/config_test.go
@@ -1,0 +1,65 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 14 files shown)


sectioneight:
@glibsm @akshayjshah the old PR was closed out in a different repo, re-opened here after cherry-picking.

modules/kafka/consumer/config_test.go
```diff
@@ -0,0 +1,65 @@
...
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestConfigValidate(t *testing.T) {
+	valid := fakeConfig()
+
+	hasHostFile := fakeConfig()
+	hasHostFile.HostFile = "/etc/uber/kafka8/clusters.yaml"
```
anuptalwalkar: We should have some tmp path here and not point to etc/uber

sectioneight:
So, what's the next step here @akshayjshah @prashantv? Should we create a new uber-go repo for kafka consumer? Would you like to be the owner of that or us? Or should we just try to merge this in uberfx?

sectioneight:
Closing until we figure out the right home for this. Meeting set up for next week.

PR #13, opened 2016-10-28
Use service WithConfiguration to not rely on global setup

sectioneight:
Again trying to move away from modifying global state in tests.

Use the WithConfiguration() option to set a configuration provider.

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -82,29 +82,6 @@ func New(options ...Option) (Owner, error) {
 		},
 	}
 
-	cfg := config.Load()
-	svc.serviceCore.configProvider = cfg
-
-	// prepend the default options.
-	// TODO: This isn't right.  Do we order the options so we make sure to use
-	// the passed in options?
-	//
-	ensureThat(WithConfiguration(cfg)(svc), "configuration")
-
-	// load standard config
-	// TODO(glib): `.GetValue("")` is a confusing interface for getting the root config node
-	svc.configProvider.GetValue("").PopulateStruct(&svc.standardConfig)
-
-	// load and configure logging
-	svc.configProvider.GetValue("logging").PopulateStruct(&svc.logConfig)
-	ulog.Configure(svc.logConfig)
-	ensureThat(WithLogger(ulog.Logger())(svc), "log configuration")
-
-	if errs := validator.Validate(svc.standardConfig); errs != nil {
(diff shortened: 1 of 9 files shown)


sectioneight:
@uber-go/service-framework

service/service.go
```diff
@@ -147,9 +153,3 @@ func New(options ...Option) (Owner, error) {
 
 	return svc, nil
 }
-
-func ensureThat(err error, description string) {
```
sectioneight: also very happy to get rid of this :)

service/service_test.go
```diff
@@ -148,12 +146,6 @@ var validServiceConfig = map[string]interface{}{
...
 
-// withConfigData sets a global config and returns a function to defer reset
-func withConfigData(data map[string]interface{}) func() {
-	oldProviders := config.Providers()
-	config.UnregisterProviders()
-	config.RegisterProviders(config.StaticProvider(data))
-	return func() {
-		config.RegisterProviders(oldProviders...)
-	}
+func withConfigOption(data map[string]interface{}) Option {
```
anuptalwalkar: rename suggestion - "withConfig"

service/testutils/service.go
```diff
@@ -23,17 +23,17 @@ package testutils
...
 // WithService is a test helper to instantiate a service
-func WithService(module service.ModuleCreateFunc, instance service.Observer, fn func(service.Owner)) {
-	WithServices([]service.ModuleCreateFunc{module}, instance, fn)
+func WithService(module service.ModuleCreateFunc, instance service.Observer, options []service.Option, fn func(service.Owner)) {
+	WithServices([]service.ModuleCreateFunc{module}, instance, options, fn)
 }
 
 // WithServices is a test helper to instantiate a service with multiple modules
-func WithServices(modules []service.ModuleCreateFunc, observer service.Observer, fn func(service.Owner)) {
+func WithServices(modules []service.ModuleCreateFunc, observer service.Observer, options []service.Option, fn func(service.Owner)) {
```
glibsm: this function name doesn't make sense. it means to imply multiple modules, but implies multiple services (which is not a thing)

sectioneight: Good call. It's actually unused now, removing.

service/service_test.go
sectioneight: Good idea. Done.

PR #15, opened 2016-10-28
Servicename in DynamicProviderFunc

anuptalwalkar:
GFM-127

### DIFF
--- a/core/config/config.go
+++ b/core/config/config.go
@@ -110,7 +110,7 @@ func GetEnvironmentPrefix() string {
 type ProviderFunc func() (ConfigurationProvider, error)
 
 // DynamicProviderFunc is used to create config providers on configuration initialization
-type DynamicProviderFunc func(config ConfigurationProvider) (ConfigurationProvider, error)
+type DynamicProviderFunc func(serviceName string, config ConfigurationProvider) (ConfigurationProvider, error)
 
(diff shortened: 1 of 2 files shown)


core/config/config.go
```diff
@@ -152,9 +152,16 @@ func Load() ConfigurationProvider {
 	}
 	baseCfg := NewProviderGroup("global", static...)
 
+	// TODO: (at) If application ID is not specified, the code lets the service initialization
```
glibsm: i think it should be a panic/hard failure if application ID is not specified. We need it for a lot places, not just config

core/config/config.go
```diff
@@ -152,9 +152,16 @@ func Load() ConfigurationProvider {
...
+	// handle the failures and skips initializing dynamic config.
+	serviceName, ok := baseCfg.GetValue("applicationID").TryAsString()
+	if !ok {
+		return baseCfg
+	}
 	var dynamic []ConfigurationProvider
+
 	for _, providerFunc := range _dynamicProviderFuncs {
-		cp, err := providerFunc(baseCfg)
+		cp, err := providerFunc(serviceName, baseCfg)
```
anuptalwalkar: Thanks. This was an abstraction to get serviceName before hand without having to call config.GetValue multiple times. I'll revert this PR and move this fetch downstream.

PR #16, opened 2016-10-29
Revert DynamicProviderfunc change

anuptalwalkar:
Reverting PR - https://github.com/uber-go/uberfx/pull/15

Moving this fetch downstreams and letting the dynamic provider initializers handle the serviceName from baseCfg.

### DIFF
--- a/core/config/config.go
+++ b/core/config/config.go
@@ -110,7 +110,7 @@ func GetEnvironmentPrefix() string {
 type ProviderFunc func() (ConfigurationProvider, error)
 
 // DynamicProviderFunc is used to create config providers on configuration initialization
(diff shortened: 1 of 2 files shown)


anuptalwalkar:
@glibsm @uber-go/service-framework

anuptalwalkar:
Thanks! I thought about that option too, but its really hard and twisted because at this point in Load() Host is not initialized. Host is initialized after config.Load() and populateStruct() is called the loaded cfg object.

sectioneight:
Ah, excellent point. Maybe we can export the applicationID const so it
could be used in fx-private just to prevent drift.

anuptalwalkar:
Ah yes, we already have config.ApplicationIDKey exported :). I will use that in private.

PR #17, opened 2016-10-29
Fix unregister providers for dynamic provider

anuptalwalkar:
We were not performing Unregister on dynamicProviderFuncs causing issues in tests.


anuptalwalkar:
@uber-go/service-framework

PR #18, opened 2016-10-31
Expose YARPC option for custom interceptors

sectioneight:
This way our users can add custom YARPC interceptors as an option
in module creation.

GFM-117

### DIFF
--- a/modules/rpc/yarpc_options_test.go
+++ b/modules/rpc/yarpc_options_test.go
@@ -21,39 +21,34 @@
 package rpc
 
 import (
-	"go.uber.org/fx/modules"
+	"testing"
+
 	"go.uber.org/fx/service"
 
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"go.uber.org/yarpc/transport"
 )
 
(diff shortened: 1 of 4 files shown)


sectioneight:
@uber-go/service-framework

modules/rpc/yarpc_options_test.go
```diff
@@ -0,0 +1,54 @@
...
+	mc := &service.ModuleCreateInfo{
+		Items: map[string]interface{}{
+			_interceptorKey: "foo",
+		},
+	}
+
+	assert.Panics(t, func() {
+		opt(mc)
+	})
+}
```
glibsm: maybe add a test for a dummy interceptor in action?

glibsm: We only test a `NopInterceptor`. I was asking if we should test with a real interceptor so we can validate it's actually executed and all that (without testing the internals of the yarpc too much)

sectioneight: Meh, I don't think there's much value in testing that YARPC interceptors work.

PR #19, opened 2016-10-31
Close tally scopes on service stop

glibsm:
Fix the root metrics scope to use the correct interface also (which is the one that has `Close()` method) and call it on shutdown.

GFM-96 #resolve

### DIFF
--- a/core/metrics/metrics_test.go
+++ b/core/metrics/metrics_test.go
@@ -66,15 +66,15 @@ func TestRegisterBadReporterPanics(t *testing.T) {
 	})
 }
 
-func goodScope(i ScopeInit) (tally.Scope, error) {
+func goodScope(i ScopeInit) (tally.RootScope, error) {
 	return tally.NoopScope, nil
 }
 
-func badScope(i ScopeInit) (tally.Scope, error) {
+func badScope(i ScopeInit) (tally.RootScope, error) {
 	return nil, errors.New("fake error")
 }
 
-func getScope() tally.Scope {
+func getScope() tally.RootScope {
 	return RootScope(scopeInit())
 }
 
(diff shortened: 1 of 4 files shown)


glibsm:
@uber-go/service-framework

core/metrics/metrics_test.go
```diff
@@ -66,15 +66,15 @@ func TestRegisterBadReporterPanics(t *testing.T) {
...
 	return tally.NoopScope, nil
 }
 
-func badScope(i ScopeInit) (tally.Scope, error) {
+func badScope(i ScopeInit) (tally.RootScope, error) {
 	return nil, errors.New("fake error")
 }
 
-func getScope() tally.Scope {
+func getScope() tally.RootScope {
```
sectioneight: Does NopScope conform to RootScope? What's the difference in the interfaces?

service/host.go
```diff
@@ -134,16 +134,18 @@ func (s *host) shutdown(err error, reason string, exitCode *int) (bool, error) {
...
+	if len(errs) > 0 {
+		for k, v := range errs {
+			ulog.Logger().Error("Failure to shut down module", "name", k.Name(), "error", v.Error())
+		}
+	}
 
-	// }
+	// Stop the metrics reporting
+	if s.scope != nil {
+		s.scope.Close()
```
sectioneight: Does this flush immediately, or do we need to wait on a channel or something?

glibsm: `Close()` is the only difference

core/metrics/metrics_test.go
glibsm: NoopScope is a RootScope

sectioneight: Sweet.

sectioneight:
Why the drop in coverage?

glibsm:
Probably due to un-commented code that does failure logging, will double check. Not sure we can test that yet

PR #20, opened 2016-10-31
Add signal handlers for SIGINT and SIGTERM

sectioneight:
This way we call Stop() on the service when it's killed.

GFM-136

This is actually all covered by the test suite, but difficult to unit
test because we can't send SIGTERM to the test process.

### DIFF
--- a/service/host.go
+++ b/service/host.go
@@ -23,8 +23,10 @@ package service
 import (
 	"fmt"
 	"os"
+	"os/signal"
 	"reflect"
 	"sync"
+	"syscall"
 	"time"
 
 	"go.uber.org/fx/core/config"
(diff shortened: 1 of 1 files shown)


service/host.go
```diff
@@ -223,20 +225,33 @@ func (s *host) Start(waitForShutdown bool) (<-chan Exit, <-chan struct{}, error)
...
 					ulog.Logger().Error("Unable to shut down modules", "initialError", e, "shutdownError", err)
 				}
-				s.shutdownMu.Lock()
-
 				return errChan, readyCh, e
 			}
 		}
 	}
 
+	s.shutdownMu.Unlock()
```
sectioneight: Changed this around to avoid a deadlock. The mutex was only there to guard `s.locked` so it's fine to unlock here.

sectioneight:
@uber-go/service-framework

PR #21, opened 2016-10-31
Add hook to register custom YARPC dispatcher

sectioneight:
This will be used internally to override the YARPC dispatcher.

GFM-112

### DIFF
--- a/modules/rpc/yarpc.go
+++ b/modules/rpc/yarpc.go
@@ -51,7 +51,7 @@ type YarpcModule struct {
 }
 
 var (
-	_dispatcherFn yarpcDispatcherFn
+	_dispatcherFn = defaultYarpcDispatcher
 
 	_ service.Module = &YarpcModule{}
 )
@@ -205,8 +205,4 @@ func RegisterDispatcher(dispatchFn yarpcDispatcherFn) {
 
 func defaultYarpcDispatcher(cfg yarpc.Config) (yarpc.Dispatcher, error) {
 	return yarpc.NewDispatcher(cfg), nil
-}
-
-func init() {
-	RegisterDispatcher(defaultYarpcDispatcher)
 }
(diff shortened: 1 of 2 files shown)


sectioneight:
@uber-go/service-framework

modules/rpc/yarpc_test.go
```diff
@@ -48,6 +48,10 @@ func TestInterceptor_OK(t *testing.T) {
 	assert.NoError(t, err)
 }
 
+func TestRegisterDispatcher_OK(t *testing.T) {
+	RegisterDispatcher(defaultYarpcDispatcher)
```
madhuravi: assert things? Eg. Start actually invokes this function when registered.

sectioneight: That's already tested elsewhere

modules/rpc/yarpc.go
```diff
@@ -183,3 +191,18 @@ func (m *YarpcModule) IsRunning() bool {
 	defer m.stateMu.RUnlock()
 	return m.rpc != nil
 }
+
+type yarpcDispatcherFn func(yarpc.Config) (yarpc.Dispatcher, error)
+
+// RegisterDispatcher allows you to override the YARPC dispatcher registration
+func RegisterDispatcher(dispatchFn yarpcDispatcherFn) {
+	_dispatcherFn = dispatchFn
```
glibsm: mutex here?

sectioneight: No, this should only ever be called once -- it's intentionally not thread-safe

glibsm: Panic on set when it's already non-nil, then? Just so we don't accidentally do it twice, or have default override the custom provided one.

sectioneight: It will always be non-nil, since we call Register with the default dispatcher on init

PR #22, opened 2016-10-31
Fix YARPC readme example

sectioneight:
[ci skip]


sectioneight:
@uber-go/service-framework

PR #23, opened 2016-10-31
 Exclude nil providers from ProviderGroup

anuptalwalkar:
GFM-59

Whenever dynamic providers are excluded via configuration, don't append to the provider group.

### DIFF
--- a/core/config/config_test.go
+++ b/core/config/config_test.go
@@ -286,16 +286,19 @@ func TestNilProvider(t *testing.T) {
 		return nil, fmt.Errorf("error creating Provider")
 	})
 	assert.Panics(t, func() { Load() }, "Can't initialize with nil provider")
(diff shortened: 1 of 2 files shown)


core/config/config_test.go
```diff
@@ -286,6 +286,12 @@ func TestNilProvider(t *testing.T) {
 		return nil, fmt.Errorf("error creating Provider")
 	})
 	assert.Panics(t, func() { Load() }, "Can't initialize with nil provider")
+	UnregisterProviders()
+
+	RegisterProviders(func() (ConfigurationProvider, error) {
+		return nil, nil
+	})
+	Load()
 	oldProviders := _staticProviderFuncs
```
sectioneight: don't you need to capture the old dynamic providers before line 289?

PR #24, opened 2016-10-31
Remove traffic reporter/tracker

sectioneight:
In favor of modules emitting metrics natively.

GFM-108


sectioneight:
@uber-go/service-framework

glibsm:
🕯

PR #25, opened 2016-11-01
Initialize dynamic before append

anuptalwalkar:
TIL appending empty slice to slice is ok, but appending uninitialized slice to slice results in <nil> append.


anuptalwalkar:
@uber-go/service-framework

madhuravi:
Add unit test.

PR #26, opened 2016-11-01
PopulateStruct to keep empty slice as nil

anuptalwalkar:
GFM-59

If a key with a slice value is not present in the *.yaml file, PopulateStruct with the parent key populates the said slice as empty slice rather than nil. PR to set fieldValue only when destSlice is not empty, and keep nil otherwise. 

Added test to evaluate the inconsistent behavior. Without the change, first assertion would fail, since empty slice != nil

### DIFF
--- a/core/config/yaml_test.go
+++ b/core/config/yaml_test.go
@@ -131,14 +131,18 @@ emptystruct:
   nonexist: true
 `)
 
-func TestPopulateMismatchedStruct(t *testing.T) {
-	provider := NewProviderGroup("global", NewYAMLProviderFromBytes(emptyyaml))
+func TestMatchEmptyStruct(t *testing.T) {
+	provider := NewProviderGroup("global", NewYAMLProviderFromBytes([]byte(``)))
 	es := emptystruct{}
+	provider.GetValue("emptystruct").PopulateStruct(&es)
 	empty := reflect.New(reflect.TypeOf(es)).Elem().Interface()
 	assert.True(t, reflect.DeepEqual(empty, es))
+}
 
-	provider = NewProviderGroup("global", NewYAMLProviderFromBytes([]byte(``)))
-	es = emptystruct{}
+func TestMatchPopulatedEmptyStruct(t *testing.T) {
+	provider := NewProviderGroup("global", NewYAMLProviderFromBytes(emptyyaml))
+	es := emptystruct{}
 	provider.GetValue("emptystruct").PopulateStruct(&es)
+	empty := reflect.New(reflect.TypeOf(es)).Elem().Interface()
 	assert.True(t, reflect.DeepEqual(empty, es))
 }
--- a/core/config/yaml_test.go
+++ b/core/config/yaml_test.go
@@ -23,6 +23,7 @@ package config
 import (
 	"bytes"
 	"io/ioutil"
+	"reflect"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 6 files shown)


sectioneight:
What's wrong with an empty slice rather than nil? We shouldn't ever be checking if a slice is `nil` explicitly -- always `if len(thing) == 0` since `nil` and `[]<type>` are both valid, but empty values.

Also, test failiruse.

anuptalwalkar:
Oh how I wish nil == empty was true for all available compares. reflect.deepequal() has a slice check for nil first, followed by length check which where the struct equals is failing. 
`
reflect: Deepequals
if v1.IsNil() != v2.IsNil() {
	return false
}`

Right now there are two ways to compare struct with slices as elements, either by comparing individual elements or reflect.Deepequals. Comparing individual elements works, but it is not forward compatible, and we'll have to be aware of any elements getting added/removed from the struct. I have seen that Deepequals is recommended whenever struct is complex with arrays and slices. PopulateStruct will likely break this existing comparison behavior and end up in bad customer experience. Thoughts?

glide.yaml
```diff
@@ -13,7 +13,7 @@ import:
 # TODO: Use stable branch for yarpc
   version: 5cb0df4c43be85f526cfe84352983966a83db29d
 - package: go.uber.org/thriftrw
-  version: dev
+  version: 6aa07c0009457cc6c7ff5497695f1581ad184350
```
sectioneight: Please don't lock to master. We need dev for uberfx-private and xyarpc

glide.lock
```diff
@@ -1,18 +1,18 @@
...
+hash: 42ff2b1ac9a75d5734d4141b56ac8c19fbd3679dfd479b34b3f43f416151af9f
+updated: 2016-11-01T16:30:17.314051732-07:00
 imports:
+- name: github.com/anmitsu/go-shlex
+  version: 53a3d8a8a362cca54fd6825caf2216aa6d8139fc
 - name: github.com/facebookgo/clock
   version: 600d898af40aa09a7a93ecb9265d87b0504b6f03
 - name: github.com/go-validator/validator
   version: 0a9835d809fb647a62611d30cb792e0b5dd65b11
-- name: github.com/golang/mock
```
sectioneight: what happened to this? we were using it from yarpc tests

glide.yaml
```diff
@@ -13,6 +13,8 @@ import:
   version: dev
 - package: go.uber.org/thriftrw
   version: dev
+- package: github.com/anmitsu/go-shlex
+- package: github.com/jessevdk/go-flags
```
sectioneight: are these really necessary? we don't import them? why did you need to add them? should they be in testImports instead?

core/config/yaml_test.go
```diff
@@ -120,3 +121,24 @@ func TestYamlNode_Callbacks(t *testing.T) {
...
+  nonexist: true
+`)
+
+func TestPopulateMismatchedStruct(t *testing.T) {
+	provider := NewProviderGroup("global", NewYAMLProviderFromBytes(emptyyaml))
+	es := emptystruct{}
+	empty := reflect.New(reflect.TypeOf(es)).Elem().Interface()
+	assert.True(t, reflect.DeepEqual(empty, es))
+
+	provider = NewProviderGroup("global", NewYAMLProviderFromBytes([]byte(``)))
```
sectioneight: This should be two different tests, since they're testing different things, right?

examples/keyvalue/Makefile
```diff
@@ -1,7 +1,9 @@
 .PHONY: deps
 deps:
-	$(ECHO_V)which thriftrw > /dev/null || go get -u go.uber.org/thriftrw
-	$(ECHO_V)which thriftrw-plugin-yarpc > /dev/null || go get -u go.uber.org/yarpc/encoding/thrift/thriftrw-plugin-yarpc
+	@echo "Installing thriftrw..."
+	$(ECHO_V)go install ../../vendor/go.uber.org/thriftrw
+	@echo "Installing thriftrw-plugin-yarpc..."
+	$(ECHO_V)go install ../../vendor/go.uber.org/yarpc/encoding/thrift/thriftrw-plugin-yarpc
```
sectioneight: 👍

glide.yaml
anuptalwalkar: These are needed to install thriftrw and yarpc plugin since the install is looking into vendor for these imports.

anuptalwalkar: Fixed.

glide.lock
anuptalwalkar: Not sure, I did glide up, and it wasn't being used?

PR #27, opened 2016-11-01
Main readme changes + moving away module specific stuff

glibsm:
I feel like readmes about the modules do not belong in the main wiki and are much better placed in their respective sections.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ func main() {
 It's common for a service to handle many different workloads. For example, a
 service may expose RPC endpoints and also ingest Kafka messages.
 
-In uberfx, there is a simpler model where we create a single binary,
+In UberFX, there is a simpler model where we create a single binary,
 but turn its modules on and off based on roles which are specified via the
 commmand line.
 
--- a/README.md
+++ b/README.md
@@ -219,7 +219,7 @@ Then running the above example will result in **Port is 3000**
 
 ## License
 
-See [LICENSE.txt](LICENSE.txt)
+[MIT](LICENSE.txt)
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -66,14 +60,12 @@ func main() {
...
 
-It's common for a service to handle many different workloads.  For example, a
-service may expose RPC endpoints and also ingest Kafka messages.  In the Python
-world, this means different deployments that run with different entry points.
-Due to Python's threading model, this was required.
+It's common for a service to handle many different workloads. For example, a
+service may expose RPC endpoints and also ingest Kafka messages.
 
-In the Go service, we can have a simpler model where we create a single binary,
+In uberfx, there is a simpler model where we create a single binary,
```
sectioneight: UberFx

README.md
```diff
@@ -1,36 +1,30 @@
-# UberFx Service Framework
+# Uber FX Service Framework
```
sectioneight: UberFx

modules/rpc/README.md
```diff
@@ -0,0 +1,42 @@
+# RPC Module
```
sectioneight: This should probably be godoc-style so it shows up on godoc.org under the package

modules/uhttp/README.md
```diff
@@ -0,0 +1,38 @@
+# HTTP Module
```
sectioneight: godoc

modules/rpc/README.md
glibsm: are you saying that there is a way to get there readmes to work with go-doc or that we should replace these reamdes with godoc?

PR #29, opened 2016-11-02
Fix travis-ci badge path in the main README.md

vors:
Fixing this
image


README.md
```diff
@@ -312,5 +312,5 @@ providers to supply override values to the component.
 [doc-img]: https://godoc.org/go.uber.org/fx?status.svg
 [cov]: https://coveralls.io/github/uber-go/uberfx?branch=master
 [cov-img]: https://coveralls.io/repos/github/uber-go/uberfx/badge.svg?branch=master
```
glibsm: Do these need to change as well?

glibsm:
Thank you for the PR, just one question before merge. 

We renamed the repository today and are still catching up with all the things that need to be done along with it

README.md
vors: Likely

sectioneight:
Thanks @vors, this totally slipped my mind when renaming yesterday.

PR #31, opened 2016-11-02
Configuration path access

anuptalwalkar:


core/config/config.go
```diff
@@ -66,7 +66,7 @@ func getConfigFiles() []string {
 
 func getResolver() FileResolver {
 	paths := []string{}
-	configDir := os.Getenv(GetEnvironmentPrefix() + configdir)
+	configDir := GetConfigurationPath()
```
sectioneight: Don't name things `Getfoo()` -- that's a Java-ism and is not idiomatic. just `ConfigurationPath`, although that stutters, so consider just `Path()` or something.

sectioneight:
Also, http://chris.beams.io/posts/git-commit/ -- imperative verb to start the commit title/PR name

core/config/config.go
```diff
@@ -96,8 +96,8 @@ func GetEnvironment() string {
 	return env
 }
 
-// GetConfigurationPath returns path to the yaml configurations
-func GetConfigurationPath() string {
+// Path returns path to the yaml configurations
+func Path() string {
 	configPath := os.Getenv(GetEnvironmentPrefix() + configdir)
```
glibsm: Ditto here with the `lol java` comment

core/config/config.go
```diff
@@ -96,6 +96,15 @@ func GetEnvironment() string {
 	return env
 }
 
+// Path returns path to the yaml configurations
+func Path() string {
+	configPath := os.Getenv(GetEnvironmentPrefix() + configdir)
```
glibsm: Ditto here with the `lol java` comment

anuptalwalkar: We have it in few more places:(. Sending a separate PR since it could be over several files, not scope of this PR.

core/config/config.go
```diff
@@ -96,6 +96,15 @@ func GetEnvironment() string {
 	return env
 }
 
+// Path returns path to the yaml configurations
```
sectioneight: Is this specific to YAML? If so, why is it in config.go and not yaml.go?

anuptalwalkar: Its not, but we only deal with yaml configurations as of now. Updating the comment with just 'configurations'

sectioneight:
Please see my comment about commit message titles

PR #32, opened 2016-11-02
Fix bench target

glibsm:
Unclear why `glide nv` is not working correctly for our use case, I suspect it's due to the fact that we have many nested packages.

In the meantime, this fixes `make bench`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,7 +67,7 @@ coveralls: $(COV_REPORT)
 .PHONY: bench
 BENCH ?= .
 bench:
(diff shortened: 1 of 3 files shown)


glibsm:
@uber-go/service-framework

Makefile
```diff
@@ -67,7 +67,7 @@ coveralls: $(COV_REPORT)
 .PHONY: bench
 BENCH ?= .
 bench:
-	$(ECHO_V)$(foreach pkg,$(PKGS),go test -bench=$(BENCH) -run="^$$" $(BENCH_FLAGS) $(pkg);)
```
glibsm: `-bench=$(BENCH)` deletion is not intentional, will restore

PR #33, opened 2016-11-02
Add doc.go generator

sectioneight:


sectioneight:
@uber-go/service-framework

modules/rpc/doc.go
```diff
@@ -0,0 +1,44 @@
...
+rpc.CreateThriftServiceFunc signature (uberfx/modules/rpc/thrift.go that
+returns a []transport.Registrant YARPC implementation from the handler:
+
+
+
+  func NewMyServiceHandler(svc service.Host) ([]transport.Registrant, error) {
+    return myservice.New(&MyServiceHandler{}), nil
+  }
+• Pass that method into the module initialization:
+
```
anuptalwalkar: out of curiosity, sometimes its single newline, vs 3 newlines. Is this intentional?

doc.go
```diff
@@ -0,0 +1,218 @@
+/*
+Package fx is the UberFx Service Framework.
+
+[][doc]
+[][cov]
+[]ci (https://travis-ci.org/uber-go/fx)
```
glibsm: any way to ditch these easily?

modules/rpc/doc.go
sectioneight: It all outputs the same. The source isn't super pretty, but this was the simplest path to code in the markdown parser, otherwise we have to keep track of state.

doc.go
sectioneight: I'm going to have to patch the upstream markdown parser, which doesn't seem to understand the syntax we're using and thinks these are "normal" text.

PR #34, opened 2016-11-02
Yaml benchmark

glibsm:
First ever crack at writing a bench, let me know if I'm doing something wrong.

Current state of things:

```
BenchmarkYAMLCreateSingleFile-8                            30000             56059 ns/op         10784 B/op        120 allocs/op
BenchmarkYAMLCreateMultiFile-8                             20000             81304 ns/op         19792 B/op        206 allocs/op
BenchmarkYAMLSimpleGetLevel1-8                          10000000               227 ns/op            64 B/op          4 allocs/op
BenchmarkYAMLSimpleGetLevel3-8                           3000000               458 ns/op           128 B/op          8 allocs/op
BenchmarkYAMLSimpleGetLevel7-8                           2000000               892 ns/op           256 B/op         16 allocs/op
BenchmarkYAMLPopulateStruct-8                             500000              2685 ns/op           720 B/op         32 allocs/op
BenchmarkYAMLPopulateStructNested-8                       300000              5543 ns/op          1408 B/op         68 allocs/op
BenchmarkYAMLPopulateStructNestedMultipleFiles-8          200000              7359 ns/op          2032 B/op         96 allocs/op
PASS
ok      go.uber.org/fx/core/config      16.390s
```

Given that providers are not created often, it's probably ok that they are slow to create (may even drop from bench). The important thing is to make sure that people can safely retrieve nested values in a hot loop without worrying about a performance hit (we're pretty far away from that right now)


glibsm:
@uber-go/service-framework

core/config/testdata/benchmark1.yaml
```diff
@@ -0,0 +1,6 @@
+api:
+  url: http://secret-api.com
+  timout: 5000
+  credentials:
+    username: guess
+    password: who
```
sectioneight: Fix your editor to have trailing newlines, please :)

core/config/yaml_benchmark_test.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func BenchmarkYAMLCreateMultiFile(b *testing.B) {
+	for n := 0; n < b.N; n++ {
+		providerTwoFiles()
+	}
+}
+
+func BenchmarkYAMLSimpleGetLevel1(b *testing.B) {
+	provider := NewYAMLProviderFromBytes([]byte(`foo: 1`))
```
sectioneight: If you just want to benchmark test the GetValue, you should reset the timer after creating the provider (here and elsewhere)

core/config/yaml_benchmark_test.go
```diff
@@ -0,0 +1,135 @@
...
+	type api struct {
+		URL         string
+		Timeout     int
+		Credentials creds
+	}
+
+	p := providerTwoFiles()
+
+	for n := 0; n < b.N; n++ {
+		s := &api{}
```
sectioneight: Do you really want to allocate a new one in the hot loop? Shouldn't we just be testing PopulateStruct?

core/config/testdata/benchmark1.yaml
glibsm: Not sure why yaml didn't have it, added ext.

glibsm:
yeah, I've been updating the benchmarks results after each change (there is no notification for body change, only title)

sectioneight:
Nicely done. We're in the microsecond territory, which really isn't too bad.

sectioneight:
The allocs are pretty atrocious though, if we do want to optimize this we should look at using object pooling.

PR #35, opened 2016-11-02
Upgrade to latest YARPC

sectioneight:
This includes using standard context instead of x/net/context.


sectioneight:
@anuptalwalkar

PR #36, opened 2016-11-03
Fixed YAML deserialization of *bool

anuptalwalkar:
GFM-128

### DIFF
--- a/core/config/yaml_test.go
+++ b/core/config/yaml_test.go
@@ -157,7 +157,7 @@ var pointerYaml = []byte(`
 pointerStruct:
   myTrueBool: true
   myFalseBool: false
-  myString: "hello"
+  myString: hello
 `)
 
 func TestPopulateStructWithPointers(t *testing.T) {
--- a/core/config/yaml_test.go
+++ b/core/config/yaml_test.go
@@ -146,3 +146,26 @@ func TestMatchPopulatedEmptyStruct(t *testing.T) {
 	empty := reflect.New(reflect.TypeOf(es)).Elem().Interface()
 	assert.True(t, reflect.DeepEqual(empty, es))
 }
+
+type pointerStruct struct {
+	MyBool *bool `yaml:"myBool"`
+}
+
+var myBoolyaml = []byte(`
+pointerStruct:
+  myBool: true
+`)
+
+func TestBoolPointer(t *testing.T) {
+	provider := NewProviderGroup("global", NewYAMLProviderFromBytes(myBoolyaml))
+	ps := pointerStruct{}
+	provider.GetValue("pointerStruct").PopulateStruct(&ps)
+	assert.True(t, *ps.MyBool)
+}
+
+func TestNonExistingBoolPointer(t *testing.T) {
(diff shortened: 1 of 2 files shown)


core/config/yaml_test.go
```diff
@@ -146,3 +146,26 @@ func TestMatchPopulatedEmptyStruct(t *testing.T) {
...
+	provider.GetValue("pointerStruct").PopulateStruct(&ps)
+	assert.True(t, *ps.MyBool)
+}
+
+func TestNonExistingBoolPointer(t *testing.T) {
+	provider := NewProviderGroup("global", NewYAMLProviderFromBytes([]byte(``)))
+	ps := pointerStruct{}
+	provider.GetValue("pointerStruct").PopulateStruct(&ps)
+	assert.Nil(t, ps.MyBool)
+}
```
sectioneight: Can you add a test with a `*string` field too? We should make sure this works for types other than just bools.

core/config/yaml_test.go
```diff
@@ -146,3 +146,26 @@ func TestMatchPopulatedEmptyStruct(t *testing.T) {
...
+pointerStruct:
+  myBool: true
+`)
+
+func TestBoolPointer(t *testing.T) {
+	provider := NewProviderGroup("global", NewYAMLProviderFromBytes(myBoolyaml))
+	ps := pointerStruct{}
+	provider.GetValue("pointerStruct").PopulateStruct(&ps)
+	assert.True(t, *ps.MyBool)
+}
```
sectioneight: for completeness sake, can we also check an explicit `false`? it'd be a shame if we messed that up

core/config/yaml_test.go
anuptalwalkar: Sure. Adding it.

core/config/yaml_test.go
```diff
@@ -146,3 +146,34 @@ func TestMatchPopulatedEmptyStruct(t *testing.T) {
...
+	MyTrueBool  *bool   `yaml:"myTrueBool"`
+	MyFalseBool *bool   `yaml:"myFalseBool"`
+	MyString    *string `yaml:"myString"`
+}
+
+var pointerYaml = []byte(`
+pointerStruct:
+  myTrueBool: true
+  myFalseBool: false
+  myString: "hello"
```
sectioneight: nit: no need to surround "hello" with quotes in YAML

core/config/yaml_test.go
```diff
@@ -146,3 +146,34 @@ func TestMatchPopulatedEmptyStruct(t *testing.T) {
 	empty := reflect.New(reflect.TypeOf(es)).Elem().Interface()
 	assert.True(t, reflect.DeepEqual(empty, es))
 }
+
+type pointerStruct struct {
+	MyTrueBool  *bool   `yaml:"myTrueBool"`
+	MyFalseBool *bool   `yaml:"myFalseBool"`
```
glibsm: Shouldn't we add `MyMissingBool`, which is not present in YAML, so we can make sure that stays nil?

anuptalwalkar: Yes, there is a separate test for checking nil *bools and other types on empty yaml.

PR #37, opened 2016-11-04
Start documenting core packages

sectioneight:
GFM-115

### DIFF
--- a/core/config/README.md
+++ b/core/config/README.md
@@ -1,41 +1,48 @@
 # Configuration Package
 
 At a high level, configuration is any data that is used in an application but
-not part of the application itself. Any reasonable complex system will need to
-have knobs to tune and not everything can have intelligent defaults.
+not part of the application itself. Any reasonably complex system needs to
+have knobs to tune, and not everything can have intelligent defaults.
 
-In UberFx, we try very hard to make sure that users only need to provide the
-minimal amount of configuration necessary to have working components, while also
-allowing them to override any field they need to if the defaults don't make
-sense for them.
+In UberFx, we try very hard to make configuring UberFx convenient. Users can:
 
-## Provider
+* Get components working with minimal configuration
+* Override any field if the default doesn't make sense for their use case
+
+## ConfigurationProvider
 
 `ConfigurationProvider` is the interface for anything that can provide values.
 We provide a few reference implementations (environment and YAML), but you are
 free to register your own providers via `config.RegisterProviders()` and
 `config.RegisterDynamicProviders`.
 
-## Dynamic Configuration Providers
+### Static configuration providers
 
-Dynamic configuration providers are treated specially because they frequently
-need some bootstrap configuration to be useful. They conform to the same
-`ConfigurationProvider` interface but are instantiated **after** the static
-providers, so that they can read static values. For example, if you were to
-implement a ZooKeeper-backed `ConfigurationProvider`, you'd likely need to
-specify (via YAML or environment variables) where your ZooKeeper nodes live.
+Static configuration providers conform to the `ConfigurationProvider` interface
+and are bootstraped first. Use these for simple providers such as file-backed or
+environment-based configuration providers.
+
+### Dynamic Configuration Providers
+
+Dynamic configuration providers frequently need some bootstrap configuration to
+be useful, so UberFx treats them specially. Dynamic configuration providers
+conform to the `ConfigurationProvider` interface, but they're instantianted
+**after** the Static `ConfigurationProvider`s on order to read bootstarp values.
+
+For example, if you were to implement a ZooKeeper-backed
+`ConfigurationProvider`, you'd likely need to specify (via YAML or environment
+variables) where your ZooKeeper nodes live.
 
 ## ConfigurationValue
 
-`ConfigurationValue` is the return type of all configuration providers'
+`ConfigurationValue` is the return type of every configuration providers'
 `GetValue(key string)` method. Under the hood, we use the empty interface
 (`interface{}`) since we don't necessarily know the structure of your
 configuration ahead of time.
 
-There are two main usages for a `ConfigurationValue`:
+You can use a `ConfigurationValue for two main purposes:
 
-* Getting a single value out
-* Populating a struct (`PopulateStruct(&myStruct)`)
+* Get a single value out of configuration.
 
 For example, if we have a YAML configuration like so:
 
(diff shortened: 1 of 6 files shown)


sectioneight:
@uber-go/service-framework

sectioneight:
@uber-go/service-framework

core/config/README.md
```diff
@@ -0,0 +1,91 @@
...
+At a high level, configuration is any data that is used in an application but
+not part of the application itself. Any reasonable complex system will need to
+have knobs to tune and not everything can have intelligent defaults.
+
+In UberFx, we try very hard to make sure that users only need to provide the
+minimal amount of configuration necessary to have working components, while also
+allowing them to override any field they need to if the defaults don't make
+sense for them.
+
+## Provider
```
Lupie: Would this be better as "ConfigurationProvider"?

Alternatively, you could have a "Providers" section with an H2 and then split it into two H3 parts: ConfigurationProvider and Dynamic Configuration Providers.

I'd like to see some distinction between this section and the next. Like if dynamic config providers are treated specially, maybe tell your reader up here that ConfigurationProvider is for static stuff too? Does that make sense?

sectioneight:
thanks @glibsm, will wait for @lupie to take a pass

core/config/README.md
```diff
@@ -0,0 +1,91 @@
...
+We provide a few reference implementations (environment and YAML), but you are
+free to register your own providers via `config.RegisterProviders()` and
+`config.RegisterDynamicProviders`.
+
+## Dynamic Configuration Providers
+
+Dynamic configuration providers are treated specially because they frequently
+need some bootstrap configuration to be useful. They conform to the same
+`ConfigurationProvider` interface but are instantiated **after** the static
+providers, so that they can read static values. For example, if you were to
```
Lupie: In second sentence, there's a lot of "they," which makes it a harder read when talking about both static and dynamic providers.

Instead, how about this:
Dynamic configuration providers frequently need some bootstrap configuration to be useful, so UberFx treats them specially. Dynamic configuration providers conform to the `ConfigurationProvider` interface, but they're instantiated **after** the static providers in order to read static values.

Is that better?

core/config/README.md
```diff
@@ -0,0 +1,91 @@
...
+implement a ZooKeeper-backed `ConfigurationProvider`, you'd likely need to
+specify (via YAML or environment variables) where your ZooKeeper nodes live.
+
+## ConfigurationValue
+
+`ConfigurationValue` is the return type of all configuration providers'
```
Lupie: of all configuration providers' --> of every configuration provider's

core/config/README.md
```diff
@@ -0,0 +1,91 @@
...
+`GetValue(key string)` method. Under the hood, we use the empty interface
+(`interface{}`) since we don't necessarily know the structure of your
+configuration ahead of time.
+
+There are two main usages for a `ConfigurationValue`:
```
Lupie: Can rephrase for clarity about who's using what, like:

You can use a `ConfigurationValue` for two main purposes:

or 

A `ConfigurationValue` can serve two main purposes:

core/config/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+* Getting a single value out
```
Lupie: Getting --> Get

Lucie question (your reader prob knows): …get value out of what?

core/config/README.md
```diff
@@ -0,0 +1,91 @@
...
+* Populating a struct (`PopulateStruct(&myStruct)`)
```
Lupie: Populating --> Populate

core/config/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+For example, if we have a YAML configuration like so:
```
Lupie: Make these split sentences easier to handle across code blocks:
"For example, say we have the following YAML configuration:"

core/config/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+You could access the value using "dotted notation":
+
+```go
+foo := provider.GetValue("one.two").AsString()
+fmt.Println(foo)
+// Output: hello
+```
+
+The `As*` method come in two variants: `TryAs*` and `As*`. The former is a
```
Lupie: come --> comes

Or better yet:
come in --> has

core/config/README.md
```diff
@@ -0,0 +1,91 @@
...
+two-value return, similar to a type assertion, where you check if the second
```
Lupie: You (the user) checks? Try not to use "you" in these public docs for anything but the human user.

core/config/README.md
```diff
@@ -0,0 +1,91 @@
...
+`bool` is true before using the value.
+
+The `As*` are similar to the `Must*` pattern in the standard library. If the
```
Lupie: are similar to -->  method is similar to…

Or are you talking about the variant? If so, say "variant" instead of "method."

core/config/README.md
```diff
@@ -0,0 +1,91 @@
...
+underlying value cannot be converted to the requested type, `As*` will `panic`.
+
+## PopulateStruct
+
+`PopulateStruct` is akin to `json.Unmarshal()`, in that it takes a pointer to a
```
Lupie: Remove comma

core/config/README.md
```diff
@@ -0,0 +1,91 @@
...
+custom struct and fills in the fields. It returns a `bool` if the requested
+value was found and populated properly.
```
Lupie: was --> is

Lupie: Might be silly question, but it doesn't return `True` in that case?

Lupie: Values aren't really populated, are they? Fields are populated with values?

core/config/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+For example, if we have a YAML file with:
```
Lupie: For example, say we have the following YAML file:

core/config/README.md
```diff
@@ -0,0 +1,91 @@
+# Configuration Package
+
+At a high level, configuration is any data that is used in an application but
+not part of the application itself. Any reasonable complex system will need to
+have knobs to tune and not everything can have intelligent defaults.
```
sectioneight: awsm

core/config/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+In UberFx, we try very hard to make sure that users only need to provide the
+minimal amount of configuration necessary to have working components, while also
+allowing them to override any field they need to if the defaults don't make
+sense for them.
```
sectioneight: SO much better.

core/config/README.md
sectioneight: Yes to the first comment.

The rest makes sense too. Updating.

core/config/README.md
sectioneight: So, so much better.

core/config/README.md
sectioneight: Done

core/config/README.md
sectioneight: Done

core/config/README.md
sectioneight: Dope, done

core/config/README.md
sectioneight: man, you're good!

core/config/README.md
sectioneight: Great suggestion. I need to stop using "you"

core/config/README.md
sectioneight: Nope, your first sentence is correct. fixing.

core/config/README.md
sectioneight: Done

core/config/README.md
sectioneight: Great catch on all three.

core/config/README.md
sectioneight: Done

PR #39, opened 2016-11-04
 Updated YamlProvider to support parsing map

anuptalwalkar:
GFM-144

Added support to populate map[string]interface{} as a ConfigurationValue on PopulateStruct. Some test refactoring and cleanup.

### DIFF
--- a/core/config/config_test.go
+++ b/core/config/config_test.go
@@ -177,7 +177,7 @@ func TestSimpleConfigValues(t *testing.T) {
 	assert.Equal(t, 1.123, provider.GetValue("float").AsFloat())
 	nested := &nested{}
 	v := provider.GetValue("nonexisting")
-	assert.False(t, v.PopulateStruct(nested))
+	assert.Nil(t, v.PopulateStruct(nested))
 }
 
 func TestNestedStructs(t *testing.T) {
@@ -215,7 +215,7 @@ func TestArrayOfStructs(t *testing.T) {
 	v := provider.GetValue("")
 
 	assert.True(t, v.HasValue())
-	assert.True(t, v.PopulateStruct(target))
+	assert.Nil(t, v.PopulateStruct(target))
 	assert.Equal(t, 0, target.Things[0].ID1)
 	assert.Equal(t, -2, target.Things[2].ID1)
 }
@@ -229,7 +229,7 @@ func TestDefault(t *testing.T) {
 	target := &nested{}
 	v := provider.GetValue("")
 	assert.True(t, v.HasValue())
-	assert.True(t, v.PopulateStruct(target))
+	assert.Nil(t, v.PopulateStruct(target))
 	assert.Equal(t, "default_name", target.Name)
 }
 
--- a/core/config/config_test.go
+++ b/core/config/config_test.go
@@ -177,7 +177,7 @@ func TestSimpleConfigValues(t *testing.T) {
 	assert.Equal(t, 1.123, provider.GetValue("float").AsFloat())
 	nested := &nested{}
 	v := provider.GetValue("nonexisting")
-	assert.Nil(t, v.PopulateStruct(nested))
+	assert.NoError(t, v.PopulateStruct(nested))
 }
 
 func TestNestedStructs(t *testing.T) {
@@ -215,7 +215,7 @@ func TestArrayOfStructs(t *testing.T) {
 	v := provider.GetValue("")
 
 	assert.True(t, v.HasValue())
-	assert.Nil(t, v.PopulateStruct(target))
+	assert.NoError(t, v.PopulateStruct(target))
 	assert.Equal(t, 0, target.Things[0].ID1)
 	assert.Equal(t, -2, target.Things[2].ID1)
 }
@@ -229,7 +229,7 @@ func TestDefault(t *testing.T) {
 	target := &nested{}
 	v := provider.GetValue("")
 	assert.True(t, v.HasValue())
-	assert.Nil(t, v.PopulateStruct(target))
+	assert.NoError(t, v.PopulateStruct(target))
 	assert.Equal(t, "default_name", target.Name)
 }
 
--- a/core/config/value.go
+++ b/core/config/value.go
@@ -336,14 +336,14 @@ func convertValue(value interface{}, targetType reflect.Type) (interface{}, erro
 }
 
 // PopulateStruct fills in a struct from configuration
-func (cv Value) PopulateStruct(target interface{}) bool {
+func (cv Value) PopulateStruct(target interface{}) error {
 	if !cv.HasValue() {
-		return false
+		return nil
 	}
 
-	_, found, _ := cv.getValueStruct(cv.key, target)
+	_, err := cv.getValueStruct(cv.key, target)
 
-	return found
+	return err
 }
 
 func (cv Value) getGlobalProvider() Provider {
@@ -353,7 +353,7 @@ func (cv Value) getGlobalProvider() Provider {
 	return cv.root
 }
 
-func (cv Value) getValueStruct(key string, target interface{}) (interface{}, bool, error) {
+func (cv Value) getValueStruct(key string, target interface{}) (interface{}, error) {
 	// walk through the struct and start asking the providers for values at each key.
 	//
 	// - for individual values, we terminate
@@ -361,7 +361,6 @@ func (cv Value) getValueStruct(key string, target interface{}) (interface{}, boo
 	// - for object values, we recurse.
 	//
 
-	found := false
 	targetValue := reflect.Indirect(reflect.ValueOf(target))
 	targetType := targetValue.Type()
 	// if b := getBucket(targetValue); b == bucketInvalid {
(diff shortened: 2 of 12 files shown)


core/config/yaml_test_data.go
```diff
@@ -0,0 +1,68 @@
...
+
+var pointerYaml = []byte(`
+pointerStruct:
+  myTrueBool: true
+  myFalseBool: false
+  myString: hello
+`)
+
+type mapStruct struct {
+	MyMap map[string]interface{} `yaml:"oneTrueMap"`
```
glibsm: i'd rest easier if this had another struct in it, which gets serialized correctly from some other fields in the yaml

anuptalwalkar: Added nested struct into existing test

core/config/value.go
```diff
@@ -487,8 +486,26 @@ func (cv Value) getValueStruct(key string, target interface{}) (interface{}, boo
...
+			if val != nil {
+				switch v := val.(type) {
+				case map[interface{}]interface{}:
+					destMap := createMap.(map[string]interface{})
+					for key, value := range v {
+						switch key := key.(type) {
+						case string:
+							destMap[key] = value
+						default:
+							panic(fmt.Sprintf("Can't parse the input yaml. Map key must be string: %v", key))
```
sectioneight: is panic() the right option here? Is there another way we can bubble up the error? Under what circumstances would we encounter this path?

core/config/value.go
```diff
@@ -487,8 +486,26 @@ func (cv Value) getValueStruct(key string, target interface{}) (interface{}, boo
...
 			}
-		}
+		case bucketMap:
+			createMap := reflect.MakeMap(fieldType).Interface()
+			val := global.GetValue(childKey).Value()
 
+			if val != nil {
+				switch v := val.(type) {
+				case map[interface{}]interface{}:
+					destMap := createMap.(map[string]interface{})
```
sectioneight: Is this type assertion always going to be correct? This is the case for `reflect.Map` but it could be a map of anything? What if the type is some other type of map? Can we add some tests around this?

sectioneight:
Hmm, is int a valid key type in YAML? In JSON keys can only be strings, and
yaml is a superset of json so they generally have similar semantics. Worth
checking into.

core/config/value.go
sectioneight: Yea, silently passing would be bad. Panicing doesn't seem like the right option. This is in `getValueStruct` which is called during `PopulateStruct`, which could be called in the middle of an app's lifecycle, not only during startup. Panic would be acceptable if it was only during app initialization, but we don't want UberFx to **ever** panic during a running (properly configured) application.

We do have the option to `return false` from PopulateStruct, which seems like the right thing to do here -- indicate that we couldn't deserialize properly. So we should probably log an error here, and then `return nil, false, errors.New("couldn't...")` (the middle value being the "ok") one so that the caller knows the PopulateStruct didn't work right.

Also we should follow up and have `PopulateStruct` return a proper `error` (or `nil`) rather than just a `bool`.

anuptalwalkar:
hmm. JSON keys can be strings, but with YAML you can put integer as map keys and have map[int]string, etc. I haven't seen it used in practice though. I am keeping it to just handle string keys for now. Let me know what you think, I can add more type assertions and handle map[type]intrface{} scenarios.

sectioneight:
let's make it a goal to always increase coverage, never decrease (unless removing unused, tested code)

core/config/config_test.go
```diff
@@ -177,7 +177,7 @@ func TestSimpleConfigValues(t *testing.T) {
 	assert.Equal(t, 1.123, provider.GetValue("float").AsFloat())
 	nested := &nested{}
 	v := provider.GetValue("nonexisting")
-	assert.False(t, v.PopulateStruct(nested))
+	assert.Nil(t, v.PopulateStruct(nested))
```
sectioneight: This should be `NoError`, not `nil` if the return type is `error`

core/config/config_test.go
```diff
@@ -215,7 +215,7 @@ func TestArrayOfStructs(t *testing.T) {
 	v := provider.GetValue("")
 
 	assert.True(t, v.HasValue())
-	assert.True(t, v.PopulateStruct(target))
+	assert.Nil(t, v.PopulateStruct(target))
```
sectioneight: NoError

core/config/config_test.go
```diff
@@ -229,7 +229,7 @@ func TestDefault(t *testing.T) {
 	target := &nested{}
 	v := provider.GetValue("")
 	assert.True(t, v.HasValue())
-	assert.True(t, v.PopulateStruct(target))
+	assert.Nil(t, v.PopulateStruct(target))
```
sectioneight: NoError

core/config/value.go
```diff
@@ -337,14 +336,14 @@ func convertValue(value interface{}, targetType reflect.Type) (interface{}, erro
...
 	if !cv.HasValue() {
-		return false
+		return nil
 	}
 
-	_, found, _ := cv.getValueStruct(cv.key, target)
+	_, err := cv.getValueStruct(cv.key, target)
 
-	return found
+	return err
```
sectioneight: Much better

core/config/value.go
```diff
@@ -436,15 +433,16 @@ func (cv Value) getValueStruct(key string, target interface{}) (interface{}, boo
 			newTarget := reflect.New(ntt)
 			if v2 := global.GetValue(childKey); v2.HasValue() {
 
-				v2.PopulateStruct(newTarget.Interface())
-
+				err := v2.PopulateStruct(newTarget.Interface())
+				if err != nil {
```
sectioneight: Block scope this:

```go
if err := v2.Populate... err != nil { 
  return nil, err
}
```

core/config/value.go
```diff
@@ -466,7 +464,10 @@ func (cv Value) getValueStruct(key string, target interface{}) (interface{}, boo
 				case bucketObject:
 					newTarget := reflect.New(elementType)
 					if v2 := global.GetValue(arrayKey); v2.HasValue() {
-						v2.PopulateStruct(newTarget.Interface())
+						err := v2.PopulateStruct(newTarget.Interface())
+						if err != nil {
```
sectioneight: block scope to not leak errors (see above comment

core/config/value.go
```diff
@@ -479,23 +480,34 @@ func (cv Value) getValueStruct(key string, target interface{}) (interface{}, boo
...
+		case bucketMap:
+			destMap := reflect.MakeMap(fieldType).Interface()
+			val := global.GetValue(childKey).Value()
 
-	if found {
-		if errs := validator.Validate(target); errs != nil {
-			return nil, true, errs
+			if val != nil {
+				switch v := val.(type) {
+				case map[interface{}]interface{}:
```
sectioneight: co case for `map[string]interface{}`? can you comment on why that is? also, a `switch` statement with a single `case` isn't very useful, perhaps this would be clearer as an if statement?

core/config/value.go
```diff
@@ -479,23 +480,34 @@ func (cv Value) getValueStruct(key string, target interface{}) (interface{}, boo
...
+					for key, value := range v {
+						switch key := key.(type) {
+						case string:
+							destMap := destMap.(map[string]interface{})
+							destMap[key] = value
+						default:
+							return nil, fmt.Errorf("Can't populate struct from the input yaml. Map key must be a string: %v", key)
```
sectioneight: nit: errors are always lowercase, to be interpolated into sentences. "can't populate struct"...

core/config/value.go
```diff
@@ -479,23 +480,34 @@ func (cv Value) getValueStruct(key string, target interface{}) (interface{}, boo
...
+					}
+					fieldValue.Set(reflect.ValueOf(destMap))
+				}
+			}
 		}
-		return target, true, nil
 	}
-	return nil, false, nil
+	errs := validator.Validate(target)
+	return target, errs
```
sectioneight: why rot just `retur target, validator.Validate(target)`?

core/config/yaml_test.go
```diff
@@ -60,7 +60,7 @@ func TestYamlStructRoot(t *testing.T) {
 
 	cs := &configStruct{}
 
-	assert.True(t, provider.GetValue("").PopulateStruct(cs))
+	assert.Nil(t, provider.GetValue("").PopulateStruct(cs))
```
sectioneight: NoError

core/config/yaml_test.go
```diff
@@ -75,7 +75,7 @@ func TestYamlStructChild(t *testing.T) {
 
 	cs := &rpcStruct{}
 
-	assert.True(t, provider.GetValue("modules.rpc").PopulateStruct(cs))
+	assert.Nil(t, provider.GetValue("modules.rpc").PopulateStruct(cs))
```
sectioneight: NoError

core/config/yaml_test.go
```diff
@@ -94,7 +94,7 @@ func TestNewYamlProviderFromReader(t *testing.T) {
 	buff := bytes.NewBuffer([]byte(yamlConfig1))
 	provider := NewYamlProviderFromReader(ioutil.NopCloser(buff))
 	cs := &configStruct{}
-	assert.True(t, provider.GetValue("").PopulateStruct(cs))
+	assert.Nil(t, provider.GetValue("").PopulateStruct(cs))
```
sectioneight: NoError

core/config/value.go
anuptalwalkar: Changed to an if statement. yamlnode is always made as map[interface{}]interface{}, this is a check for that. I'll add a comment.

PR #41, opened 2016-11-07
Updated TryAsInt to try conversion for floats

anuptalwalkar:
GFM-153

Instead of failing on TryAsInt when a float value is provided, try parsing and converting to int with appropriate type conversion.

### DIFF
--- a/core/config/config_test.go
+++ b/core/config/config_test.go
@@ -180,6 +180,19 @@ func TestSimpleConfigValues(t *testing.T) {
 	assert.NoError(t, v.PopulateStruct(nested))
 }
 
+func TestGetAsIntegerValue(t *testing.T) {
+	cv := NewValue(NewStaticProvider(nil), "key", float32(2), true, Integer, nil)
+	assert.Equal(t, 2, cv.AsInt())
+	cv = NewValue(NewStaticProvider(nil), "key", int(2), true, Integer, nil)
+	assert.Equal(t, 2, cv.AsInt())
+	cv = NewValue(NewStaticProvider(nil), "key", float64(2), true, Integer, nil)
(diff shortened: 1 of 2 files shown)


core/config/config_test.go
```diff
@@ -180,6 +180,19 @@ func TestSimpleConfigValues(t *testing.T) {
 	assert.NoError(t, v.PopulateStruct(nested))
 }
 
+func TestGetAsIntegerValue(t *testing.T) {
+	cv := NewValue(NewStaticProvider(nil), "key", float32(2), true, Integer, nil)
```
sectioneight: nit: may be better as subtests (less copy-pasta)

core/config/value.go
```diff
@@ -171,8 +171,18 @@ func (cv Value) TryAsInt() (int, bool) {
 	if val, err := convertValue(v, reflect.TypeOf(0)); v != nil && err == nil {
 		return val.(int), true
 	}
-	return 0, false
-
+	switch val := v.(type) {
+	case int32:
```
sectioneight: Can we collapse these cases?

```
switch val := val.(type) {
case int32, int64, float32, float64:
  return int(val), true
default:
  return 0, false
}
```

anuptalwalkar: We can't collapse the switch here. It errors at int(val) and the code needs type assertion.

sectioneight: lame :(

core/config/config_test.go
anuptalwalkar: subtests?

sectioneight: https://blog.golang.org/subtests

basically build up a set of test cases and range over them

anuptalwalkar: sweet

PR #43, opened 2016-11-07
Initialize map after nil value check

anuptalwalkar:
struct may have a map, but not always defined in the yaml file. initialize the map only when needed.


anuptalwalkar:
I am merging this Coverage decreased (-0.005%), since the change is really removing one blank line.

PR #45, opened 2016-11-08
Improved map parsing

anuptalwalkar:
GFM-144

### DIFF
--- a/core/config/yaml_test_data.go
+++ b/core/config/yaml_test_data.go
@@ -48,6 +48,8 @@ type nestedStruct struct {
 
 type mapStruct struct {
 	MyMap        map[string]interface{} `yaml:"oneTrueMap"`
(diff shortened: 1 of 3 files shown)


core/config/yaml_test_data.go
```diff
@@ -57,17 +59,23 @@ mapStruct:
...
+    one: uno
+    two: dos
   nestedStruct:
     additionalData: nesteddata
 `)
 
-type unsupportedMapStruct struct {
-	UnsupportedMap map[int]interface{} `yaml:"unsupportedMap"`
+type intKeyMapStruct struct {
+	IntKeyMap map[int]string `yaml:"intKeyMap"`
```
glibsm: string map with int map hint? something is fishy here

anuptalwalkar: It is just testing that [int]string would succeed, its a valid map.

PR #47, opened 2016-11-08
Fix map test with int key

sectioneight:
Not sure why this passed the previous PR, but on my mac it now works
properly. Hopefully this isn't a linux/mac thing?

### DIFF
--- a/service/host_test.go
+++ b/service/host_test.go
@@ -140,7 +140,7 @@ foo:
 	assert.Equal(t, 1, instance.ServiceConfig.Bar)
 }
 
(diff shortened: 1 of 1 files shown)


sectioneight:
@anuptalwalkar @glibsm -- not sure why this was passing for you guys, but it was definitely broken for me on HEAD but working on HEAD~, seems due to recent refactor (we now support `int` keys).

@uber-go/service-framework

sectioneight:
This is blocking other PRs. Going to self-merge for now, please take a look in the morning.

PR #49, opened 2016-11-08
Don't run coverage on tip

sectioneight:
Our tip builds are re-running the tests on `make coveralls` for some reason, causing 3+ minute delays in travis results. Plus, we only really want the coverage info from a single version since it will be more consistent that way.

On tip:

image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,4 +25,4 @@ script:
   - make test
 
 after_success:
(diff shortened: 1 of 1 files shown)


sectioneight:
After: 
image

sectioneight:
Self-merging, trivial test infra improvement.

PR #50, opened 2016-11-08
Add go report card and fix reported misspellings

sectioneight:
We have an A+, we should show it off.


sectioneight:
It looks like this:

image

Full preview here:

https://github.com/uber-go/fx/blob/723a1128c976926e918174fa47c8f1c58347b448/README.md

sectioneight:
@uber-go/service-framework

sectioneight:
Click the link and find out :)

PR #51, opened 2016-11-08
Benchcmp support in Makefile

glibsm:
Adds a few more targets to the makefile to easier generate the benchcmp after perf changes.

Workflow:

1. On master: `make benchbase`, that will create `.bench/old.txt`
2. Type type type perf changes
3. `make bench && make benchcmp` will show the diff before and after.

There are some guardrails about running `make benchbase` while not on master (should it be `origin/master?`), but can be overriden via `IGNORE_BASELINE_CHECK` (in case all you made so far are doc changes or something)


glibsm:
@uber-go/service-framework

Makefile
```diff
@@ -66,8 +66,37 @@ coveralls: $(COV_REPORT)
 
 .PHONY: bench
 BENCH ?= .
+BENCH_FILE ?= .bench/new.txt
 bench:
-	$(ECHO_V)$(foreach pkg,$(BENCH_PKGS),go test -bench=$(BENCH) -run="^$$" $(BENCH_FLAGS) $(pkg);)
+	$(ECHO_V)$(foreach pkg,$(BENCH_PKGS),go test -bench=. -run="^$$" $(BENCH_FLAGS) $(pkg) |\
```
sectioneight: nit: space betweeen `|` and `\` to make it clear they're two separate operators

Makefile
```diff
@@ -66,8 +66,37 @@ coveralls: $(COV_REPORT)
...
+		tee $(BENCH_FILE);)
+
+.PHONY: benchbase
+BASELINE_BENCH_FILE=.bench/old.txt
```
sectioneight: you usually define the PHONY right above the target. so swp this line with line 74

Makefile
```diff
@@ -66,8 +66,37 @@ coveralls: $(COV_REPORT)
...
+		exit 1; \
+	fi
+
+	@echo "$(LABEL_STYLE)Running baseline benchmark$(COLOR_RESET)"
+	@echo
+	$(ECHO_V)$(MAKE) bench BENCH_FILE=$(BASELINE_BENCH_FILE)
+
+.PHONY: benchcmp
+benchcmp:
+	$(ECHO_V)which benchcmp >/dev/null || go get -u golang.org/x/tools/cmd/benchcmp
```
sectioneight: Do we want to consider vendoring this via glide to have consistent versions? Or is it stable

Makefile
```diff
@@ -66,8 +66,37 @@ coveralls: $(COV_REPORT)
...
+	$(ECHO_V)test -s $(BASELINE_BENCH_FILE) ||\
```
sectioneight: nit: space between `||` and `\` to make it obvious they're two separate operators

Makefile
```diff
@@ -66,8 +66,37 @@ coveralls: $(COV_REPORT)
...
+		$(error "Must have a baseline bench file. *Check out master* and run `make benchbase`")
+	$(ECHO_V)test -s $(BENCH_FILE) || $(error "Must have a new benchmark file. Run `make bench`")
+	$(ECHO_V)benchcmp $(BASELINE_BENCH_FILE) $(BENCH_FILE)
+
+.PHONY: benchclean
+benchclean:
+	$(ECHO_V)if [ -a $(BASELINE_BENCH_FILE) ]; then rm $(BASELINE_BENCH_FILE); fi
```
sectioneight: -a
file exists

This is identical in effect to -e. It has been "deprecated," [1] and its use is discouraged.

http://www.tldp.org/LDP/abs/html/fto.html

Also just do rm -f $(BASELINE_BENCH_FILE) and don't bother checking it first.

.build/colors.mk
```diff
@@ -13,3 +13,6 @@ endif
 
 LABEL_STYLE = "$(COLOR_BOLD)$(COLOR_COMMAND)"
 ERROR_STYLE = "$(COLOR_BOLD)$(COLOR_ERROR)"
+
+label = echo $(COLOR_BOLD)$(COLOR_COMMAND)$(1)$(COLOR_NORMAL)
```
sectioneight: WOAH. TILabout `$(1)`. I've been wanting this forever!

Makefile
glibsm: I think it's pretty stable, we can vendor it in later if we want

sectioneight: Agreed. I'm full-on YAGNI and did this with thriftrw, which bit us eventually but would rather start everything simple and occasionally have to revisit than overdo it from the beginning.

sectioneight:
This is really, really cool Glib. Nice work.

PR #52, opened 2016-11-09
Initialize a global tracer with opentracing and jaegar in the service module

madhuravi:


madhuravi:
@uber-go/service-framework

core/tracing/tracer_test.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func TestLoadAppConfig(t *testing.T) {
+	jConfig := loadAppConfig(emptyJaegerConfig, logger)
+	assert.NotNil(t, jConfig)
+	assert.NotNil(t, jConfig.Logger)
+}
+
+func TestLoadAppConfig_JaegerConfigWithLogger(t *testing.T) {
+	jConfig := loadAppConfig(jaegerConfigWithLogger, logger)
```
anuptalwalkar: nit: may be could use helper withjaegerConfig func wrap up for the tests.

service/service.go
```diff
@@ -138,6 +139,20 @@ func New(options ...Option) (Owner, error) {
 		metrics.Freeze()
 	}
 
+	if svc.Tracer() == nil {
+		svc.configProvider.GetValue("tracing").PopulateStruct(&svc.tracerConfig)
```
anuptalwalkar: Not sure how lint passed, but you need to check for err returned on PopulateStruct

core/tracing/tracer.go
```diff
@@ -0,0 +1,78 @@
...
+)
+
+// InitGlobalTracer instantiates a new global tracer
+func InitGlobalTracer(
+	cfg *jaegerconfig.Configuration,
+	serviceName string,
+	logger ulog.Log,
+	scope tally.Scope,
+) (opentracing.Tracer, error) {
+	appCfg := cfg
```
anuptalwalkar: Why not just use cfg?

core/tracing/tracer.go
```diff
@@ -0,0 +1,78 @@
...
+
+type jaegerLogger struct {
+	log ulog.Log
+}
+
+func (jl *jaegerLogger) Error(msg string) {
+	jl.log.Error(msg)
+}
+
+func (jl *jaegerLogger) Infof(msg string, args ...interface{}) {
```
anuptalwalkar: comment on exported methods?

sectioneight:
I disabled errcheck on _test files since it's kinda annoying and not that
helpful

core/tracing/tracer.go
```diff
@@ -0,0 +1,78 @@
...
+	"go.uber.org/fx/core/ulog"
+
+	"github.com/opentracing/opentracing-go"
+	"github.com/uber-go/tally"
+	jaegerconfig "github.com/uber/jaeger-client-go/config"
+)
+
+// InitGlobalTracer instantiates a new global tracer
+func InitGlobalTracer(
+	cfg *jaegerconfig.Configuration,
```
madhuravi: Keeping this consistent with other core components like logging. Check service.go

core/tracing/tracer.go
```diff
@@ -0,0 +1,78 @@
...
+		appCfg = loadAppConfig(cfg, logger)
+		reporter = &jaegerReporter{
+			reporter: scope.Reporter(),
+		}
+	}
+	tracer, closer, err := appCfg.New(serviceName, reporter)
+	if err != nil {
+		return tracer, err
+	}
+	defer closer.Close()
```
madhuravi: This actually needs to be in service shutdown, moving it there.

core/tracing/tracer_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/uber-go/tally"
+	"github.com/uber-go/zap"
+	"github.com/uber/jaeger-client-go"
+	jaegerconfig "github.com/uber/jaeger-client-go/config"
+)
+
+var (
+	serviceName            = "serviceName"
```
madhuravi: are you saying this should be renamed to "_serviceName"? Don't see this naming elsewhere in the code.

core/tracing/tracer_test.go
madhuravi: Don't see that making these tests simpler/cleaner. Tests are simple enough and use different combinations of input arguments each time.

service/service.go
madhuravi: Good catch.

core/tracing/tracer.go
madhuravi: Done.

core/tracing/tracer.go
madhuravi: Done.

core/tracing/tracer.go
```diff
@@ -0,0 +1,78 @@
+package tracing
+
+import (
+	"time"
+
+	"go.uber.org/fx/core/ulog"
+
+	"github.com/opentracing/opentracing-go"
+	"github.com/uber-go/tally"
+	jaegerconfig "github.com/uber/jaeger-client-go/config"
```
sectioneight: why alias this? generally we should only alias a package if it conflicts with another import, but we're not importing another `config` package here, right?

core/tracing/tracer.go
```diff
@@ -0,0 +1,78 @@
...
+	var reporter *jaegerReporter
+	if cfg == nil || !cfg.Disabled {
+		appCfg = loadAppConfig(cfg, logger)
+		reporter = &jaegerReporter{
+			reporter: scope.Reporter(),
+		}
+	}
+	tracer, closer, err := appCfg.New(serviceName, reporter)
+	if err != nil {
+		return tracer, err
```
sectioneight: use `errors.Wrap` (search for other examples) whenever bubbling up errors from now on (i just added this today, so it's a new pattern)

core/tracing/tracer.go
```diff
@@ -0,0 +1,78 @@
...
+		jaegerlogger := &jaegerLogger{
+			log: logger,
+		}
+		appCfg.Logger = jaegerlogger
+	}
+	return appCfg
+}
+
+type jaegerLogger struct {
+	log ulog.Log
```
sectioneight: why not just embed this directly? then you don't have to proxy Error, Infof, etc.

core/tracing/tracer_test.go
```diff
@@ -0,0 +1,92 @@
...
+import (
+	"testing"
+
+	"go.uber.org/fx/core/ulog"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/uber-go/tally"
+	"github.com/uber-go/zap"
+	"github.com/uber/jaeger-client-go"
+	jaegerconfig "github.com/uber/jaeger-client-go/config"
```
sectioneight: just use the natural package name, config

core/tracing/tracer_test.go
sectioneight: yes, all globals at uber should be underscore-prefixed. we may have missed a few from shawn's initial prototype, but anup is correct that going forward we should fix these to conform to our style

core/tracing/tracer_test.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+var (
+	serviceName            = "serviceName"
+	logger                 = ulog.Logger()
```
sectioneight: grabbing this is a global seems weird. can you make a function that returns a ulog.Logger and use that instead? otherwise we get into weird situations where people configure the logger but you've got a reference to an older version. just generally avoid globals as much as possible, even in tests

core/ulog/in_memory_log.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/uber-go/zap"
+)
+
+// TestBuffer is a buffer used to test the zap logger
+type TestBuffer struct {
+	bytes.Buffer
+}
+
+// Sync is a noop
```
sectioneight: maybe mention that it's here to satisfy some interface?

core/ulog/in_memory_log.go
```diff
@@ -0,0 +1,41 @@
...
+	return output[:len(output)-1]
+}
+
+// Stripped returns buffer as a string without the newline
+func (b *TestBuffer) Stripped() string {
+	return strings.TrimRight(b.String(), "\n")
+}
+
+// WithInMemoryLogger creates an in-memory zap logger that can be used in tests
+func WithInMemoryLogger(t *testing.T, opts []zap.Option, f func(zap.Logger, *TestBuffer)) {
```
sectioneight: why is this exported? are you using it outside of this package? if not, please un-export. exported tests methods are a bit awkward (and yes, i've written a few, but very intentionally).

service/service.go
```diff
@@ -138,6 +139,20 @@ func New(options ...Option) (Owner, error) {
...
+		tracer, err := tracing.InitGlobalTracer(
+			&svc.tracerConfig,
+			svc.standardConfig.ServiceName,
+			svc.log,
+			svc.scope.SubScope("tracing"),
+		)
+		if err != nil {
+			return svc, err
```
sectioneight: wrap error

core/tracing/tracer_test.go
madhuravi: This is not a global though. Is this specifically for package-scope variables? Can you send me a few examples? Not able to find any code samples for this in uberfx or opengrok.

sectioneight: Anything that's not scoped to a function or a method is a global.

https://github.com/uber-go/fx/blob/master/core/config/config.go#L44-L50

Again, we haven't fixed them all, but this is the standard.

core/ulog/in_memory_log.go
madhuravi: yes, using this is ulog_test and tracer_test.

core/tracing/tracer_test.go
madhuravi: Good point. It does make tests simpler/more compact though.

madhuravi:
@sectioneight Seeing this lint error in the travis build (though it succeeds locally), any idea what it is? It has a reference to your personal repo but I dont see a dependency on this in uberfx glide. 

# github.com/sectioneight/md-to-godoc/render
../../github.com/sectioneight/md-to-godoc/render/godoc.go:30: undefined: blackfriday.CommonExtensions
../../github.com/sectioneight/md-to-godoc/render/godoc.go:42: cannot use GodocRenderer literal (type *GodocRenderer) as type blackfriday.Renderer in return argument:
	*GodocRenderer does not implement blackfriday.Renderer (missing AutoLink method)

core/tracing/tracer.go
madhuravi: The jaeger logger has a different interface, this is a conversion layer from jaeger logger to uberfx logger interface.

sectioneight:
Hmm i just merged some changes to the markdown tool, will look tomorrow.
Sorry for the interruption, it's almost certainly my fault.

core/ulog/in_memory_log.go
anuptalwalkar: We can move this to testutil at some point and use in-memory logger for validating logging messages. I am ok with exporting for now, may be add a TODO?

madhuravi: Good idea, I wasn't aware of testutils. I'll just move it there now.

sectioneight:
Yea, testutils seems like the right place (as long as we're conscious not
to make it a dumping ground). Fine to do in a future PR though.

service/host.go
```diff
@@ -149,6 +149,13 @@ func (s *host) shutdown(err error, reason string, exitCode *int) (bool, error) {
 		s.scope.Close()
 	}
 
+	// Flush tracing buffers
+	if s.tracerCloser != nil {
+		if err = s.tracerCloser.Close(); err != nil {
+			ulog.Logger().Error("Failure to close tracer", "error", err)
```
sectioneight: This is the global logger. This should use `s.Logger()` instead of the global one (there's also a mistake above which is probably why you did this, you don't need to necessarily fix that one here but we should make a note to do it later)

service/host.go
```diff
@@ -149,6 +149,13 @@ func (s *host) shutdown(err error, reason string, exitCode *int) (bool, error) {
 		s.scope.Close()
 	}
 
+	// Flush tracing buffers
+	if s.tracerCloser != nil {
```
sectioneight: Let's log.Debug that we're closing the tracer

core/tracing/tracer.go
```diff
@@ -0,0 +1,99 @@
...
+		jaegerlogger := &jaegerLogger{
+			log: logger,
+		}
+		cfg.Logger = jaegerlogger
+	}
+	return cfg
+}
+
+type jaegerLogger struct {
+	log ulog.Log
```
sectioneight: I tried to leave this before, but why not embed `log` directly, then you don't have to write the `Error` and `Infof` proxy methods yourself below

core/tracing/tracer_test.go
```diff
@@ -0,0 +1,99 @@
...
+func TestInitGlobalTracer_Disabled(t *testing.T) {
+	tracer, closer, err := InitGlobalTracer(_disabledJaegerConfig, _serviceName, getLogger(), _scope)
+	assert.NotNil(t, tracer)
+	assert.NotNil(t, closer)
+	assert.NoError(t, err)
+}
+
+func TestInitGlobalTracer_NoServiceName(t *testing.T) {
+	tracer, closer, err := InitGlobalTracer(_emptyJaegerConfig, "", getLogger(), _scope)
+	assert.NotNil(t, err)
```
sectioneight: `assert.Error(t, err)`

modules/rpc/yarpc_test.go
```diff
@@ -22,11 +22,11 @@ package rpc
 
 import (
 	"bytes"
-	"context"
 	"testing"
 
 	"go.uber.org/yarpc/encoding/raw"
 	"go.uber.org/yarpc/transport"
+	"golang.org/x/net/context"
```
sectioneight: Why this change? Did you accidentally downgrade yarpc? Newer versions of YARPC take the stdlib context.

service/host.go
```diff
@@ -140,7 +140,7 @@ func (s *host) shutdown(err error, reason string, exitCode *int) (bool, error) {
 	errs := s.stopModules()
 	if len(errs) > 0 {
 		for k, v := range errs {
-			ulog.Logger().Error("Failure to shut down module", "name", k.Name(), "error", v.Error())
+			s.Logger().Error("Failure to shut down module", "name", k.Name(), "error", v.Error())
```
sectioneight: noice

core/tracing/tracer.go
glibsm: @madhuravi I think this comment is good. We should change `log ulog.Log` to just `ulog.Log`, then you can delete a bunch of forwarding methods from here

madhuravi: I spoke to Aiden about this offline and forgot to update the comment. The methods signatures are not the same, see "Infof" versus "Info.

PR #53, opened 2016-11-09
Use updated md-to-godoc for better output

sectioneight:


doc.go
```diff
@@ -1,10 +1,10 @@
 /*
 Package fx is the UberFx Service Framework.
 
-[][doc]
-[][cov]
-[][ci]
-[]report-card (https://goreportcard.com/report/github.com/uber-go/fx)
+GoDoc (https://godoc.org/go.uber.org/fx)
```
sectioneight: @glibsm I finally fixed this stuff (or more accurately, the v2 api of the markdown parser I'm using finally recognizes these).

PR #54, opened 2016-11-09
Added benchmarks for logging

anuptalwalkar:
BenchmarkUlogWithoutFields-8                	 5000000	       242 ns/op	      48 B/op	       1 allocs/op
BenchmarkUlogWithFieldsLogIFace-8           	 1000000	      1149 ns/op	    1309 B/op	      20 allocs/op
BenchmarkUlogWithFieldsBaseLoggerStruct-8   	 2000000	      1071 ns/op	    1052 B/op	      19 allocs/op
BenchmarkUlogWithFieldsZapLogger-8          	 3000000	       577 ns/op	     513 B/op	       1 allocs/op
BenchmarkUlogLiteWithFields-8               	 3000000	       601 ns/op	     361 B/op	       8 allocs/op
BenchmarkUlogWithFieldsPreset-8             	 2000000	       902 ns/op	     819 B/op	       3 allocs/op
PASS
ok  	go.uber.org/fx/core/ulog	13.243s

### DIFF
--- a/core/ulog/log_benchmark_test.go
+++ b/core/ulog/log_benchmark_test.go
@@ -40,11 +40,16 @@ func logfields() []interface{} {
 	}
 }
 
+func discardedLogger() zap.Logger {
+	return zap.New(
+		zap.NewJSONEncoder(),
+		zap.DiscardOutput,
+	)
+}
+
 func BenchmarkUlogWithoutFields(b *testing.B) {
 	log := Logger()
-	log.SetLogger(zap.New(
-		zap.NewJSONEncoder(),
-		zap.DiscardOutput))
+	log.SetLogger(discardedLogger())
 	b.ResetTimer()
 	b.RunParallel(func(pb *testing.PB) {
 		for pb.Next() {
@@ -55,10 +60,7 @@ func BenchmarkUlogWithoutFields(b *testing.B) {
 
 func BenchmarkUlogWithFields(b *testing.B) {
 	log := Logger()
-	log.SetLogger(zap.New(
-		zap.NewJSONEncoder(),
-		zap.DiscardOutput,
-	))
+	log.SetLogger(discardedLogger())
 	b.ResetTimer()
 	b.RunParallel(func(pb *testing.PB) {
 		for pb.Next() {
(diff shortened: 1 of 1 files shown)


core/ulog/log_benchmark_test.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func BenchmarkUlogLiteWithFields(b *testing.B) {
+	log := Logger()
+	log.SetLogger(zap.New(
+		zap.NewJSONEncoder(),
+		zap.DiscardOutput,
+	))
+	b.ResetTimer()
+	b.RunParallel(func(pb *testing.PB) {
```
sectioneight: what is the difference between `RunParallel` and the usual `range` over the number of items in the `testing.B`?

core/ulog/log_benchmark_test.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+	})
+}
+
+func BenchmarkUlogWithFields(b *testing.B) {
+	log := Logger()
+	log.SetLogger(zap.New(
+		zap.NewJSONEncoder(),
+		zap.DiscardOutput,
+	))
```
sectioneight: nit: these 4 lines are exactly the same in every benchmark. extract into a function called `getDiscardLogger` or something

core/ulog/log_benchmark_test.go
anuptalwalkar: hmm. RunParallel runs benchmark in parallel by creating goroutines and distributing iterations. I guess it would end up being a little faster than ranging over tests.

sectioneight:
But are the semantics the same between serial and parallel? In other words, are we going to get the same benchmark numbers but just finish the test quicker, or are we measuring something different (e.g. how this would behave when logging from many goroutines at once)?

glibsm:
talked with @anuptalwalkar offline at length. We're going to update the suite to illustrate differences between calling log on an interface, calling log on a struct, or calling log with zap fields directly on the raw logger.

core/ulog/log_benchmark_test.go
glibsm: I think it helps to evaluate lock contention, which in this case doesn't matter since logger doesn't lock anything.

glibsm:
Could you update the PR description with the new bench?

PR #55, opened 2016-11-09
Updated yamlProvider to parse time.Duration

anuptalwalkar:
GFM-158

### DIFF
--- a/core/config/value.go
+++ b/core/config/value.go
@@ -50,6 +50,8 @@ const (
 	Dictionary
 )
 
+var _zeroTimeDuration = time.Duration(0)
+
 // GetValueType returns GO type of the provided object
 func GetValueType(value interface{}) ValueType {
 	if value == nil {
@@ -421,7 +423,7 @@ func (cv Value) getValueStruct(key string, target interface{}) (interface{}, err
 				continue
 			}
 
-			if fieldType == reflect.TypeOf(time.Duration(0)) {
+			if fieldType == reflect.TypeOf(_zeroTimeDuration) {
 				if v := global.GetValue(childKey); v.HasValue() {
 					duration, err := time.ParseDuration(v.AsString())
 					if err != nil {
--- a/core/config/value.go
+++ b/core/config/value.go
@@ -421,6 +421,17 @@ func (cv Value) getValueStruct(key string, target interface{}) (interface{}, err
 				continue
 			}
 
+			if fieldType == reflect.TypeOf(time.Duration(0)) {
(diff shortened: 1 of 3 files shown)


core/config/value.go
```diff
@@ -421,6 +421,17 @@ func (cv Value) getValueStruct(key string, target interface{}) (interface{}, err
 				continue
 			}
 
+			if fieldType == reflect.TypeOf(time.Duration(0)) {
```
sectioneight: nit: don't call type of duration in the hot path. define it outside.

core/config/yaml_test.go
```diff
@@ -209,3 +210,22 @@ func TestMapParsingMapWithNonStringKeys(t *testing.T) {
 		assert.Equal(t, "onetwothree", ik.IntKeyMap[123])
 	})
 }
+
+func TestDurationParsing(t *testing.T) {
+	withYamlBytes(t, durationYaml, func(provider Provider) {
+		ds := durationStruct{}
+		err := provider.GetValue("durationStruct").PopulateStruct(&ds)
+		assert.NoError(t, err)
+		expectedDuration, err := time.ParseDuration("30s")
```
sectioneight: maybe better to check that it's equal to 30 * time.Second for clarity

glibsm: lets also add a few tests for hour and minute durations

core/config/value.go
glibsm: i.e. move `reflect.TypeOf(time.Duration(0))` to a const

core/config/yaml_test.go
anuptalwalkar: Good idea.

core/config/value.go
```diff
@@ -421,6 +423,17 @@ func (cv Value) getValueStruct(key string, target interface{}) (interface{}, err
 				continue
 			}
 
+			if fieldType == reflect.TypeOf(_zeroTimeDuration) {
```
sectioneight: Not quite. Call `reflect.TypeOf()` outside of the hot path.

sectioneight: Yea. Reflect are the expensive calls. It's a micro-optimization, but this is some perf-critical code, so we should aim to reduce expensive calls as much as possible.

anuptalwalkar: I totally agree with being aware of writing perf critical code in hot paths. #willnothappenagain

sectioneight: #neverforget

PR #59, opened 2016-11-10
Log documentation

### LINKED ISSUE #58: Please use self-documenting signatures in the Log interface
Currently ulog.Log interface is written with methods like this:
```go
Debug(string, ...interface{})
```

* Is the first parameter a format string (like in Printf) or a plain message string?
* Are the arguments interpreted as in Printf or as alternating key-value pairs as in go-kit?

anuptalwalkar:
GFM-160
Closes #58

### DIFF
--- a/core/ulog/README.md
+++ b/core/ulog/README.md
@@ -1,13 +1,13 @@
 # Logging package
 
-ulog provides an interface and abstraction layer over the logger implementation used underneath,
-and provides simple APIs for logging. The logger is instantiated as logger with default options and can be configured
+ulog provides an API wrapper around the logging library (zap Logger)
+The logger is instantiated as logger with default options and can be configured
 via `Configure()` API and provided yaml configuration.
 
 ```go
 package main
 
-import "go.uber.org/core/ulog"
+import "go.uber.org/fx/core/ulog"
 
 func main() {
   // Initialize logger object
@@ -22,7 +22,7 @@ func main() {
 }
 ```
 
-Note that the log methods (`Info`,`Warn`, `Debug`) takes parameter as key value pairs for formatting (message, (key, value)...)
+Note that the log methods (`Info`,`Warn`, `Debug`) takes parameter as key value pairs (message, (key, value)...)
 
 ulog configuration can be defined in multiple ways, either by writing the struct yourself, or describing in the yaml
 and populating using config package.
@@ -44,7 +44,7 @@ loggingConfig := ulog.Configuration {
 ```
 
 ```go
-  var loggingConfig ulog.Configuration
+var loggingConfig ulog.Configuration
 
-  err := cfg.GetValue("logging").PopulateStruct(&loggingConfig)
+err := cfg.GetValue("logging").PopulateStruct(&loggingConfig)
 ```
--- a/core/ulog/README.md
+++ b/core/ulog/README.md
@@ -1,38 +1,39 @@
-# Ulog package
+# Logging package
 
 ulog provides an interface and abstraction layer over the logger implementation used underneath,
-and provides simple APIs for logging. The logger is instantiated as default logger and can be configured
-Configure() API and provided yaml configuration.
+and provides simple APIs for logging. The logger is instantiated as logger with default options and can be configured
+via `Configure()` API and provided yaml configuration.
 
 ```go
 package main
 
-import (
-  "go.uber.org/core/ulog"
-)
+import "go.uber.org/core/ulog"
 
 func main() {
   // Initialize logger object
   log := ulog.Logger()
 
-  // Configure logger with configuration preferred by your service
-  log.Configure(&ulog.Configuration{})
+  // Optional, configure logger with configuration preferred by your service
+  logConfig := ulog.Configuration{}
+  log.Configure(&logConfig)
 
   // Use logger in your service
   log.Info("message describing loggging reason", "key", "value")
 }
 ```
 
-Note that the log methods (Info, Warn, Debug) takes parameter as key value pairs for formatting (message, (key, value)...)
+Note that the log methods (`Info`,`Warn`, `Debug`) takes parameter as key value pairs for formatting (message, (key, value)...)
 
 ulog configuration can be defined in multiple ways, either by writing the struct yourself, or describing in the yaml
 and populating using config package.
 
 * Defining config structure:
 
+```go
 loggingConfig := ulog.Configuration {
   stdout: true,
 }
+```
 
 * Fetching configuration from yaml:
 
--- a/core/ulog/README.md
+++ b/core/ulog/README.md
@@ -1,0 +1,49 @@
+# Ulog package
+
+ulog provides an interface and abstraction layer over the logger implementation used underneath,
+and provides simple APIs for logging. The logger is instantiated as default logger and can be configured
+Configure() API and provided yaml configuration.
+
+```go
+package main
+
+import (
+  "go.uber.org/core/ulog"
+)
+
+func main() {
+  // Initialize logger object
+  log := ulog.Logger()
+
+  // Configure logger with configuration preferred by your service
+  log.Configure(&ulog.Configuration{})
+
+  // Use logger in your service
+  log.Info("message describing loggging reason", "key", "value")
+}
+```
+
+Note that the log methods (Info, Warn, Debug) takes parameter as key value pairs for formatting (message, (key, value)...)
+
+ulog configuration can be defined in multiple ways, either by writing the struct yourself, or describing in the yaml
+and populating using config package.
+
+* Defining config structure:
+
+loggingConfig := ulog.Configuration {
+  stdout: true,
+}
+
+* Fetching configuration from yaml:
+
+```yaml
+  logging:
+    stdout: true
+    level: Debug
+```
+
+```go
+  var loggingConfig ulog.Configuration
+
+  err := cfg.GetValue("logging").PopulateStruct(&loggingConfig)
+```
--- a/core/ulog/README.md
+++ b/core/ulog/README.md
@@ -38,7 +38,7 @@ loggingConfig := ulog.Configuration{
 * Configuration defined in YAML:
 
 ```yaml
-  logging:
-    stdout: true
-    level: debug
+logging:
+  stdout: true
+  level: debug
 ```
--- a/core/ulog/README.md
+++ b/core/ulog/README.md
@@ -2,7 +2,7 @@
 
 ulog provides an API wrapper around the logging library (zap Logger)
 The logger is instantiated as logger with default options and can be configured
-via `Configure()` API and provided yaml configuration.
+via `Configure()` API and provided YAML configuration.
 
 ```go
 package main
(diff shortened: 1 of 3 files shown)


core/ulog/log.go
```diff
@@ -60,20 +60,44 @@ type baselogger struct {
...
 	Check(zap.Level, string) *zap.CheckedMessage
+
+	// RawLogger returns underlying logger implementation (zap.Logger) to get around the ulog.Log interface
 	RawLogger() zap.Logger
 
+	// Log at the provided zap.Level with message, and parameters as key value pairs
 	Log(zap.Level, string, ...interface{})
+
+	// Debug logs at Debug level with message, and parameters as key value pairs
 	Debug(string, ...interface{})
```
sectioneight: I think what @yurishkuro was getting at was named parameters, e.g.
```go
Debug(msg string, args...interface{})
```

core/ulog/doc.go
```diff
@@ -0,0 +1,66 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// Package ulog is the Logging package
```
sectioneight: Did you write this by hand, or use the md-to-godoc? I don't see a `README.md` file in this diff.

core/ulog/README.md
```diff
@@ -0,0 +1,49 @@
+# Ulog package
```
sectioneight: I would call this "Logging package"

core/ulog/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+ulog provides an interface and abstraction layer over the logger implementation used underneath,
+and provides simple APIs for logging. The logger is instantiated as default logger and can be configured
+Configure() API and provided yaml configuration.
+
+```go
+package main
+
+import (
+  "go.uber.org/core/ulog"
```
sectioneight: single imports would be a single line:

```go
import "go.uber.org/fx/core/ulog"
```

also, missing **fx** in the import path

core/ulog/README.md
```diff
@@ -0,0 +1,49 @@
+# Ulog package
+
+ulog provides an interface and abstraction layer over the logger implementation used underneath,
```
sectioneight: This sentence is a bit weird. @lupie halp?

core/ulog/README.md
```diff
@@ -0,0 +1,49 @@
...
+and provides simple APIs for logging. The logger is instantiated as default logger and can be configured
```
sectioneight: is instantiated as default logger? Something is weird here

core/ulog/README.md
```diff
@@ -0,0 +1,49 @@
...
+Configure() API and provided yaml configuration.
```
sectioneight: Pit backticks around code:

```
`Configure()` API and provided YAML configuration
```

sectioneight: also missing a word. and can be configured **via the** `Configure()` API or something

core/ulog/README.md
```diff
@@ -0,0 +1,49 @@
...
+import (
+  "go.uber.org/core/ulog"
+)
+
+func main() {
+  // Initialize logger object
+  log := ulog.Logger()
+
+  // Configure logger with configuration preferred by your service
+  log.Configure(&ulog.Configuration{})
```
sectioneight: Is this line necessary if they don't need any options?

core/ulog/README.md
```diff
@@ -0,0 +1,49 @@
...
+```
+
+Note that the log methods (Info, Warn, Debug) takes parameter as key value pairs for formatting (message, (key, value)...)
+
+ulog configuration can be defined in multiple ways, either by writing the struct yourself, or describing in the yaml
+and populating using config package.
+
+* Defining config structure:
+
+loggingConfig := ulog.Configuration {
```
sectioneight: This is a code block. Surround with triple-backticks

core/ulog/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+  // Configure logger with configuration preferred by your service
+  log.Configure(&ulog.Configuration{})
+
+  // Use logger in your service
+  log.Info("message describing loggging reason", "key", "value")
+}
+```
+
+Note that the log methods (Info, Warn, Debug) takes parameter as key value pairs for formatting (message, (key, value)...)
```
sectioneight: backticks around Info, Warn, and Debug (anything that is a method name)

core/ulog/README.md
```diff
@@ -0,0 +1,49 @@
...
+* Fetching configuration from yaml:
+
+```yaml
+  logging:
+    stdout: true
+    level: Debug
+```
+
+```go
+  var loggingConfig ulog.Configuration
```
sectioneight: this shouldn't be indented (or below)

core/ulog/README.md
anuptalwalkar: Updating to logger with default configuration*

core/ulog/README.md
anuptalwalkar: This is optional. I'll add comment and redo.

core/ulog/log.go
```diff
@@ -60,21 +60,45 @@ type baselogger struct {
 type Log interface {
 	// Configure Log object with the provided log.Configuration
 	Configure(Configuration)
-	// Create a child logger with some context
-	With(...interface{}) Log
-	SetLevel(zap.Level)
-	SetLogger(zap.Logger)
-	Check(zap.Level, string) *zap.CheckedMessage
+
+	// With creates a child logger with the provided paramets as key value pairs
```
yurishkuro: s/paramets/parameters/

> with the provided paramets as key value pairs

what does that mean to the user? Is it equivalent to passing these key-value pairs to all subsequent logging calls on that logger?

core/ulog/log.go
```diff
@@ -60,21 +60,45 @@ type baselogger struct {
 type Log interface {
 	// Configure Log object with the provided log.Configuration
 	Configure(Configuration)
```
yurishkuro: I would argue that Configuration, SetLogger, and SetLevel methods should be in some logger factory or builder, not in the Log interface itself. Constructing a logger and actually logging are two different and independent concerns, they do not need to be in the same interface. It widens the API surface and opens it for abuse, e.g. you can pass a Debug logger to some library, but the library can override the log level, which it should not be able to do.

core/ulog/log.go
```diff
@@ -60,21 +60,45 @@ type baselogger struct {
...
-	Log(zap.Level, string, ...interface{})
-	Debug(string, ...interface{})
-	Info(string, ...interface{})
-	Warn(string, ...interface{})
-	Error(string, ...interface{})
-	Panic(string, ...interface{})
-	Fatal(string, ...interface{})
-	DFatal(string, ...interface{})
+	// Log at the provided zap.Level with message, and parameters as key value pairs
+	Log(level zap.Level, message string, args ...interface{})
```
anuptalwalkar: Good point. Updating it.

core/ulog/log.go
anuptalwalkar: Updated.

core/ulog/log.go
anuptalwalkar: Yes. I will add a follow up on moving the methods out. I agree that Configuration and SetLogger can be untied from Log interface, but SetLevel pretty much needs a calling logger. If we are not using it, I don't have a strong case of having SetLevel functionality on logger.

core/ulog/README.md
anuptalwalkar: changed to "ulog provides an API wrapper around the logging library (zap Logger)"

core/ulog/log.go
anuptalwalkar: created GFM-172 to track this.

yurishkuro: It's fairly natural to do something like this
```go
log := ulog.LogBuilder().WithConfig(&Configuration{}).SetLevel(..).Build()
```

Once the logger is built, it's best if it can be immutable w.r.t. configuration parameters.

core/ulog/log.go
```diff
@@ -60,21 +60,45 @@ type baselogger struct {
 type Log interface {
 	// Configure Log object with the provided log.Configuration
 	Configure(Configuration)
-	// Create a child logger with some context
-	With(...interface{}) Log
-	SetLevel(zap.Level)
-	SetLogger(zap.Logger)
-	Check(zap.Level, string) *zap.CheckedMessage
+
+	// With creates a child logger with the provided parameters as key value pairs
```
yurishkuro: please explain what that means semantically. Does the child logger appends these k-v pairs to any other k-v pairs passed to logging methods? Or does it do something different with them?

core/ulog/README.md
```diff
@@ -0,0 +1,50 @@
...
+func main() {
+  // Initialize logger object
+  log := ulog.Logger()
+
+  // Optional, configure logger with configuration preferred by your service
+  logConfig := ulog.Configuration{}
+  log.Configure(&logConfig)
+
+  // Use logger in your service
+  log.Info("message describing loggging reason", "key", "value")
```
sectioneight: nit: Message, not message

core/ulog/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+Note that the log methods (`Info`,`Warn`, `Debug`) takes parameter as key value pairs (message, (key, value)...)
+
+ulog configuration can be defined in multiple ways, either by writing the struct yourself, or describing in the yaml
+and populating using config package.
+
+* Defining config structure:
+
+```go
+loggingConfig := ulog.Configuration {
```
sectioneight: no space between `Configuration` and opening `{`

core/ulog/README.md
```diff
@@ -0,0 +1,50 @@
...
+  stdout: true,
```
sectioneight: This should be `Stdout`, right?

core/ulog/README.md
```diff
@@ -0,0 +1,50 @@
...
+}
+```
+
+* Fetching configuration from yaml:
+
+```yaml
+  logging:
+    stdout: true
+    level: Debug
```
sectioneight: lowercase `debug`

core/ulog/README.md
```diff
@@ -0,0 +1,50 @@
...
+```
+
+```go
+var loggingConfig ulog.Configuration
+
+err := cfg.GetValue("logging").PopulateStruct(&loggingConfig)
```
sectioneight: not sure we should show this example, we do this internally only, right?

core/ulog/README.md
anuptalwalkar: I believe either way works. updated.

core/ulog/README.md
anuptalwalkar: Well, Configure is exposed for external use as well. I'll remove it for now.

core/ulog/README.md
```diff
@@ -0,0 +1,44 @@
...
+```go
+loggingConfig := ulog.Configuration{
+  Stdout: true,
+}
+```
+
+* Configuration defined in YAML:
+
+```yaml
+  logging:
```
sectioneight: nit: de-dent these two the beginning of the line (and the next two lines 2 spaces in as well)

PR #62, opened 2016-11-11
Add service docs

sectioneight:


core/config/doc.go
```diff
@@ -33,15 +33,15 @@
 //
 // Provider
 //
-// Provider is the interface for anything that can provide values.
+//  Provider is the interface for anything that can provide values.
```
anuptalwalkar: Why extra space at the beginning?

sectioneight: It's a bug I just fixed (see #76). Will regen.

sectioneight:
Ok, ready for review new @uber-go/service-framework

sectioneight:
also @lupie

service/README.md
```diff
@@ -0,0 +1,33 @@
+# Service lifecycle
```
Lupie: The main README uses title case, so let's stick with that (i.e., Service Lifecycle).

service/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+Service, being a bit of an overloaded term, requires some
+specific care to explain the various components in the `service`
+package in UberFx.
+
+## Instantiation
+
+Generally, you create a service in one of two ways:
+
+* The builder pattern, e.g. `service.WithModules(...).Build()`
```
Lupie: Quick fix: "The builder pattern (e.g., `service.WithModules(…).Build()`)"

Longer fix: Is this the way everyone always uses the builder pattern? It seems like "in this case" or "that is" might be less confusing to readers than "e.g."

service/README.md
```diff
@@ -0,0 +1,33 @@
...
+* Calling `service.New()` directly.
+
+The former is generally much easier, and used in all the examples, but `New` is exported
```
Lupie: If you want to get more straightforward:
"The former is generally easier. We use the builder pattern in all examples, but…"

service/README.md
```diff
@@ -0,0 +1,33 @@
...
+in case you'd like extra control over how your service is instantiated.
+
+If you **do** call `service.New()`, you will need to call `AddModules(...)` to configure
```
Lupie: Nitpicking:
- **do** --> choose to
- you will need --> you need

service/README.md
```diff
@@ -0,0 +1,33 @@
...
+which modules you'd like to serve.
+
+## Options
+
+Both the builder pattern and the `New()` functions take a variadic `Options`
```
Lupie: Should "functions" be "function," or does the builder pattern count as a function as well?

service/README.md
```diff
@@ -0,0 +1,33 @@
...
+pattern, allowing you to pick and choose which components you'd like to
+override. As with many of the goals of UberFx, specify zero options should give
```
Lupie: - "As with many of the goals of UberFx" --> "As a common theme in UberFx" OR "As you'll notice across UberFx" OR "As usual with UberFx"
- specify --> specifying

service/README.md
```diff
@@ -0,0 +1,33 @@
...
+you a fully working application.
+
+Once you have a service, you generally want to call `.Start()` on it.
```
Lupie: Any time you say something's generally a good idea, provide a reason, even if it seems obvious. If your audience needs to be told to call `.Start()`, they also probably need to be reminded of why.

Could you add a few words at the end of the sentence, like "to make sure that it runs" or whatever the reason is?

Also, could we just tell them to call `.Start()` instead of saying "generally want to"? If they don't want to, they can disobey.

service/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+`Start(bool)` comes in two variants: a blocking, and a non-blocking version. In
```
Lupie: - blocking --> blocking version
- Remove comma

service/README.md
```diff
@@ -0,0 +1,33 @@
...
+our sample apps, we choose to use the blocking version (`svc.Start(true)`) and
```
Lupie: Delete "choose to"

sectioneight:
Awesome feedback, thanks Lucie!

PR #64, opened 2016-11-11
Fixed doc with PopulateStruct with return err

anuptalwalkar:
nit fix


sectioneight:
need to run `make gendoc`?

PR #65, opened 2016-11-14
Fixed default logger to be json

anuptalwalkar:
GFM-146

### DIFF
--- a/core/ulog/log_test.go
+++ b/core/ulog/log_test.go
@@ -134,6 +134,24 @@ func TestConfiguredLoggerWithTextFormatter(t *testing.T) {
 	})
 }
 
+func TestConfiguredLoggerWithTextFormatter_NonDev(t *testing.T) {
+	dev := development
+	development = false
+	defer func() {
+		development = dev
+	}()
+	withLogger(t, func(tmpDir string, logFile string) {
+		log := Logger()
+		txt := true
+		log.Configure(Configuration{
+			Level:         "debug",
+			TextFormatter: &txt,
+		})
+		zaplogger := log.RawLogger()
+		assert.Equal(t, zap.DebugLevel, zaplogger.Level())
+	})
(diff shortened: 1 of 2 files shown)


core/ulog/log_test.go
```diff
@@ -134,6 +134,24 @@ func TestConfiguredLoggerWithTextFormatter(t *testing.T) {
 	})
 }
 
+func TestConfiguredLoggerWithTextFormatter_NonDev(t *testing.T) {
+	dev := development
+	development = false
```
sectioneight: `development` should probably be a `_development` global to indicate it's a global, right?

core/ulog/log_test.go
```diff
@@ -134,6 +134,24 @@ func TestConfiguredLoggerWithTextFormatter(t *testing.T) {
...
+	}()
+	withLogger(t, func(tmpDir string, logFile string) {
+		log := Logger()
+		txt := true
+		log.Configure(Configuration{
+			Level:         "debug",
+			TextFormatter: &txt,
+		})
+		zaplogger := log.RawLogger()
+		assert.Equal(t, zap.DebugLevel, zaplogger.Level())
```
sectioneight: this makes me sort of nervous. i'm thinking we need to have an in-built set of defaults (e.g. YAML byte slices) that contain the configuration defaults per-environment and then we can load those explicitly. the more checks i see in our tests and production code `if environment == ...` the more brittle our software becomes. perhaps we can talk in person about this?

anuptalwalkar: Yes, you are right. My goal was to simplify it a bit, but its clearly not happening with the current state.  I am refactoring this code instead, and moving to Builder pattern. Its more clean for user access and testing as well.

anuptalwalkar:
Closing this PR in favor of grater good.

PR #66, opened 2016-11-14
Bench improvement

glibsm:
I must have been drunk or something when I set this up originally, but these were done wrong. Now it actually appends all the bench results into the same file, rather than truncating it every time.

Benchcmp also doesn't run `make bench` anymore, it just uses the result of it. 

Also fixes improper setup for `IGNORE_BASELINE_CHECK` flag, it was not really handled correctly before.

Also removes a redundant cast

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,7 +93,7 @@ benchbase:
 benchcmp:
 	$(ECHO_V)which benchcmp >/dev/null || go get -u golang.org/x/tools/cmd/benchcmp
 	$(ECHO_V)test -s $(BASELINE_BENCH_FILE) || \
-		$(call die,Baseline benchmark file missing. Check out master and run \'make bench\')
+		$(call die,Baseline benchmark file missing. Check out master and run \'make benchbase\')
 	$(ECHO_V)test -s $(BENCH_FILE) || \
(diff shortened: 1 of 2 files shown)


glibsm:
@uber-go/service-framework

Makefile
```diff
@@ -69,13 +69,14 @@ BENCH_FILE ?= .bench/new.txt
...
+	$(ECHO_V)rm $(BENCH_FILE)
 	$(ECHO_V)$(foreach pkg,$(BENCH_PKGS),go test -bench=. -run="^$$" $(BENCH_FLAGS) $(pkg) | \
-		tee $(BENCH_FILE);)
+		tee -a $(BENCH_FILE);)
 
 BASELINE_BENCH_FILE=.bench/old.txt
 .PHONY: benchbase
 benchbase:
-	$(ECHO_V)if [ -z "$(IGNORE_BASELINE_CHECK)" ] && [ -z "$(git diff master)" ]; then \
+	$(ECHO_V)if [ -z $(IGNORE_BASELINE_CHECK) ] && [ -z "$(git diff master)" ]; then \
```
sectioneight: why remove the quotes? seems safer that way

glibsm:
@sectioneight actually fixed now.

PR #67, opened 2016-11-15
Refactored ulog into Builder pattern

anuptalwalkar:
GFM-146
GFM-172

I have updated ulog to follow builder pattern. With ulog.Builder, SetLogger, SetLevel and Configure can only be called on LogBuilder. This simplifies the Log interface, and makes Log object more robust. Also cleaned up how Configure() works for development and non-development setup. 

Removed initFields() for performance, which were mostly used on Logger initialization.


anuptalwalkar:
@yurishkuro @uber-go/service-framework

core/ulog/log.go
```diff
@@ -104,96 +71,29 @@ type Log interface {
...
 	return &baselogger{
-		log:        zap.New(zap.NewJSONEncoder()),
-		initFields: nil,
+		log: zap.New(zap.NewJSONEncoder()),
 	}
 }
 
 // Logger returns the package-level logger configured for the service
-func Logger(fields ...interface{}) Log {
+func Logger() Log {
```
anuptalwalkar: I am keeping this method for now, even though ulog can be created using NewBuilder.Build(). This method is used in most of the places in fx, and warrants a separate PR. I will add a TODO here.

core/ulog/builder.go
```diff
@@ -0,0 +1,169 @@
...
+type LogBuilder struct {
+	log       zap.Logger
+	logConfig Configuration
+	level     zap.Level
+}
+
+var _development = strings.Contains(config.GetEnvironment(), "development")
+
+// NewBuilder creates an empty builder for building ulog.Log object
+func NewBuilder(keyvals ...interface{}) *LogBuilder {
```
yurishkuro: why `keyvals` if they are not used?

core/ulog/builder.go
```diff
@@ -0,0 +1,169 @@
...
+	return &LogBuilder{
+		level: -1,
```
yurishkuro: what does -1 mean?

core/ulog/builder.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+	return &baselogger{
+		log: log,
+	}
+}
+
+func (lb *LogBuilder) devLogger() zap.Logger {
+	options := make([]zap.Option, 0, 3)
+	options = append(options, zap.DebugLevel)
+	return zap.New(zap.NewTextEncoder(), options...)
```
yurishkuro: why not just `return zap.New(zap.NewTextEncoder(), zap.DebugLevel)`?

core/ulog/builder.go
```diff
@@ -0,0 +1,169 @@
...
+	options := make([]zap.Option, 0, 3)
+	options = append(options, zap.InfoLevel, zap.Output(zap.AddSync(os.Stdout)))
+	return zap.New(zap.NewJSONEncoder(), options...)
+}
+
+// Configure Log object with the provided log.Configuration
+func (lb *LogBuilder) Configure() zap.Logger {
+	lb.log = lb.defaultLogger()
+
+	options := make([]zap.Option, 0, 3)
```
yurishkuro: can do simply `var options []zap.Option`

core/ulog/builder.go
```diff
@@ -0,0 +1,169 @@
...
+func (lb *LogBuilder) devLogger() zap.Logger {
+	options := make([]zap.Option, 0, 3)
+	options = append(options, zap.DebugLevel)
+	return zap.New(zap.NewTextEncoder(), options...)
+}
+
+func (lb *LogBuilder) defaultLogger() zap.Logger {
+	options := make([]zap.Option, 0, 3)
+	options = append(options, zap.InfoLevel, zap.Output(zap.AddSync(os.Stdout)))
+	return zap.New(zap.NewJSONEncoder(), options...)
```
yurishkuro: same as above - why obfuscate a simple call with variadic args?

core/ulog/builder.go
```diff
@@ -0,0 +1,169 @@
...
+	prefixWithFileLine *bool `yaml:"prefix_with_fileline"`
+	TextFormatter      *bool // use TextFormatter (default json)
+}
+
+// FileConfiguration describes the properties needed to log to a file
+type FileConfiguration struct {
+	Enabled   bool
+	Directory string
+	FileName  string
+}
```
yurishkuro: who is responsible for rotating the log?

core/ulog/builder.go
yurishkuro: I think this can be called simply `Builder()`. It is not crucial for the end user if the builder is "new" or reused, and makes for more concise code: `ulog.Builder().Build()`

anuptalwalkar: makes sense.

core/ulog/builder.go
anuptalwalkar: Fixed both

core/ulog/builder.go
anuptalwalkar: I wanted to keep it unset. Updated the Build method to set log level based on how it is defined.

core/ulog/log_benchmark_test.go
```diff
@@ -46,8 +45,7 @@ func BenchmarkUlogWithoutFields(b *testing.B) {
 }
 
 func BenchmarkUlogWithFieldsLogIFace(b *testing.B) {
-	log := Logger()
-	log.SetLogger(discardedLogger())
+	log := Builder().SetLogger(discardedLogger()).Build()
```
glibsm: make a `WithDiscardedLogger` helper?

core/ulog/builder.go
```diff
@@ -0,0 +1,170 @@
...
+	level     zap.Level
+	setlevel  *bool
+}
+
+var _development = strings.Contains(config.GetEnvironment(), "development")
+
+// Builder creates an empty builder for building ulog.Log object
+func Builder() *LogBuilder {
+	return &LogBuilder{
+		level: -2,
```
glibsm: what's a minus two

core/ulog/builder.go
```diff
@@ -0,0 +1,170 @@
...
+
+// LogBuilder is the struct containing logger
+type LogBuilder struct {
+	log       zap.Logger
+	logConfig Configuration
+	level     zap.Level
+	setlevel  *bool
+}
+
+var _development = strings.Contains(config.GetEnvironment(), "development")
```
glibsm: Just checking: are we not worried about the import order here? We're assuming that any and all static configuration is going to be done before this logger is imported, yes?

core/ulog/builder_test.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+		log := Builder().WithConfiguration(cfg).Build()
+		zaplogger := log.RawLogger()
+		assert.Equal(t, zap.DebugLevel, zaplogger.Level())
+	})
+}
+
+func withLogger(t *testing.T, f func(*LogBuilder, string, string)) {
+	dev := _development
+	_development = false
```
glibsm: false should be a param, no?

core/ulog/builder.go
glibsm: good question

sectioneight:
please don't ever check the environment key at the top level global. people
should be able to lazily set the env prefix

core/ulog/builder.go
anuptalwalkar: my baad. It's not used.

core/ulog/builder.go
anuptalwalkar: We don't handle log rotation in the app. It is handled by external tools, (or internal tools in our case). I can talk offline.

service/options_test.go
```diff
@@ -43,7 +43,7 @@ func TestAddModules_Errors(t *testing.T) {
 }
 
 func TestWithLogger_OK(t *testing.T) {
-	logger := ulog.Logger()
+	logger := ulog.Builder().Build()
```
glibsm: looks like we can provide a package level `Build()` function which will return an empty built builder with no parameters. Would reduce the typing in all the tests and things.

core/testutils/test_helper.go
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
glibsm: this is an oddly named file..

core/ulog/builder_test.go
```diff
@@ -0,0 +1,144 @@
...
+			},
+		}
+		log := Builder().WithConfiguration(cfg).Build()
+		zaplogger := log.RawLogger()
+		assert.Equal(t, zap.DebugLevel, zaplogger.Level())
+	})
+}
+
+func withLogger(t *testing.T, f func(*LogBuilder, string, string)) {
+	defer testutils.EnvOverride(t, config.GetEnvironmentPrefix()+_environment, "madeup_env")()
```
glibsm: `config.GetEnvironmentPrefix()+_environment, "madeup_env"` is ugly

lets make a `config.Env(env stirng)` function that auto-prefixes

core/config/config.go
```diff
@@ -91,11 +96,16 @@ func EnvProvider() ProviderFunc {
...
 	if env == "" {
-		env = "development"
+		env = _devEnv
 	}
 	return env
 }
 
+// IsDevelopment returns true if the current environment is set to development
+func IsDevelopment() bool {
+	return strings.Contains(GetEnvironment(), _devEnv)
```
glibsm: lets drop them gets

core/ulog/builder_test.go
anuptalwalkar: Env sounds very generic and not what API is returning. Perhaps EnvKey?

core/config/config.go
anuptalwalkar: dropped

service/options_test.go
anuptalwalkar: We could keep the ulog.Logger(), but its good to follow the Builder().Build() suite in the long run. Providing ulog.Build() function sounds like breaking the established builder pattern.

glibsm: Here is a revolutionary idea: how about `log.New()` which just does an empty builder?

anuptalwalkar: much like.

core/config/config.go
```diff
@@ -87,18 +92,23 @@ func EnvProvider() ProviderFunc {
 	}
 }
 
-// GetEnvironment returns current environment setup for the service
-func GetEnvironment() string {
-	env := os.Getenv(GetEnvironmentPrefix() + environment)
+// Environment returns current environment setup for the service
+func Environment() string {
+	env := os.Getenv(EnvironmentKey())
 	if env == "" {
```
glibsm: does it ever return nil?

PR #68, opened 2016-11-16
Updated log documentation with Builder pattern

anuptalwalkar:
GFM-172

PR blocked on changes reflected in #67, though I don't expect any changes in the builder pattern.

### DIFF
--- a/core/ulog/README.md
+++ b/core/ulog/README.md
@@ -1,8 +1,8 @@
 # Logging package
 
-ulog provides an API wrapper around the logging library (zap Logger).
-ulog uses builder pattern to instantiate the logger. Using `LogBuilder` user can set up
-configuration, inject logger and log level prior to log initialization.
+`package ulog` provides an API wrapper around the logging library (zap Logger).
+`ulog` uses builder pattern to instantiate the logger. Using `LogBuilder` user can set up
+configuration, inject logger and log level prior to the logger initialization.
 
 ```go
 package main
@@ -28,8 +28,8 @@ func main() {
 Note that the log methods (`Info`,`Warn`, `Debug`) takes parameter as key value
 pairs (message, (key, value)...)
 
-ulog configuration can be defined in multiple ways, either by writing the struct
-yourself, or describing in the YAML and populating using config package.
+`ulog.Configuration` can be setup in multiple ways, either by initializing the struct,
+or describing in the YAML and populating using `config` package.
 
 * Defining config structure:
 
@@ -57,7 +57,7 @@ func setupMyZapLogger(zaplogger zap.Logger) ulog.Log {
 }
 ```
 
-* Benchmarks
+* Current benchmarks
 
 Current performance benchmark data with `ulog interface`, `ulog baselogger struct`, and `zap.Logger`
 
--- a/core/ulog/README.md
+++ b/core/ulog/README.md
@@ -1,8 +1,8 @@
 # Logging package
 
-ulog provides an API wrapper around the logging library (zap Logger)
-The logger is instantiated as logger with default options and can be configured
-via `Configure()` API and provided YAML configuration.
+ulog provides an API wrapper around the logging library (zap Logger).
+ulog uses builder pattern to instantiate the logger. Using `LogBuilder` user can set up
+configuration, inject logger and log level prior to log initialization.
 
 ```go
 package main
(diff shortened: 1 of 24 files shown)


anuptalwalkar:
@lupie

core/ulog/README.md
```diff
@@ -1,8 +1,8 @@
 # Logging package
 
-ulog provides an API wrapper around the logging library (zap Logger)
-The logger is instantiated as logger with default options and can be configured
-via `Configure()` API and provided YAML configuration.
+`package ulog` provides an API wrapper around the logging library (zap Logger).
```
Lupie: - Remove "Logger" and maybe add a link to [zap](https://github.com/uber-go/zap) OR make it zap.Logger (I can't tell which one is more relevant from zap's docs)
- Add some high-level benefit for users up here (why should anyone read about an API wrapper around zap?)

core/ulog/README.md
```diff
@@ -1,8 +1,8 @@
...
+`ulog` uses builder pattern to instantiate the logger. Using `LogBuilder` user can set up
```
Lupie: - builder pattern --> the builder pattern

Looks like some people capitalize builder pattern. I don't think we need to, but let's put "the" in front of it whenever it comes up.

- Is ``LogBuilder`` familiar to readers, or do you need to explain what that is too?
- Proposed rewrite of second sentence: "With ``LogBuilder``, you can set up configuration, inject the logger, and…"
- I'm not sure if "log level prior to logger initialization" means you're logging the level OR injecting the log level before initialization

Basically, you're saying that you can set up the logger and apply it to whatever you want before logging begins, right? If so, I'd say something like this:

"``LogBuilder`` is a configuration tool for ``ulog``. With ``LogBuilder``, you can configure the logger, inject it wherever you need it, and set the log level before logging begins."

core/ulog/README.md
```diff
@@ -11,21 +11,25 @@ import "go.uber.org/fx/core/ulog"
...
+  // build ulog.Log from logBuilder
+  log := lobBuilder.Build()
 
   // Use logger in your service
   log.Info("Message describing loggging reason", "key", "value")
 }
 ```
 
-Note that the log methods (`Info`,`Warn`, `Debug`) takes parameter as key value pairs (message, (key, value)...)
+Note that the log methods (`Info`,`Warn`, `Debug`) takes parameter as key value
```
Lupie: - takes --> take
- parameter --> parameters

core/ulog/README.md
```diff
@@ -11,21 +11,25 @@ import "go.uber.org/fx/core/ulog"
...
+pairs (message, (key, value)...)
 
-ulog configuration can be defined in multiple ways, either by writing the struct yourself, or describing in the YAML
-and populating using config package.
+`ulog.Configuration` can be setup in multiple ways, either by initializing the struct,
```
Lupie: - setup --> set up
- multiple --> one of two

core/ulog/README.md
```diff
@@ -11,21 +11,25 @@ import "go.uber.org/fx/core/ulog"
...
+or describing in the YAML and populating using `config` package.
```
Lupie: Describing what in the YAML? The configuration details?

core/ulog/README.md
```diff
@@ -42,3 +46,25 @@ logging:
   stdout: true
   level: debug
 ```
+
+User can initialize their own `zap.Logger` implementation and inject into ulog.
```
Lupie: Are you guys alright with using "you" instead of "user," since these docs are for the user?

If so…
- User --> You
- their --> your
- ulog --> ``ulog``

core/ulog/README.md
```diff
@@ -42,3 +46,25 @@ logging:
...
+To configure and inject `zap.Logger`, setup the logger prior to building
```
Lupie: - setup --> set up

core/ulog/README.md
```diff
@@ -42,3 +46,25 @@ logging:
...
+the `ulog.Log` object
```
Lupie: Add period

core/ulog/doc.go
```diff
@@ -20,32 +20,37 @@
 
 // Package ulog is the Logging package.
```
Lupie: Where does this line come from?

- If we're going to monospace and lowercase ``package ulog``, let's do it in doc.go as well
- No reason to capitalize "logging"
- Instead: "``package ulog`` is UberFx's logging package" for better SEO and reading experience

core/ulog/README.md
anuptalwalkar: How about - 
"With `LogBuilder` user can do pre-initialization set up by injecting configuration, custom logger and log level prior to building the usable `ulog.Log` object."

core/ulog/README.md
anuptalwalkar: changed to "describing necessary `logging` configuration in the YAML "

core/ulog/doc.go
anuptalwalkar: This language is coming from gendoc, by compiling "Logging package" into Package ulog is the Logging package. We will need to change the tool to recompile this line into more friendlier language.

PR #71, opened 2016-11-18
Added filter chain for HTTP module and filters for tracer and error handling

madhuravi:
Added filter chain for HTTP module and filters for tracer and error handling
Setting tracer in the http request context

### DIFF
--- a/modules/rpc/yarpc.go
+++ b/modules/rpc/yarpc.go
@@ -121,7 +121,6 @@ func (m *YarpcModule) Start(readyCh chan<- struct{}) <-chan error {
 
 	interceptor := yarpc.Interceptors(m.interceptors...)
 
-	// TODO(ai/madhu) pass option for opentracing to NewDispatcher
 	m.rpc, err = _dispatcherFn(m.Host(), yarpc.Config{
 		Name: m.config.AdvertiseName,
 		Inbounds: []transport.Inbound{
--- a/modules/uhttp/filters.go
+++ b/modules/uhttp/filters.go
@@ -22,11 +22,13 @@ package uhttp
 
 import (
 	"fmt"
+	"log"
 	"net/http"
 
-	"github.com/opentracing/opentracing-go"
-
 	"golang.org/x/net/context"
+
+	"github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/ext"
 )
 
 // Filter applies filters on requests, request contexts or responses such as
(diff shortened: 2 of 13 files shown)


modules/rpc/yarpc.go
```diff
@@ -121,7 +121,6 @@ func (m *YarpcModule) Start(readyCh chan<- struct{}) <-chan error {
 
 	interceptor := yarpc.Interceptors(m.interceptors...)
 
-	// TODO(ai/madhu) pass option for opentracing to NewDispatcher
 	m.rpc, err = _dispatcherFn(m.Host(), yarpc.Config{
```
sectioneight: I meant to comment on the other diff. Since we're embedding `ModuleBase` directly (e.g. composing it), we should access the tracer directly instead of calling `.ModuleBase.Tracer()`

e.g..

```go
Tracer: m.Tracer(),
```

It's slightly cleaner and makes it easier to refactor

modules/uhttp/filters.go
```diff
@@ -0,0 +1,94 @@
...
+
+	"github.com/opentracing/opentracing-go"
+
+	"golang.org/x/net/context"
+)
+
+// Filter applies filters on requests, request contexts or responses such as
+// adding tracing to the context
+type Filter interface {
+	Apply(w http.ResponseWriter, r *http.Request, next http.Handler)
```
sectioneight: We should chat with @anuptalwalkar about his custom context. If we're going to be threading through our own context, we can't attach it to `http.Request` and will need to make it an explicit first argumen.

modules/uhttp/http.go
```diff
@@ -61,6 +61,7 @@ const (
 )
 
 var _ service.Module = &Module{}
+var tracingKey = struct{}{}
```
sectioneight: nit: group these `var`s:

```
var (
   _ service.Module = &Module{}

   tracingKey = struct{}{}
}
```

also, why is tracingKey the empty struct? shouldn't it have some value?

modules/uhttp/filters.go
```diff
@@ -0,0 +1,105 @@
...
+type tracerFilter struct {
+	tracer opentracing.Tracer
+}
+
+func (t tracerFilter) Apply(w http.ResponseWriter, r *http.Request, next http.Handler) {
+	operationName := r.Method
+	carrier := opentracing.HTTPHeadersCarrier(r.Header)
+	spanCtx, err := t.tracer.Extract(opentracing.HTTPHeaders, carrier)
+	if err != nil && err != opentracing.ErrSpanContextNotFound {
+		log.Printf("Malformed inbound tracing context: %s", err.Error())
```
sectioneight: This shouldn't be using stdlib log. Use `ulog.Logger()` or preferrably a module-scoped logger

modules/uhttp/filters.go
```diff
@@ -0,0 +1,105 @@
...
+	ctx := r.Context()
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	ctx = opentracing.ContextWithSpan(ctx, span)
+	r = r.WithContext(ctx)
+	next.ServeHTTP(w, r)
+}
+
+// errorFilter handles any panics and return an error
```
sectioneight: can we call this panicFilter? It's really not about errors

modules/uhttp/filters_test.go
```diff
@@ -0,0 +1,64 @@
...
+	assert.True(t, strings.Contains(response.Body.String(), "filters ok"))
+}
+
+func TestExecutionChainFilters(t *testing.T) {
+	chain := newExecutionChain(
+		[]Filter{tracerFilter{tracer: noopTracer}, FilterFunc(errorFilter)},
+		getNoopHandler(),
+	)
+	response := testServeHTTP(chain)
+	assert.True(t, strings.Contains(response.Body.String(), "filters ok"))
```
sectioneight: you can also do `assert.Contains()` to get better failure messages

modules/uhttp/filters_test.go
```diff
@@ -0,0 +1,64 @@
...
+	response := httptest.NewRecorder()
+	chain.ServeHTTP(response, request)
+	return response
+}
+
+func getNoopHandler() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		io.WriteString(w, "filters ok")
+	})
+}
```
sectioneight: Can you add a test that ensures the panic filter is still working? Or is that covered elsewhere?

madhuravi: this is already covered elsewhere. Since we are testing behavior and behavior did not change, I did not have to change that test. Check TestHTTPModule_Panic_OK in http_test

PR #77, opened 2016-11-19
Propogate core.Context through http module

anuptalwalkar:
GFM-126

### DIFF
--- a/examples/simple/handlers.go
+++ b/examples/simple/handlers.go
@@ -25,13 +25,14 @@ import (
 	"io"
 	"net/http"
 
+	"go.uber.org/fx/core"
 	"go.uber.org/fx/modules/uhttp"
 	"go.uber.org/fx/service"
 )
 
 type exampleHandler struct{}
 
-func (exampleHandler) ServeHTTP(ctx service.Context, w http.ResponseWriter, r *http.Request) {
+func (exampleHandler) ServeHTTP(ctx core.Context, w http.ResponseWriter, r *http.Request) {
 	io.WriteString(w, fmt.Sprintf("Headers: %+v", r.Header))
 }
 
--- a/examples/simple/handlers.go
+++ b/examples/simple/handlers.go
@@ -31,7 +31,7 @@ import (
 
 type exampleHandler struct{}
 
-func (exampleHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+func (exampleHandler) ServeHTTP(ctx service.Context, w http.ResponseWriter, r *http.Request) {
 	io.WriteString(w, fmt.Sprintf("Headers: %+v", r.Header))
 }
 
@@ -43,6 +43,6 @@ func enforceHeader(r uhttp.Route) uhttp.Route {
 func registerHTTPers(service service.Host) []uhttp.RouteHandler {
 	handler := &exampleHandler{}
 	return []uhttp.RouteHandler{
-		uhttp.NewRouteHandler("/", handler, enforceHeader),
+		uhttp.NewRouteHandler("/", handler),
 	}
 }
--- a/modules/uhttp/filters_test.go
+++ b/modules/uhttp/filters_test.go
@@ -28,6 +28,7 @@ import (
 	"strings"
 	"testing"
 
+	"go.uber.org/fx/core"
 	"go.uber.org/fx/service"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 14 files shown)


modules/uhttp/filters_test.go
```diff
@@ -21,18 +21,16 @@
 package uhttp
 
 import (
+	gcontext "context"
```
sectioneight: Why gcontext? This doesn't clash with anything so we shouldn't alias it.

modules/uhttp/http.go
```diff
@@ -160,19 +158,17 @@ func (m *Module) Initialize(host service.Host) error {
...
 		handle := newExecutionChain(m.filters, h.Handler)
-		m.router.Handle(h.Path, handle)
+		m.router.AddPatternRoute(h.Path, handle)
 	}
 
-	// Debug is opt-out
-	// TODO(madhu): Apply filters to the debug handler too?
 	if m.config.Debug == nil || *m.config.Debug {
-		m.router.PathPrefix("/debug/pprof").Handler(http.DefaultServeMux)
+		// TODO(anup): introduce debug handler
```
sectioneight: This should be part of this PR.

modules/uhttp/router.go
```diff
@@ -0,0 +1,107 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package uhttp
+
+import (
+	gcontext "context"
```
sectioneight: again, don't alias unless it clashes, which this doesn't

modules/uhttp/router.go
```diff
@@ -0,0 +1,107 @@
...
+	filter.Apply(ctx, w, r, HandlerFunc(h.serveHTTP))
+}
+
+// serveHTTP routes to the appropriate handler based on the incoming path
+func (h *Router) serveHTTP(ctx service.Context, w http.ResponseWriter, r *http.Request) {
+	var handler Handler
+
+	h.mux.RLock()
+	for idx := range h.routes {
+		if h.routes[idx].matches(r) {
```
sectioneight: Seems like we're re-inventing gorilla mux here. Can we not just wrap handlers in closures that take a service.Context?

modules/uhttp/router.go
```diff
@@ -0,0 +1,107 @@
...
+		filter = tracerFilter{}
+	}
+	filter.Apply(ctx, w, r, HandlerFunc(h.serveHTTP))
+}
+
+// serveHTTP routes to the appropriate handler based on the incoming path
+func (h *Router) serveHTTP(ctx service.Context, w http.ResponseWriter, r *http.Request) {
+	var handler Handler
+
+	h.mux.RLock()
```
sectioneight: This may be prohibitively expensive. We should consider freezing registration and then not having to care about thread-safety for adding routes while serving HTTP requests.

examples/simple/handlers.go
```diff
@@ -31,7 +31,7 @@ import (
 
 type exampleHandler struct{}
 
-func (exampleHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+func (exampleHandler) ServeHTTP(ctx service.Context, w http.ResponseWriter, r *http.Request) {
```
sectioneight: Open for discussion: is `service` the right package for this? Isn't it more of a "request context"?

modules/uhttp/router.go
```diff
@@ -0,0 +1,50 @@
...
+
+// NewRouter creates a new empty router
+func NewRouter(host service.Host) *Router {
+	return &Router{
+		host: host,
+	}
+}
+
+// HandleRoute wraps and calls the http.Handler underneath
+func (h *Router) HandleRoute(host service.Host, path string, handler Handler) {
```
sectioneight: Why not just call this `Handle` and then change the next line to `h.Router.Handle()`?

modules/uhttp/http.go
```diff
@@ -160,17 +158,15 @@ func (m *Module) Initialize(host service.Host) error {
...
 	// Do something unrelated to annotations
-	m.router = mux.NewRouter()
+	m.router = NewRouter(m.Host())
 
 	m.mux.Handle("/", m.router)
 
 	for _, h := range m.handlers {
 		handle := newExecutionChain(m.filters, h.Handler)
-		m.router.Handle(h.Path, handle)
+		m.router.HandleRoute(m.Host(), h.Path, handle)
```
sectioneight: Why do we have to pass the host as an argument when the router already has a reference to the host?

PR #80, opened 2016-11-21
Rename config.GetValue to config.Get

anuptalwalkar:
GFM-203

`cfg.GetValue().Value()`, stutters, and no `go` style method naming.
I wanted to get this in earlier as more code we write, we'll have to update in more places.


glibsm:
sweeeeeet

PR #81, opened 2016-11-21
Update ulog with zap.DynamicLevel implementation

anuptalwalkar:


core/ulog/builder.go
```diff
@@ -125,19 +111,21 @@ func (lb *LogBuilder) Configure() zap.Logger {
 	lb.log = lb.defaultLogger()
 
 	var options []zap.Option
+	dlevel := zap.DynamicLevel()
```
jcorbin: ideally this could be reduced to:
```
if lb.logConfig.Verbose {
  options = append(options, zap.DebugLevel)
}
// ...
else {
  options = append(options, lv)
}
```

Since it doesn't appear that you need the dynamism provided by `zap.DynamicLevel`.

Doing it this way will actually be µFaster since it'll eliminate an atomic load from every log site.

jcorbin:
Yup, this works but could be simpler.  The real use case for `zap.DynamicLevel` is when you actually want to change the logging level at runtime (e.g. with an RPC request).

PR #82, opened 2016-11-22
Fix format and context propagation in uhttp filters

madhuravi:
Set http request context, create default client with timeout, fix logging format

### DIFF
--- a/modules/uhttp/http_test.go
+++ b/modules/uhttp/http_test.go
@@ -41,6 +41,9 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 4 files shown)


madhuravi:
@uber-go/service-framework

modules/uhttp/http_test.go
```diff
@@ -41,6 +41,9 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// Custom default client since http's defaultClient does not set timeout
+var defaultClient = &http.Client{Timeout: 5 * time.Second}
```
anuptalwalkar: _defaultClient

PR #83, opened 2016-11-22
Log and emit metrics for HTTP panic recovers

sectioneight:
GFM-133


sectioneight:
@uber-go/service-framework

PR #86, opened 2016-11-22
Update glide and move testify to require

sectioneight:
"glide up" was complaining about testify being a testImport and a main
import. Since we're technically importing it in our application code
(_test.go files), it makes sense to have here.


sectioneight:
@uber-go/service-framework

sectioneight:
Glide was complaining that we had it in both testImports and a main import (due to a transitive dependency). Basically "glide up" no longer worked.

PR #87, opened 2016-11-22
Fix host state handling

glibsm:
Add better handling for service state


glibsm:
@uber-go/service-framework

PR #89, opened 2016-11-22
Pin tally to pre-refactor

glibsm:
Tally went through some rather large interface changes, so I'm pinning that dependency for now.
Going to work on adopting the new api, but I don't want anyone's service to break in the mean time.


glibsm:
@uber-go/service-framework

sectioneight:
next time maybe leave a TODO or comment in the YAML about why it's pinned to a sha

PR #90, opened 2016-11-23
Add uhttp client with tracing filter for outbound context propagation

madhuravi:
Add uhttp client with tracing filter for outbound context propagation
Add client filters and client execution chain that executed filters in order before http client method

### DIFF
--- a/core/context.go
+++ b/core/context.go
@@ -26,7 +26,7 @@ import (
 	"go.uber.org/fx/service"
 )
 
-// Context embeds Host and go context for use
+// Context embeds Host and Go stdlib context for use
 type Context interface {
 	gcontext.Context
 	service.Host
--- a/core/context.go
+++ b/core/context.go
@@ -26,15 +26,28 @@ import (
 	"go.uber.org/fx/service"
 )
 
-// Context embedds Host and go context for use
+// Context embeds Host and go context for use
 type Context interface {
 	gcontext.Context
 	service.Host
+	WithContext(ctx gcontext.Context) *context
 }
 
 type context struct {
 	gcontext.Context
 	service.Host
+}
+
+// WithContext returns a shallow copy of c with its context changed to ctx.
+// The provided ctx must be non-nil. Follows from net/http Request WithContext.
+func (c *context) WithContext(ctx gcontext.Context) (newC *context) {
+	if ctx == nil {
+		panic("nil context")
+	}
+	newC = new(context)
+	*newC = *c
+	newC.Context = ctx
+	return newC
 }
 
 var _ Context = &context{}
--- a/modules/uhttp/client.go
+++ b/modules/uhttp/client.go
@@ -20,8 +20,9 @@ type Client struct {
 // NewClient creates a new instance of uhttp Client
 func NewClient(client http.Client, filters ...ClientFilter) *Client {
 	if filters == nil {
-		filters = []ClientFilter{ClientFilterFunc(tracingClientFilter)}
+		filters = []ClientFilter{}
 	}
+	filters = append(filters, ClientFilterFunc(tracingClientFilter))
 	return &Client{Client: client, filters: filters}
 }
 
@@ -29,8 +30,9 @@ func NewClient(client http.Client, filters ...ClientFilter) *Client {
 func (c *Client) Do(ctx core.Context, req *http.Request) (resp *http.Response, err error) {
 	filters := c.filters
 	if filters == nil {
-		filters = []ClientFilter{ClientFilterFunc(tracingClientFilter)}
+		filters = []ClientFilter{}
 	}
+	filters = append(filters, ClientFilterFunc(tracingClientFilter))
 	execChain := newClientExecutionChain(filters, BasicClientFunc(c.do))
 	return execChain.Do(ctx, req)
 }
--- a/modules/uhttp/client.go
+++ b/modules/uhttp/client.go
@@ -1,0 +1,100 @@
+package uhttp
+
+import (
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"golang.org/x/net/context/ctxhttp"
+
+	"go.uber.org/fx/core"
+)
+
+// Client wraps around a http client
+type Client struct {
+	http.Client
+	filters []ClientFilter
+}
+
+// NewClient creates a new instance of uhttp Client
+func NewClient(client http.Client, filters ...ClientFilter) *Client {
+	if filters == nil {
+		filters = []ClientFilter{ClientFilterFunc(tracingClientFilter)}
+	}
+	return &Client{Client: client, filters: filters}
+}
+
+// Do is a context-aware, filter-enabled extension of Do() in http.Client
+func (c *Client) Do(ctx core.Context, req *http.Request) (resp *http.Response, err error) {
+	filters := c.filters
+	if filters == nil {
+		filters = []ClientFilter{ClientFilterFunc(tracingClientFilter)}
+	}
+	execChain := newClientExecutionChain(filters, BasicClientFunc(c.do))
+	return execChain.Do(ctx, req)
+}
+
+func (c *Client) do(ctx core.Context, req *http.Request) (resp *http.Response, err error) {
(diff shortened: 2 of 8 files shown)


madhuravi:
@uber-go/service-framework

modules/uhttp/client.go
```diff
@@ -0,0 +1,100 @@
...
+
+// Client wraps around a http client
+type Client struct {
+	http.Client
+	filters []ClientFilter
+}
+
+// NewClient creates a new instance of uhttp Client
+func NewClient(client http.Client, filters ...ClientFilter) *Client {
+	if filters == nil {
```
sectioneight: Don't check for nil, check for `len(filters) == 0`

modules/uhttp/client.go
```diff
@@ -0,0 +1,100 @@
...
+		filters = []ClientFilter{ClientFilterFunc(tracingClientFilter)}
+	}
+	return &Client{Client: client, filters: filters}
```
sectioneight: Would we ever not allow a tracing filter? Shouldn't we just unilaterally append it?

modules/uhttp/client_filters.go
```diff
@@ -0,0 +1,81 @@
+package uhttp
+
+import (
+	"net/http"
+
+	"github.com/opentracing/opentracing-go"
+
```
sectioneight: nit: no newline here. put line 9 up here and separate it with a space from opentracing

modules/uhttp/client.go
madhuravi: Yep that makes sense. Clients can pass custom filters but they wouldn't be passing the tracing filter. Any suggestions on how I can restrict struct initialization to just the new method? I basically have duplicate filter checks in NewClient and Do methods since they can initialize the struct method directly. I cannot have the exported NewClient method return an unexported client struct. I cannot make an interface either since the http.Client is a struct.

madhuravi:
Also make gendoc removes license headers and shows up all the files in the diff. Any tricks on how to get around that?

sectioneight:
go get -u sectioneight/github/md-to-godoc

madhuravi:
Ah ok. It wasnt the docs that was causing the failure, some other lint errors that I've fixed since and tests pass now.

core/context.go
```diff
@@ -26,17 +26,30 @@ import (
 	"go.uber.org/fx/service"
 )
 
-// Context embedds Host and go context for use
+// Context embeds Host and go context for use
```
sectioneight: nit: Go or stdlib context not `go contet`

core/context.go
```diff
@@ -26,17 +26,30 @@ import (
...
 type Context interface {
 	gcontext.Context
 	service.Host
+	WithContext(ctx gcontext.Context) *context
```
sectioneight: why are we import aliasing gocontext? it doesn't clash? we should just import it as `context` and use `context.Context` here unless I'm missing something

modules/uhttp/client.go
```diff
@@ -0,0 +1,122 @@
...
+
+import (
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"golang.org/x/net/context/ctxhttp"
+
+	"go.uber.org/fx/core"
```
sectioneight: nit: this should go above golang.org, which should be last

modules/uhttp/client_filters_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+	"go.uber.org/fx/core"
+
+	"github.com/stretchr/testify/assert"
+)
+
+var (
+	_respOK   = &http.Response{StatusCode: http.StatusOK}
+	_req      = httptest.NewRequest("", "http://localhost", nil)
+	errClient = errors.New("Client error")
```
sectioneight: errors should always be lowercaed. `"client error"` although ideally we're be more descriptive as to waht type of client error

modules/uhttp/client_test.go
```diff
@@ -0,0 +1,123 @@
...
+package uhttp
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"go.uber.org/fx/core"
+	"go.uber.org/fx/service"
+	"golang.org/x/net/context"
```
sectioneight: Move this down with stretchr -- go.uber.org/fx should be on its own

modules/uhttp/http_test.go
```diff
@@ -42,7 +42,7 @@ import (
 )
 
 // Custom default client since http's defaultClient does not set timeout
-var _defaultClient = &http.Client{Timeout: 5 * time.Second}
+var _defaultHTTPClient = &http.Client{Timeout: 5 * time.Second}
```
sectioneight: hmm, TODO for future, this 5 second timeout seems like it should be configurable. maybe 2 seconds to start out with?

modules/uhttp/http_test.go
```diff
@@ -202,13 +202,14 @@ func getURL(m *Module) string {
 
 func makeRequest(m *Module, method, url string, body io.Reader, fn func(r *http.Response)) {
 	base := getURL(m)
+	fmt.Printf("Method: %v URL: %v Body: %v\n", method, base+url, body)
```
sectioneight: Remove fmt.Printf

core/context.go
madhuravi: Not sure what you mean. We are embedding the golang context here.

core/context.go
madhuravi: I was following the existing file's convention but tried this out just now. It causes an interface type loop since we have a struct named context in the same file.

sectioneight: Ah, makes sense

core/context.go
sectioneight: my point is that "go" should always be "Go' and that we should maybe say "Go stdlib context"

modules/uhttp/client.go
sectioneight: This still neds fixing

sectioneight: Basically, never check a slice for nil. A nil slice and an empty slice are different things, but functionaly should be treeated the same.

core/context.go
```diff
@@ -26,17 +26,30 @@ import (
...
 }
 
 type context struct {
 	gcontext.Context
 	service.Host
 }
 
+// WithContext returns a shallow copy of c with its context changed to ctx.
+// The provided ctx must be non-nil. Follows from net/http Request WithContext.
+func (c *context) WithContext(ctx gcontext.Context) (newC *context) {
```
glibsm: that's a lot of context on one line..

modules/uhttp/client.go
```diff
@@ -0,0 +1,120 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package uhttp
```
glibsm: why is this not in package client under uhttp? would make things like `ClientFilters` just be `Filters`

glibsm:
Seems like there are a lot of `Client*` things in here, which to me suggests the need for another package

madhuravi:
You're right Glib. I have been thinking about another package as well. Will do that in a separate diff soon.

core/context.go
madhuravi: :) other ideas?

glibsm: well, the only thing i see is `WithContext` that can be changed to `With`, but other than that not much to be done :(

PR #93, opened 2016-11-23
Move http client into its own package

madhuravi:


madhuravi:
@uber-go/service-framework

madhuravi:
@glibsm

modules/uhttp/client/client_test.go
```diff
@@ -32,10 +33,11 @@ import (
...
 )
 
-var _defaultUHTTPClient = NewClient(*_defaultHTTPClient)
+var _defaultHTTPClient = &http.Client{Timeout: 2 * time.Second}
+var _defaultUHTTPClient = New(*_defaultHTTPClient)
 
-func TestNewClient(t *testing.T) {
-	uhttpClient := NewClient(*_defaultHTTPClient)
+func TestNew(t *testing.T) {
+	uhttpClient := New(*_defaultHTTPClient)
```
glibsm: this is odd. why do we not take `*http.Client` as param to avoid this?

modules/uhttp/client/client_test.go
```diff
@@ -32,10 +33,11 @@ import (
 	"golang.org/x/net/context"
 )
 
-var _defaultUHTTPClient = NewClient(*_defaultHTTPClient)
+var _defaultHTTPClient = &http.Client{Timeout: 2 * time.Second}
```
glibsm: depending on your answer to the other comment you can remove the & here if all you need is a value struct

modules/uhttp/client/client_test.go
madhuravi: Read a bit about this and pointed embed makes sense in this case based on http://stackoverflow.com/questions/28501976/embedding-in-go-with-pointer-or-with-value. Changed.

PR #96, opened 2016-11-24
Add noop logger

sectioneight:
This will make testing easier if you don't care about logs, especially for service owners.

GFM-232

### DIFF
--- a/core/ulog/noop_logger.go
+++ b/core/ulog/noop_logger.go
@@ -1,0 +1,32 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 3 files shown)


core/ulog/noop_logger.go
```diff
@@ -0,0 +1,32 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package ulog
+
+import "github.com/uber-go/zap"
+
+func noopLogger() Log {
+	return &baselogger{
+		log: zap.New(zap.NullEncoder()),
```
anuptalwalkar: minor- What does NullEncoder do?

You can do this as well - 
`return zap.New(
		zap.NewJSONEncoder(),
		zap.DiscardOutput,
)`

sectioneight: NullUncoder is faster -- it doesn't even create the JSON to discard, it literally runs no code

PR #100, opened 2016-11-30
Fix package shadowing in examples

sectioneight:
We we previously shadowing the `service` package which is bad practice.


sectioneight:
@uber-go/service-framework

PR #101, opened 2016-12-01
Simplify service Startup API

anuptalwalkar:
GFM-235

### DIFF
--- a/service/host.go
+++ b/service/host.go
@@ -234,14 +234,14 @@ func (s *host) start() Control {
 		s.shutdownMu.Unlock()
 		return Control{
 			ReadyChan:    readyCh,
-			ServiceError: errors.New("error shutting down the service"),
+			ServiceError: errors.New("shutting down the service"),
 		}
 	} else if s.IsRunning() {
 		s.shutdownMu.Unlock()
 		return Control{
 			ExitChan:     s.closeChan,
 			ReadyChan:    readyCh,
-			ServiceError: nil,
+			ServiceError: errors.New("service is already running"),
 		}
 	} else {
 		if s.observer != nil {
--- a/service/host.go
+++ b/service/host.go
@@ -201,11 +201,20 @@ func (s *host) AddModules(modules ...ModuleCreateFunc) error {
 	return nil
 }
 
-// Start runs the serve loop. If Shutdown() was called then the shutdown reason
-// will be returned.
-func (s *host) Start(waitForShutdown bool) (<-chan Exit, <-chan struct{}, error) {
-	// TODO(glib): this function has a lot of different exits and different
-	// mutex unlock spots, maybe a rewrite is in order?
+// StartAsync service is used as a non-blocking the call on service start. StartAsync will
+// return the call to the caller with a Control to listen on channels
+// and trigger manual shutdowns.
+func (s *host) StartAsync() Control {
+	return s.start(false)
+}
+
+// Start service is used for blocking the call on service start. Start will block the
+// call and yield the control to the service lifecyce manager.
+func (s *host) Start() Control {
+	return s.start(true)
+}
+
+func (s *host) start(waitForShutdown bool) Control {
 	var err error
 	s.locked = true
 	s.shutdownMu.Lock()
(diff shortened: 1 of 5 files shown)


service/service.go
```diff
@@ -55,10 +55,24 @@ type Owner interface {
 
 	AddModules(modules ...ModuleCreateFunc) error
 
-	Start(waitForExit bool) (exit <-chan Exit, ready <-chan struct{}, err error)
+	// Start service is used for blocking the call on service start. Start will block the
+	// call and yield the control to the service lifecyce manager.
+	Start() Control
```
glibsm: if the start is blocking, why do we return control, if it's going to block?
unless i'm misunderstanding something

anuptalwalkar: No, you are right. We are not doing anything with the Control on blocking call. I'll update.

glibsm:
I'll take another look after rebase (sorry for conflict!)

service/host.go
```diff
@@ -201,11 +201,20 @@ func (s *host) AddModules(modules ...ModuleCreateFunc) error {
...
+// StartAsync service is used as a non-blocking the call on service start. StartAsync will
+// return the call to the caller with a Control to listen on channels
+// and trigger manual shutdowns.
+func (s *host) StartAsync() Control {
+	return s.start(false)
+}
+
+// Start service is used for blocking the call on service start. Start will block the
+// call and yield the control to the service lifecyce manager.
+func (s *host) Start() {
```
glibsm: I think it's important to note that *no code* will be executed after `.Start()`. The way things stand right now, `os.Exit` is called when service is terminated via signal, or some other measure. So literally everything quits and nothing in main after Start will ever run

service/host.go
```diff
@@ -201,11 +201,20 @@ func (s *host) AddModules(modules ...ModuleCreateFunc) error {
...
+	s.start(true)
+}
+
+func (s *host) start(waitForShutdown bool) Control {
```
glibsm: since we are splitting control into async and blocking, i think all the code that relates to blocking should be removed from this method and places into `Start()`. This will make the method a little less cluttered

anuptalwalkar: Good point.

service/host_test.go
```diff
@@ -35,16 +35,16 @@ import (
 func TestOnCriticalError_NoObserver(t *testing.T) {
 	err := errors.New("Blargh")
 	sh := makeHost()
-	closeCh, ready, err := sh.Start(false)
-	require.NoError(t, err, "Expected no error starting up")
+	Control := sh.StartAsync()
```
sectioneight: nit: local variables should never be capitalized, especially since you're shadowing a type here

anuptalwalkar: oh good catch.

service/host.go
```diff
@@ -217,15 +230,25 @@ func (s *host) Start(waitForShutdown bool) (<-chan Exit, <-chan struct{}, error)
 	}()
```
glibsm: this is very specific to StartAsync() and has no impact on `Start` -> should also be moved out of this function

service/host.go
```diff
@@ -217,15 +230,25 @@ func (s *host) Start(waitForShutdown bool) (<-chan Exit, <-chan struct{}, error)
...
 	if s.inShutdown {
 		s.shutdownMu.Unlock()
-		return nil, readyCh, fmt.Errorf("errShuttingDown")
+		return Control{
+			ReadyChan:    readyCh,
+			ServiceError: fmt.Errorf("errShuttingDown"),
```
glibsm: errors.New

service/host.go
```diff
@@ -217,15 +230,25 @@ func (s *host) Start(waitForShutdown bool) (<-chan Exit, <-chan struct{}, error)
...
+		}
 	} else if s.IsRunning() {
 		s.shutdownMu.Unlock()
-		return s.closeChan, readyCh, nil
+		return Control{
```
glibsm: return error `service already running`, maybe?

service/host.go
```diff
@@ -217,15 +230,25 @@ func (s *host) Start(waitForShutdown bool) (<-chan Exit, <-chan struct{}, error)
...
+			ServiceError: nil,
+		}
 	} else {
 		if s.observer != nil {
 			if err := s.observer.OnInit(s); err != nil {
 				s.shutdownMu.Unlock()
-				return nil, readyCh, err
+				return Control{
+					ReadyChan:    readyCh,
+					ServiceError: err,
```
glibsm: errors.Wrap(err, "failed to initialize the observer")

service/host.go
```diff
@@ -247,19 +270,23 @@ func (s *host) Start(waitForShutdown bool) (<-chan Exit, <-chan struct{}, error)
 				if _, err := s.shutdown(e, "", nil); err != nil {
 					s.Logger().Error("Unable to shut down modules", "initialError", e, "shutdownError", err)
 				}
-				return errChan, readyCh, e
+				return Control{
+					ExitChan:     errChan,
+					ReadyChan:    readyCh,
+					ServiceError: e,
```
glibsm: `e` is used to range over errors, but we exit with the first cycle. We should log all the errors, then return the concatenation of all the errors with a wrap saying `failed to start the modules because: a b and c`

service/host.go
glibsm: actually i'm not even sure what the right course of action is here...

service/host.go
anuptalwalkar: Current behavior seems to be the correct behavior to me. Returning the first error you get is probably essential for debugging rather than knowing all errors a, b, c...

service/host.go
```diff
@@ -201,11 +203,24 @@ func (s *host) AddModules(modules ...ModuleCreateFunc) error {
...
+// StartAsync service is used as a non-blocking the call on service start. StartAsync will
+// return the call to the caller with a Control to listen on channels
+// and trigger manual shutdowns.
+func (s *host) StartAsync() Control {
+	return s.start()
+}
+
+// Start service is used for blocking the call on service start. Start will block the
+// call and yield the control to the service lifecyce manager. Start will not yield back
+// the control to the caller, so no code will be executed after calling Star()
```
glibsm: Start()

glibsm:
Lets make sure tests are passing before merging, though

PR #102, opened 2016-12-02
Upgrade rpc to 1.0.0-rc1

glibsm:
Took forever to dig into why tests were failing after the upgrade, but finally figured out that we weren't creating a test service properly.

Rest was mostly renaming things

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -9,9 +9,9 @@ import:
 - package: github.com/gorilla/context
   version: ^1.1.0
 - package: go.uber.org/yarpc
-  version: 7c6a4c5618cb38228ef9f0eb85ea71d8fc5af2c4
+  version: v1.0.0-rc1
 - package: go.uber.org/thriftrw
-  version: 6aa07c0009457cc6c7ff5497695f1581ad184350
+  version: ^1
 - package: github.com/go-validator/validator
   version: v2
 - package: github.com/pkg/errors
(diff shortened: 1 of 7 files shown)


glibsm:
@uber-go/service-framework

glibsm:
FYI @breerly

glide.yaml
```diff
@@ -9,9 +9,9 @@ import:
 - package: github.com/gorilla/context
   version: ^1.1.0
 - package: go.uber.org/yarpc
-  version: 7c6a4c5618cb38228ef9f0eb85ea71d8fc5af2c4
+  version: v1.0.0-rc1
```
breerly: nice

glide.yaml
```diff
@@ -9,9 +9,9 @@ import:
...
 - package: go.uber.org/thriftrw
-  version: 6aa07c0009457cc6c7ff5497695f1581ad184350
+  version: ^1
```
breerly: double nice - now that thriftrw is 1.x, you won't have to worry about this guy for awhile :)

modules/rpc/yarpc.go
```diff
@@ -39,12 +39,12 @@ import (
...
-	config       yarpcConfig
-	log          ulog.Log
-	stateMu      sync.RWMutex
-	interceptors []transport.Interceptor
+	rpc                yarpc.Dispatcher
+	register           registerServiceFunc
+	config             yarpcConfig
+	log                ulog.Log
+	stateMu            sync.RWMutex
+	inboundMiddlewares []transport.UnaryInboundMiddleware
```
breerly: We've renamed interceptors and filters to inboundMiddleware and outboundMiddleware. It's a bit more verbose, but people we're getting confused with the previous terminology.

PR #104, opened 2016-12-02
Add runtime metrics reporting and set it to start by default

madhuravi:
Add runtime metrics reporting and set it to start by default

### DIFF
--- a/metrics/runtime_metrics.go
+++ b/metrics/runtime_metrics.go
@@ -29,6 +29,9 @@ import (
 	"github.com/uber-go/tally"
 )
 
+// _numGCThreshold comes from the PauseNs buffer size https://golang.org/pkg/runtime/#MemStats
+var _numGCThreshold = uint32(256)
+
 // StartCollectingRuntimeMetrics starts generating runtime metrics under given metrics scope with
 // given frequency.
 // Recommended usage:
@@ -71,7 +74,7 @@ type RuntimeCollector struct {
 
 // RuntimeConfig contains configuration for initializing runtime metrics
 type RuntimeConfig struct {
-	Disabled bool `yaml:"enabled"`
+	Disabled bool `yaml:"disabled"`
 }
 
 // NewRuntimeCollector creates a new RuntimeCollector.
@@ -150,9 +153,9 @@ func (r *RuntimeCollector) generate() {
 	lastNum := atomic.SwapUint32(&r.metrics.lastNumGC, num) // reset for the next iteration
 	if delta := num - lastNum; delta > 0 {
 		r.metrics.numGC.Inc(int64(delta))
-		if delta > 255 {
-			// too many GCs happened, the timestamps buffer got wrapped around. Generate only the last 256
-			lastNum = num - 256
+		// Match the MemStats buffer and generate only the last _numGCThreshold
+		if delta >= _numGCThreshold {
+			lastNum = num - _numGCThreshold
 		}
 		for i := lastNum; i != num; i++ {
 			pause := memStats.PauseNs[i%256]
--- a/metrics/runtime_metrics.go
+++ b/metrics/runtime_metrics.go
@@ -22,6 +22,7 @@ package metrics
 
 import (
 	"runtime"
+	"sync"
 	"sync/atomic"
 	"time"
 
@@ -63,7 +64,8 @@ type RuntimeCollector struct {
 	scope           tally.Scope
 	collectInterval time.Duration
 	metrics         runtimeMetrics
-	started         bool
+	startedMutex    sync.RWMutex
+	started         bool // protected by startedMutex
 	quit            chan struct{}
 }
 
@@ -100,14 +102,19 @@ func NewRuntimeCollector(
 
 // IsRunning returns true if the collector has been started and false if not.
 func (r *RuntimeCollector) IsRunning() bool {
+	r.startedMutex.RLock()
+	defer r.startedMutex.RUnlock()
 	return r.started
 }
 
 // Start starts the collector thread that periodically emits metrics.
 func (r *RuntimeCollector) Start() {
+	r.startedMutex.RLock()
 	if r.started {
+		r.startedMutex.RUnlock()
 		return
 	}
+	r.startedMutex.RUnlock()
 	go func() {
 		ticker := time.NewTicker(r.collectInterval)
 		for {
@@ -120,7 +127,9 @@ func (r *RuntimeCollector) Start() {
 			}
 		}
 	}()
+	r.startedMutex.Lock()
 	r.started = true
+	r.startedMutex.Unlock()
 }
 
 // generate sends runtime metrics to the local metrics collector.
--- a/metrics/runtime_metrics.go
+++ b/metrics/runtime_metrics.go
@@ -28,21 +28,18 @@ import (
 	"github.com/uber-go/tally"
 )
 
-// StartReportingRuntimeMetrics starts reporting runtime metrics under given metrics scope with
+// StartCollectingRuntimeMetrics starts generating runtime metrics under given metrics scope with
 // given frequency.
-// This is just a convenience wrapper around creating and starting a new reporter manually.
 // Recommended usage:
-//     StartReportingRuntimeMetrics(rootScope.Scope("runtime"), time.Second)
-// If you want to stop the reporter later:
-//     rmr := StartReportingRuntimeMetrics(rootScope.Scope("runtime"), time.Second)
+//     rmr := StartCollectingRuntimeMetrics(rootScope.Scope("runtime"), time.Second)
 //     ...
 //     rmr.Close()
-func StartReportingRuntimeMetrics(
-	scope tally.Scope, reportInterval time.Duration,
-) *RuntimeReporter {
-	runtimeReporter := NewRuntimeReporter(scope, reportInterval)
-	runtimeReporter.Start()
-	return runtimeReporter
+func StartCollectingRuntimeMetrics(
+	scope tally.Scope, collectInterval time.Duration,
+) *RuntimeCollector {
+	runtimeCollector := NewRuntimeCollector(scope, collectInterval)
+	runtimeCollector.Start()
+	return runtimeCollector
 }
 
 type runtimeMetrics struct {
@@ -58,24 +55,24 @@ type runtimeMetrics struct {
 	lastNumGC       uint32
 }
 
-// RuntimeReporter is a struct containing the state of the RuntimeMetricsReporter.
-type RuntimeReporter struct {
-	scope          tally.Scope
-	reportInterval time.Duration
-	metrics        runtimeMetrics
-	started        bool
-	quit           chan struct{}
+// RuntimeCollector is a struct containing the state of the runtimeMetrics.
+type RuntimeCollector struct {
+	scope           tally.Scope
+	collectInterval time.Duration
+	metrics         runtimeMetrics
+	started         bool
+	quit            chan struct{}
 }
 
-// NewRuntimeReporter creates a new RuntimeMetricsReporter.
-func NewRuntimeReporter(
-	scope tally.Scope, reportInterval time.Duration,
-) *RuntimeReporter {
+// NewRuntimeCollector creates a new RuntimeCollector.
+func NewRuntimeCollector(
+	scope tally.Scope, collectInterval time.Duration,
+) *RuntimeCollector {
 	var memstats runtime.MemStats
 	runtime.ReadMemStats(&memstats)
-	return &RuntimeReporter{
-		scope:          scope,
-		reportInterval: reportInterval,
+	return &RuntimeCollector{
+		scope:           scope,
+		collectInterval: collectInterval,
 		metrics: runtimeMetrics{
 			numGoRoutines:   scope.Gauge("num-goroutines"),
 			goMaxProcs:      scope.Gauge("gomaxprocs"),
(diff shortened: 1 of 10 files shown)


metrics/runtime_metrics.go
```diff
@@ -0,0 +1,154 @@
...
+func StartReportingRuntimeMetrics(
+	scope tally.Scope, reportInterval time.Duration,
+) *RuntimeReporter {
+	runtimeReporter := NewRuntimeReporter(scope, reportInterval)
+	runtimeReporter.Start()
+	return runtimeReporter
+}
+
+type runtimeMetrics struct {
+	NumGoRoutines   tally.Gauge
```
sectioneight: Why are most of these exported except for the `lastNumGC`? It seems like, since this struct isn't exported, these should all be unexported fields as well.

metrics/runtime_metrics.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+// IsRunning returns true if the reporter has been started and false if not.
+func (r *RuntimeReporter) IsRunning() bool {
+	return r.started
+}
+
+// Start starts the reporter thread that periodically emits metrics.
+func (r *RuntimeReporter) Start() {
+	if r.started {
```
sectioneight: Intentionally not thread-safe? Should we leave a note about this?

metrics/runtime_metrics_test.go
```diff
@@ -0,0 +1,62 @@
...
+func TestStartRuntimeReporter(t *testing.T) {
+	reporter := StartReportingRuntimeMetrics(tally.NoopScope, time.Millisecond)
+	assert.True(t, reporter.IsRunning())
+
+	runtime.GC()
+	time.Sleep(5 * time.Millisecond)
+	reporter.report()
+	time.Sleep(5 * time.Millisecond)
+	reporter.report()
+	reporter.Close()
```
sectioneight: What is this actually asserting? And what's with the magic `5` milliseconds?

modules/uhttp/client/client.go
```diff
@@ -43,6 +43,9 @@ func New(client *http.Client, filters ...Filter) *Client {
 	return &Client{Client: client, filters: filters}
 }
 
+// DefaultClient is the default uhttp client with defaults for http client and filters.
+var DefaultClient = New(http.DefaultClient)
```
sectioneight: This seems unrelated. Did you accidentally add this to this PR?

modules/uhttp/filters.go
```diff
@@ -55,7 +55,7 @@ func tracingServerFilter(ctx fx.Context, w http.ResponseWriter, r *http.Request,
 	ext.HTTPUrl.Set(span, r.URL.String())
 	defer span.Finish()
 
-	gctx := fx.NewContext(opentracing.ContextWithSpan(ctx, span), ctx)
+	gctx := ctx.WithContext(opentracing.ContextWithSpan(ctx, span))
```
sectioneight: again, unrelated?

service/core.go
```diff
@@ -64,19 +66,20 @@ type SetContainerer interface {
...
+	scopeMux        sync.Mutex
+	scope           tally.RootScope
+	runtimeReporter *metrics.RuntimeReporter
+	observer        Observer
+	resources       map[string]interface{}
+	logConfig       ulog.Configuration
+	log             ulog.Log
+	tracerConfig    jaegerconfig.Configuration
+	tracer          opentracing.Tracer
+	tracerCloser    io.Closer
```
sectioneight: if we're going to be messing up history here can we alphabetize these? this is getting pretty big and hard to read

service/service.go
```diff
@@ -139,6 +141,10 @@ func New(options ...Option) (Owner, error) {
 		metrics.Freeze()
 	}
 
+	if svc.RuntimeMetrics() == nil {
```
sectioneight: We should provide a way to opt-out. Somebody should be able to explicitly say "I don't want to report runtime metrics" (as an option).

madhuravi:
@yurishkuro since you wrote the original code.

madhuravi:
@uber-go/service-framework

metrics/runtime_metrics.go
madhuravi: You're right, changed all to unexported.

service/core.go
madhuravi: Alphabetized and grouped some fields into embedded structs.

modules/uhttp/filters.go
madhuravi: removed.

modules/uhttp/client/client.go
madhuravi: removed.

metrics/runtime_metrics_test.go
madhuravi: This was copied over but nothing magic about 5 afaik. Runtime metrics are reported every millsecond, even 1 millisecond should do it.

metrics/runtime_metrics.go
```diff
@@ -0,0 +1,154 @@
...
+import (
+	"runtime"
+	"sync/atomic"
+	"time"
+
+	"github.com/uber-go/tally"
+)
+
+// StartReportingRuntimeMetrics starts reporting runtime metrics under given metrics scope with
+// given frequency.
```
glibsm: this wrapping is all out of whack. Let stick to 80 char doc strings, if we can

service/core.go
```diff
@@ -63,20 +65,49 @@ type SetContainerer interface {
...
+	tracer       opentracing.Tracer
+	tracerCloser io.Closer
+	tracerConfig jaegerconfig.Configuration
+}
+
+func (tc *tracerCore) Tracer() opentracing.Tracer {
+	return tc.tracer
+}
+
+type loggingCore struct {
```
glibsm: what's all this for?

metrics/runtime_metrics.go
```diff
@@ -0,0 +1,154 @@
...
+	memoryHeapIdle  tally.Gauge
+	memoryHeapInuse tally.Gauge
+	memoryStack     tally.Gauge
+	numGC           tally.Counter
+	gcPauseMs       tally.Timer
+	lastNumGC       uint32
+}
+
+// RuntimeReporter is a struct containing the state of the RuntimeMetricsReporter.
+type RuntimeReporter struct {
```
glibsm: this is going to get into hand-waving territory, but calling it a Reporter is very confusing, since it takes in a tally.Scope, which has a Reporter to send metrics over.

I'd prefer we call it something along the lines of "runtime metrics collector", since that's what it actually does.

service/service.go
madhuravi: Good point. Why do we want this only for runtime metrics? We don't seem to have it for standard metrics. It is running in a subroutine, should not cause a latency impact.

metrics/runtime_metrics.go
madhuravi: I thought we decided on 100 char?

service/core.go
madhuravi: Just grouping to make it more readable/manageable. Feedback is welcome.

metrics/runtime_metrics.go
madhuravi: It does actually report though. It delegates to the tally reporter but that's an implementation detail.

glibsm: I don't follow. Where does the delegation happen?

madhuravi: never mind, terminology mixup. Will change it.

sectioneight:
Somebody could pass in a noop metrics scope which would be non nil. But
there's no way for somebody to disable this behavior as is.

madhuravi:
@sectioneight if they pass in a noop metrics scope, runtime will not get reported either since it uses the same root scope. It does still collect metrics though. I can create a NoopRuntimeCollector that noops on all methods but I feel like we can punt on that until we have real use cases where they would want to disable this altogether. What do you think?

sectioneight:
It's still doing work that could be undesired from a service. We can make
an Option that disables it, no need to create a stub implementation. Just
check if the Option was passed or if a config field is set to explicitly
disable it.

metrics/runtime_metrics.go
madhuravi: Changed my mind, makes sense to add one here. Done.

metrics/runtime_metrics.go
```diff
@@ -0,0 +1,168 @@
...
+	"sync/atomic"
+	"time"
+
+	"github.com/uber-go/tally"
+)
+
+// StartCollectingRuntimeMetrics starts generating runtime metrics under given metrics scope with
+// given frequency.
+// Recommended usage:
+//     rmr := StartCollectingRuntimeMetrics(rootScope.Scope("runtime"), time.Second)
```
anuptalwalkar: This would be a nice addition in readme, doc.go

metrics/runtime_metrics.go
```diff
@@ -0,0 +1,168 @@
...
+	r.metrics.memoryStack.Update(int64(memStats.StackInuse))
+
+	// memStats.NumGC is a perpetually incrementing counter (unless it wraps at 2^32)
+	num := memStats.NumGC
+	lastNum := atomic.SwapUint32(&r.metrics.lastNumGC, num) // reset for the next iteration
+	if delta := num - lastNum; delta > 0 {
+		r.metrics.numGC.Inc(int64(delta))
+		if delta > 255 {
+			// too many GCs happened, the timestamps buffer got wrapped around. Generate only the last 256
+			lastNum = num - 256
```
anuptalwalkar: why 256? make it a package variable?

service/core.go
anuptalwalkar: We can go a little further and move logging initialization code here or under /service/core package along with other *Core stuff, and use function call in the service New. service.New is about 100+ lines of code right now.

service/service.go
```diff
@@ -139,6 +141,17 @@ func New(options ...Option) (Owner, error) {
 		metrics.Freeze()
 	}
 
+	if svc.RuntimeMetricsCollector() == nil {
+		var runtimeMetricsConfig metrics.RuntimeConfig
+		err := svc.configProvider.Get("runtimeMetrics").PopulateStruct(&runtimeMetricsConfig)
```
glibsm: we already have `metrics` top-level struct for config. Perhaps we should nest `runtime` under that?

service/core.go
glibsm: I'm not sure we should do this refactor as part of this diff, it would be much easier to iterate on and discuss as part of a separate diff

metrics/runtime_metrics_test.go
```diff
@@ -0,0 +1,77 @@
...
+func TestRuntimeCollector(t *testing.T) {
+	collector := NewRuntimeCollector(tally.NoopScope, time.Millisecond)
+	defer closeCollector(t, collector)
+
+	assert.False(t, collector.IsRunning())
+	collector.Start()
+	assert.True(t, collector.IsRunning())
+	runtime.GC()
+	time.Sleep(5 * time.Millisecond)
+	collector.generate()
```
glibsm: after collector.generate, can we validate anything?

metrics/runtime_metrics.go
madhuravi: Good callout. Looks like metrics doesn't have docs yet. Will add this as part of a separate diff.

service/service.go
madhuravi: Who manages top level metrics? I don't want to accidentally override a preconfigured key for eg.

metrics/runtime_metrics_test.go
glibsm: No need for mocks here either.

Just right above in this test, we are using `tally.Noopscope`. Instead of that we should have a `memReporter` that just creates a hash of reported metrics (I think we should already have that somewhere, and if not that's pretty easy to build and we can rebase onto that diff).

Then you don't have to mock anything, you can assert on that reporter directly to make sure it has the correct metrics.

service/core.go
madhuravi: This was originally from Aiden's comment about the struct getting bloated. Since it was a result of adding a new field to this struct, I think it's ok to have it as part of this diff. 
Good idea Anup, don't want to add all that refactor as part of this though. Will make it part of a new diff.

service/service.go
sectioneight: We manage it, we can make it whatever we want. I agree with Glib, this should be nested `metrics.runtime`

madhuravi: Done

metrics/runtime_metrics_test.go
madhuravi: Based on offline discussion with @sectioneight @glibsm @anuptalwalkar the conclusion is we only test our library's functionality. The memStatsReporter is for testing that tally has recorded these metrics (outside uberfx functionality), so decided to not do this.

metrics/runtime_metrics.go
```diff
@@ -0,0 +1,171 @@
...
+			lastNum = num - _numGCThreshold
+		}
+		for i := lastNum; i != num; i++ {
+			pause := memStats.PauseNs[i%256]
+			r.metrics.gcPauseMs.Record(time.Duration(pause))
+		}
+	}
+}
+
+// Close stops collecting runtime metrics. It cannot be started again after it's
```
glibsm: @anuptalwalkar maybe we should consider this strategy for service start/stop... might make handling states and cleanup a lot easier. If a service is stopped - that's it. Re-create it

metrics/runtime_metrics_test.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func TestStartRuntimeCollector(t *testing.T) {
+	collector := StartCollectingRuntimeMetrics(tally.NoopScope, time.Millisecond, _emptyRuntimeConfig)
+	defer closeCollector(t, collector)
+
+	assert.True(t, collector.IsRunning())
+	runtime.GC()
+	time.Sleep(time.Millisecond)
+	collector.generate()
```
glibsm: Are these generate calls actually necessary? it looks like the collector works on a ticker, which means every time we sleep for `+-time.Millisecond` it will run on it's own (runtime_metrics.go:126).

I think the output of this test is that the collection has been done four times. If only there was a way to find out how many metrics ended up in tally ;)

service/service_test.go
```diff
@@ -121,6 +121,22 @@ func TestServiceWithRoles(t *testing.T) {
...
+metrics:
+  runtime:
+    disabled: true
+`)
+	cfgOpt := WithConfiguration(config.NewYAMLProviderFromBytes(data))
+
+	svc, err := New(cfgOpt)
+	require.NoError(t, err)
+
+	assert.Nil(t, svc.RuntimeMetricsCollector())
```
glibsm: nice!

metrics/runtime_metrics_test.go
madhuravi: Good point. Removed.

PR #105, opened 2016-12-03
Fail make test on data race condition

anuptalwalkar:
Current make test with runs $(COV_REPORT), but exits with status 0 on data race. Added separate race check, and test timeout.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ ALL_SRC := $(shell find . -name "*.go" | grep -v -e vendor \
 test:
 	$(ECHO_V)$(MAKE) -C examples/keyvalue/ kv/types.go ECHO_V=$(ECHO_V)
 	$(ECHO_V)go test $(RACE) -timeout=1s $(PKGS)
-	$(MAKE) $(COV_REPORT)
+	$(ECHO_V) $(MAKE) $(COV_REPORT)
 
 TEST_IGNORES = vendor .git
 COVER_IGNORES = $(TEST_IGNORES) examples testutils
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -20,6 +20,7 @@ ALL_SRC := $(shell find . -name "*.go" | grep -v -e vendor \
 
 .PHONY: test
 test: $(COV_REPORT)
```
glibsm: remove $(COV_REPORT) from here

Makefile
```diff
@@ -20,6 +20,7 @@ ALL_SRC := $(shell find . -name "*.go" | grep -v -e vendor \
...
+	$(ECHO_V)go test -race -timeout=1s $(PKGS)
```
glibsm: add $(MAKE) $(COV_REPORT) on the next line, it will produce much better output in travis (won't run coveralls when the tests fail_

glibsm: Also, we don't want to run with `-race` locally by default. It uses a much higher amount of CPU and memory, so we only want to do that on CI, which is why we have the $(RACE) flag

Makefile
```diff
@@ -48,7 +51,7 @@ $(COV_REPORT): $(PKG_FILES) $(ALL_SRC)
 		-ignore "$(OVERALLS_IGNORE)" \
 		-covermode=atomic \
 		$(DEBUG_FLAG) -- \
-		$(TEST_FLAGS) $(RACE) $(TEST_VERBOSITY_FLAG) | \
+		$(TEST_FLAGS) $(RACE) -timeout=1s $(TEST_VERBOSITY_FLAG) | \
```
glibsm: lets move timeout into an actual flag `$(TEST_TIMEOUT)` since it's used in more than one place.

PR #106, opened 2016-12-03
Fix race on host.state

glibsm:
we were racing for `host.state`, added a lock to make sure transition finishes before another begins.

Removed a troublesome test that was trying to test service.Stop. After @anuptalwalkar lands the small refactor, I'll get back to it. As of right now `.Start(true)` function never passes control as part of stop `os.Exit` is called, so it's hard to test. Need a bit of a re-think


glibsm:
@uber-go/service-framework

PR #108, opened 2016-12-06
Add link to examples

sectioneight:
I get a lot of questions about what UberFx looks like. I think this
belongs near the top since most people want to see code, not words.


sectioneight:
Preview here: https://github.com/uber-go/fx/blob/53fa65b462c4c3c94b2c3891fdb1a06ae8edd295/README.md#examples

PR #112, opened 2016-12-07
Upgrade tally and updates to match StatsReporter, Gauge, Closer changes

madhuravi:
NOTE: I have a TODO for the stats reporter. Will add that in a later PR

### DIFF
--- a/service/host.go
+++ b/service/host.go
@@ -155,7 +155,7 @@ func (s *host) shutdown(err error, reason string, exitCode *int) (bool, error) {
 	}
 
 	// Stop the metrics reporting
-	if s.scope != nil {
+	if s.scope != nil && s.metricsCloser != nil {
 		if err = s.metricsCloser.Close(); err != nil {
 			s.Logger().Error("Failure to close metrics", "error", err)
 		}
--- a/service/host.go
+++ b/service/host.go
@@ -156,7 +156,7 @@ func (s *host) shutdown(err error, reason string, exitCode *int) (bool, error) {
 
 	// Stop the metrics reporting
 	if s.scope != nil {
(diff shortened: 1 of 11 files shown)


service/host.go
```diff
@@ -156,7 +156,9 @@ func (s *host) shutdown(err error, reason string, exitCode *int) (bool, error) {
 
 	// Stop the metrics reporting
 	if s.scope != nil {
-		s.scope.Close()
+		if err = s.metricsCloser.Close(); err != nil {
```
sectioneight: small possibility for an upstream bug to affect us -- `s.metricsCloser` could be `nil` even if `s.scope` is non-nil. Perhaps we should be defensive and change line 158 to also check if `s.metricsCloser` is non-nil?

tracing/tracer.go
```diff
@@ -41,8 +41,9 @@ func InitGlobalTracer(
 	var reporter *jaegerReporter
 	if cfg == nil || !cfg.Disabled {
 		cfg = loadAppConfig(cfg, logger)
+		// TODO: Change to use the right stats reporter
 		reporter = &jaegerReporter{
-			reporter: scope.Reporter(),
+			reporter: tally.NullStatsReporter,
```
sectioneight: oh, is this not possible anymore? are you going to address that in this PR or a future one?

service/host.go
madhuravi: good catch.

tracing/tracer.go
madhuravi: see description, to be addressed in future PR.

madhuravi: sorry I tend to interchange the two. Fixed description.

PR #113, opened 2016-12-07
Add more unit tests for host shutdown

madhuravi:


service/host_test.go
```diff
@@ -186,6 +189,84 @@ func TestShutdownWithError_ReturnsError(t *testing.T) {
...
+		closeChan: make(chan Exit, 1), // Indicates service is running
+		serviceCore: serviceCore{
+			metricsCore: metricsCore{
+				scope:            tally.NoopScope,
+				metricsCloser:    testutils.NoopCloser{},
+				runtimeCollector: metrics.NewRuntimeCollector(tally.NoopScope, time.Millisecond),
+			},
+		},
+	}
+	checkShutdown(t, sh, false)
```
madhuravi: The shutdown function itself doesnt seem to care about any of these states. These are testing that when the shutdown function is called, that the metrics, runtime, tracer closers are closed correctly and if there is an error, they return the error. This was not covered by tests before.

madhuravi:
@sectioneight Good call out. I think DRY applies to tests too, other people will be reading/modifying them as well. Made a few changes to that effect.

PR #115, opened 2016-12-07
Yaml parsing with nested struct and custom types

anuptalwalkar:
GFM-244

### DIFF
--- a/config/value.go
+++ b/config/value.go
@@ -418,9 +418,9 @@ func (cv Value) valueStruct(key string, target interface{}) (interface{}, error)
 							fieldValue.Set(reflect.New(fieldType.Elem()))
 						}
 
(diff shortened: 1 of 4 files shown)


config/value.go
```diff
@@ -417,7 +417,23 @@ func (cv Value) valueStruct(key string, target interface{}) (interface{}, error)
 						if fieldValue.IsNil() {
 							fieldValue.Set(reflect.New(fieldType.Elem()))
 						}
-						fieldValue.Elem().Set(reflect.ValueOf(val))
+
+						// We cannot assign reflect.ValueOf(val) to fieldValue as is, when field is a custom type
```
sectioneight: instead of "custom type" I'd maybe say "user-defined named type" to align with the official documentation on assignability

PR #117, opened 2016-12-08
Fix make

peter-edge:
`npm install` prints some output to the lint log, which then makes `make lint` fail. `.DEFAULT_GOAL` is also set to `all`.


sectioneight:
Thanks Peter!

PR #119, opened 2016-12-08
Change metrics func to get a stats reporter and propagate to the tracer

madhuravi:
Tally API upgrade removed access to the stats reporter from the scope. Tracer needs access to the stats reporter. This is to propagate and store stats reporter in service.Host so it can be used by the tracer.

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"io"
 	"sync"
+	"time"
 
 	"go.uber.org/fx/config"
 
@@ -31,9 +32,9 @@ import (
 )
 
 var (
-	_scopeFunc ScopeFunc
-	_mu        sync.Mutex
-	_frozen    bool
+	_repFunc StatsReporterFunc
+	_mu      sync.Mutex
+	_frozen  bool
 
 	// DefaultReporter does not do anything
 	// TODO(glib): add a logging reporter and use it by default, rather than noop
@@ -47,8 +48,8 @@ type ScopeInit interface {
 	Config() config.Provider
 }
 
-// ScopeFunc is used during service init time to register the reporter
-type ScopeFunc func(i ScopeInit) (tally.Scope, io.Closer, error)
+// StatsReporterFunc is used during service init time to register the reporter
+type StatsReporterFunc func(i ScopeInit) (tally.StatsReporter, error)
 
 // Freeze ensures that after service is started, no other metrics manipulations can be done
 //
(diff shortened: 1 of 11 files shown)


madhuravi:
@uber-go/service-framework

metrics/metrics.go
```diff
@@ -72,33 +73,33 @@ func ensureNotFrozen() {
...
 
-	if _scopeFunc != nil {
-		rep, closer, err := _scopeFunc(i)
+	if _repFunc != nil {
+		rep, err := _repFunc(i)
 		if err != nil {
 			panic(fmt.Sprintf("Failed to initialize metrics reporter %v", err))
 		}
-		return rep, closer
+		scope, closer := tally.NewRootScope("", nil, rep, time.Second)
```
glibsm: This is not what we want, as I was trying to explain earlier. We want to give control over creating this scope to the service owners, if they choose to do so.

The parameters to this function is something that has to be configurable: prefix, tags, and the tick duration.

This is how this file initially looked, by taking in a reporter, but it's a lot less flexible than allowing service owners to create their own metrics structure that might work better for them. We can't make those sort of decisions (especially not the tags they want to include on all metrics events) for them.

glibsm:
See #7 for more details

metrics/metrics.go
madhuravi: Service owners can still configure using WithMetrics options. From further conversation with Glib, he suggested we might want to add special configuration for uber internal services. For this reason, changed to allowing scope and reporter to be passed in and use the pre-configured metrics scope.

metrics/metrics.go
```diff
@@ -84,21 +83,20 @@ func RegisterRootScope(rep ScopeFunc) {
...
+		scope, reporter, closer, err := _scopeFunc(i)
 		if err != nil {
 			panic(fmt.Sprintf("Failed to initialize metrics reporter %v", err))
 		}
-		return rep, closer
+		return scope, reporter, closer
 	}
-
-	return nil, nil
+	return nil, nil, nil
```
anuptalwalkar: This is a strange func call to deal with. We should wrap the return values in a struct, and return it with possible error as second parameter.

madhuravi: What makes it strange? In other words, what's the advantage of wrapping this in a struct? We treat the error as a panic here, why would we want to return the error?

anuptalwalkar: Well, I am talking about dealing with individual return parameters is strange for the caller. Right now you have to check for each return variable in service.go and assign null scope, add check for each nil/not nil, and so on. encapsulating what rootScope returns would make the code concise and readable. 

For error I am not talking about panic, but instead of returning nil, nil, nil, it would make more sense for the caller to simply check for err message (something along the lines that no scope func is registered) 
The method would look like this - 
`
func RootScope(i ScopeInit) (scopeRegistry, error) {
  if _scopeFunc != nil {
    return scopeRegistry{}, nil
  }
  return nil, errors.New("no scope functions registered")
}
`

anuptalwalkar: I am not good with naming, but you can call scopeRegistry encapsulation rootScope or just scope..

service/host_test.go
```diff
@@ -195,7 +195,7 @@ func TestHostShutdown_RunningService(t *testing.T) {
 func TestHostShutdown_CloseSuccessful(t *testing.T) {
 	sh := makeRunningHost()
 	sh.serviceCore.metricsCore = metricsCore{
-		scope:            tally.NoopScope,
+		metrics:          tally.NoopScope,
```
glibsm: nit: not sure i like renaming metrics to scope. Yes, we expose this publicly through `Metrics()` API, but internally we don't need to mix and match things, i.e. `call it like you see it`. It's a tally scope.

service/host_mock.go
```diff
@@ -41,7 +41,7 @@ func NullHost() Host {
 			log: ulog.NoopLogger,
 		},
 		metricsCore: metricsCore{
-			scope: tally.NoopScope,
+			metrics: tally.NoopScope,
```
glibsm: add a null stats reporter here as well, please otherwise some tests might crash due to runtime metrics trying to use it

metrics/metrics.go
madhuravi: I don't think a scope registry is necessary here. This exchange happens literally in one place and the caller (service.go) assigns it to a struct. Creating a new struct seems redundant. Go has strongly typed multiple return values, so the caller knows exactly what to expect. 
With errors, if this is not a panic, I completely agree. There is no use for the returned error if we are handling it within.

service/host_mock.go
madhuravi: Won't crash because of the noop tracer but added it anyway.

service/host_test.go
madhuravi: I see what you mean. The reason I finally went for this is if you look at service, we'll have
if svc.Metrics() == nil {
  svc.scope = tally.Scope
}

Checking Metrics() and setting scope appears unrelated to someone reading it. That's why I decided to keep them uniform.

PR #122, opened 2016-12-09
Remove metrics nil checks and return the noop values from metrics package

madhuravi:


metrics/metrics.go
```diff
@@ -98,5 +98,5 @@ func RootScope(i ScopeInit) (tally.Scope, tally.StatsReporter, io.Closer) {
 		}
 		return scope, reporter, closer
 	}
-	return nil, nil, nil
+	return tally.NoopScope, tally.NullStatsReporter, nil
```
sectioneight: Can we return a noop closer as well so people don't have to nil-check this?

madhuravi: Good point. However there is no default noop closer and tally does not provide a noop closer. I have one in testutils that I created for unit tests.
It seems contrived to create a Noop closer for this purpose in uberfx. We return nil in a lot of places and don't create noop object everytime.

sectioneight:
I meant literally return https://golang.org/pkg/io/ioutil/#NopCloser passing
in `nil`. That way you can call Close() without a potential nil-deref

madhuravi:
@sectioneight The signature for NopCloser is
func NopCloser(r io.Reader) io.ReadCloser

It takes in a reader. I think it would be confusing to construct a reader here and pass it in just to be able to use this.

sectioneight:
A ReadCloser is a superset of a Closer. In other words, anything that takes an `io.Closer`, you can pass `ioutil.NopCloser(nil)`:

https://play.golang.org/p/GqeXIzTZdE

madhuravi:
I don't mind use that. As a reader of this code I still think it's confusing, looking at metrics
func RootScope(i ScopeInit) (tally.Scope, tally.StatsReporter, io.Closer) {
  ...
  return tally.NoopScope, tally.NullStatsReporter, ioutil.NopCloser(nil)
}

What is ioutil doing here, pop over the method and you'll see "NopCloser returns a ReadCloser with a no-op Close method wrapping the provided Reader r." And this really has nothing to do with metrics. Seems nicer to have readability over an extra nil check.

sectioneight:
As producers of APIs we should be providing ones that are as friendly to use as possible. Particularly when returning interfaces in a non-error case, it is extremely awkward to have to check if the interface return is `nil` or not, particularly due to the semantics of nil interfaces in Go. You don't have to use `ioutil.NopCloser` if you don't like how it reads, but having something that you can call `Close()` on (just like you have two other interfaces you can call methods on that do nothing) is much friendlier to developers and reads much more naturally than an extra `nil`.

metrics/metrics_test.go
```diff
@@ -36,9 +37,9 @@ func TestRegisterReporter_OK(t *testing.T) {
 	defer cleanup()
 
 	scope, reporter, closer := getScope()
-	assert.Nil(t, scope)
-	assert.Nil(t, reporter)
-	assert.Nil(t, closer)
+	assert.Equal(t, scope, tally.NoopScope)
+	assert.Equal(t, reporter, tally.NullStatsReporter)
+	assert.Equal(t, ioutil.NopCloser(nil), closer)
```
sectioneight: This is an implementation detail. I'd do this instead:

```
assert.NoError(t, closer.Close())
```

PR #123, opened 2016-12-12
Only install dependencies once for examples

### LINKED ISSUE #118: make lint test does glide install twice
`make lint test` calls into the makefile in examples/keyvalue, which does it's own dependency management. It would be nice if this was somehow merged, so that glide isn't called twice. I'd honestly just put the example makefile goals into the main makefile, including any extra dependencies. This makes `make lint test` much slower.

sectioneight:
Fixes #118

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,8 +22,7 @@ ALL_SRC := $(shell find . -name "*.go" | grep -v -e vendor \
 TEST_TIMEOUT := "-timeout=3s"
 
 .PHONY: test
-test:
-	$(ECHO_V)$(MAKE) -C examples/keyvalue/ kv/types.go ECHO_V=$(ECHO_V)
+test: examples
 	$(ECHO_V)go test $(RACE) $(TEST_TIMEOUT) $(PKGS)
 	$(ECHO_V)$(MAKE) $(COV_REPORT)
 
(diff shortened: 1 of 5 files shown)


sectioneight:
@uber-go/service-framework

examples/keyvalue/Makefile
```diff
@@ -1,13 +1,4 @@
-.PHONY: deps
-deps:
-	@echo "Installing UberFx dependencies..."
-	$(ECHO_V)$(MAKE) -C ../.. libdeps
-	@echo "Installing thriftrw..."
-	$(ECHO_V)go install ../../vendor/go.uber.org/thriftrw
```
sectioneight: Yea, this was a mistake. Fixing now.

Makefile
```diff
@@ -121,8 +117,18 @@ gendoc:
 		xargs -I% md-to-godoc -input=%
 
 .PHONY: clean
-clean::
+clean:
```
madhuravi: Why change this to single colon?

sectioneight: All the other targets are single. Double is useful if it's going to be extended by some other file, but these make rules are all self-contained, so semantically single-colon is correct. I think this was added by accident (or copy-pasted from some internal makefiles, which may have had some extensibility concerns we don't have here).

madhuravi: Have we thought about setting up convenient wrappers for makefile targets? We have an internal repo that teams use today and include in their Makefiles.

sectioneight:
I think people will continue to use the existing build infrastructure. Are
there pieces here you'd like to see extracted internally?

PR #124, opened 2016-12-12
Added fx.Context for rpc handlers with codegen

anuptalwalkar:
GFM-227

1. wrappers from thrift.Unary/Oneway handlers to inject fx.Context
2. code generation for fx.Context on top of yarpc codegen, with a TODO  to merging with yarpc plugin.


breerly:
Oh? Merged already? Great.

sectioneight:
We have had extensive conversations about this. Meeting some POC code with
hopes of upstreaming but end-user use case is interest number 1 and code
layout follows from that.

breerly:
We have a shared interest in this - I don't want UberFX forcefully going around the APIs and generated code we've laid out without giving us a single voice in the matter.

breerly:
This is a dangerous precedent to be setting. Working around each-other will fragment the experience for developers, instead, we should work on aligning our solutions.

sectioneight:
We've had plenty of docs and emails about it with y'all. This code isn't
set in stone but is here to present the type of experience we'd like to
provide to devs.

breerly:
Sure, your team has had conversations and emails. We have yet to be involved, and this PR does exactly the thing we said not to do: copy and paste the plugin that generates the code. 

We don't want this code wrapped or modified - if UberFX thinks this is needed, we can talk about what you're trying to accomplish. Going around us like this is no way to work together.

sectioneight:
We definitely have involved your team and have an open google doc about
this exact issue where we asked for comment. I'll link you when i get into
the office.

sectioneight:
And again, this is just a POC so we have something to look at and talk
about for today's meeting. The goal is definitely to work with your team to
merge the two.

I'd caution you to err on the side of assuming good intent rather than
malintent when missing context, particularly on public forums.

breerly:
I watch this repo so that I can vet our integration - catching this before you guys go too far down this path is in both of our interests. We share the same customers, so I absolutely *do* assume good intent! I want UberFX to kick butt, which is why I'm getting involved now.

Excited to jam later.

sectioneight:
Glad to hear. The conversation you missed in person is that having side
band code generation is not a good idea nor is it something we want to be
in the business of maintaining; this PR was simply so we could get some
code to compile so we could play around with what it feels like to inject
custom context.

Looking forward to working with your team to align on how to do this best--
this is pre-alpha software here and we're focused on user experiences first
and technology second. Since your team is passionate about the technologies
Im hopeful we can get your support in providing these experiences in a
clean, idiomatic, not-copy-pasted manner.

breerly:
👍

PR #125, opened 2016-12-13
Refactored service initialization for readability

anuptalwalkar:
service.New is out of control. Refactoring the methods out as setup* to make it more readable and know exactly what it is doing. (ideally I would want no if/else except for error checks, but its a start)

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -21,14 +21,8 @@
 package service
 
 import (
-	"time"
-
 	"go.uber.org/fx/config"
-	"go.uber.org/fx/metrics"
-	"go.uber.org/fx/tracing"
-	"go.uber.org/fx/ulog"
 
-	"github.com/go-validator/validator"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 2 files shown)


service/service.go
```diff
@@ -120,73 +114,27 @@ func New(options ...Option) (Owner, error) {
...
-			svc.statsReporter,
-		)
-		if err != nil {
-			return svc, errors.Wrap(err, "unable to initialize global tracer")
-		}
-		svc.tracer = tracer
-		svc.tracerCloser = closer
+	if err := svc.setupRuntimeMetricsCollector(svc.Config()); err != nil {
+		return nil, err
+	} else if err := svc.setupTracer(svc.Config()); err != nil {
```
sectioneight: nit: drop the `else if` and just put an `if` on the next line

service/service_setup.go
```diff
@@ -0,0 +1,110 @@
...
+
+func (svc *serviceCore) setupMetrics() {
+	if svc.Metrics() == nil {
+		svc.metrics, svc.statsReporter, svc.metricsCloser = metrics.RootScope(svc)
+		metrics.Freeze()
+	}
+}
+
+func (svc *serviceCore) setupRuntimeMetricsCollector(cfg config.Provider) error {
+	if svc.RuntimeMetricsCollector() == nil {
```
sectioneight: nit: prefer early returns:

```go
if svc.RuntimeMetricsColector() != nil {
  return nil
}

// logic to set up collector
```

this is more idiomatic and reduced indentation and increases readability.

service/service_setup.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+		svc.runtimeCollector = metrics.StartCollectingRuntimeMetrics(
+			svc.metrics.SubScope("runtime"), time.Second, runtimeMetricsConfig,
+		)
+	}
+	return nil
+}
+
+func (svc *serviceCore) setupTracer(cfg config.Provider) error {
+	if svc.Tracer() == nil {
```
sectioneight: early returns

PR #127, opened 2016-12-15
Use yarpc inbound middleware for context injection

anuptalwalkar:
GFM-254

### DIFF
--- a/modules/rpc/handler.go
+++ b/modules/rpc/handler.go
@@ -58,7 +58,7 @@ func (f OnewayHandlerFunc) HandleOneway(ctx fx.Context, reqMeta yarpc.ReqMeta, b
 }
 
 // WrapUnary wraps the unary handler and returns implementation of thrift.UnaryHandler for yarpc calls
-// TODO(anup): fix wrapUnary signature to remove host once update yarpc plugin is imported in the repo
+// TODO(anup): Remove host parameter once updated yarpc plugin is imported in the repo
 func WrapUnary(host service.Host, unaryHandlerFunc UnaryHandlerFunc) thrift.UnaryHandler {
 	return &unaryHandlerWrapper{
 		Host:             host,
@@ -78,7 +78,7 @@ func (hw *unaryHandlerWrapper) Handle(ctx context.Context, reqMeta yarpc.ReqMeta
 }
 
 // WrapOneway wraps the oneway handler and returns implementation of thrift.OnewayHandler for yarpc calls
-// TODO(anup): fix wrapOneway signature to remove host once update yarpc plugin is imported in the repo
+// TODO(anup): Remove host parameter once updated yarpc plugin is imported in the repo
 func WrapOneway(host service.Host, onewayHandlerFunc OnewayHandlerFunc) thrift.OnewayHandler {
 	return &onewayHandlerWrapper{
 		Host:              host,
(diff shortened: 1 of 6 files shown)


breerly:
Might make sense to specify "inbound middleware" here - we support both inbound and outbound middleware.

anuptalwalkar:
Good point. Updated the title. I noticed 'oneway inbound middleware' was missing in the repo, so added that to PR

anuptalwalkar:
@uber-go/service-framework

modules/rpc/handler.go
```diff
@@ -57,6 +58,7 @@ func (f OnewayHandlerFunc) HandleOneway(ctx fx.Context, reqMeta yarpc.ReqMeta, b
 }
 
 // WrapUnary wraps the unary handler and returns implementation of thrift.UnaryHandler for yarpc calls
+// TODO(anup): Remove host parameter once updated yarpc plugin is imported in the repo
```
sectioneight: Is there a task ID associated with this? In general we should strive to associate TODOs with tasks, although I know that can be burdensome

modules/rpc/handler_test.go
```diff
@@ -88,3 +89,33 @@ func TestWrapOneway_error(t *testing.T) {
...
+	unary := fxContextUnaryInboundMiddleware{
+		Host: service.NullHost(),
+	}
+	err := unary.Handle(context.Background(), &transport.Request{}, nil, &fakeUnaryHandler{})
+	assert.Equal(t, "dummy", err.Error())
+}
+
+type fakeUnaryHandler struct{}
+
+func (_m *fakeUnaryHandler) Handle(ctx context.Context, _param1 *transport.Request, _param2 transport.ResponseWriter) error {
```
sectioneight: What's with the `_m` here? If you're not using the receiver you can elide it:

```go
func (*fakeUnaryHandler) Handle(...) {
  // as before
}
```

modules/rpc/handler.go
anuptalwalkar: Yes, I'll add task ID here. (GFM-255)

anuptalwalkar:
The only APIs that are available to users are `WithUnaryInboundMiddleware` and `WithOnewayInboundMiddleware` I think calling them `WithInboundMiddleware` for regular interceptors and `WithOnewayInboundMiddleware` for one way interceptors sounds better.

breerly:
We've borrowed the term Unary from gRPC, which represents the Request/Response message pattern. Oneway came from Thrift.

We needed a way to differentiate between these RPC Types. I'd rather not see new jargon layered ontop of ours if possible - getting these terms instilled into developers minds is already going to be difficult.

Up until this point - most types of messages *were* request/response messages, so it's understandable that these terms are new and alien. I'd like to lean on educating instead of hiding.

If you have ideas, we'd love to hear them - we still have a couple weeks before we lock these terms down.

anuptalwalkar:
@breely In practice and from users perspective, I have seen that its a more of a common pattern that bi-directional protocols are defined with default names and prefix/suffix is explicitly added for other types. 
For example,  a `Task` is just a task that you would expect to return on complete, and an `AsyncTask` is what you would expect to return on receive. Its generally not a good readable and understandable form to have explicit naming on all APIs.

sectioneight:
Right, I'm not suggesting inventing new jargon. I'm wondering if there are
existing terms we could use that people already understand instead of
introducing yet another thing.

anuptalwalkar:
I am merging the PR. Feel free to open an issue if we have better ideas on naming.

sectioneight:
Agreed, we can iterate on naming but that shouldn't block making forward
progress. Nice work on this Anup!

breerly:
Same here, +1

PR #128, opened 2016-12-15
Slim down Context and propogate through uhttp

### LINKED ISSUE #126: RFC: fx.Context as a struct
Following up our discussion yesterday: I would like to propose the following:

-   `fx.Context` is a struct that embeds `context.Context` rather than an
    interface.
-   UberFx values stored on `fx.Context` actually just use
    `context.WithValue` rather than wrapping a `context.Context` in a new
    type.  Value accessors will be defined for all UberFx values.
-   Corresponding methods will be declared on `fx.Context` which just call the
    accessor on itself.
    
Specifically,

```go
package fx

type Context struct{ context.Context }

func (ctx Context) Logger() Logger {
    return GetLogger(ctx)
}

func GetLogger(ctx context.Context) Logger {
    // ..
}
```

This has a few advantages:

-   Discoverability by having methods on `fx.Context`.
-   Full compatibility with `context.Context` and `WithValue`. You won't lose
    UberFx information if you do,
    
    ```go
    fxctx := fx.NewContext(ctx, host)
    ctx := fxctx.WithValue(...)
    ```
-   Downgrading to `context.Context` and upgrading back to `fx.Context` just
    works.
    
    ```go
    fxctx := fx.NewContext(ctx, host)
    var ctx2 context.Context = fxctx
    fxctx2 := fx.Context{ctx2}
    fmt.Println(fxctx.Logger() == fxctx2.Logger())  // true
    ```

The only thing to keep in mind is that accessors have to have reasonable
behavior for the case where a `context.Context` does not have UberFx
information on it.

Thoughts?

CC @breerly @kriskowal @willhug @sectioneight

anuptalwalkar:
GFM-248, GFM-256
Fixes #126

### DIFF
--- a/context.go
+++ b/context.go
@@ -31,5 +31,4 @@ type Context interface {
 	gcontext.Context
 
 	Logger() ulog.Log
-	WithContext(ctx gcontext.Context) Context
 }
--- a/context.go
+++ b/context.go
@@ -23,66 +23,13 @@ package fx
 import (
 	gcontext "context"
 
-	"go.uber.org/fx/service"
 	"go.uber.org/fx/ulog"
 )
-
-var _logger = "logger"
 
 // Context embeds Host and Go stdlib context for use
 type Context interface {
 	gcontext.Context
 
-	// unexported func ensures only fx package implements Context
-	fx()
 	Logger() ulog.Log
-	WithContext(ctx gcontext.Context) *context
-}
-
-type context struct {
-	gcontext.Context
-}
-
-// Convert context.Context to fx.Context
-func Convert(ctx gcontext.Context) Context {
-	fx, ok := ctx.(Context)
-	if ok {
-		return fx
-	}
-	return &context{
-		Context: ctx,
-	}
-}
-
-var _ Context = &context{}
-
-// NewContext always returns fx.Context for use in the service
-func NewContext(ctx gcontext.Context, host service.Host) Context {
-	return &context{
-		Context: gcontext.WithValue(ctx, _logger, host.Logger()),
-	}
-}
-
-func (c *context) fx() {
-	//noop
-}
-
-func (c *context) Logger() ulog.Log {
-	return c.getLogger()
-}
-
-// WithContext returns a shallow copy of c with its context changed to ctx.
-// The provided ctx must be non-nil. Follows from net/http Request WithContext.
-func (c *context) WithContext(ctx gcontext.Context) (newC *context) {
-	if ctx == nil {
-		panic("nil context")
-	}
-	newC = new(context)
-	*newC = *c
-	newC.Context = ctx
-	return newC
-}
-
-func (c *context) getLogger() ulog.Log {
-	return c.Context.Value(_logger).(ulog.Log)
+	WithContext(ctx gcontext.Context) Context
 }
--- a/context.go
+++ b/context.go
@@ -24,18 +24,51 @@ import (
 	gcontext "context"
 
 	"go.uber.org/fx/service"
+	"go.uber.org/fx/ulog"
 )
+
+var _logger = "logger"
 
 // Context embeds Host and Go stdlib context for use
 type Context interface {
 	gcontext.Context
-	service.Host
+
+	// unexported func ensures only fx package implements Context
+	fx()
+	Logger() ulog.Log
 	WithContext(ctx gcontext.Context) *context
 }
 
 type context struct {
 	gcontext.Context
-	service.Host
+}
+
+// Convert context.Context to fx.Context
+func Convert(ctx gcontext.Context) Context {
+	fx, ok := ctx.(Context)
+	if ok {
+		return fx
+	}
+	return &context{
+		Context: ctx,
+	}
+}
+
+var _ Context = &context{}
+
+// NewContext always returns fx.Context for use in the service
+func NewContext(ctx gcontext.Context, host service.Host) Context {
+	return &context{
+		Context: gcontext.WithValue(ctx, _logger, host.Logger()),
+	}
+}
+
+func (c *context) fx() {
+	//noop
+}
+
+func (c *context) Logger() ulog.Log {
+	return c.getLogger()
 }
 
 // WithContext returns a shallow copy of c with its context changed to ctx.
@@ -50,12 +83,6 @@ func (c *context) WithContext(ctx gcontext.Context) (newC *context) {
 	return newC
 }
 
-var _ Context = &context{}
-
-// NewContext always returns fx.Context for use in the service
-func NewContext(ctx gcontext.Context, host service.Host) Context {
-	return &context{
-		Context: ctx,
-		Host:    host,
-	}
+func (c *context) getLogger() ulog.Log {
+	return c.Context.Value(_logger).(ulog.Log)
 }
--- a/internal/fxcontext/context.go
+++ b/internal/fxcontext/context.go
@@ -28,54 +28,39 @@ import (
 	"go.uber.org/fx/ulog"
 )
 
-var _logger = "logger"
+type contextKey int
 
-// Convert context.Context to fx.Context
-func Convert(ctx gcontext.Context) fx.Context {
-	fxctx, ok := ctx.(fx.Context)
-	if ok {
-		return fxctx
-	}
-	return New(ctx, nil)
-}
+const (
+	contextLogger contextKey = iota
+)
 
-type context struct {
+var _ fx.Context = &Context{}
+
+// Context embeds Host and Go stdlib context for use
+type Context struct {
 	gcontext.Context
 }
 
-var _ fx.Context = &context{}
-
-// New always returns fx.Context for use in the service
+// New always returns Context for use in the service
 func New(ctx gcontext.Context, host service.Host) fx.Context {
 	if host != nil {
-		return &context{
-			Context: gcontext.WithValue(ctx, _logger, host.Logger()),
+		return &Context{
+			gcontext.WithValue(ctx, contextLogger, host.Logger()),
 		}
 	}
-	return &context{
+	return &Context{
 		Context: ctx,
 	}
 }
 
-func (c *context) Logger() ulog.Log {
+// Logger returns context based logger
+func (c *Context) Logger() ulog.Log {
 	return c.getLogger()
 }
 
-// WithContext returns a shallow copy of c with its context changed to ctx.
-// The provided ctx must be non-nil. Follows from net/http Request WithContext.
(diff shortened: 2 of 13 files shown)


context.go
```diff
@@ -24,18 +24,51 @@ import (
...
 	gcontext.Context
-	service.Host
+}
+
+// Convert context.Context to fx.Context
+func Convert(ctx gcontext.Context) Context {
+	fx, ok := ctx.(Context)
+	if ok {
+		return fx
+	}
```
anuptalwalkar: What do you think about type assertion before wrapping? I am not sure of the performance overhead of a check vs additional allocation.

sectioneight:
Wait. This looks like we're exporting a wrap function. I thought we
explicitly decided against that.

anuptalwalkar:
We need Convert/Upgrade function (I am just calling it Convert from now), in the handler wrappers and filters that we are using. Moving all that functionality inside Context was not the right way to do it. I moved the Context inside an internal package to avoid exposing `fxcontext.Convert` to users, which seems to be a nice middle ground between exporting/unexporting. 

@abhinav What do you think? This aligns with the suggestion in your RFC.

internal/fxcontext/context.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+	gcontext "context"
+
+	"go.uber.org/fx"
+	"go.uber.org/fx/service"
+	"go.uber.org/fx/ulog"
+)
+
+var _logger = "logger"
```
madhuravi: Might be a good idea to prefix all our keys with fx. So this would be "fx-logger". This will prevent our keys being overriden by other teams if they also happen to be setting these values.

prashantv: Avoid strings, if you want to make sure these never clash with any other value in the context, you should create a custom type and use values of that type,
see https://github.com/uber/tchannel-go/blob/dev/context.go#L31

Article on this:
https://medium.com/@matryer/context-keys-in-go-5312346a868d#.t2wme3z2b

Since it does an equals comparison, a custom int type is probably the best option.

abhinav:
Hey, no this is not what we discussed in the https://github.com/uber-go/fx/issues/126.

`ctx.(fx.Context)` will *not* work. See https://github.com/uber-go/fx/issues/126#issuecomment-267157743 for why.

@sectioneight agreed that we should go for the original suggestion in the RFC (outlined here: https://github.com/uber-go/fx/issues/126#issue-195618785) unless there are performance issues in which case you can switch to the Upgrade-based solution (outlined here: https://github.com/uber-go/fx/issues/126#issuecomment-267130611) without breaking anything. An Upgrade function is not needed in the first case because one can simply do `fx.Context{ctx}`.

anuptalwalkar:
Hmm. My only worry with the first comment approach is that `fx.Context{ctx}` conversion is available for access. I guess we can keep it inside an internal package.

internal/fxcontext/context.go
anuptalwalkar: Thanks for the pointer. I added contextLogger as a custom contextType to for WithValue access.

abhinav:
That's a valid concern with the struct approach. It's not completely backwards
compatible to switch to the interface-based approach if anyone uses
`Context{ctx}` rather than `Upgrade(ctx)`. In that case it could be an
unexported attribute rather than a pubicly embedded attribute.

anuptalwalkar:
+1 I have added `Context{ctx}`, but struct is part of an internal package and fx.Context is the only interface exposed to users.  We can switch to Upgrade(ctx) easily with just unexporting the struct and providing the function access if/when there are performance concerns.

modules/uhttp/filters_test.go
```diff
@@ -29,36 +29,40 @@ import (
...
 	response := httptest.NewRecorder()
-	ctx := fx.NewContext(context.Background(), service.NullHost())
+	ctx := fxcontext.New(context.Background(), service.NullHost())
 	chain.ServeHTTP(ctx, response, request)
 	return response
 }
 
-func getNoopHandler() Handler {
-	return HandlerFunc(func(ctx fx.Context, w http.ResponseWriter, r *http.Request) {
+func getNoopFilter(host service.Host) HandlerFunc {
```
madhuravi: Minor: I think the previous handler naming actually helped distinguish the final handler. I prefer that for readability.

internal/fxcontext/context.go
```diff
@@ -0,0 +1,64 @@
...
+
+// New always returns Context for use in the service
+func New(ctx gcontext.Context, host service.Host) fx.Context {
+	if host != nil {
+		return &Context{
+			gcontext.WithValue(ctx, _contextLogger, host.Logger()),
+		}
+	}
+	return &Context{
+		Context: ctx,
```
abhinav: Reduce duplication. This can be compressed down to,

```go
if host != nil {
    ctx = gcontext.WithValue(...)
}
return &Context{Context: ctx}
```

internal/fxcontext/context.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+}
+
+// Logger returns context based logger
+func (c *Context) Logger() ulog.Log {
+	return c.getLogger()
+}
+
+func (c *Context) getLogger() ulog.Log {
```
abhinav: Why does this function exist separately?

If you're trying to go for the suggestion from the original issue, it was
actually a function that operates on the Google Context.

```go
func GetLogger(ctx context.Context) ulog.Log { ... }
```

Allowing the `Logger` method for `fx.Context` to be,

```go
func (c *Context) Logger() ulog.Log {
    return GetLogger(c.Context)
}
```

internal/fxcontext/context.go
```diff
@@ -0,0 +1,64 @@
...
+	if c.Context.Value(_contextLogger) == nil {
+		c.Context = gcontext.WithValue(c.Context, _contextLogger, ulog.Logger())
```
abhinav: Ideally the getter doesn't modify the context and just returns a default value of some kind.

internal/fxcontext/context.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	return c.Context.Value(_contextLogger).(ulog.Log)
```
abhinav: Don't call `ctx.Value` twice here. Store the value in a local variable.
`ctx.Value(..)` is `O(n)` where `n` is the number of values stored in the
context with `WithValue`.

Because of that reason, you may also want to use a single struct for all
UberFx data you store on the struct rather than one `WithValue` for each piece
of information.

internal/fxcontext/context_test.go
```diff
@@ -18,37 +18,60 @@
...
+	gcontext "context"
 	"testing"
 
-	"go.uber.org/fx/service"
-
 	"github.com/stretchr/testify/assert"
 
-	gcontext "context"
+	"go.uber.org/fx/service"
+	"go.uber.org/fx/ulog"
```
abhinav: Is this the import group ordering you use? Generally we use,

-   Standard library
-   Local package imports
-   Third party package imports

modules/uhttp/client/client_filters_test.go
```diff
@@ -26,9 +26,13 @@ import (
 	"net/http/httptest"
 	"testing"
 
+	"golang.org/x/net/context"
+
 	"go.uber.org/fx"
 
 	"github.com/stretchr/testify/assert"
+	"go.uber.org/fx/internal/fxcontext"
+	"go.uber.org/fx/service"
```
abhinav: import group ordering is upside down

modules/uhttp/client/client_filters_test.go
```diff
@@ -26,9 +26,13 @@ import (
 	"net/http/httptest"
 	"testing"
 
+	"golang.org/x/net/context"
```
abhinav: Is this the `context` package you meant to import?

modules/uhttp/filters.go
```diff
@@ -33,61 +36,83 @@ import (
...
-	operationName := r.Method
-	carrier := opentracing.HTTPHeadersCarrier(r.Header)
-	spanCtx, err := ctx.Tracer().Extract(opentracing.HTTPHeaders, carrier)
-	if err != nil && err != opentracing.ErrSpanContextNotFound {
-		ctx.Logger().Info("Malformed inbound tracing context: ", "error", err.Error())
+// contextFilter updates context to fx.Context
+func contextFilter(host service.Host) FilterFunc {
+	return func(ctx context.Context, w http.ResponseWriter, r *http.Request, next Handler) {
+		fxctx := fxcontext.New(ctx, host)
+		next.ServeHTTP(fxctx, w, r)
```
abhinav: IMO this and the other filters should be types that implement `Filter` instead
of using closures to hold the attributes.

```
type contextFilter struct{ Host service.Host }

func (c *contextFilter) Apply(...) {
    fxctx := fxcontext.New(ctx, host)
    next.ServeHTTP(..., w, r)
}
```

internal/fxcontext/context.go
anuptalwalkar: No this won't work since new is returning fx.Context.

internal/fxcontext/context.go
anuptalwalkar: I was going for storing logger inside the context.Context as a value, which is what c.getLogger() is doing. I don't need it to be a separate function, since Logger() can achieve that.

internal/fxcontext/context.go
abhinav: What do you mean? This is the same logic you have without doing `&Context{..}`
twice. You're still returning an `fx.Context`.

internal/fxcontext/context.go
abhinav: Yes, I mean either you don't need a separate `getLogger` method and you inline
the logic into the `Logger` method, or you have a top-level `GetLogger`
function which operates on `context.Context`s (using `.Value(..)`), and the
`Logger` method simply calls that on the underlying `context.Context`.

There is currently no benefit to the separate `getLogger` function.

internal/fxcontext/context.go
anuptalwalkar: I see.  I got an impression that ctx.Value is a map O(1), I will update it to single struct to avoid perf issues.

internal/fxcontext/context.go
anuptalwalkar: right. Im removing getLogger() and keeping it in Logger(). Having just a default value doesn't buy us much in terms of maintaining the context (when WithValue is called).

anuptalwalkar:
Thanks @abhinav, I will send a follow up PR since this PR is already merged.

PR #129, opened 2016-12-15
Stop running tests twice

sectioneight:
COV_REPORT is already the target that runs the tests. Prior to this
`make test` was running the tests with `go test` and then again for
coverage.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,9 +22,7 @@ ALL_SRC := $(shell find . -name "*.go" | grep -v -e vendor \
 TEST_TIMEOUT := "-timeout=3s"
 
 .PHONY: test
-test: examples
-	$(ECHO_V)go test $(RACE) $(TEST_TIMEOUT) $(PKGS)
-	$(ECHO_V)$(MAKE) $(COV_REPORT)
+test: examples $(COV_REPORT)
 
 TEST_IGNORES = vendor .git
 COVER_IGNORES = $(TEST_IGNORES) examples testutils


sectioneight:
@anuptalwalkar

Makefile
```diff
@@ -22,9 +22,7 @@ ALL_SRC := $(shell find . -name "*.go" | grep -v -e vendor \
 TEST_TIMEOUT := "-timeout=3s"
 
 .PHONY: test
-test: examples
-	$(ECHO_V)go test $(RACE) $(TEST_TIMEOUT) $(PKGS)
```
anuptalwalkar: I added this because of strange behavior on race check. 

`go test $(RACE)` fails the build on race condition and `make test` returns 2. 
`$(COV_REPORT)`, on the other hand results in an error and `make test` returns 0

Any other way to fix `$(COV_REPORT)` that results in `make test` to actually error out?

sectioneight: I'll let you pick up this PR then, but we shouldn't be running the tests twice.

anuptalwalkar: This test fails on race condition which is skipped: https://github.com/uber-go/fx/blob/master/service/host_test.go#L275

anuptalwalkar: I tried multiple ways to get make test it working with data race check to fail the build, but we have data race in few more places (service package and few tests in sentry). Lets get this in for now to have faster builds and work on fixing data race conditions as a task item.

sectioneight:
@anuptalwalkar want to merge then? I can't approve this since this is my PR technically

sectioneight:
Thanks!

PR #131, opened 2016-12-16
Benchmarks for context conversion

anuptalwalkar:
Type conversion benchmarks for wrapping context.Context with fx.Context struct vs using Upgrade/Convert function with type assertion.

Looks like with type assertion when no WithValue is called, saves us 1 alloc and ~7 ns/op.  When WithValue is called, type assertion is 8 ns/op expensive. 

```
BenchmarkContext_StructWrapper-8                	30000000	        55.3 ns/op	      32 B/op	       3 allocs/op
BenchmarkContext_StructWrapperWithValue-8       	30000000	        43.5 ns/op	      32 B/op	       3 allocs/op
BenchmarkContext_WithTypeAssertion-8            	30000000	        47.8 ns/op	      16 B/op	       2 allocs/op
BenchmarkContext_WithTypeAssertionWithValue-8   	20000000	        52.8 ns/op	      32 B/op	       3 allocs/op
```

### DIFF
--- a/internal/fxcontext/context_benchmark_test.go
+++ b/internal/fxcontext/context_benchmark_test.go
@@ -1,0 +1,78 @@
+package fxcontext
+
+import (
+	"context"
+	"testing"
+
+	"go.uber.org/fx"
+	"go.uber.org/fx/service"
+	"go.uber.org/fx/ulog"
+)
+
+func withContext(t *testing.B, f func(fxctx fx.Context)) {
+	f(New(context.Background(), service.NullHost()))
+}
+
+func BenchmarkContext_StructWrapper(b *testing.B) {
+	withContext(b, func(fxctx fx.Context) {
+		b.ResetTimer()
+		b.RunParallel(func(pb *testing.PB) {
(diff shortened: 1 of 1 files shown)


internal/fxcontext/context_benchmark_test.go
```diff
@@ -0,0 +1,98 @@
...
+	"go.uber.org/fx"
+	"go.uber.org/fx/service"
+	"go.uber.org/fx/ulog"
+)
+
+func withContext(t *testing.B, f func(fxctx fx.Context)) {
+	f(New(context.Background(), service.NullHost()))
+}
+
+func BenchmarkContext_StructWrapper(b *testing.B) {
```
madhuravi: It's confusing that the test StructWrapper is calling the function typeConversion on line 41 whereas the test WithTypeAssertion is calling the function typeUpgrade. Fix and clarify naming.

sectioneight:
Numbers look good. Nanoseconds are totally fine. Allocs seem small. I don't think we need to worry about optimizing this for now.

PR #132, opened 2016-12-16
 Followup nit fixes on fx.Context

anuptalwalkar:
Addressed feedback on #131 

Updated benchmark after addressing feedback-
```
BenchmarkContext_TypeConversion-8               	50000000	        36.9 ns/op	      24 B/op	       2 allocs/op
BenchmarkContext_TypeConversionWithValue-8      	50000000	        34.5 ns/op	      24 B/op	       2 allocs/op
BenchmarkContext_WithTypeAssertion-8            	50000000	        30.8 ns/op	       8 B/op	       1 allocs/op
BenchmarkContext_WithTypeAssertionWithValue-8   	50000000	        39.0 ns/op	      24 B/op	       2 allocs/op
PASS
```

### DIFF
--- a/internal/fxcontext/context.go
+++ b/internal/fxcontext/context.go
@@ -46,7 +46,9 @@ type Context struct {
 // New always returns Context for use in the service
 func New(ctx gcontext.Context, host service.Host) fx.Context {
 	if host != nil {
-		ctx = gcontext.WithValue(ctx, _fxContextStore, store{host.Logger()})
+		ctx = gcontext.WithValue(ctx, _fxContextStore, store{
+			log: host.Logger(),
+		})
 	}
 	return &Context{
 		Context: ctx,
--- a/internal/fxcontext/context.go
+++ b/internal/fxcontext/context.go
@@ -30,7 +30,11 @@ import (
 
 type contextKey int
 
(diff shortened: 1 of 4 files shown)


breerly:
cc @abhinav @prashantv

sectioneight:
Really appreciate the extra reviews, thanks RPC friends! I haven't had time
to code review with my managerial duties and am excited on this new
collaboration

internal/fxcontext/context.go
```diff
@@ -42,23 +46,24 @@ type Context struct {
...
-}
-
-func (c *Context) getLogger() ulog.Log {
-	if c.Context.Value(_contextLogger) == nil {
-		c.Context = gcontext.WithValue(c.Context, _contextLogger, ulog.Logger())
+	fxctxStore := c.Context.Value(_fxContextStore)
+	if fxctxStore == nil {
+		fxctxStore = store{
+			log: ulog.Logger(),
+		}
```
abhinav: I would add a private `getStore()` function which takes care of replacing the
context with a default store so that `Logger()` is just,

    return c.getStore().log

That way you don't have to repeat this when you add a new property.

PR #133, opened 2016-12-20
make: Always use vendored thriftrw and plugins

abhinav:
This changes `make examples` to always use the vendored `thriftrw` and
`thriftrw-yarpc-plugin`. It took me a while to realize that `make test` was
failing to build because it was using my globally installed version of
`thriftrw` and `thriftrw-yarpc-plugin`, both of which are newer versions than
what UberFx is using.

The `vendor-build.sh` script added here is also usd in yarpc-go to do
something similar. It is a simplified version of an internal tool which we are
planning on open sourcing in the future.


breerly:
Thanks, this saved me some time (managing the global install of vendored bins is annoying).

:shipit:

.build/vendor-build.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
```
anuptalwalkar: nit, filename: vendor_build.sh, or build_vendor.sh to conform with .build file naming?

anuptalwalkar: Also, some context on usage would be nice.

abhinav:
To use a different version of the plugin, you simply have it on your `$PATH`
and call `go generate` in that directory manually rather than relying on `make
examples`. Alternatively, we can add a flag to disable this later if needed.

PR #136, opened 2016-12-21
Integrate auth interface

anuptalwalkar:
GFM-231
This is a non-failing version of auth integration with uhttp


modules/uhttp/client/client.go
```diff
@@ -95,11 +114,19 @@ func (c *Client) PostForm(
 
 // Head is a context-aware, filter-enabled extension of Head() in http.Client
 func (c *Client) Head(ctx fx.Context, url string) (resp *http.Response, err error) {
+	authClient := uauth.Client()
+	fxctx, err := authClient.Authenticate(ctx)
+	if err != nil {
+		ctx.Logger().Error(uauth.ErrAuthentication, "error", err)
+	}
```
sectioneight: This seems repetitive (third time in this file). Any way to DRY?

modules/uhttp/filters.go
```diff
@@ -31,6 +31,7 @@ import (
 
 	"github.com/opentracing/opentracing-go"
 	"github.com/opentracing/opentracing-go/ext"
+	"go.uber.org/fx/uauth"
```
sectioneight: nit: grouping (above)

modules/uhttp/filters.go
```diff
@@ -80,6 +81,21 @@ func tracingServerFilter(host service.Host) FilterFunc {
...
+func authFilter(host service.Host) FilterFunc {
+	return func(ctx context.Context, w http.ResponseWriter, r *http.Request, next Handler) {
+		fxctx := &fxcontext.Context{
+			Context: ctx,
+		}
+		authClient := uauth.Client()
+		err := authClient.Authorize(fxctx)
+		if err != nil {
+			fxctx.Logger().Error(uauth.ErrAuthorization, "error", err)
+		}
```
sectioneight: Should we also emit some metrics here?

modules/uhttp/http.go
```diff
@@ -128,7 +128,11 @@ func newModule(
 	module := &Module{
 		ModuleBase: *modules.NewModuleBase(ModuleType, mi.Name, mi.Host, []string{}),
 		handlers:   handlers,
-		filters:    []Filter{contextFilter(mi.Host), tracingServerFilter(mi.Host), panicFilter(mi.Host)},
+		filters: []Filter{
+			contextFilter(mi.Host),
+			tracingServerFilter(mi.Host),
+			authFilter(mi.Host),
+			panicFilter(mi.Host)},
```
sectioneight: nit: prefer trailing-comma for line-diff attribution:

```go
  authFilter(mi.Host),
  panicFilter(mi,Host),
}
```

service/core.go
```diff
@@ -35,6 +36,7 @@ import (
 
 // A Host represents the hosting environment for a service instance
 type Host interface {
+	AuthClient() uauth.AuthClient
```
sectioneight: TBD: is `Auth` too short? Do we mean that we encapsulate both Authorization **and** Authentication? If so, this is probably appropriate. If not, we should be explicit.

service/core.go
```diff
@@ -103,7 +105,16 @@ func (lc *loggingCore) Logger() ulog.Log {
 	return lc.log
 }
 
+type authCore struct {
```
sectioneight: nit: I don't like the term `core` -- it seems like a placeholder for something more descriptive

service/service_setup.go
```diff
@@ -111,3 +112,11 @@ func (svc *serviceCore) setupObserver(cfg config.Provider) {
 		}
 	}
 }
+
+func (svc *serviceCore) setupAuthClient(cfg config.Provider) {
```
sectioneight: This doesn't seem to be called anywhere in this PR. Is that intentional? Also, a note about thread-safety?

anuptalwalkar:
@drdoakz This is an interface integrated with the open source side, which allows us to integrate any auth client with the service to authorize/authenticate requests based on the provided context.

modules/uhttp/filters.go
anuptalwalkar: hmm. I don't have access to metrics, or host here. I added metrics for authorization.

service/core.go
anuptalwalkar: I actually don't need core here. Cleaned up.

sectioneight:
Don't have access to host? Isn't it in the function signature? Or is GH
email lying to me

anuptalwalkar:
Sorry, I just reread your comment. Yes, Added Metrics for Authorization. we don't have access for Authentication in client filters.

modules/uhttp/client/client.go
```diff
@@ -61,11 +64,15 @@ func (c *Client) do(ctx fx.Context, req *http.Request) (resp *http.Response, err
 
 // Get is a context-aware, filter-enabled extension of Get() in http.Client
 func (c *Client) Get(ctx fx.Context, url string) (resp *http.Response, err error) {
+	authctx := authenticate(ctx)
```
madhuravi: Why not make this a filter similar to the tracerFilter? Makes it more consistent and cleaner.

modules/uhttp/filters.go
```diff
@@ -80,6 +81,22 @@ func tracingServerFilter(host service.Host) FilterFunc {
 	}
 }
 
+// authFilter authorizes services based on configuration
+func authFilter(host service.Host) FilterFunc {
```
madhuravi: Rename to authorizeFilter so as not to confuse with authenticate.

modules/uhttp/filters.go
```diff
@@ -80,6 +81,22 @@ func tracingServerFilter(host service.Host) FilterFunc {
...
+	return func(ctx context.Context, w http.ResponseWriter, r *http.Request, next Handler) {
+		fxctx := &fxcontext.Context{
+			Context: ctx,
+		}
+		authClient := uauth.Client()
+		err := authClient.Authorize(fxctx)
+		if err != nil {
+			host.Metrics().Counter("http.authorize.fail").Inc(1)
+			fxctx.Logger().Error(uauth.ErrAuthorization, "error", err)
```
sectioneight: Should this `return` and / or write some status code rather than continuing on to `next` if auth fails?

uauth/uauth.go
```diff
@@ -0,0 +1,83 @@
...
+
+	if _registerFunc != nil {
+		_std = _registerFunc(info)
+	} else {
+		_std = stubClient()
+	}
+}
+
+// AuthClient is an interface to perform authorization and authentication
+type AuthClient interface {
```
sectioneight: nit: the package name already has "auth" in it so this stutters: `uauth.AuthClient` -- why not just `uauth.Client`?

modules/uhttp/client/client.go
anuptalwalkar: duh, good idea. I didn't think of adding client filter.

modules/uhttp/client/client_filters.go
```diff
@@ -81,6 +82,20 @@ func tracingFilter(
...
+) (resp *http.Response, err error) {
+	authClient := uauth.Instance()
+	authctx, err := authClient.Authenticate(ctx)
+	if err != nil {
+		ctx.Logger().Error(uauth.ErrAuthentication, "error", err)
+		return nil, err
+	}
+	return next.Do(&fxcontext.Context{
+		Context: authctx,
+	}, req)
```
madhuravi: nit: This reads a little weird to me. Alternative:
next.Do(
    &fxcontext.Context{Context: authctx},
    req
)

modules/uhttp/client/client_filters.go
```diff
@@ -81,6 +82,20 @@ func tracingFilter(
 	return resp, err
 }
 
+func authenticationFilter(
+	ctx fx.Context, req *http.Request, next BasicClient,
+) (resp *http.Response, err error) {
+	authClient := uauth.Instance()
+	authctx, err := authClient.Authenticate(ctx)
+	if err != nil {
+		ctx.Logger().Error(uauth.ErrAuthentication, "error", err)
```
anuptalwalkar: We don't have access to host here. I'll add a todo since it requires wiring service.Host.

uauth/auth_failure_sub.go
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
madhuravi: typo in filename?

madhuravi: also, should this be in the testutils package?

uauth/uauth.go
```diff
@@ -0,0 +1,90 @@
...
+import (
+	"context"
+	"sync"
+
+	"go.uber.org/fx/config"
+)
+
+var (
+	_registerFunc RegisterFunc
+	_setupMux     sync.Mutex
```
madhuravi: nit: We've used _nameMu and _nameMux for mutex variables. We should pick one and be consistent. I prefer _setupMu. (Mux makes me think of multiplexing)

uauth/localauth.go
```diff
@@ -0,0 +1,54 @@
...
+
+package uauth
+
+import "context"
+
+var _std = defaultAuth()
+
+var _ Client = &base{}
+
+type base struct {
```
madhuravi: nit: Rename to defaultClient, something of that sort. base is very generic.

uauth/auth_stub.go
```diff
@@ -0,0 +1,44 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package uauth
+
+import "context"
+
+var _ Client = &stub{}
+
+type stub struct {
```
madhuravi: Why do we need this? Not sure why defaultClient uses this. Is this just a NoopClient? stub feels very generic, rename.

modules/uhttp/client/client_filters.go
madhuravi: sorry I meant to add this to uhttp/filters.go, we have access to the host there.

uauth/localauth.go
```diff
@@ -0,0 +1,54 @@
...
+
+func defaultAuth() Client {
+	return &base{
+		authClient: stubClient(nil),
+	}
+}
+
+// Instance returns initialized instance of auth client
+func Instance() Client {
+	return _std
```
madhuravi: We need a mutex for _std too. Instance() is called in the filters and _std may get over-written with SetupClient in auth.go

anuptalwalkar: I don't like the idea of having mutex here since its in the request path. Thinking about this more, better solution would be to initialize it once on setup call.

modules/uhttp/client/client_filters.go
anuptalwalkar: ok, It's already added

modules/uhttp/filters.go
```diff
@@ -80,6 +81,25 @@ func tracingServerFilter(host service.Host) FilterFunc {
...
+// authorizationFilter authorizes services based on configuration
+func authorizationFilter(host service.Host) FilterFunc {
+	return func(ctx context.Context, w http.ResponseWriter, r *http.Request, next Handler) {
+		fxctx := &fxcontext.Context{
+			Context: ctx,
+		}
+		authClient := uauth.Instance()
+		err := authClient.Authorize(fxctx)
+		if err != nil {
+			host.Metrics().Counter("http.authorize.fail").Inc(1)
```
alsamylkin: Use scopes instead of dots for separation?

modules/uhttp/client/client.go
```diff
@@ -95,11 +105,23 @@ func (c *Client) PostForm(
...
+	}, req)
+}
+
+func authenticate(ctx fx.Context) context.Context {
+	authClient := uauth.Client()
+	authctx, err := authClient.Authenticate(ctx)
+	if err != nil {
+		ctx.Logger().Error(uauth.ErrAuthentication, "error", err)
+	}
+	return authctx
```
anuptalwalkar: We are not. Updated.

modules/uhttp/client/client_filters.go
madhuravi: Adding a TODO here sounds good.

uauth/localauth.go
madhuravi: That works. Glib used a freeze variable for metrics. Feel free to check that out as well.

uauth/localauth.go
```diff
@@ -0,0 +1,54 @@
...
+
+var _ Client = &defaultClient{}
+
+type defaultClient struct {
+	authClient Client
+}
+
+func defaultAuth() Client {
+	return &defaultClient{
+		authClient: noopClient(nil),
```
madhuravi: Should there be a TODO here? I'm guessing this is a placeholder.

PR #141, opened 2016-12-22
Add more sentry configuration params

alsamylkin:
It will let to specify params directly in a config, not thorough the code.

### DIFF
--- a/ulog/sentry/sentry.go
+++ b/ulog/sentry/sentry.go
@@ -44,6 +44,16 @@ var _zapToRavenMap = map[zap.Level]raven.Severity{
 	zap.FatalLevel: raven.FATAL,
 }
 
+// Configuration provides sentry DSN and other optional parameters for logging
+type Configuration struct {
+	DSN               string
+	MinLevel          *string                `yaml:"min_level,omitempty"`
+	TraceEnabled      *bool                  `yaml:"trace_enabled,omitempty"`
+	TraceSkipFrames   *int                   `yaml:"trace_skip_frames,omitempty"`
+	TraceContextLines *int                   `yaml:"trace_context_lines,omitempty"`
+	Fields            map[string]interface{} `yaml:",omitempty"`
+}
+
 // Hook wraps the default raven-go client for some out-of-box awesomeness
 // and tight integration with ulog
 type Hook struct {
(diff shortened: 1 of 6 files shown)


ulog/sentry/sentry.go
```diff
@@ -44,6 +44,16 @@ var _zapToRavenMap = map[zap.Level]raven.Severity{
 	zap.FatalLevel: raven.FATAL,
 }
 
+// Configuration provides sentry DSN and other optional parameters for logging
+type Configuration struct {
+	DSN               string
+	MinLevel          *string                `yaml:"min_level,omitempty"`
+	TraceEnabled      *bool                  `yaml:"trace_enabled,omitempty"`
+	TraceSkipFrames   *int                   `yaml:"trace_skip_frames,omitempty"`
```
alsamylkin: Yep, make sense, thanks!

ulog/sentry/sentry_test.go
```diff
@@ -158,6 +158,70 @@ func TestPacketSending(t *testing.T) {
...
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			conf, err := FromConfig(tc.conf)
+			if tc.isErr {
+				assert.Error(t, err)
+				return
+			}
+
+			assert.Nil(t, err)
```
sectioneight: require.NoError

ulog/sentry/sentry.go
```diff
@@ -44,6 +44,18 @@ var _zapToRavenMap = map[zap.Level]raven.Severity{
 	zap.FatalLevel: raven.FATAL,
 }
 
+// Configuration provides sentry DSN and other optional parameters for logging
+type Configuration struct {
+	DSN      string
+	MinLevel *string                `yaml:"min_level,omitempty"`
```
sectioneight: why all the omit empty tags? iirc those are only useful if we serialize to YAML which we'd never do -- only deser for config

alsamylkin: Right, I've read the docs and assumed both ways, but tested and it is unnecessary for parsing.

ulog/sentry/sentry.go
```diff
@@ -92,6 +104,43 @@ func New(dsn string, options ...Option) (*Hook, error) {
 	return sh, nil
 }
 
+// FromConfig returns a new Sentry Hook based on SentryConfiguration.
+func FromConfig(c Configuration) (*Hook, error) {
```
anuptalwalkar: nit: Make it less verbose, `sentry.Configure(c Configuration)`

ulog/sentry/sentry_test.go
```diff
@@ -158,6 +159,70 @@ func TestPacketSending(t *testing.T) {
...
+			}, false},
+		{"ParseError", Configuration{MinLevel: &str}, nil, true},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			conf, err := FromConfig(tc.conf)
+			if tc.isErr {
+				assert.Error(t, err)
+				return
```
anuptalwalkar: nit: prefer else/

madhuravi:
Fix ulog/doc.go, it currently has instructions to write code to disable traces. Change to use config the way you specified.

ulog/sentry/sentry_test.go
alsamylkin: Just reducing nesting: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

PR #142, opened 2016-12-22
Add a root node string

alsamylkin:
Remove a confusing interface for getting the root node.

### DIFF
--- a/config/provider.go
+++ b/config/provider.go
@@ -25,6 +25,9 @@ import "fmt"
 // ConfigurationChangeCallback is called for updates of configuration data
 type ConfigurationChangeCallback func(key string, provider string, configdata interface{})
 
+// Root marks the root node in a Provider
+const Root = ""
+
 // A Provider provides a unified interface to accessing
 // configuration systems.
 type Provider interface {
--- a/config/README.md
+++ b/config/README.md
@@ -126,6 +126,14 @@ fmt.Println(foo)
 // Output: hello
 ```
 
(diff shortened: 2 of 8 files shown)


anuptalwalkar:
Can you update the readme with reference to Get("") as well?

modules/uhttp/client/client.go
```diff
@@ -39,7 +39,7 @@ type Client struct {
 
 // New creates a new instance of uhttp Client
 func New(client *http.Client, filters ...Filter) *Client {
-	filters = append(filters, FilterFunc(tracingFilter))
```
anuptalwalkar: hmm.. you probably need to merge master and push in your branch.

config/provider.go
```diff
@@ -25,6 +25,9 @@ import "fmt"
 // ConfigurationChangeCallback is called for updates of configuration data
 type ConfigurationChangeCallback func(key string, provider string, configdata interface{})
 
+// Root marks the root node in a Provider
+const Root = ""
```
anuptalwalkar: We should think about adding user config types instead of string Root = "", which can be used everywhere. (e.g. Get(RPCModule), Get(LogConfig), etc..

alsamylkin: You mean actual types or or just const strings? I am not aware how can we do the former..

anuptalwalkar: Something like - 
```go
type ConfigType string

const (
  Root = ""
)
```

PR #145, opened 2016-12-23
Authorize requests with rpc

anuptalwalkar:
GFM-231

### DIFF
--- a/modules/rpc/handler.go
+++ b/modules/rpc/handler.go
@@ -26,6 +26,7 @@ import (
 	"go.uber.org/fx"
 	"go.uber.org/fx/internal/fxcontext"
 	"go.uber.org/fx/service"
+	"go.uber.org/fx/uauth"
 	"go.uber.org/thriftrw/wire"
 	"go.uber.org/yarpc"
 	"go.uber.org/yarpc/encoding/thrift"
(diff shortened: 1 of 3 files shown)


modules/rpc/handler.go
```diff
@@ -26,6 +26,7 @@ import (
 	"go.uber.org/fx"
 	"go.uber.org/fx/internal/fxcontext"
 	"go.uber.org/fx/service"
+	"go.uber.org/fx/uauth"
```
breerly: Why not just `auth` here? Might make sense to keep the branding at the top-level, e.g. FX, but then have all the parts *inside* FX be named in a boring and straightforward way.

modules/rpc/handler.go
```diff
@@ -97,3 +98,41 @@ func (f fxContextOnewayInboundMiddleware) HandleOneway(ctx context.Context, req
...
 }
+
+type authInboundMiddleware struct {
+	service.Host
+}
+
+func (a authInboundMiddleware) Handle(ctx context.Context, req *transport.Request, resw transport.ResponseWriter, handler transport.UnaryHandler) error {
+	fxctx, err := authorize(ctx, a.Host)
+	if err != nil {
+		return err
```
breerly: This is going to surface as a server-side error... we should define a first-class error in YARPC so that we can handle AuthZ errors seperately from server error.

Created https://github.com/yarpc/yarpc-go/issues/653 - you should probably put a TODO here for now.

modules/rpc/handler.go
anuptalwalkar: Yeah, you are right. I was hoping for uber specific authentication to be plugged in eventually, may be open sourced, but `auth` seems pretty generic as well. I'll get a rename PR once this change gets in. I don't want to add unrelated stuff to this PR.

modules/rpc/handler.go
anuptalwalkar: cool. I will add a TODO. Thanks for the review!

breerly: Just chatted with @abhinav - you can actually return a `yarpc.api.transport.BadRequestError` which will get treated as a 4xx style error. This is important because Authz/n doesn't mean the server is unhealthy, it means the client is err'ing.

modules/rpc/handler.go
kriskowal: I’m tempted to advise in the other direction—package names are “global” for the purposes of tools that automatically infer imports. I’d be tempted to prefix or suffix FX modules with fx to give tools a push, the way we have "yarpc/yarpctest" and not merely "yarpc/test".

breerly: Agreed when the term overlaps with standard lib. In this case, it doesn't, and you wouldn't be using 2 auth libs at the same time.

modules/rpc/handler.go
anuptalwalkar: Interesting. BadRequestError is an internal package to YARPC (with the version I have with fx, before /api/transport) I will add a TODO once upgrade happens, to move the error to BadRequestError.

breerly: 👍

modules/rpc/handler.go
anuptalwalkar: I am keeping it as is. Let's get an issue open to discuss naming. `u` in `uauth`, `uhttp` or `ulog` comes from `uberFx`.  We would need a generic naming overhaul

anuptalwalkar:
> why do we update metrics in host, but log in context?

Logging keeps track of the entire request context. Metrics on the other hand is derived from global entity (aka host). We can use metrics on the request, but it doesn't have any request specific data. We decided to keep request specific data in the context, and rest of the stuff derived from host.

PR #149, opened 2017-01-06
Auth interface with Logger and SetAttribute methods

anuptalwalkar:
Updated CreateAuthInfo with Logger, for logging during Auth.Client creation.
SetAttribute method to set necessary baggage from service (service name, UUID, etc) during authentication.

### DIFF
--- a/auth/auth_failure_client.go
+++ b/auth/auth_failure_client.go
@@ -44,6 +44,6 @@ func (*failureClient) Authorize(ctx context.Context) error {
 	return errors.New(ErrAuthorization)
 }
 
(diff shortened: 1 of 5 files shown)


auth/auth_failure_client.go
```diff
@@ -43,3 +43,7 @@ func (*failureClient) Authenticate(ctx context.Context) (context.Context, error)
 func (*failureClient) Authorize(ctx context.Context) error {
 	return errors.New(ErrAuthorization)
 }
+
+func (*failureClient) SetAttribute(ctx context.Context, key string, value string) context.Context {
```
sectioneight: tiny nit: `key, value string`

PR #151, opened 2017-01-09
Prepare unstuttering of configuration change callback

glibsm:
One private hooks are fixed, we can drop the old name
and change the alias definition to be the source of truth

### DIFF
--- a/config/provider.go
+++ b/config/provider.go
@@ -22,8 +22,12 @@ package config
 
 import "fmt"
 
+// ChangeCallback is a temporary step to un-stutter config.ConfigurationChangeCallback
+// TODO(glib): remove this after direct usage of ConfigurationChangeCallback has been removed
+type ChangeCallback ConfigurationChangeCallback
+
 // ConfigurationChangeCallback is called for updates of configuration data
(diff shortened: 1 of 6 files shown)


config/provider.go
```diff
@@ -22,8 +22,12 @@ package config
 
 import "fmt"
 
+// ChangeCallback is a temporary step to un-stutter config.ConfigurationChangeCallback
+// TODO(glib): remove this after direct usage of ConfigurationChangeCallback has been removed
+type ChangeCallback ConfigurationChangeCallback
+
 // ConfigurationChangeCallback is called for updates of configuration data
-type ConfigurationChangeCallback func(key string, provider string, configdata interface{})
+type ConfigurationChangeCallback func(key string, provider string, configData interface{})
```
madhuravi: Why don't we just rename this?

glibsm: We could, but I made the decision to alias for two reasons:

1. Because it would immediately break the private hooks and I'm not sure that's the best thing to do right now - fixing them. We can fix them when the time is right to touch that code.

2. I think this is a good practice for the future, as we won't just be able to rename things willy-nilly, we would be bound by the major versions.

PR #152, opened 2017-01-09
Log the trace id and span id in http filters

madhuravi:
Set trace id and span id as logger fields in the http filter chain.

### DIFF
--- a/internal/fxcontext/context.go
+++ b/internal/fxcontext/context.go
@@ -58,15 +58,28 @@ func New(ctx gcontext.Context, host service.Host) fx.Context {
 // Logger returns context based logger. If logger is absent from the context,
 // the function updates the context with a new context based logger
 func (c *Context) Logger() ulog.Log {
-	return c.getStore().log
+	store := c.getStore()
+	if store.log == nil {
+		store.log = ulog.Logger()
+	}
+	return store.log
 }
 
+// WithLogger returns a new context with the input logger
+func (c *Context) WithLogger(logger ulog.Log) *Context {
+	store := c.getStore()
+	store.log = logger
+	ctx := gcontext.WithValue(c, _fxContextStore, store)
+	return &Context{
+		Context: ctx,
+	}
+}
+
+// getStore returns a non-nil store instance
 func (c *Context) getStore() store {
 	fxctxStore := c.Context.Value(_fxContextStore)
 	if fxctxStore == nil {
-		fxctxStore = store{
-			log: ulog.Logger(),
-		}
+		fxctxStore = store{}
 		c.Context = gcontext.WithValue(c.Context, _fxContextStore, fxctxStore)
 	}
 	return fxctxStore.(store)
--- a/internal/fxcontext/context_test.go
+++ b/internal/fxcontext/context_test.go
@@ -24,6 +24,8 @@ import (
 	gcontext "context"
 	"testing"
 
+	"golang.org/x/net/context"
+
 	"go.uber.org/fx/service"
 	"go.uber.org/fx/ulog"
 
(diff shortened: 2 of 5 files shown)


madhuravi:
@uber-go/service-framework

modules/uhttp/filters.go
```diff
@@ -68,11 +69,16 @@ func tracingServerFilter(host service.Host) FilterFunc {
...
 			fxctx.Logger().Info("Malformed inbound tracing context: ", "error", err.Error())
 		}
-
 		span := opentracing.GlobalTracer().StartSpan(operationName, ext.RPCServerOption(spanCtx))
 		ext.HTTPUrl.Set(span, r.URL.String())
 		defer span.Finish()
 
+		jSpanCtx, ok := span.Context().(jaeger.SpanContext)
+		if ok {
+			fxctx.Logger().Info(
```
anuptalwalkar: I thought this task meant to add it to the logger (using With()), and not actually output the log, so that whenever user calls info or error, Logger would populate with these preset fields.

madhuravi: Check description. This is not directly related to the task. This is for debugging.

internal/fxcontext/context.go
```diff
@@ -58,15 +58,28 @@ func New(ctx gcontext.Context, host service.Host) fx.Context {
 // Logger returns context based logger. If logger is absent from the context,
 // the function updates the context with a new context based logger
 func (c *Context) Logger() ulog.Log {
-	return c.getStore().log
+	store := c.getStore()
+	if store.log == nil {
```
alsamylkin: why do you want to always check for nil? I thought store is an internal type..

modules/uhttp/filters.go
```diff
@@ -68,13 +69,19 @@ func tracingServerFilter(host service.Host) FilterFunc {
 		if err != nil && err != opentracing.ErrSpanContextNotFound {
 			fxctx.Logger().Info("Malformed inbound tracing context: ", "error", err.Error())
 		}
-
 		span := opentracing.GlobalTracer().StartSpan(operationName, ext.RPCServerOption(spanCtx))
 		ext.HTTPUrl.Set(span, r.URL.String())
 		defer span.Finish()
 
+		jSpanCtx, ok := span.Context().(jaeger.SpanContext)
+		if ok {
```
alsamylkin: Nit: merge with the line above?

modules/uhttp/filters.go
```diff
@@ -68,13 +69,19 @@ func tracingServerFilter(host service.Host) FilterFunc {
...
+			traceLogger := fxctx.Logger().With(
+				"trace id", jSpanCtx.TraceID(), "span id", jSpanCtx.SpanID(),
```
alsamylkin: Can we test it?

internal/fxcontext/context_test.go
```diff
@@ -24,6 +24,8 @@ import (
 	gcontext "context"
 	"testing"
 
+	"golang.org/x/net/context"
```
anuptalwalkar: use gcontext

modules/uhttp/filters.go
madhuravi: What if its not a jaeger tracer? uberfx can be used with other types of tracers.

modules/uhttp/filters.go
madhuravi: I mentioned to Anup offline. I am writing tests, expect an update soon.

modules/uhttp/filters.go
alsamylkin: I meant this:
```go
if jSpanCtx, ok := span.Context().(jaeger.SpanContext); ok {
    traceLogger := fxctx.Logger().With(
 				"trace id", jSpanCtx.TraceID(), "span id", jSpanCtx.SpanID(),
 				)
    fxctx = fxctx.WithLogger(traceLogger)
}
```
Should be ok?

internal/fxcontext/context.go
madhuravi: Store can be created without a logger. That doesn't make much sense right now but in future once we add more things to the store object for eg. auth etc. It is possible to have one set and not the other.

PR #154, opened 2017-01-09
Insert service name for authentication

anuptalwalkar:
GFM-277

### DIFF
--- a/auth/localauth.go
+++ b/auth/localauth.go
@@ -22,7 +22,7 @@ package auth
 
 import "context"
 
-var _std = defaultAuth()
+var _std Client
 
 var _ Client = &defaultClient{}
 
@@ -40,6 +40,9 @@ func defaultAuth() Client {
 
 // Instance returns initialized instance of auth client
 func Instance() Client {
+	if _std == nil {
+		_std = defaultAuth()
+	}
 	return _std
 }
 
--- a/auth/localauth.go
+++ b/auth/localauth.go
@@ -32,7 +32,7 @@ type defaultClient struct {
 
 // defaultAuth is a placeholder auth client when no auth client is registered
 // TODO(anup): add configurable authentication, whether a service needs one or not
-func defaultAuth() Client {
+func defaultAuth(info CreateAuthInfo) Client {
 	return &defaultClient{
 		authClient: noopClient(nil),
 	}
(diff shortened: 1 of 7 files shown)


modules/uhttp/client/client.go
```diff
@@ -27,18 +27,22 @@ import (
 	"strings"
 
 	"go.uber.org/fx"
+	"go.uber.org/fx/config"
 
 	"golang.org/x/net/context/ctxhttp"
 )
 
+var serviceName string
```
alsamylkin: Nit: I thought we are using _ prefix for globals?

modules/uhttp/client/client_test.go
```diff
@@ -27,15 +27,20 @@ import (
...
+applicationID: test
+`)
+
 var _defaultHTTPClient = &http.Client{Timeout: 2 * time.Second}
-var _defaultUHTTPClient = New(_defaultHTTPClient)
+var _defaultUHTTPClient = New(config.NewYAMLProviderFromBytes(testYaml), _defaultHTTPClient)
 
 func TestNew(t *testing.T) {
-	uhttpClient := New(_defaultHTTPClient)
+	uhttpClient := New(config.NewYAMLProviderFromBytes(testYaml), _defaultHTTPClient)
```
alsamylkin: "Document" panic in a test with a config without applicationID?:)

auth/localauth.go
```diff
@@ -40,6 +40,9 @@ func defaultAuth() Client {
 
 // Instance returns initialized instance of auth client
 func Instance() Client {
+	if _std == nil {
```
alsamylkin: do we care about thread safety here?

anuptalwalkar: Good point. removing this check actually. I actually don't need default instance here, and I don't want lock in the request path.

modules/uhttp/client/client_filters.go
```diff
@@ -86,6 +86,10 @@ func authenticationFilter(
 	ctx fx.Context, req *http.Request, next BasicClient,
 ) (resp *http.Response, err error) {
 	authClient := auth.Instance()
+
+	// Client needs to know what service it is to authenticate
+	authClient.SetAttribute(ctx, auth.ServiceAuth, _serviceName)
```
alsamylkin: Do we need to set it up for every request? It doesn't seem to change..

anuptalwalkar: Yes, we need to set it every time on the context in order for authenticate to work. (ctx is new per request)

PR #155, opened 2017-01-10
Move logger context population so it can be reused by YARPC

madhuravi:


madhuravi:
@sectioneight

internal/fxcontext/context.go
```diff
@@ -23,6 +23,9 @@ package fxcontext
 import (
 	gcontext "context"
 
+	"github.com/opentracing/opentracing-go"
+	"github.com/uber/jaeger-client-go"
+
```
sectioneight: nit: these go below fx imports

internal/fxcontext/context.go
```diff
@@ -65,10 +68,17 @@ func (c *Context) Logger() ulog.Log {
...
-// WithLogger returns a new context with the input logger
-func (c *Context) WithLogger(logger ulog.Log) *Context {
+// WithContextAwareLogger returns a new context with a context-aware logger
+func (c *Context) WithContextAwareLogger(span opentracing.Span) *Context {
 	store := c.getStore()
-	store.log = logger
+	// Note that opentracing.Tracer does not expose the tracer id
+	// We only inject tracing information for jaeger.Tracer
+	if jSpanCtx, ok := span.Context().(jaeger.SpanContext); ok {
+		traceLogger := c.Logger().With(
```
sectioneight: let's use the non-sugared version here for perf in the tight loop

sectioneight:
Let's also add some documentation to read me/godoc about this behavior so people don't have to dig through code to discover it

internal/fxcontext/context.go
madhuravi: ulog.Logger With has some sentry magic. It adds the fields to sentry as well.

sectioneight: Hmm,  we should fix that. People should be able to add zap `Field`s directly and still get the same Sentry benefits. Particularly in a path like this (on the per-request loop) where we know the fields ahead of time, we **really** shouldn't be incurring the overhead of the `interface{}` conversion since we know it's expensive.

In other words, this should be something like:

```go
traceLogger := c.Logger().Typed().With(
  zap.StringField("traceID", traceID),
  zap.StringField("spanID", spanID),
)
```

madhuravi:
@sectioneight Do you mean documentation to uhttp or context? Use for WithContextAwareLogger is only within UberFx. And for http, users will not have to set anything. It should just work. Do we want to just say that this is a thing we have with the http module and you can expect to see it in the logs?

sectioneight:
@madhuravi I meant that although it does "just work", we should document somewhere that we're auto-inserting trace/span IDs into the logs and the format they're in, where they come from, and why you might want to use them.

internal/fxcontext/context.go
sectioneight: GFM-278

PR #156, opened 2017-01-10
Merge objects

alsamylkin:
We will be able to get access not only to leaf nodes of the parse tree and look up is going to be faster.

The changes are done only for yaml provider. We can update ProviderGroups later.

### DIFF
--- a/config/yaml.go
+++ b/config/yaml.go
@@ -54,7 +54,8 @@ func newYAMLProviderCore(files ...io.ReadCloser) Provider {
 		root: &yamlNode{
 			nodeType: objectNode,
 			key:      Root,
-			value:    root},
+			value:    root,
+		},
 		vCache: make(map[string]Value),
 	}
 }
@@ -74,7 +75,7 @@ func mergeMaps(dst interface{}, src interface{}) interface{} {
 	case map[interface{}]interface{}:
 		dstMap, ok := dst.(map[interface{}]interface{})
 		if !ok {
-			panic(fmt.Sprintf("Expected map[interface{}]interface{}, actual: %+v", dstMap))
+			panic(fmt.Sprintf("Expected map[interface{}]interface{}, actual: %T", dstMap))
 		}
 
 		for k, v := range s {
@@ -123,8 +124,8 @@ func NewYamlProviderFromReader(readers ...io.ReadCloser) Provider {
 // As above, all the objects are going to be merged and arrays/values overridden in the order of the yamls.
 func NewYAMLProviderFromBytes(yamls ...[]byte) Provider {
 	closers := make([]io.ReadCloser, len(yamls))
-	for i := range yamls {
-		closers[i] = ioutil.NopCloser(bytes.NewReader(yamls[i]))
+	for i, yml := range yamls {
+		closers[i] = ioutil.NopCloser(bytes.NewReader(yml))
 	}
 
 	return newYAMLProviderCore(closers...)
--- a/config/yaml.go
+++ b/config/yaml.go
@@ -29,6 +29,7 @@ import (
 	"strings"
 
 	"gopkg.in/yaml.v2"
+	"strconv"
 )
 
 type yamlConfigProvider struct {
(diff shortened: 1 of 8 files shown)


config/yaml.go
```diff
@@ -32,31 +32,67 @@ import (
...
 
 var _ Provider = &yamlConfigProvider{}
 
 func newYAMLProviderCore(files ...io.ReadCloser) Provider {
-	roots := make([]*yamlNode, len(files))
+	root := &yamlNode{
+		nodeType: objectNode,
+		key:      Root,
+		value:    make(map[interface{}]interface{}),
+		children: nil,
```
sectioneight: nit: this is superfluous

config/yaml.go
```diff
@@ -32,31 +32,67 @@ import (
...
+	}
 
-	for n, v := range files {
-		if root, err := newyamlNode(v); err != nil {
+	for _, v := range files {
+		tmp := make(map[interface{}]interface{})
```
sectioneight: nit: `tmp` isn't a create name for a variable

config/yaml.go
```diff
@@ -32,31 +32,67 @@ import (
...
+		panic("Destination node is nil")
+	}
+
+	if src == nil {
+		return src
+	}
+
+	switch s := src.(type) {
+	case map[interface{}]interface{}:
+		d, ok := dst.(map[interface{}]interface{})
```
sectioneight: I'm all for brevity but I think `d` may be a little too short given how far down it's used here.

config/yaml.go
```diff
@@ -76,40 +112,29 @@ func NewYAMLProviderFromFiles(mustExist bool, resolver FileResolver, files ...st
 	return newYAMLProviderCore(readers...)
 }
 
-// NewYamlProviderFromReader creates a configuration provider from an
-// io.ReadCloser
-func NewYamlProviderFromReader(reader io.ReadCloser) Provider {
-	return newYAMLProviderCore(reader)
+// NewYamlProviderFromReader creates a configuration provider from an io.ReadClosers.
```
sectioneight: nit: `an` suggests a single `ReadCloser`, but you take multiple. maybe:

```
creates a configuration provider from a list of `ReadClosers`
```
or something to indicate plurality

config/yaml.go
```diff
@@ -76,40 +112,29 @@ func NewYAMLProviderFromFiles(mustExist bool, resolver FileResolver, files ...st
...
+// Same as above all the objects are going to be merged and arrays/values overridden in the order of the files.
```
sectioneight: weird english, needs punctuation. "As above, all of the objects are going..."

config/yaml.go
```diff
@@ -76,40 +112,29 @@ func NewYAMLProviderFromFiles(mustExist bool, resolver FileResolver, files ...st
...
+func NewYamlProviderFromReader(reader ...io.ReadCloser) Provider {
```
sectioneight: call this `readers` not `reader` since it's plural

config/yaml.go
```diff
@@ -285,3 +274,27 @@ func (n *yamlNode) Children() []*yamlNode {
...
 	return nodes
 }
+
+func unmarshalYamlValue(reader io.ReadCloser, value interface{}) error {
+	if data, err := ioutil.ReadAll(reader); err != nil {
+		return err
+	} else if err = yaml.Unmarshal(data, value); err != nil {
+		return err
+	}
+	if err := reader.Close(); err != nil {
```
sectioneight: line 284 through 288 can just be `return reader.Close()`

config/yaml_test.go
```diff
@@ -258,3 +259,33 @@ func TestGrumpyTextUnMarshallerParsing(t *testing.T) {
...
+
+	ms := mapStruct{}
+	assert.NoError(t, provider.Get("mapStruct").PopulateStruct(&ms))
+
+	assert.NotNil(t, ms.MyMap)
+	assert.NotZero(t, len(ms.MyMap))
+
+	found := false
+	switch p := ms.MyMap["policy"].(type) {
+	case map[interface{}]interface{}:
```
sectioneight: what is this case statement doing in the test? seems like a table test is a better option so that it's clear exactly what behavior you're asserting

config/yaml.go
```diff
@@ -29,34 +29,71 @@ import (
...
+
+	if src == nil {
+		return src
+	}
+
+	switch s := src.(type) {
+	case map[interface{}]interface{}:
+		dstMap, ok := dst.(map[interface{}]interface{})
+		if !ok {
+			panic(fmt.Sprintf("Expected map[interface{}]interface{}, actual: %+v", dstMap))
```
sectioneight: i believe `%T` will give you the type if that's actually what you want to present to the user

config/yaml.go
```diff
@@ -29,34 +29,71 @@ import (
...
+
+		root = mergeMaps(root, curr)
 	}
 
 	return &yamlConfigProvider{
-		roots:  roots,
+		root: &yamlNode{
+			nodeType: objectNode,
+			key:      Root,
+			value:    root},
```
sectioneight: trailing comma and closing brace on its own line please

config/yaml.go
```diff
@@ -29,34 +29,71 @@ import (
...
+				dstMap[k] = v
+			} else {
+				dstMap[k] = mergeMaps(oldVal, v)
+			}
+		}
+	default:
+		dst = src
+	}
+
+	return dst
```
sectioneight: Maybe doesn't matter depending on how often this is called, but should we just modify the maps in place instead of allocating and returning a new one?

config/yaml.go
```diff
@@ -76,40 +113,29 @@ func NewYAMLProviderFromFiles(mustExist bool, resolver FileResolver, files ...st
...
-func NewYAMLProviderFromBytes(yaml []byte) Provider {
-	reader := bytes.NewReader(yaml)
-	node, err := newyamlNode(ioutil.NopCloser(reader))
-	if err != nil {
-		panic(err)
+// NewYAMLProviderFromBytes creates a config provider from a byte-backed YAML blobs.
+// As above, all the objects are going to be merged and arrays/values overridden in the order of the yamls.
+func NewYAMLProviderFromBytes(yamls ...[]byte) Provider {
+	closers := make([]io.ReadCloser, len(yamls))
+	for i := range yamls {
```
sectioneight:
```
for i, yml := range yamls {
```
then you don't have to re-index into `yaml[i]` on the next line

alsamylkin: Hmm, I thought, it is going to copy the temp variable array and use later, but it is not. Thanks!

PR #157, opened 2017-01-10
Add filters when client is initialized without New

anuptalwalkar:
Covered by existing tests.

### DIFF
--- a/modules/uhttp/client/client.go
+++ b/modules/uhttp/client/client.go
@@ -32,28 +32,35 @@ import (
 	"golang.org/x/net/context/ctxhttp"
 )
 
-var _serviceName string
+var (
+	_serviceName string
+	_defaultTrue = true
+)
 
 // Client wraps around a http client
 type Client struct {
 	*http.Client
-	filters []Filter
+	filters           []Filter
+	setDefaultFilters *bool
 }
 
 // New creates a new instance of uhttp Client
 func New(cfg config.Provider, client *http.Client, filters ...Filter) *Client {
 	_serviceName = cfg.Get(config.ApplicationIDKey).AsString()
 	filters = append(filters, FilterFunc(tracingFilter), FilterFunc(authenticationFilter))
-	return &Client{Client: client, filters: filters}
+	return &Client{
+		Client:            client,
+		filters:           filters,
+		setDefaultFilters: &_defaultTrue,
+	}
 }
 
 // Do is a context-aware, filter-enabled extension of Do() in http.Client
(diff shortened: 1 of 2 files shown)


modules/uhttp/client/client.go
```diff
@@ -32,28 +32,35 @@ import (
...
+	}
 }
 
 // Do is a context-aware, filter-enabled extension of Do() in http.Client
 func (c *Client) Do(ctx fx.Context, req *http.Request) (resp *http.Response, err error) {
 	filters := c.filters
-	// TODO: Need a way to handle the case where Client is initialized without the NewClient method
-	// and some filters are set. Need to always include the tracing filter
-	if len(filters) == 0 {
+	if c.setDefaultFilters == nil || *c.setDefaultFilters == false {
```
madhuravi: nil check not necessary. it defaults to false.

modules/uhttp/client/client.go
```diff
@@ -32,28 +32,35 @@ import (
...
+	_serviceName string
+	_defaultTrue = true
+)
 
 // Client wraps around a http client
 type Client struct {
 	*http.Client
-	filters []Filter
+	filters           []Filter
+	setDefaultFilters *bool
```
madhuravi: This name is a bit confusing. Maybe defaultFilters added? Also add a comment indicating what it does.

modules/uhttp/client/client.go
```diff
@@ -32,28 +32,35 @@ import (
 	"golang.org/x/net/context/ctxhttp"
 )
 
-var _serviceName string
+var (
+	_serviceName string
+	_defaultTrue = true
```
madhuravi: nit: default seems to indicate that the default value is true. The default value is actually false. Not sure it needs a variable, you can just set inline.

modules/uhttp/client/client.go
```diff
@@ -32,28 +32,34 @@ import (
...
+	}
 }
 
 // Do is a context-aware, filter-enabled extension of Do() in http.Client
 func (c *Client) Do(ctx fx.Context, req *http.Request) (resp *http.Response, err error) {
 	filters := c.filters
-	// TODO: Need a way to handle the case where Client is initialized without the NewClient method
-	// and some filters are set. Need to always include the tracing filter
-	if len(filters) == 0 {
+	if c.defaultFiltersAdded == false {
```
alsamylkin: Do we care that this is not thread safe?

anuptalwalkar: well, not really. Its not exported and only set/changed once.

PR #158, opened 2017-01-10
Remove Initialize method from Module

madhuravi:
We use New methods to initialize modules. The Initialize method is a Noop for existing modules and appears redundant. Removed the Initialize method.

### DIFF
--- a/service/stub_module.go
+++ b/service/stub_module.go
@@ -33,14 +33,8 @@ type StubModule struct {
 var _ Module = &StubModule{}
 
 // NewStubModule generates a Module for use in testing
-func NewStubModule() *StubModule {
-	return &StubModule{}
-}
-
-// Initialize fakes an init call on the module
-func (s *StubModule) Initialize(host Host) error {
-	s.Host = host
-	return s.InitError
+func NewStubModule(host Host) *StubModule {
+	return &StubModule{Host: host}
 }
 
 // Start mimics startup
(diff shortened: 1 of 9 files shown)


madhuravi:
@uber-go/service-framework

service/stub_module.go
```diff
@@ -33,14 +33,8 @@ type StubModule struct {
...
-func NewStubModule() *StubModule {
-	return &StubModule{}
-}
-
-// Initialize fakes an init call on the module
-func (s *StubModule) Initialize(host Host) error {
-	s.Host = host
-	return s.InitError
+func NewStubModule(host Host) *StubModule {
+	return &StubModule{Host: host}
```
glibsm: Hm. Now that I'm reading this, is the host fully finished initializing by the time we add these modules?

sectioneight: That is the conversation we had over chat yesterday. I expressed concerned around bootstrapping, where the service Builder is instantiating modules (with New) before the service owner has time to necessarily configure all the components. This is primarily a concern for user-defined modules (e.g. modules that don't live in this repository).

`AddModules` is called during `Build()` which is generally the first line of a user's `main` func. Between Build() and Start() they may hydrate more objects, connect to databases, etc.

@madhuravi wanted to remove it for now, and re-add it if we run into cases where we need a separate initialization phase.

madhuravi: To answer the question, today the host is always initialized by the time we add these modules. We don't have use cases today for where it will not be. When it comes up, we can revisit and see if Initialize is the solution or we should handle it some other way.

PR #160, opened 2017-01-11
Fix auth setup

anuptalwalkar:


sectioneight:
Seems like a test is in order?

modules/uhttp/client/client_filters.go
```diff
@@ -88,13 +88,21 @@ func authenticationFilter(
...
+	authctx := authClient.SetAttribute(ctx, auth.ServiceAuth, _serviceName)
 
-	authctx, err := authClient.Authenticate(ctx)
+	authctx, err = authClient.Authenticate(authctx)
 	if err != nil {
 		ctx.Logger().Error(auth.ErrAuthentication, "error", err)
 		return nil, err
 	}
+
+	span := opentracing.SpanFromContext(authctx)
```
madhuravi: We should just do this once after all the filters. You can make a spanInjectFilter that just does this and execute tracingFilter, authFilter, spanInjectFilter. So neither tracing nor auth needs to inject it.

madhuravi: Spoke offline, maybe best to refactor to a function instead of a third filter since they are so tightly coupled.

modules/uhttp/client/client_filters.go
```diff
@@ -88,16 +85,32 @@ func authenticationFilter(
...
+	if err := injectHeadersIntoSpanContext(req.Header, span); err != nil {
+		ctx.Logger().Error("Error injecting auth context", "error", err)
+		return nil, err
+	}
+
 	return next.Do(&fxcontext.Context{Context: authctx}, req)
 }
 
+func injectHeadersIntoSpanContext(header http.Header, span opentracing.Span) error {
```
madhuravi: It's the other way, injectSpanIntoHeaders

anuptalwalkar: oh..  yes.

modules/uhttp/client/client_filters_test.go
```diff
@@ -67,6 +74,47 @@ func TestExecutionChainFiltersError(t *testing.T) {
...
 }
 
+func withOpentracingSetup(t *testing.T, fn func(tracer opentracing.Tracer)) {
+	tracer, closer, err := tracing.InitGlobalTracer(&jconfig.Configuration{}, "Test", ulog.NoopLogger, tally.NullStatsReporter)
+	defer closer.Close()
+	assert.NotNil(t, closer)
+	require.NoError(t, err)
+
+	_serviceName = "test_service"
+	auth.UnregisterClient()
```
madhuravi: Redundant? Wouldn't SetupClient over-write it.

modules/uhttp/client/client_filters_test.go
```diff
@@ -67,6 +74,47 @@ func TestExecutionChainFiltersError(t *testing.T) {
...
+		assert.NoError(t, err)
+		assert.Equal(t, _respOK, resp)
+	})
+}
+
+func getContextPropogationClient(t *testing.T) BasicClient {
+	return BasicClientFunc(
+		func(ctx fx.Context, req *http.Request) (resp *http.Response, err error) {
+			span := opentracing.SpanFromContext(ctx)
+			assert.NotNil(t, span)
```
madhuravi: why are all these asserts necessary? Wouldnt the authFilter fail if this were the case?

anuptalwalkar: This assert is just making sure that span is never nil. In case if we use getContextPropogationClient in other tests. second one is just verifying that baggage is present.

modules/uhttp/client/client_filters_test.go
```diff
@@ -67,6 +74,45 @@ func TestExecutionChainFiltersError(t *testing.T) {
 	assert.Nil(t, resp)
 }
 
+func withOpentracingSetup(t *testing.T, fn func(tracer opentracing.Tracer)) {
+	tracer, closer, err := tracing.InitGlobalTracer(&jconfig.Configuration{}, "Test", ulog.NoopLogger, tally.NullStatsReporter)
+	defer closer.Close()
+	assert.NotNil(t, closer)
+	require.NoError(t, err)
```
madhuravi: defer tracing.InitGlobalTracer(tracing.NoopTracer)

madhuravi:
We probably need more tests to get the coverage up.

modules/uhttp/client/client_filters_test.go
anuptalwalkar: I need to set/unset clients for different types of tests

PR #159, opened 2017-01-11
Move http docs to README

madhuravi:

PR #162, opened 2017-01-11
Upgrade yarpc to 1.0.0

glibsm:


examples/keyvalue/gen.go
```diff
@@ -20,4 +20,4 @@
 
 package keyvalue
 
-//go:generate thriftrw --plugin=fx kv.thrift
+//go:generate thriftrw "--plugin=yarpc --context-import-path=go.uber.org/fx --unary-handler-wrapper=go.uber.org/fx/modules/rpc.WrapUnary --oneway-handler-wrapper=go.uber.org/fx/modules/rpc.WrapOneway" kv.thrift
```
anuptalwalkar: @abhinav We should document quotations with go generate in thriftrw-go/plugin. Currently it works on having double quotes around entire --plugin parameter

modules/rpc/handler.go
```diff
@@ -28,56 +28,55 @@ import (
...
 
 // Handle calls the caller HandlerFunc.
-func (f UnaryHandlerFunc) Handle(ctx fx.Context, reqMeta yarpc.ReqMeta, body wire.Value) (thrift.Response, error) {
-	return f(ctx, reqMeta, body)
+func (f UnaryHandlerFunc) Handle(ctx fx.Context, body wire.Value) (thrift.Response, error) {
+	return f(ctx, body)
 }
 
 // WrapUnary wraps the unary handler and returns a thrift.UnaryHandlerFunc for yarpc calls
 // TODO(anup): GFM-255 Remove host parameter once updated yarpc plugin is imported in the repo
```
sectioneight: is this comment still valid?

modules/rpc/yarpc.go
```diff
@@ -128,20 +128,23 @@ func (m *YarpcModule) Start(readyCh chan<- struct{}) <-chan error {
...
 
-	channel, err := tchannel.NewChannel(m.config.AdvertiseName, nil)
+	interceptor := yarpc.UnaryInboundMiddleware(m.inboundMiddlewares...)
+	onewayInterceptor := yarpc.OnewayInboundMiddleware(m.onewayInboundMiddlewares...)
+
+	tchTransport, err := tch.NewChannelTransport(
+		tch.ServiceName(m.config.AdvertiseName),
+		tch.ListenAddr(m.config.Bind),
+	)
 	ret := make(chan error, 1)
```
sectioneight: unrelated nit (i think this is my code anyway): move this above line 134 so the err check is right after we potentially get the error

modules/uhttp/doc.go
```diff
@@ -30,7 +30,6 @@
 //     "io"
 //     "net/http"
 //
-//     "go.uber.org/fx"
```
anuptalwalkar: Remove this file.. Looks like these are remains from https://github.com/uber-go/fx/pull/159

modules/rpc/yarpc_options.go
```diff
@@ -52,22 +52,22 @@ func WithOnewayInboundMiddleware(i ...transport.OnewayInboundMiddleware) modules
...
 }
 
-func inboundMiddlewaresFromCreateInfo(mci service.ModuleCreateInfo) []transport.UnaryInboundMiddleware {
+func inboundMiddlewaresFromCreateInfo(mci service.ModuleCreateInfo) []middleware.UnaryInbound {
 	items, ok := mci.Items[_interceptorKey]
 	if !ok {
 		return nil
 	}
 
 	// Intentionally panic if programmer adds non-interceptor slice to the data
```
sectioneight: can we update this comment to not mention "interceptor" since that language was apparently changed in yarpc?

modules/uhttp/doc.go
```diff
@@ -30,7 +30,6 @@
 //     "io"
```
sectioneight: Did you mean to commit this? You may need to rebase, I think @madhuravi just changed this

modules/rpc/handler.go
anuptalwalkar: Yes, this is fixed with the plugin update. We can remove/cleanup GFM-255 in remaining places.

glibsm: I'd like to leave that for @anuptalwalkar in another PR. Don't want to conflate many things together, there is some work needed to remove all those comments

glide.yaml
```diff
@@ -11,7 +11,7 @@ import:
 - package: github.com/gorilla/context
   version: ^1.1.0
 - package: go.uber.org/yarpc
-  version: v1.0.0-rc1
+  version: ^v1.0.0
```
breerly: This can be just `^1`

PR #163, opened 2017-01-12
Export auth NoopClient for downstream use

anuptalwalkar:


auth/auth_stub.go
```diff
@@ -27,6 +27,9 @@ var _ Client = &noop{}
 type noop struct {
 }
 
+// NoopClient is used for testing and no-op integration
+var NoopClient = noopClient(nil)
```
sectioneight: Maybe make a group with line 25?

```go
var (
  _ Client = &noop{}

  NoopClient = noopClient(nil)
)
```

auth/auth_stub.go
```diff
@@ -22,7 +22,12 @@ package auth
 
 import "context"
 
-var _ Client = &noop{}
+var (
+	// NoopClient is used for testing and no-op integration
+	NoopClient = noopClient(nil)
```
glibsm: yarpc does `Nop`, we do `Noop`, and we also have `NullHost`....

How do we consolidate?

sectioneight:
I think Nop is the correct term from the stdlib (e.g. ioutil.NopCloser). We
should clean up.

auth/auth_stub.go
madhuravi: +1 This has been bothering me as well. We need to decide on a convention for uberfx atleast.

anuptalwalkar: Nullop? jk, consolidating to Nop sounds good to me.

anuptalwalkar: https://github.com/uber-go/fx/pull/164

PR #165, opened 2017-01-12
Cleanup and simplify auth.Client initialization

anuptalwalkar:
auth.Instance() global instance is harder to manage in code, and more so in tests. 
I removed any instance of auth.Instance() usage, and pull `auth.Client` from the service host itself. 
For tests, we can now use `host.NopHost()` for simple nop auth, and `host.NopHostAuthFailure()` to invoke auth failures.

### DIFF
--- a/modules/uhttp/client/client.go
+++ b/modules/uhttp/client/client.go
@@ -61,6 +61,7 @@ func New(info auth.CreateAuthInfo, client *http.Client, filters ...Filter) *Clie
 func (c *Client) Do(ctx fx.Context, req *http.Request) (resp *http.Response, err error) {
 	filters := c.filters
 	if c.defaultFiltersAdded == false {
+		// TODO(anup): GFM-289 Update uhttp client to not return exported struct
 		filters = append(filters, tracingFilter(), authenticationFilter(c.info))
 		c.defaultFiltersAdded = true
 	}
--- a/modules/uhttp/client/client.go
+++ b/modules/uhttp/client/client.go
@@ -27,6 +27,7 @@ import (
 	"strings"
 
 	"go.uber.org/fx"
+	"go.uber.org/fx/auth"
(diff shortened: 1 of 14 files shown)


anuptalwalkar:
@uber-go/service-framework

modules/uhttp/client/client.go
```diff
@@ -39,16 +40,18 @@ var (
 // Client wraps around a http client
 type Client struct {
 	*http.Client
+	info                auth.CreateAuthInfo
```
alsamylkin: can we save it in the auth filter?

anuptalwalkar: If we save it in auth filter, we will not have access to it if Client is created independently (without using new) Few lines below in Do, we inject default filters for this edge case.

modules/uhttp/client/client.go
```diff
@@ -58,7 +61,7 @@ func New(cfg config.Provider, client *http.Client, filters ...Filter) *Client {
 func (c *Client) Do(ctx fx.Context, req *http.Request) (resp *http.Response, err error) {
 	filters := c.filters
 	if c.defaultFiltersAdded == false {
-		filters = append(filters, FilterFunc(tracingFilter), FilterFunc(authenticationFilter))
+		filters = append(filters, tracingFilter(), authenticationFilter(c.info))
```
alsamylkin: I think this line adds an edge case with infinite number of filters :)

modules/uhttp/client/client.go
alsamylkin: does Client have to be public/struct? What other options do we have?

anuptalwalkar: Yes, (sadly) otherwise http client will not be accessible outside for use @madhuravi

anuptalwalkar: GFM-289

PR #166, opened 2017-01-12
Add test env and move env helper to be usable everywhere

glibsm:


config/config.go
```diff
@@ -101,11 +102,19 @@ func Environment() string {
...
+// IsDevelopmentEnv returns true if the current environment is set to development
+func IsDevelopmentEnv() bool {
 	return strings.Contains(Environment(), _devEnv)
 }
 
+// IsTestEnv returns true if the current environment is set to test
+//
+// Useful, for example, to enable alternate functionality of some components
+// during testing that might normally require network connectivity or a database
+func IsTestEnv() bool {
```
alsamylkin: There could be different environments with different purposes, e.g. staging, internal, etc. Shouldn't test be controlled by the service config as well?

glibsm: Based on the earlier discussion, I'll make some changes to this diff and remove this functionality.

config/config.go
```diff
@@ -51,7 +51,8 @@ var (
 )
 
 var (
-	_devEnv = "development"
+	_devEnv  = "development"
+	_testEnv = "test"
```
sectioneight: unused?

glibsm: interesting. did go compiler sneeze while processing this line? I thought unused vars were an error

sectioneight:
Unused globals aren't errors :(

PR #168, opened 2017-01-14
Hold on to http.Server in prep for graceful shutdown

sectioneight:
In Go 1.8, we'll be able to call Shutdown(ctx) on the *http.Server
object.

Also fix a minor issue in error handling.

### DIFF
--- a/modules/uhttp/http.go
+++ b/modules/uhttp/http.go
@@ -77,6 +77,7 @@ type Module struct {
 	config   Config
 	log      ulog.Log
 	mux      *http.ServeMux
+	srv      *http.Server
 	router   *Router
 	listener net.Listener
 	handlers []RouteHandler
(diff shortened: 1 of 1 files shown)


modules/uhttp/http.go
```diff
@@ -189,14 +190,19 @@ func (m *Module) Start(ready chan<- struct{}) <-chan error {
...
 	m.listenMu.Unlock()
 
 	go func() {
 		listener := m.accessListener()
 		ready <- struct{}{}
-		ret <- nil
-		if err := http.Serve(listener, m.mux); err != nil {
+		if err := m.srv.Serve(listener); err != nil {
 			m.log.Error("HTTP Serve error", "error", err)
+			ret <- err
```
alsamylkin: Add test for it?

modules/uhttp/http.go
```diff
@@ -189,14 +190,19 @@ func (m *Module) Start(ready chan<- struct{}) <-chan error {
...
+		} else {
+			ret <- nil
```
alsamylkin: Remove else and simply ret<-err?

modules/uhttp/http.go
sectioneight: I tried, but it's really hard to induce an error now. I could artificially close the `listener` but it wouldn't really be testing a code path that would be hit in production, since we don't expose a mutator to the listener.

modules/uhttp/http.go
sectioneight: Consolidated, good call.

PR #169, opened 2017-01-17
Use custom RoundTripper interface for http.Client

alsamylkin:
uhttp client is implemented as a custom struct right now and users have to to create it with New, provide an http.Client, there are also some thread safety issues.

http.Client is actually a high level facade and uses Transport to send requests, so we can provide a custom implementation of http.RoundTripper interface and filter requests before it will be send by an actual transport.

### DIFF
--- a/modules/uhttp/client/client.go
+++ b/modules/uhttp/client/client.go
@@ -21,20 +21,20 @@
 package client
 
 import (
-	"io"
 	"net/http"
-	"net/url"
-	"strings"
 
 	"go.uber.org/fx"
 	"go.uber.org/fx/auth"
 	"go.uber.org/fx/config"
 
 	"golang.org/x/net/context/ctxhttp"
+	"sync"
 )
 
 var (
-	_serviceName string
+	_serviceName   string
+	_uFilters      sync.Mutex
+	_filterTripper filterTripper
 )
 
 // Client wraps around a http client
@@ -43,6 +43,27 @@ type Client struct {
 	info                auth.CreateAuthInfo
 	filters             []Filter
 	defaultFiltersAdded bool
+}
+
+type filterTripper struct {
+	oldTripper http.RoundTripper
+	chain      executionChain
+}
+
+// UpdateDefaultFilters updates filters for the default RoundTripper
+func UpdateDefaultFilters(filters ...Filter) {
+	_uFilters.Lock()
+	_filterTripper.chain = newExecutionChain(filters, _filterTripper.oldTripper)
+	_uFilters.Unlock()
+}
+
+func (t *filterTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+
+	return t.oldTripper.RoundTrip(req)
+}
+
+func init() {
+	http.DefaultTransport = &filterTripper{oldTripper: http.DefaultTransport}
 }
 
 // New creates a new instance of uhttp Client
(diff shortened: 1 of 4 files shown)


modules/uhttp/client/client.go
```diff
@@ -21,20 +21,20 @@
...
 	"net/http"
-	"net/url"
-	"strings"
 
 	"go.uber.org/fx"
 	"go.uber.org/fx/auth"
 	"go.uber.org/fx/config"
 
 	"golang.org/x/net/context/ctxhttp"
+	"sync"
```
sectioneight: should be at the top, stdlib import

modules/uhttp/client/client.go
```diff
@@ -45,6 +45,27 @@ type Client struct {
...
+type filterTripper struct {
+	oldTripper http.RoundTripper
+	chain      executionChain
+}
+
+// UpdateDefaultFilters updates filters for the default RoundTripper
+func UpdateDefaultFilters(filters ...Filter) {
+	_uFilters.Lock()
+	_filterTripper.chain = newExecutionChain(filters, _filterTripper.oldTripper)
+	_uFilters.Unlock()
```
anuptalwalkar: do `defer _uFilters.Unlock()` on line 55

modules/uhttp/client/client.go
```diff
@@ -45,6 +45,27 @@ type Client struct {
...
+}
+
+func (t *filterTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+
```
anuptalwalkar: nit: extra new line

modules/uhttp/client/client.go
```diff
@@ -57,79 +78,14 @@ func New(info auth.CreateAuthInfo, client *http.Client, filters ...Filter) *Clie
...
 type BasicClientFunc func(ctx fx.Context, req *http.Request) (resp *http.Response, err error)
 
-// Do implements Do from the BasicClient interface
-func (f BasicClientFunc) Do(
-	ctx fx.Context, req *http.Request,
-) (resp *http.Response, err error) {
-	return f(ctx, req)
+// RoundTrip implements RoundTrip from the http.RoundTripper interface
+func (f BasicClientFunc) RoundTrip(req *http.Request) (resp *http.Response, err error) {
+	return f(req.Context().(fx.Context), req)
```
anuptalwalkar: Can we do type checking before typecasting? Or you can simply do - 
```
ctx := fxcontext.Context {
  Context: req.Context(),
}
```

modules/uhttp/client/client_filters.go
```diff
@@ -27,30 +27,31 @@ import (
 	"go.uber.org/fx/auth"
 	"go.uber.org/fx/internal/fxcontext"
 
+	"fmt"
```
anuptalwalkar: stdlib import at the top

modules/uhttp/client/client.go
```diff
@@ -45,6 +45,27 @@ type Client struct {
 	defaultFiltersAdded bool
 }
 
+type filterTripper struct {
+	oldTripper http.RoundTripper
+	chain      executionChain
+}
+
+// UpdateDefaultFilters updates filters for the default RoundTripper
+func UpdateDefaultFilters(filters ...Filter) {
```
anuptalwalkar: Where is this used? Tests?

modules/uhttp/client/client.go
```diff
@@ -45,6 +45,27 @@ type Client struct {
 	defaultFiltersAdded bool
 }
 
+type filterTripper struct {
+	oldTripper http.RoundTripper
```
anuptalwalkar: what is oldTripper?

modules/uhttp/client/client_filters.go
```diff
@@ -123,16 +125,20 @@ func newExecutionChain(
...
+	finalClient   http.RoundTripper
 }
 
-func (ec executionChain) Do(
-	ctx fx.Context, req *http.Request,
-) (resp *http.Response, err error) {
+func (ec *executionChain) RoundTrip(req *http.Request) (resp *http.Response, err error) {
 	if ec.currentFilter < len(ec.filters) {
 		filter := ec.filters[ec.currentFilter]
 		ec.currentFilter++
```
anuptalwalkar: Oh wait, instead of fmt.Errorf, return errors.New("")

modules/uhttp/client/client_filters.go
```diff
@@ -123,16 +125,20 @@ func newExecutionChain(
...
-		return filter.Apply(ctx, req, ec)
+
+		if ctx, ok := req.Context().(fx.Context); ok {
+			return filter.Apply(ctx, req, ec)
+		}
+
+		return nil, fmt.Errorf("Expected fx.Context, but received: %T", req.Context())
```
anuptalwalkar: instead of fmt.Errorf, return errors.New("")

modules/uhttp/client/client.go
alsamylkin: Why? newExecutionChain doesn't panic..

modules/uhttp/client/client_filters.go
alsamylkin: I need the type of the req.Context()...

alsamylkin:
This is actually a preview, I am looking for a feedback on this approach.
Do we really need a custom http client?

modules/uhttp/client/client.go
```diff
@@ -21,115 +21,53 @@
...
-	return &Client{
-		Client:              client,
-		info:                info,
-		filters:             filters,
-		defaultFiltersAdded: true,
+// New creates an http.Client that includes 2 extra filters: tracing and auth
+// they are going to be applied in following order: tracing, auth, remaining filters
+// and only if all of them passed the request is going to be send.
+func New(info auth.CreateAuthInfo, filters ...Filter) *http.Client {
+	extra := make([]Filter, len(filters)+2)
```
alsamylkin: This is super ugly, but I don't know if there is a nicer solution :(

madhuravi: why not just filters = append(filters, tracing, auth)

alsamylkin: Filter order matters, we want to have tracing first, then auth, and user filters next.

madhuravi: How about -
defaultFilters = []Filter{tracingFilter(), authenticationFilter(info)}
filters = append(defaultFilters, filters)

modules/uhttp/client/client_filters.go
```diff
@@ -25,32 +25,48 @@ import (
...
 	"github.com/opentracing/opentracing-go"
 	"github.com/opentracing/opentracing-go/ext"
 )
 
-// Filter applies filters on client requests, request contexts such as
-// adding tracing to the context
+// Sender sends the http request. It is a simple wrapper around http.RoundTripper, so
+// Send must be safe to use by multiple go routines
+type Sender interface {
+	Send(ctx fx.Context, r *http.Request) (resp *http.Response, err error)
```
madhuravi: I prefer Do instead of Send, it matches the http naming. Less mental overload on engineers. What do you think?

modules/uhttp/client/client.go
```diff
@@ -21,115 +21,53 @@
...
-
-// The BasicClientFunc type is an adapter to allow the use of ordinary functions as BasicClient.
-type BasicClientFunc func(ctx fx.Context, req *http.Request) (resp *http.Response, err error)
-
-// Do implements Do from the BasicClient interface
-func (f BasicClientFunc) Do(
-	ctx fx.Context, req *http.Request,
-) (resp *http.Response, err error) {
-	return f(ctx, req)
+func (ec executionChain) RoundTrip(r *http.Request) (resp *http.Response, err error) {
```
madhuravi: Why do we need a RoundTrip method on executionChain? Better to stick to the naming from Sender interface. Only execChain finalTransport will invoke the RoundTrip function. Everywhere else, we can have Send or Do

alsamylkin: Because http.Client expects the http.RoundTripper interface and executionChain is our implementation that we provide.

modules/uhttp/client/client_filters.go
alsamylkin: Talked offline and use Execute for this review for now. We will, probably, rename it later anyway.

madhuravi:
Looks good to me! Can you add more tests to bump up the coverage? I see some Execute and error conditions in client_filters are bringing the coverage down.

alsamylkin:
I removed lots of code, that's why our coverage decreased :)
Created ticket GFM-305 to improve it for tracing.

modules/uhttp/client/client.go
```diff
@@ -21,115 +21,56 @@
...
-		c.defaultFiltersAdded = true
+// New creates an http.Client that includes 2 extra filters: tracing and auth
+// they are going to be applied in following order: tracing, auth, remaining filters
+// and only if all of them passed the request is going to be send.
+func New(info auth.CreateAuthInfo, filters ...Filter) *http.Client {
+	defaultFilters := []Filter{tracingFilter(), authenticationFilter(info)}
+	defaultFilters = append(defaultFilters, filters...)
+	return &http.Client{
+		Transport: newExecutionChain(defaultFilters, http.DefaultTransport),
+		Timeout:   2 * time.Minute,
```
anuptalwalkar: Can we make this configurable? I know 2 minutes is a long time, but needs user modifiable.

alsamylkin: User still can modify it, client.Timeout. 
I just didn't want to use the default infinite timeout from http.Client

PR #170, opened 2017-01-18
Allow user filter registration and setup on http

anuptalwalkar:
GFM-75

This PR is for Bulk user filter registration for http on top of defaultFilters defined by the framework.

### DIFF
--- a/examples/simple/handlers.go
+++ b/examples/simple/handlers.go
@@ -21,6 +21,7 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"io"
 	"net/http"
@@ -45,5 +46,16 @@ func registerHTTPers(service service.Host) []uhttp.RouteHandler {
 	handler := &exampleHandler{}
 	return []uhttp.RouteHandler{
 		uhttp.NewRouteHandler("/", handler),
+	}
+}
+
+func simpleFilter() uhttp.FilterFunc {
+	return func(ctx context.Context, w http.ResponseWriter, r *http.Request, next uhttp.Handler) {
+		fxctx, ok := ctx.(fx.Context)
+		if ok {
+			next.ServeHTTP(fxctx, w, r)
+		} else {
+			uhttp.Wrap(service.NopHost(), next).ServeHTTP(w, r)
+		}
 	}
 }
--- a/examples/simple/handlers.go
+++ b/examples/simple/handlers.go
@@ -21,7 +21,6 @@
 package main
 
 import (
-	"context"
 	"fmt"
 	"io"
 	"net/http"
@@ -49,13 +48,10 @@ func registerHTTPers(service service.Host) []uhttp.RouteHandler {
 	}
 }
 
-func simpleFilter() uhttp.FilterFunc {
-	return func(ctx context.Context, w http.ResponseWriter, r *http.Request, next uhttp.Handler) {
-		fxctx, ok := ctx.(fx.Context)
-		if ok {
-			next.ServeHTTP(fxctx, w, r)
-		} else {
-			uhttp.Wrap(service.NopHost(), next).ServeHTTP(w, r)
-		}
-	}
+type simpleFilter struct {
+}
+
+func (simpleFilter) Apply(ctx fx.Context, w http.ResponseWriter, r *http.Request, next uhttp.Handler) {
+	io.WriteString(w, "Going through simpleFilter")
+	next.ServeHTTP(ctx, w, r)
 }
--- a/modules/uhttp/filterchain_builder.go
+++ b/modules/uhttp/filterchain_builder.go
@@ -54,7 +54,7 @@ func (fc filterChain) ServeHTTP(ctx fx.Context, w http.ResponseWriter, r *http.R
 type FilterChainBuilder interface {
 	// AddFilter is used to add the next filter to the chain during construction time.
 	// The calls to AddFilter can be chained.
-	AddFilter(filter ...Filter) FilterChainBuilder
+	AddFilters(filter ...Filter) FilterChainBuilder
 
 	// Build creates an immutable FilterChain.
 	Build(finalHandler Handler) filterChain
@@ -69,9 +69,7 @@ type filterChainBuilder struct {
 
 func defaultFilterChainBuilder(host service.Host) FilterChainBuilder {
 	fcb := NewFilterChainBuilder(host)
-	return fcb.AddFilter(contextFilter(host)).
-		AddFilter(tracingServerFilter(host)).
-		AddFilter(authorizationFilter(host))
+	return fcb.AddFilters(contextFilter(host), tracingServerFilter(host), authorizationFilter(host))
 }
 
 // NewFilterChainBuilder creates an empty filterChainBuilder for setup
@@ -81,15 +79,15 @@ func NewFilterChainBuilder(host service.Host) FilterChainBuilder {
 	}
 }
 
-func (f filterChainBuilder) AddFilter(filters ...Filter) FilterChainBuilder {
+func (f filterChainBuilder) AddFilters(filters ...Filter) FilterChainBuilder {
 	for _, filter := range filters {
 		f.filters = append(f.filters, filter)
 	}
-	f.filters = append(f.filters, panicFilter(f.Host))
 	return f
 }
 
 func (f filterChainBuilder) Build(finalHandler Handler) filterChain {
+	f.filters = append(f.filters, panicFilter(f.Host))
 	return filterChain{
 		filters:      f.filters,
 		finalHandler: finalHandler,
--- a/modules/uhttp/filterchain_builder.go
+++ b/modules/uhttp/filterchain_builder.go
@@ -33,13 +33,6 @@ type filterChain struct {
 	filters       []Filter
 }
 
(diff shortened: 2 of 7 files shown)


examples/simple/handlers.go
```diff
@@ -47,3 +48,14 @@ func registerHTTPers(service service.Host) []uhttp.RouteHandler {
 		uhttp.NewRouteHandler("/", handler),
 	}
 }
+
+func simpleFilter() uhttp.FilterFunc {
```
sectioneight: why wrap this in a closure? in other words, why doesn't `simpleFilter` take ctx, w, r, next

modules/uhttp/filterchain_builder.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+}
+
+func (f filterChainBuilder) AddFilter(filter Filter) FilterChainBuilder {
+	f.filters = append(f.filters, filter)
+	return f
+}
+
+func (f filterChainBuilder) Build(finalHandler Handler) filterChain {
+	fc := filterChain{}
```
alsamylkin: can we simply write
`fc.filters = f.filters`?
If one is going to call AddFilter after Build, lets `copy()` then.

modules/uhttp/http.go
```diff
@@ -79,7 +79,7 @@ type Module struct {
 	listener net.Listener
 	handlers []RouteHandler
 	listenMu sync.RWMutex
-	filters  []Filter
+	fcb      FilterChainBuilder
```
alsamylkin: why do we need fcb as a field? We can use a local variable in New..

modules/uhttp/http.go
```diff
@@ -96,9 +96,9 @@ type Config struct {
 type GetHandlersFunc func(service service.Host) []RouteHandler
 
 // New returns a new HTTP module
-func New(hookup GetHandlersFunc, options ...modules.Option) service.ModuleCreateFunc {
+func New(hookup GetHandlersFunc, filters []Filter, options ...modules.Option) service.ModuleCreateFunc {
```
alsamylkin: Should filters be Options?

modules/uhttp/filterchain_builder.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// NewFilterChainBuilder creates an empty filterChainBuilder for setup
+func NewFilterChainBuilder(host service.Host) FilterChainBuilder {
+	return &filterChainBuilder{
+		Host: host,
+	}
+}
+
+func (f filterChainBuilder) AddFilter(filter Filter) FilterChainBuilder {
```
alsamylkin: Could you write a variadic version?
`AddFilter(filters... Filter)` so one will not write `.AddFilter().AddFilter()...`

examples/simple/handlers.go
anuptalwalkar: good point. I can implement uhttp.Filter with Apply(ctx, w, r, next) method. That might be better example.

modules/uhttp/http.go
anuptalwalkar: The call wrapping happens in Start, which is where filter chain is getting built.

modules/uhttp/filterchain_builder.go
```diff
@@ -0,0 +1,97 @@
...
+	f.filters = append(f.filters, filter)
```
alsamylkin: I think the order matters here:
e.g. if we'll add a filter after panic filter and that filter panics, then our code will panic.
Either add in the beginning of the slice or reverse the order of filter calls..

modules/uhttp/http.go
anuptalwalkar: hmm. Filters are bound to uhttp, not modules, having it as parameter makes sense here.

modules/uhttp/filterchain_builder.go
anuptalwalkar: I agree. We have filters added in order (for example, authFilter needs tracing, and context, etc). I think it makes sense to add panicFilter at the end of all custom filters. So the order will be context-tracing-auth-user_filters-panic filter

anuptalwalkar: That was a good catch btw, we don't want service to panic on panics in user filters:)

modules/uhttp/filterchain_builder.go
```diff
@@ -0,0 +1,97 @@
...
+type filterChainBuilder struct {
+	service.Host
+
+	finalHandler Handler
+	filters      []Filter
+}
+
+func defaultFilterChainBuilder(host service.Host) FilterChainBuilder {
+	fcb := NewFilterChainBuilder(host)
+	return fcb.AddFilter(contextFilter(host)).
```
alsamylkin: Use a single call to AddFilters?

modules/uhttp/filterchain_builder.go
```diff
@@ -0,0 +1,97 @@
...
+	return &filterChainBuilder{
+		Host: host,
+	}
+}
+
+func (f filterChainBuilder) AddFilter(filters ...Filter) FilterChainBuilder {
+	for _, filter := range filters {
+		f.filters = append(f.filters, filter)
+	}
+	f.filters = append(f.filters, panicFilter(f.Host))
```
alsamylkin: So if someone calls AddFilter multiple times, we are going to have multiple panic filters in the chain?
Why not to revers order processing/add it in the Build?

modules/uhttp/http.go
alsamylkin: But it is not used anywhere else, so why keep it?

modules/uhttp/filterchain_builder.go
anuptalwalkar: Moved it to build. I explained earlier that reverse order processing is not the right way.

glibsm:
Do we want to talk about terminology consistency here, or do we not want to bring that up again? In other words, `yarpc` has changed the terminology from `Filter` to `InboundMiddleware` (Unary/Oneway flavor).

glibsm:
Yarpc changelog link for reference: https://github.com/yarpc/yarpc-go/blob/dev/CHANGELOG.md#v100-rc1-2016-11-23

anuptalwalkar:
@glibsm HTTP Filters and server filterChain is a pretty common terminology across multiple languages, I don't see why we want to introduce new term here. I see InboundMiddleware naming specific to YARPC, but I am happy to discuss the naming scheme.

modules/uhttp/filterchain_builder.go
alsamylkin: I meant to reverse the addition too and process from the end of the chain.

anuptalwalkar: Ah, I see. Any issue with its current order? I like to keep appends simple, and make it easy to understand.

alsamylkin: Well, calling Build() twice produces 2 panic filters..
Personally it doesn't feel right.

modules/uhttp/filters_test.go
```diff
@@ -76,24 +76,23 @@ func TestTracingFilterWithLogs(t *testing.T) {
...
 func TestFilterChainFilters_AuthFailure(t *testing.T) {
 	host := service.NopHostAuthFailure()
-	chain := newFilterChain([]Filter{
+	chain := NewFilterChainBuilder(host).AddFilters(
 		contextFilter(host),
 		tracingServerFilter(host),
 		authorizationFilter(host),
-		panicFilter(host)},
-		getNopHandler(host))
+		panicFilter(host)).Build(getNopHandler(host))
```
alsamylkin: Here we have 2 panic filters :)

anuptalwalkar: uh :( tests.

modules/uhttp/filterchain_builder.go
```diff
@@ -0,0 +1,94 @@
...
+		fc.finalHandler.ServeHTTP(ctx, w, r)
+	} else {
+		filter := fc.filters[fc.currentFilter]
+		fc.currentFilter++
+		filter.Apply(ctx, w, r, fc)
+	}
+}
+
+// FilterChainBuilder builds a filterChain object with added filters
+type FilterChainBuilder interface {
```
alsamylkin: Why do we need a public interface?

anuptalwalkar: Will remove. I was thinking if this can be used by users to build user filters.

modules/uhttp/filterchain_builder.go
```diff
@@ -0,0 +1,94 @@
...
+type filterChainBuilder struct {
+	service.Host
+
+	finalHandler Handler
+	filters      []Filter
+}
+
+func defaultFilterChainBuilder(host service.Host) FilterChainBuilder {
+	fcb := NewFilterChainBuilder(host)
+	return fcb.AddFilters(contextFilter(host), tracingServerFilter(host), authorizationFilter(host), panicFilter(host))
```
alsamylkin: I thought that panicFilter should be always either first or last and here custom filters can be added after it, so if e.g. tracing filter panics we'll panic too..

anuptalwalkar: good point. Moved it after context filter.

PR #172, opened 2017-01-18
Add log line after jaeger span finish for debugging

madhuravi:


modules/uhttp/filters.go
```diff
@@ -70,7 +70,10 @@ func tracingServerFilter(host service.Host) FilterFunc {
 		}
 		span := opentracing.GlobalTracer().StartSpan(operationName, ext.RPCServerOption(spanCtx))
 		ext.HTTPUrl.Set(span, r.URL.String())
-		defer span.Finish()
+		defer func() {
+			span.Finish()
+			fxctx.Logger().Info("Span finished")
```
alsamylkin: Add info which span is finished?

modules/uhttp/filters_test.go
```diff
@@ -67,10 +67,13 @@ func TestTracingFilterWithLogs(t *testing.T) {
 		response := testServeHTTP(chain, host)
 		assert.Contains(t, response.Body.String(), "filters ok")
 		assert.True(t, len(buf.Lines()) > 0)
+		var spanFinished bool
 		for _, line := range buf.Lines() {
 			assert.Contains(t, line, "traceID")
 			assert.Contains(t, line, "spanID")
+			spanFinished = strings.Contains(line, "Span finished")
```
alsamylkin: We want all spans to be finished, so assert should be moved here?

glibsm:
If it's for debugging, why is it an `info` log and not `debug`?

modules/uhttp/filters.go
madhuravi: That information is already added to the logger by the time defer is called. We add trace id and span id to the logger.

modules/uhttp/filters_test.go
madhuravi: Yes but it's not in all of the log lines emitted ;)

alsamylkin: Hmm, so we are checking the last line only? This seems to be an overkill then :)

madhuravi: It's minor redundancy in a test Alex :P However the reason is we can't rely on the buf lines order, that would make the test fragile if it's not exactly in the last line and an extra log was added somewhere else.

alsamylkin: Well, if it will not be in the last line the test output, the test will fail :)

madhuravi: True! Changed to use logical or.

PR #174, opened 2017-01-19
Allow int and floats when config values are fetched as float64

madhuravi:
Wish there was a better way than switch statements but oh well.. Copying this format from TryAsInt

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -206,6 +206,22 @@ func TestGetAsIntegerValue(t *testing.T) {
 	}
 }
 
+func TestGetAsFloatValue(t *testing.T) {
+	testCases := []struct {
+		value interface{}
+	}{
+		{float32(2)},
+		{float64(2)},
+		{int(2)},
+		{int32(2)},
+		{int64(2)},
+	}
+	for _, tc := range testCases {
+		cv := NewValue(NewStaticProvider(nil), "key", tc.value, true, Float, nil)
+		assert.Equal(t, float64(2), cv.AsFloat())
+	}
+}
+
 func TestNestedStructs(t *testing.T) {
 	provider := NewProviderGroup(
 		"test",
--- a/config/value.go
+++ b/config/value.go
@@ -49,9 +49,13 @@ const (
 	Slice
 	// Dictionary contains words and their definitions
 	Dictionary
+	// Zero constants
+	_float64Zero = float64(0)
 )
 
-var _typeTimeDuration = reflect.TypeOf(time.Duration(0))
+var (
+	_typeTimeDuration = reflect.TypeOf(time.Duration(0))
+)
 
 // GetType returns GO type of the provided object
 func GetType(value interface{}) ValueType {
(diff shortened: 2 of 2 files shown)


config/value.go
```diff
@@ -206,7 +206,18 @@ func (cv Value) TryAsFloat() (float64, bool) {
...
+	case int:
+		return float64(val), true
+	case int32:
+		return float64(val), true
+	case int64:
+		return float64(val), true
+	case float32:
+		return float64(val), true
+	default:
+		return f, false
```
anuptalwalkar: Not related to PR, but I wonder if f should be defined as constant `_floatZero`?rather than having to allocate float64(0) on every TryAsFloat call

config/value.go
```diff
@@ -49,9 +49,13 @@ const (
 	Slice
 	// Dictionary contains words and their definitions
 	Dictionary
+	// Zero constants
+	_float64Zero = float64(0)
```
anuptalwalkar: nit, this can be combined with _typeTimeDuration

madhuravi: _typeTimeDuration is a var

config/value.go
alsamylkin: It is a builtin primitive type, probably even hurt perf, because we need to fetch value from the global variable :)

config/config_test.go
```diff
@@ -206,6 +206,22 @@ func TestGetAsIntegerValue(t *testing.T) {
 	}
 }
 
+func TestGetAsFloatValue(t *testing.T) {
+	testCases := []struct {
+		value interface{}
+	}{
+		{float32(2)},
+		{float64(2)},
+		{int(2)},
```
alsamylkin: Are we interested in unsigned types?:)

config/config_test.go
```diff
@@ -206,6 +206,22 @@ func TestGetAsIntegerValue(t *testing.T) {
...
+		{int32(2)},
+		{int64(2)},
```
alsamylkin: Could you also test an error path?
E.g. time.Duration which is alias to int.

config/value.go
```diff
@@ -201,12 +203,22 @@ func (cv Value) TryAsBool() (bool, bool) {
...
 	}
-	return f, false
+	switch val := v.(type) {
+	case int:
+		return float64(val), true
+	case int32:
+		return float64(val), true
+	case int64:
+		return float64(val), true
+	case float32:
```
glibsm: you can combine all these in one line `case int, int32, int64, float32, uint etc...`

config/config_test.go
madhuravi: There is already a test for the error paths "TestInvalidConfigFailures".

config/config_test.go
madhuravi: Good question. Though about this too but not sure if config values can be sent as unsigned types. We might want to do TryAsUnsignedInt but that's different from what you're asking.
What do you think @anuptalwalkar

config/value.go
madhuravi: That doesn't seem to work in this case Glib. "cannot convert val (type interface {}) to type int: need type assertion"

PR #175, opened 2017-01-19
Add secrets file to config provider

anuptalwalkar:
GFM-301

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -42,6 +42,11 @@ const (
 	config      = "config"
 )
 
+const (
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -42,6 +42,11 @@ const (
 	config      = "config"
 )
 
+const (
+	baseFile    = "base"
```
glibsm: underscore

anuptalwalkar: https://github.com/uber-go/fx/pull/178

PR #177, opened 2017-01-19
Making Metrics available for auth clients

anuptalwalkar:


alsamylkin:
Could you provide more context why do we need metrics in the CreateAuthInfo?

anuptalwalkar:
Forgot to add description. For private clients, we want to emit metrics based on how they are configured.  (for example, client emits metrics on specific auth failures)... 
It doesn't make sense to have that configuration setup in public repo as we already have some basic configuration setup integrated for internal clients. We can talk offline since its not a public forum topic.

PR #179, opened 2017-01-19
Fix jaeger logger to make arguments even before invoking zap

madhuravi:


madhuravi:
@anuptalwalkar

tracing/tracer.go
```diff
@@ -77,7 +77,13 @@ func (jl *jaegerLogger) Error(msg string) {
 
 // Infof logs an info message with args as key value pairs
 func (jl *jaegerLogger) Infof(msg string, args ...interface{}) {
-	jl.log.Info(msg, args...)
+	// TODO (madhu) Zap logger requires even number of arguments, add a key to each arg
+	// This is a hack in place till jaeger directly takes a zap logger
+	var logArgs []interface{}
```
alsamylkin: preallocate the array?

modules/uhttp/filters_test.go
```diff
@@ -66,9 +73,11 @@ func TestTracingFilterWithLogs(t *testing.T) {
 		assert.True(t, len(buf.Lines()) > 0)
 		var spanFinished bool
 		for _, line := range buf.Lines() {
-			assert.Contains(t, line, "traceID")
-			assert.Contains(t, line, "spanID")
-			spanFinished = spanFinished || strings.Contains(line, "Span finished")
+			if strings.Contains(line, "Span finished") {
```
alsamylkin: Why is it ok to not have traceID/spanID in some lines now?

madhuravi: I enabled logging from the jaeger library for this test. This logger is assigned on creating the tracer which is at service startup and hence will not have traceID or spanID. Essentially, we now have lines logged by the jaeger library.

PR #180, opened 2017-01-19
Allocate size for jaeger log args

madhuravi:


madhuravi:
@alsamylkin

tracing/tracer.go
```diff
@@ -79,11 +79,12 @@ func (jl *jaegerLogger) Error(msg string) {
 func (jl *jaegerLogger) Infof(msg string, args ...interface{}) {
 	// TODO (madhu) Zap logger requires even number of arguments, add a key to each arg
 	// This is a hack in place till jaeger directly takes a zap logger
-	var logArgs []interface{}
+	targetLen := len(args) * 2
+	var targetArgs = make([]interface{}, targetLen)
```
alsamylkin: I think it should be 
`make([]interface{}, 0, targetLen)`

sectioneight: no need var `var`, you can do `targetArgs := make(...)`

madhuravi: Alex, if capacity is not specified, it defaults to len. Just specifying length should do.

alsamylkin: I think you are appending to the end of the slice, so targetArgs would have len == 2*targetLen ;)

madhuravi: oops you're right.

PR #181, opened 2017-01-20
Automatically emits metrics with log counts by level

glibsm:
The idea here is that we'll be getting a strong signal post deploy
if the distribution of `error` logs shoots up - probably revert is 
on the cards.

GFM-154

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -114,6 +114,11 @@ func New(options ...Option) (Owner, error) {
 		svc.configProvider = config.Load()
 	}
 
+	// load standard config
+	if err := svc.setupStandardConfig(); err != nil {
+		return nil, err
+	}
+
 	// Initialize metrics. If no metrics reporters were Registered, do nop
 	// TODO(glib): add a logging reporter and use it by default, rather than nop
 	svc.setupMetrics()
@@ -121,11 +126,6 @@ func New(options ...Option) (Owner, error) {
 	svc.setupLogging()
 
 	svc.setupAuthClient()
-
-	// load standard config
-	if err := svc.setupStandardConfig(); err != nil {
-		return nil, err
-	}
 
 	if err := svc.setupRuntimeMetricsCollector(); err != nil {
 		return nil, err
--- a/service/service.go
+++ b/service/service.go
@@ -114,30 +114,30 @@ func New(options ...Option) (Owner, error) {
 		svc.configProvider = config.Load()
 	}
 
-	svc.setupLogging(svc.configProvider)
-
-	svc.setupAuthClient(svc.configProvider)
-
-	// load standard config
-	if err := svc.setupStandardConfig(svc.Config()); err != nil {
-		return nil, err
-	}
-
 	// Initialize metrics. If no metrics reporters were Registered, do nop
 	// TODO(glib): add a logging reporter and use it by default, rather than nop
 	svc.setupMetrics()
 
-	if err := svc.setupRuntimeMetricsCollector(svc.Config()); err != nil {
+	svc.setupLogging()
+
+	svc.setupAuthClient()
+
+	// load standard config
+	if err := svc.setupStandardConfig(); err != nil {
 		return nil, err
 	}
 
-	if err := svc.setupTracer(svc.Config()); err != nil {
+	if err := svc.setupRuntimeMetricsCollector(); err != nil {
+		return nil, err
+	}
+
+	if err := svc.setupTracer(); err != nil {
 		return nil, err
 	}
 
 	// if we have an observer, look for a property called "config" and load the service
 	// node into that config.
-	svc.setupObserver(svc.Config())
+	svc.setupObserver()
 
 	// Put service into Initialized state
 	svc.transitionState(Initialized)
--- a/service/service_setup.go
+++ b/service/service_setup.go
@@ -35,13 +35,12 @@ import (
 
 func (svc *serviceCore) setupLogging() {
 	if svc.log == nil {
-		logBuilder := ulog.Builder().WithScope(svc.metrics)
-
 		err := svc.configProvider.Get("logging").PopulateStruct(&svc.logConfig)
 		if err != nil {
 			ulog.Logger().Info("Logging configuration is not provided, setting to default logger", "error", err)
 		}
 
+		logBuilder := ulog.Builder().WithScope(svc.metrics)
 		svc.log = logBuilder.WithConfiguration(svc.logConfig).Build()
 	} else {
 		svc.log.Debug("Using custom log provider due to service.WithLogger option")
--- a/service/service_setup.go
+++ b/service/service_setup.go
@@ -33,7 +33,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-func (svc *serviceCore) setupLogging(cfg config.Provider) {
+func (svc *serviceCore) setupLogging() {
 	if svc.log == nil {
 		logBuilder := ulog.Builder()
 		// load and configure logging
@@ -47,8 +47,8 @@ func (svc *serviceCore) setupLogging(cfg config.Provider) {
 	}
 }
 
-func (svc *serviceCore) setupStandardConfig(cfg config.Provider) error {
-	if err := cfg.Get(config.Root).PopulateStruct(&svc.standardConfig); err != nil {
+func (svc *serviceCore) setupStandardConfig() error {
+	if err := svc.configProvider.Get(config.Root).PopulateStruct(&svc.standardConfig); err != nil {
 		return errors.Wrap(err, "unable to load standard configuration")
 	}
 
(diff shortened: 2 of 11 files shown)


service/service_setup.go
```diff
@@ -33,22 +33,23 @@ import (
 	"github.com/pkg/errors"
 )
 
-func (svc *serviceCore) setupLogging(cfg config.Provider) {
+func (svc *serviceCore) setupLogging() {
 	if svc.log == nil {
```
sectioneight: semi-related nit: prefer early returns and reducing nesting:

```go
if svc.log != nil {
  svc.log.Debug("Using custom...")
  return
}

// the rest
logBuilder := ...
```

service/service_setup.go
```diff
@@ -33,22 +33,23 @@ import (
...
+
 		svc.log = logBuilder.WithConfiguration(svc.logConfig).Build()
 	} else {
 		svc.log.Debug("Using custom log provider due to service.WithLogger option")
 	}
 }
 
-func (svc *serviceCore) setupStandardConfig(cfg config.Provider) error {
-	if err := cfg.Get(config.Root).PopulateStruct(&svc.standardConfig); err != nil {
+func (svc *serviceCore) setupStandardConfig() error {
```
sectioneight: why the change to rely on the member field instead of arguments? personal preference?

testutils/metrics/metrics.go
```diff
@@ -0,0 +1,90 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
sectioneight: 2017

testutils/metrics/metrics.go
```diff
@@ -0,0 +1,90 @@
...
+	"time"
+
+	"github.com/uber-go/tally"
+)
+
+// TestStatsReporter allows for easy testing of any metrics related functionality
+type TestStatsReporter struct {
+	m sync.Mutex
+
+	Cw sync.WaitGroup
```
sectioneight: I don't like thesenames. They convey absolutely zero meaning. Additionally the casing implies it's an abbreviated word. I'd suggest `CountersWG` or similar

service/service_setup.go
```diff
@@ -33,22 +33,23 @@ import (
 	"github.com/pkg/errors"
 )
 
-func (svc *serviceCore) setupLogging(cfg config.Provider) {
+func (svc *serviceCore) setupLogging() {
 	if svc.log == nil {
-		logBuilder := ulog.Builder()
-		// load and configure logging
+		logBuilder := ulog.Builder().WithScope(svc.metrics)
```
sectioneight: This needs to be configurable. It should be on by default, but there needs to be a boolean flag somebody can set that turns off the metrics reporting for log messages.

ulog/builder.go
```diff
@@ -87,6 +96,7 @@ func (lb *LogBuilder) WithSentryHook(hook *sentry.Hook) *LogBuilder {
 }
 
 // Build the ulog logger for use
+// TODO: build should return `(Log, error)` in case we can't properly instantiate
```
sectioneight: Why a TODO and not part of this PR? What are the cases where we can't properly instantiate?

ulog/builder_test.go
```diff
@@ -147,6 +148,22 @@ func TestConfiguredLoggerWithSentryError(t *testing.T) {
...
+	defer env.Override(t, config.EnvironmentKey(), "wat?")()
+
+	s, r := metrics.NewTestScope()
+	l := Builder().WithScope(s).Build()
+
+	r.Cw.Add(1)
+	l.Warn("Warning log!")
+	r.Cw.Wait()
+
+	assert.Equal(t, 1, len(r.Counters))
```
sectioneight: Can we check that the counter used the right bucket?

ulog/metrics/metrics.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"github.com/uber-go/tally"
+	"github.com/uber-go/zap"
+)
+
+// Hook counts the number of logging messages per level
+func Hook(s tally.Scope) zap.Hook {
+	return zap.Hook(func(e *zap.Entry) error {
+		// TODO: consider not using .Counter method which does a map lookup
+		// and just create objects for a counter of each type and retain in a struct
```
sectioneight: huh?

service/service.go
```diff
@@ -114,30 +114,30 @@ func New(options ...Option) (Owner, error) {
 		svc.configProvider = config.Load()
 	}
 
-	svc.setupLogging(svc.configProvider)
+	// Initialize metrics. If no metrics reporters were Registered, do nop
+	// TODO(glib): add a logging reporter and use it by default, rather than nop
+	svc.setupMetrics()
```
alsamylkin: Shouldn't we setup standard config first(it has service name, owner, all the important things), all metrics and then logger?

service/service_setup.go
```diff
@@ -66,13 +67,13 @@ func (svc *serviceCore) setupMetrics() {
...
 
-func (svc *serviceCore) setupRuntimeMetricsCollector(cfg config.Provider) error {
+func (svc *serviceCore) setupRuntimeMetricsCollector() error {
 	if svc.RuntimeMetricsCollector() != nil {
 		return nil
 	}
 
 	var runtimeMetricsConfig metrics.RuntimeConfig
-	err := cfg.Get("metrics.runtime").PopulateStruct(&runtimeMetricsConfig)
+	err := svc.configProvider.Get("metrics.runtime").PopulateStruct(&runtimeMetricsConfig)
```
alsamylkin: I think it should be .Scope("metrics").Get("runtime").

ulog/metrics/metrics.go
```diff
@@ -0,0 +1,36 @@
...
+package metrics
+
+import (
+	"github.com/uber-go/tally"
+	"github.com/uber-go/zap"
+)
+
+// Hook counts the number of logging messages per level
+func Hook(s tally.Scope) zap.Hook {
+	return zap.Hook(func(e *zap.Entry) error {
```
alsamylkin: check e != nil?

service/service_setup.go
glibsm: I agree. I did spend some time thinking about it, but I am not sure with the current state of things the implementation is going to be the cleanest. 

Right now, the thing that would fit best is to include a `disabled:` flag in config and allow user to specify that. However, I don't think that mixing and matching `enabled` and `disabled` in configuration is a good thing (I've found that the positive bools tend to work out better in the long run).

I think that a better fit for this is going to be the default configuration, which is currently in the works by @alsamylkin. Then, we can include `metrics: logging: enabled: true` or something like that, and allow to easily override in service config.

Basically it's a long way of saying I think we should wait to add that in the proper way. Thoughts?

service/service_setup.go
glibsm: Mostly simplifying for three reasons:

1. the parameters being passed in are totally redundant on a method call.
2. most of these functions were not even using the parameters and used the object directly
3. needed access to the metrics scope in the logging function and rather than adding an additional parameter, simplified the whole approach.

I would be ok to restoring parameter passing, but in that case I think these methods should be changed to function and take service as a parameter. Kind of like the options pattern

testutils/metrics/metrics.go
glibsm: need to update the license binary i guess

testutils/metrics/metrics.go
glibsm: Point taken, i don't like them either. Will clean up

ulog/builder_test.go
glibsm: That is tested elsewhere, I think it would be redundant. The test here is to make sure that we set up the hook at all

ulog/metrics/metrics.go
glibsm: engrish.

I did a horrid job of saying that we should essentially cache the counters, since there is a finite number of them: one per each level, rather than calling `Counter()` method on the scope, which does some locking and map lookup

testutils/metrics/metrics.go
```diff
@@ -0,0 +1,90 @@
...
+
+import (
+	"sync"
+	"time"
+
+	"github.com/uber-go/tally"
+)
+
+// TestStatsReporter allows for easy testing of any metrics related functionality
+type TestStatsReporter struct {
```
glibsm: It's used by two different packages in uberfx

service/service.go
glibsm: Yes, that's a great point.

glibsm:
Fixing the license added a bit of fluff to the diff, can move out into another if there are objections.

ulog/builder.go
glibsm: Didn't want to bloat the PR and touch that code, but will at some point. We do some unmarhsaling, we read configs and do some other things that produce errors in logger build. When encountering error, we log an error message and fall back to default value. I think we should probably just return those errors to make sure service owners don't misconfigure logging by accident

ulog/builder_test.go
```diff
@@ -147,6 +148,49 @@ func TestConfiguredLoggerWithSentryError(t *testing.T) {
...
+		{
+			true,
+			[]byte("logging:\n  disableMetrics: true"),
+		},
+	}
+
+	for _, tt := range tests {
+		logCfg := Configuration{}
+		logYamlCfg := config.NewYAMLProviderFromBytes(tt.data)
+		logYamlCfg.Get("logging").PopulateStruct(&logCfg)
```
alsamylkin: Could you test actual behavior, not a serialization?

ulog/builder_test.go
```diff
@@ -147,6 +148,49 @@ func TestConfiguredLoggerWithSentryError(t *testing.T) {
...
+			false,
+			[]byte("logging:\n  disableMetrics: false"),
+		},
+		{
+			true,
+			[]byte("logging:\n  disableMetrics: true"),
+		},
+	}
+
+	for _, tt := range tests {
```
alsamylkin: use subtests?

ulog/builder_test.go
glibsm: i'll remove the test because you're right it doesn't help. But testing behavior right now is not easy - since zap doesn't let you peek into the hooks or options that were used to create it. I'll separate the `Build()` method into two parts, so we can assert which options zap is about to be created with.

PR #182, opened 2017-01-20
Documentation for auth package and its usage

anuptalwalkar:
GFM-304

### DIFF
--- a/auth/README.md
+++ b/auth/README.md
@@ -31,12 +31,12 @@ Example implementation of userAuthClient:
 ```
 _ Client = &userAuthClient{}
 
-ype userAuthClient struct {
-  // embed user auth client
+type userAuthClient struct {
+  // embed back-end security service client here
 }
 
 func userAuthClient(info CreateAuthInfo) auth.Client {
-	return &nouserAuthClientp{}
+	return &userAuthClient{}
 }
 
 func (*userAuthClient) Name() string {
@@ -45,6 +45,19 @@ func (*userAuthClient) Name() string {
 ```
 
 ### Implement custom auth APIs with `auth.Client` by delegating calls to your service's client.
+```
+func (u *userAuthClient) Authenticate(ctx context.Context) (context.Context, error) {
+  // authenticate with back-end security server
+  ctx, err := u.Client.Authenticate(ctx)
+	return ctx, err
+}
+
+func (u *userAuthClient) Authorize(ctx context.Context) error {
+  // authorize with back-end security server
+  err := u.Client.Authorize(ctx)
+	return err
+}
+```
 
 ### Register custom implementation construct with `fx`,
 The last step is to integrate the user auth client with the framework. This can be done by implementing init
--- a/auth/README.md
+++ b/auth/README.md
@@ -16,7 +16,7 @@ identify the service/user and grant certificate for further access.
 
 Authentication:
 Authentication API is called by calling entity to authenticate itself. The Authenticate call
-returns context, which must be populated by the backend service with signed certificate that is valid for a timeframe.
+returns context, which must be populated by the back-end service with signed certificate that is valid for a time frame.
 
 Authorization:
 Authorization API is called by a service entity to authorize its callers. context provided by a
@@ -24,7 +24,7 @@ request must have a signed certificate which caller received on authentication.
 
 ## Integrating custom auth service
 `Auth package` just provides an interface and API integration with existing modules. Users can define
-their own backend security framework and integrate its clients with the service framework by following simple steps:
+their own back-end security framework and integrate its clients with the service framework by following simple steps:
 
 ### Implement `auth.Client` interface for custom security service
 Example implementation of userAuthClient:
@@ -44,7 +44,7 @@ func (*userAuthClient) Name() string {
 }
 ```
 
-### Impleemt custom auth APIs with `auth.Client` by delegating calls to your service's client.
+### Implement custom auth APIs with `auth.Client` by delegating calls to your service's client.
 
 ### Register custom implementation construct with `fx`,
 The last step is to integrate the user auth client with the framework. This can be done by implementing init
--- a/auth/README.md
+++ b/auth/README.md
@@ -1,0 +1,56 @@
+# Auth Package
+
+Auth package is used for request authentication and authorization for user to service
+and service to service communication. Auth is used for scenarios that require stricter
+validations, and usage restrictions. `auth.Client` provided by the package provides seamless
+integration with presently written modules in the service framework. As a middleware, Auth
+provides additional restrictions on who can access a service, and who can be
+authenticated to access the service.
+The Auth package doesn't dictate how authentication and authorization should work, or which
+algorithm the security service should use, but only allows client integration with the service framework.
+
+## Auth calls
+SetAttribute:
+SetAttribute sets necessary request attributes for authentication. By setting attributes, security service can
+identify the service/user and grant certificate for further access.
+
+Authentication:
+Authentication API is called by calling entity to authenticate itself. The Authenticate call
+returns context, which must be populated by the backend service with signed certificate that is valid for a timeframe.
+
+Authorization:
+Authorization API is called by a service entity to authorize its callers. context provided by a
+request must have a signed certificate which caller received on authentication.
+
+## Integrating custom auth service
+`Auth package` just provides an interface and API integration with existing modules. Users can define
+their own backend security framework and integrate its clients with the service framework by following simple steps:
+
+### Implement `auth.Client` interface for custom security service
+Example implementation of userAuthClient:
+```
+_ Client = &userAuthClient{}
+
+ype userAuthClient struct {
+  // embed user auth client
+}
+
+func userAuthClient(info CreateAuthInfo) auth.Client {
+	return &nouserAuthClientp{}
+}
+
+func (*userAuthClient) Name() string {
+	return "userAuthClient"
+}
+```
+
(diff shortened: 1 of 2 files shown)


anuptalwalkar:
@Lupie

auth/README.md
```diff
@@ -0,0 +1,69 @@
+# Auth Package
+
+Auth package is used for request authentication and authorization for user to service
```
Lupie: The conventional Golang style is (lowercase) "package foo," right?
Unless you have a counterargument, let's look for all the "Foo Package," "Foo package," and "`Foo package`" spots in the docs and change them to "package foo."

So this will be package auth throughout.

auth/README.md
```diff
@@ -0,0 +1,69 @@
...
+func userAuthClient(info CreateAuthInfo) auth.Client {
+	return &userAuthClient{}
+}
+
+func (*userAuthClient) Name() string {
+	return "userAuthClient"
+}
+```
+
+### Implement custom auth APIs with `auth.Client` by delegating calls to your service's client.
```
Lupie: Remove period

auth/README.md
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func (u *userAuthClient) Authorize(ctx context.Context) error {
+  // authorize with back-end security server
+  err := u.Client.Authorize(ctx)
+	return err
+}
+```
+
+### Register custom implementation construct with `fx`,
```
Lupie: Is there a reason for the comma at the end?

auth/README.md
Lupie: user-to-service

auth/README.md
```diff
@@ -0,0 +1,69 @@
+# Auth Package
+
+Auth package is used for request authentication and authorization for user to service
+and service to service communication. Auth is used for scenarios that require stricter
```
Lupie: service-to-service

auth/README.md
Lupie: Instead of "Auth package is used for…" you could say more directly, "Use package auth for…" or even "Package auth authenticates and authorizes requests for user-to-service…."

auth/README.md
Lupie: Again here, instead of "Auth is used for," try "Use auth for" or "Reserve auth for." It clarifies who uses auth and when.

auth/README.md
```diff
@@ -0,0 +1,69 @@
...
+validations, and usage restrictions. `auth.Client` provided by the package provides seamless
```
Lupie: Remove comma.

Lupie: Alternatives (to avoid provided…provides):
- `auth.Client` comes with the package to provide
- `auth.Client`, included in the package, provides

auth/README.md
```diff
@@ -0,0 +1,69 @@
...
+integration with presently written modules in the service framework. As a middleware, Auth
+provides additional restrictions on who can access a service, and who can be
+authenticated to access the service.
+The Auth package doesn't dictate how authentication and authorization should work, or which
```
Lupie: Nit: or --> nor

auth/README.md
```diff
@@ -0,0 +1,69 @@
...
+algorithm the security service should use, but only allows client integration with the service framework.
```
Lupie: should use, but only --> should use. It only

auth/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Auth calls
+SetAttribute:
+SetAttribute sets necessary request attributes for authentication. By setting attributes, security service can
+identify the service/user and grant certificate for further access.
```
Lupie: Because service and user aren't interchangeable here, say "service and user" instead.

If you're worried about the "service and user and grant," you can change "and grant" to "as well as grant." It's not necessary—up to you.

auth/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+Authentication:
+Authentication API is called by calling entity to authenticate itself. The Authenticate call
```
Lupie: - The authentication API
- The first sentence isn't quite right; how about "Access the authentication API by calling an entity to authenticate itself"?

^ That's still a little confusing, but it's enough to keep the reader going.

Lupie: Do we need to capitalize "Authenticate call"?

auth/README.md
```diff
@@ -0,0 +1,69 @@
...
+returns context, which must be populated by the back-end service with signed certificate that is valid for a time frame.
```
Lupie: - back-end --> backend
- Which backend service?
- Readers might want to know how long that time frame is

auth/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+Authorization:
+Authorization API is called by a service entity to authorize its callers. context provided by a
```
Lupie: - Same comments as above in the authentication API
- context --> Context

auth/README.md
```diff
@@ -0,0 +1,69 @@
...
+request must have a signed certificate which caller received on authentication.
```
Lupie: certificate which caller --> certificate, which the caller

auth/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Integrating custom auth service
+`Auth package` just provides an interface and API integration with existing modules. Users can define
+their own back-end security framework and integrate its clients with the service framework by following simple steps:
```
Lupie: back-end --> backend

auth/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Implement `auth.Client` interface for custom security service
```
Lupie: You might want to number these steps.

auth/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+func (*userAuthClient) Name() string {
+	return "userAuthClient"
+}
+```
+
+### Implement custom auth APIs with `auth.Client` by delegating calls to your service's client.
+```
+func (u *userAuthClient) Authenticate(ctx context.Context) (context.Context, error) {
+  // authenticate with back-end security server
```
Lupie: back-end --> backend

auth/README.md
```diff
@@ -0,0 +1,69 @@
...
+  ctx, err := u.Client.Authenticate(ctx)
+	return ctx, err
+}
+
+func (u *userAuthClient) Authorize(ctx context.Context) error {
+  // authorize with back-end security server
```
Lupie: back-end --> backend

auth/README.md
```diff
@@ -0,0 +1,69 @@
...
+`Auth package` just provides an interface and API integration with existing modules. Users can define
+their own back-end security framework and integrate its clients with the service framework by following simple steps:
+
+### Implement `auth.Client` interface for custom security service
+Example implementation of userAuthClient:
+```
+_ Client = &userAuthClient{}
+
+type userAuthClient struct {
+  // embed back-end security service client here
```
Lupie: back-end --> backend

auth/README.md
```diff
@@ -0,0 +1,69 @@
...
+func (u *userAuthClient) Authorize(ctx context.Context) error {
+  // authorize with back-end security server
+  err := u.Client.Authorize(ctx)
+	return err
+}
+```
+
+### Register custom implementation construct with `fx`,
+The last step is to integrate the user auth client with the framework. This can be done by implementing init
+function and registering client with `fx`.
```
Lupie: registering client --> registering the client

auth/README.md
anuptalwalkar: I like your suggestion for "as well as grant.". Updating.

auth/README.md
```diff
@@ -1,38 +1,39 @@
 # Auth Package
 
-Auth package is used for request authentication and authorization for user to service
-and service to service communication. Auth is used for scenarios that require stricter
-validations, and usage restrictions. `auth.Client` provided by the package provides seamless
+Use `package auth` for request authentication and authorization for user-to-service
```
Lupie: We have package auth without backticks later in the doc. Let's choose one. It seems like we can just say "package auth" in plain text.

auth/README.md
```diff
@@ -1,38 +1,39 @@
...
+and service-to-service communication. Use auth for scenarios that require stricter
+validations and usage restrictions. `auth.Client`, included in the package provides seamless
```
Lupie: Comma after "included in the package"

auth/README.md
```diff
@@ -1,38 +1,39 @@
...
 integration with presently written modules in the service framework. As a middleware, Auth
 provides additional restrictions on who can access a service, and who can be
 authenticated to access the service.
-The Auth package doesn't dictate how authentication and authorization should work, or which
-algorithm the security service should use, but only allows client integration with the service framework.
+The Auth package doesn't dictate how authentication and authorization should work, nor which
```
Lupie: Auth --> auth
(Unless you think there's a strong reason to capitalize packages. I'm still learning Go's style conventions, but I think lowercase is OK.)

auth/README.md
Lupie: For example:
- Line 3: `package auth`
- Line 26: `Auth package`
- Line 4: auth
- Line 6: Auth
- Line 9: The Auth package

We should be somewhat consistent, like capital A or lowercase a, backticks or no backticks? I think "package auth" is fine, but I'm open to arguments.

anuptalwalkar: Good point. using `package auth` makes sense, i'll maintain it everywhere else.

auth/README.md
anuptalwalkar: Fixed from above.

PR #183, opened 2017-01-24
Create async task execution module with backend and encoding options

madhuravi:
Async task execution module allows service owners to register any function that can be then called asynchronously.
It can be configured with different backends and encoding options.
The function is validated and transmitted using reflect libraries.

### DIFF
--- a/modules/task/backend.go
+++ b/modules/task/backend.go
@@ -36,42 +36,42 @@ type Backend interface {
 type NopBackend struct{}
 
 // Publish implements the Backend interface
-func (b *NopBackend) Publish(message []byte, userContext map[string]string) error {
+func (b NopBackend) Publish(message []byte, userContext map[string]string) error {
 	return nil
 }
 
 // Consume  implements the Backend interface
-func (b *NopBackend) Consume() error {
+func (b NopBackend) Consume() error {
 	return nil
 }
 
 // Encoder implements the Backend interface
-func (b *NopBackend) Encoder() Encoding {
+func (b NopBackend) Encoder() Encoding {
 	return &NopEncoding{}
 }
 
 // Type implements the Backend interface
-func (b *NopBackend) Type() string {
+func (b NopBackend) Type() string {
 	return ""
 }
 
 // Name implements the Backend interface
-func (b *NopBackend) Name() string {
+func (b NopBackend) Name() string {
 	return ""
 }
 
 // Start implements the Backend interface
-func (b *NopBackend) Start(ready chan<- struct{}) <-chan error {
+func (b NopBackend) Start(ready chan<- struct{}) <-chan error {
 	return make(chan error)
 }
 
 // Stop implements the Backend interface
-func (b *NopBackend) Stop() error {
+func (b NopBackend) Stop() error {
 	return nil
 }
 
 // IsRunning implements the Backend interface
-func (b *NopBackend) IsRunning() bool {
+func (b NopBackend) IsRunning() bool {
 	return true
 }
 
@@ -81,13 +81,13 @@ type InMemBackend struct {
 }
 
 // Publish implements the Backend interface
-func (b *InMemBackend) Publish(message []byte, userContext map[string]string) error {
+func (b InMemBackend) Publish(message []byte, userContext map[string]string) error {
 	b.bufQueue <- message
 	return nil
 }
 
 // Consume  implements the Backend interface
-func (b *InMemBackend) Consume() error {
+func (b InMemBackend) Consume() error {
 	select {
 	case v, ok := <-b.bufQueue:
 		if ok {
@@ -100,31 +100,31 @@ func (b *InMemBackend) Consume() error {
 }
 
 // Encoder implements the Backend interface
-func (b *InMemBackend) Encoder() Encoding {
+func (b InMemBackend) Encoder() Encoding {
 	return gobEncoding
 }
 
 // Type implements the Backend interface
-func (b *InMemBackend) Type() string {
+func (b InMemBackend) Type() string {
 	return ""
 }
 
 // Name implements the Backend interface
-func (b *InMemBackend) Name() string {
+func (b InMemBackend) Name() string {
 	return ""
 }
 
 // Start implements the Backend interface
-func (b *InMemBackend) Start(ready chan<- struct{}) <-chan error {
+func (b InMemBackend) Start(ready chan<- struct{}) <-chan error {
 	return make(chan error)
 }
 
 // Stop implements the Backend interface
-func (b *InMemBackend) Stop() error {
+func (b InMemBackend) Stop() error {
 	return nil
 }
 
 // IsRunning implements the Backend interface
-func (b *InMemBackend) IsRunning() bool {
+func (b InMemBackend) IsRunning() bool {
 	return true
 }
--- a/modules/task/backend.go
+++ b/modules/task/backend.go
@@ -52,12 +52,12 @@ func (b NopBackend) Encoder() Encoding {
 
 // Type implements the Backend interface
 func (b NopBackend) Type() string {
-	return ""
+	return "nop"
 }
 
 // Name implements the Backend interface
 func (b NopBackend) Name() string {
-	return ""
+	return "nop"
 }
 
 // Start implements the Backend interface
@@ -75,19 +75,24 @@ func (b NopBackend) IsRunning() bool {
 	return true
 }
 
-// InMemBackend is a noop implementation of the Backend interface
-type InMemBackend struct {
+// inMemBackend is an in-memory implementation of the Backend interface
+type inMemBackend struct {
 	bufQueue chan []byte
 }
 
+// NewInMemBackend creates a new in memory backend, designed for use in tests
+func NewInMemBackend() Backend {
+	return &inMemBackend{make(chan []byte, 2)}
+}
+
 // Publish implements the Backend interface
-func (b *InMemBackend) Publish(message []byte, userContext map[string]string) error {
+func (b *inMemBackend) Publish(message []byte, userContext map[string]string) error {
 	b.bufQueue <- message
 	return nil
 }
 
 // Consume  implements the Backend interface
-func (b *InMemBackend) Consume() error {
+func (b *inMemBackend) Consume() error {
 	select {
 	case v, ok := <-b.bufQueue:
 		if ok {
@@ -100,31 +105,31 @@ func (b *InMemBackend) Consume() error {
 }
 
 // Encoder implements the Backend interface
-func (b *InMemBackend) Encoder() Encoding {
+func (b *inMemBackend) Encoder() Encoding {
 	return gobEncoding
 }
 
 // Type implements the Backend interface
-func (b *InMemBackend) Type() string {
-	return ""
+func (b *inMemBackend) Type() string {
+	return "inMem"
 }
 
 // Name implements the Backend interface
-func (b *InMemBackend) Name() string {
-	return ""
+func (b *inMemBackend) Name() string {
+	return "inMem"
 }
 
 // Start implements the Backend interface
-func (b *InMemBackend) Start(ready chan<- struct{}) <-chan error {
+func (b *inMemBackend) Start(ready chan<- struct{}) <-chan error {
 	return make(chan error)
 }
 
 // Stop implements the Backend interface
-func (b *InMemBackend) Stop() error {
+func (b *inMemBackend) Stop() error {
 	return nil
 }
 
 // IsRunning implements the Backend interface
-func (b *InMemBackend) IsRunning() bool {
+func (b *inMemBackend) IsRunning() bool {
 	return true
 }
--- a/modules/task/backend.go
+++ b/modules/task/backend.go
@@ -18,7 +18,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-package utask
+package task
 
 import "go.uber.org/fx/service"
 
--- a/modules/task/backend.go
+++ b/modules/task/backend.go
@@ -20,7 +20,11 @@
 
 package task
 
-import "go.uber.org/fx/service"
+import (
+	"fmt"
+
+	"go.uber.org/fx/service"
+)
 
 var gobEncoding = &GobEncoding{}
 
@@ -50,27 +54,27 @@ func (b NopBackend) Encoder() Encoding {
 	return &NopEncoding{}
 }
 
-// Type implements the Backend interface
+// Type implements the Module interface
 func (b NopBackend) Type() string {
 	return "nop"
 }
 
-// Name implements the Backend interface
+// Name implements the Module interface
 func (b NopBackend) Name() string {
 	return "nop"
 }
 
-// Start implements the Backend interface
+// Start implements the Module interface
 func (b NopBackend) Start(ready chan<- struct{}) <-chan error {
 	return make(chan error)
 }
 
-// Stop implements the Backend interface
+// Stop implements the Module interface
 func (b NopBackend) Stop() error {
 	return nil
 }
 
-// IsRunning implements the Backend interface
+// IsRunning implements the Module interface
 func (b NopBackend) IsRunning() bool {
 	return true
 }
@@ -98,6 +102,7 @@ func (b *inMemBackend) Consume() error {
 		if ok {
 			return Run(v)
 		}
+		return fmt.Errorf("The bufQueue channel has been closed")
 	default:
 		// No value ready in channel, moving on
 	}
@@ -109,27 +114,28 @@ func (b *inMemBackend) Encoder() Encoding {
 	return gobEncoding
 }
 
-// Type implements the Backend interface
+// Type implements the Module interface
 func (b *inMemBackend) Type() string {
 	return "inMem"
 }
 
-// Name implements the Backend interface
+// Name implements the Module interface
 func (b *inMemBackend) Name() string {
 	return "inMem"
 }
 
-// Start implements the Backend interface
+// Start implements the Module interface
 func (b *inMemBackend) Start(ready chan<- struct{}) <-chan error {
 	return make(chan error)
 }
 
-// Stop implements the Backend interface
+// Stop implements the Module interface
 func (b *inMemBackend) Stop() error {
+	close(b.bufQueue)
 	return nil
 }
 
-// IsRunning implements the Backend interface
+// IsRunning implements the Module interface
 func (b *inMemBackend) IsRunning() bool {
 	return true
 }
--- a/modules/task/backend.go
+++ b/modules/task/backend.go
@@ -81,13 +81,13 @@ type InMemBackend struct {
 }
 
 // Publish implements the Backend interface
-func (b InMemBackend) Publish(message []byte, userContext map[string]string) error {
+func (b *InMemBackend) Publish(message []byte, userContext map[string]string) error {
 	b.bufQueue <- message
 	return nil
 }
 
 // Consume  implements the Backend interface
-func (b InMemBackend) Consume() error {
+func (b *InMemBackend) Consume() error {
 	select {
 	case v, ok := <-b.bufQueue:
 		if ok {
@@ -100,31 +100,31 @@ func (b InMemBackend) Consume() error {
 }
 
 // Encoder implements the Backend interface
-func (b InMemBackend) Encoder() Encoding {
+func (b *InMemBackend) Encoder() Encoding {
 	return gobEncoding
 }
 
 // Type implements the Backend interface
-func (b InMemBackend) Type() string {
+func (b *InMemBackend) Type() string {
 	return ""
 }
 
 // Name implements the Backend interface
-func (b InMemBackend) Name() string {
+func (b *InMemBackend) Name() string {
 	return ""
 }
 
 // Start implements the Backend interface
-func (b InMemBackend) Start(ready chan<- struct{}) <-chan error {
+func (b *InMemBackend) Start(ready chan<- struct{}) <-chan error {
 	return make(chan error)
 }
 
 // Stop implements the Backend interface
-func (b InMemBackend) Stop() error {
+func (b *InMemBackend) Stop() error {
 	return nil
 }
 
 // IsRunning implements the Backend interface
-func (b InMemBackend) IsRunning() bool {
+func (b *InMemBackend) IsRunning() bool {
 	return true
 }
--- a/modules/task/backend.go
+++ b/modules/task/backend.go
@@ -21,7 +21,7 @@
 package task
 
 import (
-	"fmt"
+	"errors"
 
 	"go.uber.org/fx/service"
 )
@@ -81,12 +81,13 @@ func (b NopBackend) IsRunning() bool {
 
 // inMemBackend is an in-memory implementation of the Backend interface
 type inMemBackend struct {
-	bufQueue chan []byte
+	bufQueue  chan []byte
+	isRunning bool
 }
 
 // NewInMemBackend creates a new in memory backend, designed for use in tests
 func NewInMemBackend() Backend {
-	return &inMemBackend{make(chan []byte, 2)}
+	return &inMemBackend{bufQueue: make(chan []byte, 2)}
 }
 
 // Publish implements the Backend interface
@@ -102,7 +103,7 @@ func (b *inMemBackend) Consume() error {
 		if ok {
 			return Run(v)
 		}
-		return fmt.Errorf("The bufQueue channel has been closed")
+		return errors.New("The bufQueue channel has been closed")
 	default:
 		// No value ready in channel, moving on
 	}
@@ -126,16 +127,18 @@ func (b *inMemBackend) Name() string {
 
 // Start implements the Module interface
 func (b *inMemBackend) Start(ready chan<- struct{}) <-chan error {
+	b.isRunning = true
 	return make(chan error)
 }
 
 // Stop implements the Module interface
 func (b *inMemBackend) Stop() error {
+	b.isRunning = false
 	close(b.bufQueue)
 	return nil
 }
 
 // IsRunning implements the Module interface
 func (b *inMemBackend) IsRunning() bool {
-	return true
+	return b.isRunning
 }
--- a/modules/task/backend_test.go
+++ b/modules/task/backend_test.go
@@ -24,7 +24,7 @@ import "testing"
 import "github.com/stretchr/testify/assert"
 
 var backends = []Backend{
-	&NopBackend{}, &InMemBackend{make(chan []byte, 2)},
+	&NopBackend{}, NewInMemBackend(),
 }
 
 func TestBackends(t *testing.T) {
@@ -34,8 +34,8 @@ func TestBackends(t *testing.T) {
 }
 
 func testBackendMethods(t *testing.T, backend Backend) {
-	assert.NotNil(t, backend.Name())
-	assert.NotNil(t, backend.Type())
+	assert.NotEmpty(t, backend.Name())
+	assert.NotEmpty(t, backend.Type())
 	assert.NotNil(t, backend.Encoder())
 	assert.NotNil(t, backend.Start(make(chan struct{})))
 	assert.True(t, backend.IsRunning())
--- a/modules/task/backend_test.go
+++ b/modules/task/backend_test.go
@@ -18,7 +18,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-package utask
+package task
 
 import "testing"
 import "github.com/stretchr/testify/assert"
--- a/modules/task/encoding.go
+++ b/modules/task/encoding.go
@@ -18,7 +18,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-package utask
+package task
 
 import (
 	"bytes"
--- a/modules/task/encoding.go
+++ b/modules/task/encoding.go
@@ -39,17 +39,17 @@ type NopEncoding struct {
 }
 
 // Register implements the Encoding interface
-func (g *NopEncoding) Register(obj interface{}) error {
+func (g NopEncoding) Register(obj interface{}) error {
 	return nil
 }
 
 // Marshal implements the Encoding interface
-func (g *NopEncoding) Marshal(obj interface{}) ([]byte, error) {
+func (g NopEncoding) Marshal(obj interface{}) ([]byte, error) {
 	return []byte{}, nil
 }
 
 // Unmarshal implements the Encoding interface
-func (g *NopEncoding) Unmarshal(data []byte, obj interface{}) error {
+func (g NopEncoding) Unmarshal(data []byte, obj interface{}) error {
 	return nil
 }
 
@@ -58,13 +58,13 @@ type GobEncoding struct {
 }
 
 // Register implements the Encoding interface
-func (g *GobEncoding) Register(obj interface{}) error {
+func (g GobEncoding) Register(obj interface{}) error {
 	gob.Register(obj)
 	return nil
 }
 
 // Marshal encodes an object into bytes
-func (g *GobEncoding) Marshal(obj interface{}) ([]byte, error) {
+func (g GobEncoding) Marshal(obj interface{}) ([]byte, error) {
 	var buf bytes.Buffer
 	enc := gob.NewEncoder(&buf)
 	if err := enc.Encode(obj); err != nil {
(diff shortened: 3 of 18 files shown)


madhuravi:
Build succeeds locally, failures are likely due to a bug in older glide versions (https://github.com/Masterminds/glide/issues/745). Will follow up tomorrow.

modules/utask/encoding.go
```diff
@@ -0,0 +1,85 @@
...
+func (g *GobEncoding) Register(obj interface{}) error {
+	gob.Register(obj)
+	return nil
+}
+
+// Marshal encodes an object into bytes
+func (g *GobEncoding) Marshal(obj interface{}) ([]byte, error) {
+	var buf bytes.Buffer
+	enc := gob.NewEncoder(&buf)
+	if err := enc.Encode(obj); err != nil {
```
sectioneight: This probably isn't the right place to log this -- the caller should be expecting that the task library does error handling but not logging, otherwise it would pass it a logger. In other words, just do:

```go
err := enc.Encode(obj)
return buf.Bytes(), err
```

modules/utask/encoding.go
```diff
@@ -0,0 +1,85 @@
...
+		ulog.Logger().Error("Encode error:", "error", err)
+		return nil, err
+	}
+	return buf.Bytes(), nil
+}
+
+// Unmarshal decodes a byte array into the passed in object
+func (g *GobEncoding) Unmarshal(data []byte, obj interface{}) error {
+	dec := gob.NewDecoder(bytes.NewBuffer(data))
+	if err := dec.Decode(obj); err != nil {
```
sectioneight: Same comment. Propagate errors up, but don't assume people want their library logging to a global logger

modules/utask/encoding.go
```diff
@@ -0,0 +1,85 @@
...
+		ulog.Logger().Error("Decode error:", err)
```
sectioneight: also, this would be a runtime error due to wrong number of args :(

modules/utask/encoding.go
```diff
@@ -0,0 +1,85 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
sectioneight: Wrong year. Can somebody please fix this?

modules/utask/encoding_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNopEncoding(t *testing.T) {
+	testEncodingMethods(t, &NopEncoding{}, false)
+}
+
+func TestGobEncoding(t *testing.T) {
+	testEncodingMethods(t, &GobEncoding{}, true)
```
sectioneight: looks like a table test written out like a non-table test. suggest converting to a table test to encourage others to add new cases

modules/utask/execution.go
```diff
@@ -0,0 +1,141 @@
...
+func (s *Signature) Execute() ([]reflect.Value, error) {
+	var targetArgs []reflect.Value
+	for _, arg := range s.Args {
+		targetArgs = append(targetArgs, reflect.ValueOf(arg))
+	}
+	if fn, ok := fnNameMap[s.FnName]; ok {
+		fnValue := reflect.ValueOf(fn)
+		return fnValue.Call(targetArgs), nil
+	}
+	return nil, fmt.Errorf("Function: %s not found. Did you forget to register?", s.FnName)
```
sectioneight: nit: errors should start with lowercase. also use `%q` for function name to auto-surround with quotes so it's clear where the name starts and ends

modules/utask/execution.go
```diff
@@ -0,0 +1,141 @@
...
+		return fmt.Errorf("Expected %d function arg(s) but found %d\n", fnType.NumIn(), len(args))
+	}
+	var argValues []reflect.Value
+	for i := 0; i < fnType.NumIn(); i++ {
+		arg := reflect.ValueOf(args[i])
+		argType := reflect.TypeOf(args[i])
+		if !argType.AssignableTo(fnType.In(i)) {
+			// TODO(madhu): Is it useful to show the arg index or the arg value in the error msg?
+			return fmt.Errorf(
+				"Cannot assign function argument: %d from type: %s to type: %s\n",
```
sectioneight: lowercase

modules/utask/execution.go
```diff
@@ -0,0 +1,141 @@
...
+			)
+		}
+		argValues = append(argValues, arg)
+	}
+	fnName := getFunctionName(fn)
+	s := Signature{FnName: fnName, Args: args}
+
+	sBytes, err := GlobalBackend().Encoder().Marshal(s)
+	if err != nil {
+		return errors.Wrap(err, "unable to encode the function")
```
sectioneight: enqueue the function call? enqueue the arguments?

modules/utask/execution_test.go
```diff
@@ -0,0 +1,161 @@
...
+)
+
+func init() {
+	_backend = &InMemBackend{}
+}
+
+func TestEnqueueNonFunction(t *testing.T) {
+	err := Register("I am not a function")
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "Expected a func as input but was")
```
sectioneight: suggest being more lenient in string tests. i'd just check for "a func as input"

modules/utask/execution_test.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func TestEnqueueWithMultipleReturnValues(t *testing.T) {
+	fn := func() (string, error) { return "", nil }
+	err := Register(fn)
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "Expected function to return only error but found")
```
sectioneight: just check "return only error"

modules/utask/task.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"go.uber.org/fx/modules"
+	"go.uber.org/fx/service"
+)
+
+// ModuleType represents the utask module type
+const ModuleType = "utask"
+
+var (
+	_backendRegisterFn backendRegisterFn
```
sectioneight: Does this mean we can only have one backend **registration** function available?

modules/utask/backend.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+// IsRunning implements the Backend interface
+func (b *NopBackend) IsRunning() bool {
+	return true
+}
+
+// InMemBackend is a noop implementation of the Backend interface
+type InMemBackend struct {
+	bufQueue [][]byte
```
alsamylkin: Could we use buffered channels instead? They are more idiomatic..

modules/utask/backend.go
```diff
@@ -0,0 +1,125 @@
...
+		err := Run(b.bufQueue[0])
+		b.bufQueue = b.bufQueue[1:]
+		return err
+	}
+	return nil
+}
+
+// Encoder implements the Backend interface
+func (b *InMemBackend) Encoder() Encoding {
+	return &GobEncoding{}
```
alsamylkin: Make it a const?

modules/utask/encoding.go
```diff
@@ -0,0 +1,85 @@
...
+	Marshal(interface{}) ([]byte, error)
+	Unmarshal([]byte, interface{}) error
+}
+
+// NopEncoding is a noop encoder
+type NopEncoding struct {
+}
+
+// Register implements the Encoding interface
+func (g *NopEncoding) Register(obj interface{}) error {
```
alsamylkin: Any particular reason, why do you prefer a pointer receiver over a value here?

modules/utask/task.go
```diff
@@ -0,0 +1,88 @@
...
+	config  Config
+}
+
+// Config contains config for task backends
+type Config struct {
+	broker  string `yaml:"broker"`
+	timeout string `yaml:"timeout"`
+}
+
+type backendRegisterFn func(service.Host, Config) (Backend, error)
```
anuptalwalkar: Why do you need Config struct to be passed? Can't it be extracted from service.Host.Config()?

modules/utask/encoding.go
madhuravi: agree, originally added for debugging. Do we not want to errors.Wrap though? what's a good rule of thumb for when to wrap?

modules/utask/task.go
```diff
@@ -0,0 +1,88 @@
...
+const ModuleType = "utask"
+
+var (
+	_backendRegisterFn backendRegisterFn
+	_backend           Backend = &NopBackend{}
+)
+
+// GlobalBackend returns global instance of the backend
+// TODO (madhu): Make work with multiple backends
+func GlobalBackend() Backend {
```
alsamylkin: Do we really want a global object and all concurrency issues that come with it?

modules/utask/execution.go
madhuravi: Can't say at this point, depends on what the error is, can clarify to say function or args.

modules/utask/execution_test.go
madhuravi: I think too little can fail to check for the right test case since error messages can be similar in wording. Will try to strike a balance.

modules/utask/task.go
madhuravi: For now, added a TODO to support multiple backends considering it will be a very small percent of asks if at all. Wanted to get this in first.

modules/utask/task.go
```diff
@@ -0,0 +1,88 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package utask
+
+import (
+	"sync"
+
+	"github.com/pkg/errors"
```
anuptalwalkar: imports ordering nit: go.uber.org imports before github imports.

modules/utask/task.go
```diff
@@ -0,0 +1,88 @@
...
+		return []service.Module{mod}, nil
+	}
+}
+
+func newAsyncModule(mi service.ModuleCreateInfo) (service.Module, error) {
+	backend, err := _backendRegisterFn(mi.Host, Config{})
+	if err != nil {
+		return nil, err
+	}
+	_backend = backend
```
anuptalwalkar: This is strange. Why are you setting up global variable and returning it as part of struct?

anuptalwalkar: Also, does this need locking?

modules/utask/task.go
```diff
@@ -0,0 +1,88 @@
...
+		Backend: backend,
+		modBase: *modules.NewModuleBase(ModuleType, "task", mi.Host, []string{}),
+	}, nil
+}
+
+// AsyncModule denotes the asynchronous task queue module
+type AsyncModule struct {
+	Backend
+	modBase modules.ModuleBase
+	stateMu sync.RWMutex
```
anuptalwalkar: Is this used anywhere?

modules/utask/task.go
```diff
@@ -0,0 +1,88 @@
...
+type Config struct {
+	broker  string `yaml:"broker"`
+	timeout string `yaml:"timeout"`
+}
+
+type backendRegisterFn func(service.Host, Config) (Backend, error)
+
+// RegisterAsyncBackend registers the backend for the async task module
+func RegisterAsyncBackend(backendRegisterFn backendRegisterFn) {
+	_backendRegisterFn = backendRegisterFn
```
anuptalwalkar: locking?

modules/utask/execution.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+	retValues, err := s.Execute()
+	if err != nil {
+		return err
+	}
+	return castToError(retValues[0])
+}
+
+// validateAsyncFn verifies that the type is a function type that returns only an error
+func validateAsyncFn(fnType reflect.Type) error {
```
alsamylkin: Could you rename it to validateFnReturnType? We don'e validate that the function is actually async :)

modules/utask/execution.go
```diff
@@ -0,0 +1,152 @@
...
+			// TODO(madhu): Is it useful to show the arg index or the arg value in the error msg?
+			return fmt.Errorf(
+				"Cannot assign function argument: %d from type: %s to type: %s\n",
+				i+1, argType, fnType.In(i),
+			)
+		}
+		argValues = append(argValues, arg)
+	}
+	fnName := getFunctionName(fn)
+	s := Signature{FnName: fnName, Args: args}
```
anuptalwalkar: Signature sounds more close to signed data. may be call it a `Function`

modules/utask/execution.go
```diff
@@ -0,0 +1,152 @@
...
+type Signature struct {
+	FnName string
+	Args   []interface{}
+}
+
+// Execute executes the function
+func (s *Signature) Execute() ([]reflect.Value, error) {
+	var targetArgs []reflect.Value
+	for _, arg := range s.Args {
+		targetArgs = append(targetArgs, reflect.ValueOf(arg))
```
alsamylkin: Why do we store the reflection types, not actual values? It seems like you're converting to/from in Enqueue/Execute.

modules/utask/execution.go
```diff
@@ -0,0 +1,152 @@
...
+	fnLookup.RLock()
+	defer fnLookup.RUnlock()
+	if fn, ok := fnLookup.fnNameMap[s.FnName]; ok {
+		fnValue := reflect.ValueOf(fn)
+		return fnValue.Call(targetArgs), nil
+	}
+	return nil, fmt.Errorf("function: %q not found. Did you forget to register?", s.FnName)
+}
+
+// Register registers a function for async tasks
```
alsamylkin: Why do we need to register a function before executing? Can we enqueue it with args?

modules/utask/backend.go
madhuravi: good idea

modules/utask/task.go
madhuravi: Aiden and I discussed this a bit and decided to have the global object with the focus a simple API and experience for engineers. This is a tradeoff but it is also only ever set on module start.

alsamylkin: How is it simpler than a backend per module? Seems reverse actually: testing is easier, extending to multiple backends(by using multiple modules) is easier too..

modules/utask/execution.go
```diff
@@ -0,0 +1,152 @@
...
+func Register(fn interface{}) error {
+	fnType := reflect.TypeOf(fn)
+	if err := validateAsyncFn(fnType); err != nil {
+		return err
+	}
+	for i := 0; i < fnType.NumIn(); i++ {
+		gob.Register(reflect.Zero(fnType.In(i)).Interface())
```
alsamylkin: What is going to happen if the function we want to call changed it's signature, but still has the same name? E.g. 2 different versions exist at the same time during deployment..

modules/utask/execution.go
```diff
@@ -0,0 +1,152 @@
...
+func (s *Signature) Execute() ([]reflect.Value, error) {
+	var targetArgs []reflect.Value
+	for _, arg := range s.Args {
+		targetArgs = append(targetArgs, reflect.ValueOf(arg))
+	}
+	fnLookup.RLock()
+	defer fnLookup.RUnlock()
+	if fn, ok := fnLookup.fnNameMap[s.FnName]; ok {
+		fnValue := reflect.ValueOf(fn)
+		return fnValue.Call(targetArgs), nil
```
alsamylkin: What is going to happen in case of panics?

modules/utask/execution.go
madhuravi: It was about the format of what a valid async function is. I can see how that can be confusing. Changed to validateFnFormat.

modules/utask/execution.go
madhuravi: panics should just propagate just like everywhere else. This shouldn't be any different from a regular function call.

modules/utask/execution.go
madhuravi: The caveat for using our async task framework is if engineers are changing function signature, they handle it in a backwards compatible manner. Not the best approach but just keeping it simple for now. I'd love to handle this on our end somehow in the longer term. If you have ideas, I'd love to hear them, feel free to ping me offline.

modules/utask/execution.go
madhuravi: I went back and forth on this. Registering in advance is not required, it's mostly for perf. The fnNameMap gets populated on register and if this happens on service startup, writes to the map will be contained there and enqueueing the function will only be the read path.

modules/utask/execution.go
madhuravi: I used that as FunctionSignature :D It doesn't actually store the function object so Function would be confusing too I feel. Renamed to fnSignature.

modules/utask/execution.go
madhuravi: We don't store the types. It is passed as an interface and reflect.ValueOf gets the underlying value. I only want to convert once though. This brought my attention to some old redundant code I had. Removed now.

sectioneight:
Perhaps we should have an in-person (team) review of this? Seems to be a
lot of great discussion, I'm worried GH may not be the best place when
we're all in the same office.

modules/utask/task.go
madhuravi: A backend is tied to an async module. This global variable is a user experience optimization, users will mostly have only one async module. So they would be able to do
utask.Enqueue() 
from anywhere in the code without having to pass around the module everywhere.

modules/utask/task.go
breerly: Agreed here - suggest using plain 'ol dependency injection to inject the backend, as opposed to having a global.

modules/utask/encoding.go
madhuravi: It's not necessary, I find it annoying to have to pick and just fall back to pointer received. Changed these to value receivers.

modules/utask/task.go
madhuravi: I think it will be in future. Removed it for now, will add later when the need arises.

modules/utask/task.go
madhuravi: With typical use, it shouldn't. The only place it can be set is on module initialization. But that's probably being over optimistic.I will add it for the time being and revisit when doing benchmarks.

modules/task/backend.go
```diff
@@ -0,0 +1,130 @@
...
+	return true
+}
+
+// InMemBackend is a noop implementation of the Backend interface
+type InMemBackend struct {
+	bufQueue chan []byte
+}
+
+// Publish implements the Backend interface
+func (b InMemBackend) Publish(message []byte, userContext map[string]string) error {
```
alsamylkin: This seems should have a pointer receiever

modules/task/execution.go
```diff
@@ -0,0 +1,149 @@
...
+// fnSignature represents a function and its arguments
+type fnSignature struct {
+	FnName string
+	Args   []interface{}
+}
+
+// Execute executes the function
+func (s *fnSignature) Execute() ([]reflect.Value, error) {
+	var targetArgs []reflect.Value
+	for _, arg := range s.Args {
```
alsamylkin: Prealloc

modules/task/execution.go
```diff
@@ -0,0 +1,149 @@
...
+		targetArgs = append(targetArgs, reflect.ValueOf(arg))
+	}
+	fnLookup.RLock()
+	defer fnLookup.RUnlock()
+	if fn, ok := fnLookup.fnNameMap[s.FnName]; ok {
+		fnValue := reflect.ValueOf(fn)
+		return fnValue.Call(targetArgs), nil
```
alsamylkin: timeouts?

madhuravi: I thought about this some more and came to the conclusion that timeouts don't make too much sense here. We usually have timeouts for cross-service RPC style communication. This is calling a function on a service, the service owner should have tests to make sure they don't have infinite loops etc.

breerly: If the call leaves the process, it's a network call and has to deal with same class of issues that an RPC does. This is true when calling functions on other services and on "the same" service.

You'll want a timeout at some layer, even if that isn't exposed to the user.

madhuravi: Agreed. We do have timeouts on the cross-network layer. The specific component that Alex and I are discussing are within the service.

modules/task/backend_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+import "testing"
+import "github.com/stretchr/testify/assert"
+
+var backends = []Backend{
+	&NopBackend{}, &InMemBackend{make(chan []byte, 2)},
+}
+
+func TestBackends(t *testing.T) {
+	for _, b := range backends {
```
alsamylkin: Same for TestEncoding..

alsamylkin: Also use t.Parallel()?

modules/task/execution.go
alsamylkin: @madhuravi Our goal is to make their life easier and think about local timeouts as well :)

modules/task/execution.go
```diff
@@ -0,0 +1,171 @@
...
+var fnLookup = fnRegister{fnNameMap: make(map[string]interface{})}
+
+// fnSignature represents a function and its arguments
+type fnSignature struct {
+	FnName string
+	Args   []interface{}
+}
+
+// Execute executes the function
+func (s *fnSignature) Execute() ([]reflect.Value, error) {
```
alsamylkin: Why do we want to return a slice, when we know there will be only one Value?

modules/task/execution.go
```diff
@@ -0,0 +1,171 @@
...
+// Enqueue sends a func before sending to the task queue
+func Enqueue(fn interface{}, args ...interface{}) error {
+	// Validate function
+	fnType := reflect.TypeOf(fn)
+	if err := validateFnFormat(fnType); err != nil {
+		return err
+	}
+	fnName := getFunctionName(fn)
+	// Register
+	if err := register(fn, fnType, fnName); err != nil {
```
alsamylkin: What is the function registration story?
Can we consume functions that were not registered? E.g. in case of a distributed task system if a function was registered on one machine, but not on another, how would we deal with it? Looks like consume can't be called before all the functions are registered, how can we prevent it?

modules/task/execution_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+func TestEnqueueMapFn(t *testing.T) {
+	fn := func(map[string]string) error { return nil }
+	err := Enqueue(fn, make(map[string]string))
+	assert.NoError(t, err)
+	err = GlobalBackend().Consume()
+	assert.NoError(t, err)
+}
+
+func TestEnqueueComplexFnWithError(t *testing.T) {
```
alsamylkin: Could you also add a test with a closure, not a pure function?

modules/task/execution.go
```diff
@@ -0,0 +1,171 @@
...
+	// Validate function against arguments
+	if err := validateFnAgainstArgs(fnType, args); err != nil {
+		return err
+	}
+	// Publish function to the backend
+	s := fnSignature{FnName: fnName, Args: args}
+
+	sBytes, err := GlobalBackend().Encoder().Marshal(s)
+	if err != nil {
+		return errors.Wrap(err, "unable to encode the function or args")
```
alsamylkin: why do you wrap the error here, but not in line 75/71?

modules/task/execution.go
```diff
@@ -0,0 +1,171 @@
...
+
+// Run decodes the message and executes as a task
+func Run(message []byte) error {
+	var s fnSignature
+	if err := GlobalBackend().Encoder().Unmarshal(message, &s); err != nil {
+		return errors.Wrap(err, "unable to decode the message")
+	}
+	retValues, err := s.Execute()
+	if err != nil {
+		return err
```
alsamylkin: I don't understand the wrapping strategy :)

modules/task/execution_test.go
alsamylkin: Also a function with variadic arguments..

modules/task/task.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+}
+
+func newAsyncModule(mi service.ModuleCreateInfo) (service.Module, error) {
+	backend, err := _backendRegisterFn(mi.Host, Config{})
+	if err != nil {
+		return nil, err
+	}
+	_globalBackendMu.Lock()
+	_globalBackend = backend
```
alsamylkin: So if we are going to create 2 async modules one after another, we are going to overwrite the global backend? This seems dangerous.

modules/task/task.go
```diff
@@ -0,0 +1,97 @@
...
+// ModuleType represents the task module type
+const ModuleType = "task"
+
+type globalBackend struct {
+	backend Backend
+	sync.RWMutex
+}
+
+var (
+	_backendRegisterFn backendRegisterFn
```
alsamylkin: Shouldn't it has a default value, if the function is never registered?

modules/task/task.go
```diff
@@ -0,0 +1,97 @@
...
+	defer _globalBackendMu.RUnlock()
+	return _globalBackend
+}
+
+// NewModule creates an async task queue module
+func NewModule() service.ModuleCreateFunc {
+	return func(mi service.ModuleCreateInfo) ([]service.Module, error) {
+		mod, err := newAsyncModule(mi)
+		if err != nil {
+			return nil, errors.Wrap(err, "unable to instantiate async task module")
```
alsamylkin: errors.Wrap is noop on nil errors, so you can replace this if.

modules/task/execution.go
madhuravi: We should also balance simplicity and adding new features. Not ruling this out but let's get this out, see how people use it and add it down the lane when we have a better understanding of how folks are using it.

modules/task/backend.go
```diff
@@ -0,0 +1,130 @@
...
+func (b NopBackend) Consume() error {
+	return nil
+}
+
+// Encoder implements the Backend interface
+func (b NopBackend) Encoder() Encoding {
+	return &NopEncoding{}
+}
+
+// Type implements the Backend interface
```
madhuravi: Yes, we want Start, Stop and the like in backends as well since they will have initialization/setup and clean up actions like closing channels etc.

modules/task/backend.go
```diff
@@ -0,0 +1,130 @@
...
+func (b NopBackend) Type() string {
+	return ""
```
madhuravi: It has name and type methods. It doesn't really matter what string we return. Changed this to return "nop"

modules/task/backend.go
```diff
@@ -0,0 +1,130 @@
...
+	return nil
+}
+
+// Consume  implements the Backend interface
+func (b *InMemBackend) Consume() error {
+	select {
+	case v, ok := <-b.bufQueue:
+		if ok {
+			return Run(v)
+		}
```
madhuravi: it's a valid state if channel does not have anything available

modules/task/backend.go
```diff
@@ -0,0 +1,130 @@
...
+	default:
+		// No value ready in channel, moving on
```
madhuravi: This would be done in start, roughly

go func() {
  Consume()
  sleep
}

modules/task/encoding.go
```diff
@@ -0,0 +1,83 @@
...
+import (
+	"bytes"
+	"encoding/gob"
+
+	"github.com/pkg/errors"
+)
+
+// Encoding is capable of encoding and decoding objects
+type Encoding interface {
+	Register(interface{}) error
```
madhuravi: unfortunate but what's the alternative. I don't want if statements for gob in the code. other encodings can make this a noop.

modules/task/execution.go
madhuravi: Execute doesnt need to know that, it simply runs the function. We validate this upstream.

modules/task/execution.go
madhuravi: Good catch. This is another reason they need to be explicitly registered before they are enqueued. Going back to earlier state where Register needs to be called.

modules/task/backend_test.go
```diff
@@ -0,0 +1,47 @@
...
+		testBackendMethods(t, b)
+	}
+}
+
+func testBackendMethods(t *testing.T, backend Backend) {
+	assert.NotNil(t, backend.Name())
```
madhuravi: changed to notempty

modules/task/execution.go
madhuravi: they have a detailed error message in the same file, wrapping is not necessary. Wrapping is usually to add context when you receive an error from a downstream library or package so you know in what context that error is thrown.

modules/task/execution.go
madhuravi: Explained above.

modules/task/task.go
madhuravi: I don't think so, we don't do this anywhere in UberFx and it makes sense. We should fail if it's not set.

modules/task/task.go
madhuravi: yes but we return different values for the modules also based on the if.

modules/task/task.go
madhuravi: Like I said before, it does not support multiple backends right now. This behavior will change a bit with multiple backends. Add to a slice etc.

modules/task/backend_test.go
madhuravi: I'm following Dave Cheney's blog on table driven tests https://dave.cheney.net/2013/06/09/writing-table-driven-tests-in-go

glibsm: @alsamylkin `t.Run` is sub-tests, not table tests.
@madhuravi since the post in 2013, they introduced a new concept called sub-tests https://blog.golang.org/subtests.

We should decide for ourselves which one to use, but subtests make the output very nice

modules/task/execution_test.go
madhuravi: Good catch! Variadic arguments don't work. I think that's an ok restriction to have for now. Will add a task to follow up on that.

modules/task/backend_test.go
madhuravi: Agreed, idea of subtests is nice. This is literally testing two structs and I don't foresee very many different cases being added. Feels like a lot of overhead for simple tests. I'll keep this in mind for tests where test cases will build up.

modules/task/backend.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// Consume  implements the Backend interface
+func (b *inMemBackend) Consume() error {
+	select {
+	case v, ok := <-b.bufQueue:
+		if ok {
+			return Run(v)
+		}
+		return fmt.Errorf("The bufQueue channel has been closed")
```
alsamylkin: errors.New()?

modules/task/backend.go
```diff
@@ -0,0 +1,141 @@
...
+
+// Stop implements the Module interface
+func (b *inMemBackend) Stop() error {
+	close(b.bufQueue)
+	return nil
+}
+
+// IsRunning implements the Module interface
+func (b *inMemBackend) IsRunning() bool {
+	return true
```
alsamylkin: Nit, Even after stop?:)

modules/task/execution.go
```diff
@@ -0,0 +1,183 @@
...
+
+// Enqueue sends a func before sending to the task queue
+func Enqueue(fn interface{}, args ...interface{}) error {
+	// Is function registered
+	fnName := getFunctionName(fn)
+	fnLookup.RLock()
+	_, ok := fnLookup.fnNameMap[fnName]
+	fnLookup.RUnlock()
+	if !ok {
+		return fmt.Errorf("function: %q not found. Did you forget to register?", fnName)
```
alsamylkin: nit: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

modules/task/execution.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+func castToError(value reflect.Value) error {
+	if value.IsNil() {
+		return nil
+	}
+	if v, ok := value.Interface().(error); ok {
+		return v
+	}
+	return fmt.Errorf("Expected return value to be error but found: %s", value.Interface())
```
alsamylkin: Nit, error formatting.

modules/task/execution_test.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+type Car struct {
+	Brand string
+	Year  int
+}
+
+func Complex(car Car) error {
+	if car.Brand == "infinity" {
+		return fmt.Errorf("Complex error")
```
alsamylkin: Same error nit

alsamylkin: Also use errors.New?

modules/task/execution_test.go
```diff
@@ -0,0 +1,201 @@
...
+	wg.Add(1)
+	go func() {
+		i := 1
+		defer wg.Done()
+		fn = func() error {
+			i = i + 1
+			if i == 2 {
+				return nil
+			}
+			return fmt.Errorf("Unexpected i")
```
alsamylkin: nit, formatting and errors.New?

modules/task/backend.go
madhuravi: Added a boolean member

modules/task/execution.go
madhuravi: not sure what you're referring to, it's already lowercase.

alsamylkin: No punctuation marks in the end

madhuravi: Feels a bit literal. I think a question mark here makes sense for what the message. It will still play well with any following colons or anything else. I see most of this section as guidelines and see how far we should take them in a given context. Open to discussion on this though. We can chat tomorrow if you feel strongly about it.

modules/task/backend_test.go
glibsm: @alsamylkin My point was to mention that table driven test is something that came first, later to be extended and improved by subtests.

PR #185, opened 2017-01-24
Update make timeout for sentry tests

anuptalwalkar:
GFM-280
Sentry tests are slow and causing test timeout while running on travis. (try `make test RACE=-race` on local machine)


glibsm:
Yikes, those tests are my fault. I'll see what is wrong and how we can speed that up

sectioneight:
See attached flame graph on associated ticket. It's doing some crypto.

PR #187, opened 2017-01-25
Fix jaeger to zap logger conversion

madhuravi:
Earlier I was translating the formatted string as a structured zap message. Converting it into a string and logging that is way cleaner.

### DIFF
--- a/tracing/tracer_test.go
+++ b/tracing/tracer_test.go
@@ -94,12 +94,12 @@ func TestJaegerLogger(t *testing.T) {
 	testutils.WithInMemoryLogger(t, nil, func(zapLogger zap.Logger, buf *testutils.TestBuffer) {
 		loggerWithZap := ulog.Builder().SetLogger(zapLogger).Build()
 		jLogger := jaegerLogger{log: loggerWithZap}
-		jLogger.Infof("info message", "oddArg")
-		jLogger.Infof("info message", "value1", "value2")
+		jLogger.Infof("info message: %s", "oddArg")
+		jLogger.Infof("info message: %s %s", "value1", "value2")
 		jLogger.Error("error message")
 		assert.Equal(t, []string{
-			`{"level":"info","msg":"info message","arg":"oddArg"}`,
-			`{"level":"info","msg":"info message","arg":"value1","arg":"value2"}`,
+			`{"level":"info","msg":"info message: oddArg"}`,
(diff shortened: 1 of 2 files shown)


tracing/tracer_test.go
```diff
@@ -94,12 +94,12 @@ func TestJaegerLogger(t *testing.T) {
 	testutils.WithInMemoryLogger(t, nil, func(zapLogger zap.Logger, buf *testutils.TestBuffer) {
 		loggerWithZap := ulog.Builder().SetLogger(zapLogger).Build()
 		jLogger := jaegerLogger{log: loggerWithZap}
-		jLogger.Infof("info message", "oddArg")
-		jLogger.Infof("info message", "value1", "value2")
+		jLogger.Infof("info message: %s", "oddArg")
```
alsamylkin: Could you log a message without args?

sectioneight:
Wait, this is better how? Structured logging is what we should be preferring. Can you explain the rationale behind this?

sectioneight:
@yurishkuro thoughts?

madhuravi:
@sectioneight jaeger does not use structured logging, We were force-converting a formatted log to a structured log and the end result would not look like what it was intended for.

Eg. say jaeger does
log.Infof("Span closed: %d", span.ID())

Before this PR (converting to structured logs):
{"Span closed: %d", "arg", "1234"}

After this PR (formatted log):
{"Span closed: 1234"}

@akshayjshah originally pointed out this bug in the issue here https://github.com/uber/jaeger-client-go/issues/77#issuecomment-273906390. That should give you more context.

sectioneight:
IMO the ideal is this:

```
{"Span closed", "spanID", 1234}
```

is that not possible with the jaeger APIs? In other words, string interpolation in logs is almost never a good idea.

madhuravi:
Agreed but seems like jaeger APIs do not follow this today. The issue I referenced was for a similar sentiment. Yuri can comment further.

akshayjshah:
Even just looking at the test cases, this seems like an immediate change for the better. We shouldn't intentionally be writing JSON messages with duplicate keys. Even if we number the fields, ("arg1", "arg2", etc.) we'll definitely break ElasticSearch indexing (since those keys won't have values with a consistent type).

Agreed that the ideal is proper structured logging. If Yuri thinks that's worth the interface change, great; if not, this is fine. Realistically, we're going to interact with lots of libraries that use string interpolation in their logs.

PR #188, opened 2017-01-25
Fix examples build

glibsm:
Examples makefile had the wrong syntax for the default target
and wasn't building the server component.

After fixing that, there was a small chain reaction with examples
not bothering to update the system thriftrw.

When running uberfx `make test`, run through full examples
makefile, rather than codegen to find failures early.

### DIFF
--- a/.build/deps.mk
+++ b/.build/deps.mk
@@ -4,8 +4,8 @@ libdeps:
 	$(ECHO_V)glide --version 2>/dev/null ||	go get -v github.com/Masterminds/glide
 	$(ECHO_V)glide install
 
-.PHONY: dependencies
-dependencies: libdeps
+.PHONY: deps
+deps: libdeps
 	@$(call label,Installing test dependencies...)
 	$(ECHO_V)go install ./vendor/github.com/axw/gocov/gocov
 	$(ECHO_V)go install ./vendor/github.com/mattn/goveralls
--- a/examples/keyvalue/Makefile
+++ b/examples/keyvalue/Makefile
@@ -1,3 +1,5 @@
+.DEFAULT_GOAL := all
+
 kv/types.go: kv.thrift
 	$(ECHO_V)go generate
 
@@ -6,5 +8,3 @@ server/server: $(wildcard server/*.go) kv/types.go
 
 .PHONY: all
 all: server/server
-
-.DEFAULT_GOAL: all
--- a/examples/keyvalue/Makefile
+++ b/examples/keyvalue/Makefile
@@ -1,6 +1,6 @@
 .DEFAULT_GOAL := all
 
-kv/types.go: kv.thrift
+kv/types.go: deps kv.thrift
 	$(ECHO_V)go generate
 
 server/server: $(wildcard server/*.go) kv/types.go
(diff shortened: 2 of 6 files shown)


examples/keyvalue/Makefile
```diff
@@ -7,4 +9,7 @@ server/server: $(wildcard server/*.go) kv/types.go
 .PHONY: all
 all: server/server
 
-.DEFAULT_GOAL: all
+.PHONY: deps
+deps:
+	$(ECHO_V)go get -v go.uber.org/thriftrw
```
anuptalwalkar: Wait. I though we were installing from ../../vendor/ and not do go get? This will lead to failures once yarpc update happens.

.build/deps.mk
```diff
@@ -4,8 +4,8 @@ libdeps:
 	$(ECHO_V)glide --version 2>/dev/null ||	go get -v github.com/Masterminds/glide
 	$(ECHO_V)glide install
 
-.PHONY: dependencies
-dependencies: libdeps
+.PHONY: deps
```
sectioneight: why this change?

examples/keyvalue/Makefile
```diff
@@ -7,4 +9,7 @@ server/server: $(wildcard server/*.go) kv/types.go
...
+	$(ECHO_V)go get -v go.uber.org/yarpc/encoding/thrift/thriftrw-plugin-yarpc
```
sectioneight: This is wrong. We need to use the version that matches what we have in glide.yaml, otherwise the type signatures are going to mismatch. The YARPC team noticed and fixed this for us while you were away in december.

examples/keyvalue/Makefile
glibsm: yarpc is stable now, no?

.build/deps.mk
glibsm: `make deps` is in examples, and `make dependencies` is annoying to type, especially since it doesn't autocomplete from the included .mk file :(

examples/keyvalue/Makefile
glibsm: ooookaaaay. @anuptalwalkar made the same point

examples/keyvalue/Makefile
sectioneight: It is now, but what happens when 2.0 comes out and we're using 1.0 in glide, but this fetches the latest (2.0)?

glide.yaml
```diff
@@ -26,6 +26,9 @@ import:
   subpackages:
   - assert
   - require
+- package: github.com/golang/mock
```
anuptalwalkar: Does this mean anyone who needs thriftrw, will also need mock package?

glibsm: By the looks of it, yes. It doesn't work locally if you remove the `mock` package from system path. 

Looks like just including thriftrw doesn't pull in the mock in the lockfile.

/cc @breerly

sectioneight:
We shouldn't need mocks in production code, right?

glide.yaml
anuptalwalkar: yikes. I was hoping glide up would pull the package. Oh well.

glibsm:
@sectioneight thrift codegen uses mocks, so yeah we need it in production by the looks of it. Unless something is wrong with our yarpc setup. I cc'd @breerly here so we can comment as well.

/bcc @abhinav

abhinav:
Starting 1.1, the YARPC ThriftRW plugin generates gomock mocks by default in a
separate subpackage (`${serviceName}test`). To disable this, you can use the
`-no-gomock` flag on the plugin.

    thriftrw -p 'yarpc -no-gomock' foo.thrift

Generally, it shouldn't be a problem to have that code in your repo though.
You don't actually have a dependency on `gomock` until you use that package.
`glide` won't even add the dependency to your glide.lock unless you're
actually importing the test package.

sectioneight:
Right, so we shouldn't be having to add gomock as an explicit dependency,
right?

abhinav:
Sounds right.

glide.yaml
sectioneight: According to the convo with @abhinav this shouldn't be necessary. Please convince me it is. I don't like this.

PR #189, opened 2017-01-26
Added metrics around http filters and handles

anuptalwalkar:
GFM-247

Some cleanup around how filters are initialized, so we don't have to `fxcontext.Context{}` everywhere. 
```
Basic timers added - 
http.METHOD.time
http.request.auth

Basic Counter  - http.request.status.STATUS_CODE
```

### DIFF
--- a/modules/module_tags.go
+++ b/modules/module_tags.go
@@ -21,12 +21,16 @@
 package modules
 
 var (
-	moduleTag = "Module"
-	typeTag   = "Type"
+	//TagModule is module tag for metrics
+	TagModule = "module"
+	// TagType is either request or response
+	TagType = "type"
+	// TagStatus is status of the request
+	TagStatus = "status"
 )
 
 // HTTPTags creates metrics scope with defined tags
 var HTTPTags = map[string]string{
-	moduleTag: "http",
-	typeTag:   "request",
+	TagModule: "http",
+	TagType:   "request",
 }
--- a/modules/module_tags.go
+++ b/modules/module_tags.go
@@ -20,7 +20,7 @@
 
 package modules
 
-var (
+const (
 	//TagModule is module tag for metrics
 	TagModule = "module"
 	// TagType is either request or response
--- a/modules/module_tags.go
+++ b/modules/module_tags.go
@@ -1,0 +1,32 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package modules
+
+var (
+	moduleTag = "Module"
+	typeTag   = "Type"
+)
+
+// HTTPTags creates metrics scope with defined tags
+var HTTPTags = map[string]string{
+	moduleTag: "http",
+	typeTag:   "request",
+}
--- a/modules/uhttp/filterchain_builder.go
+++ b/modules/uhttp/filterchain_builder.go
@@ -52,7 +52,7 @@ type filterChainBuilder struct {
 
 func defaultFilterChainBuilder(host service.Host) filterChainBuilder {
 	fcb := newFilterChainBuilder(host)
-	return fcb.AddFilters(contextFilter(host), panicFilter(host), tracingServerFilter(host), authorizationFilter(host))
+	return fcb.AddFilters(panicFilter{host}, metricsFilter{host}, tracingServerFilter{host}, authorizationFilter{host})
 }
 
 // NewFilterChainBuilder creates an empty filterChainBuilder for setup
--- a/modules/uhttp/filterchain_builder.go
+++ b/modules/uhttp/filterchain_builder.go
@@ -24,6 +24,7 @@ import (
 	"net/http"
 
 	"go.uber.org/fx"
+	"go.uber.org/fx/modules"
 	"go.uber.org/fx/service"
 )
 
@@ -52,7 +53,14 @@ type filterChainBuilder struct {
 
 func defaultFilterChainBuilder(host service.Host) filterChainBuilder {
 	fcb := newFilterChainBuilder(host)
-	return fcb.AddFilters(panicFilter{host}, metricsFilter{host}, tracingServerFilter{host}, authorizationFilter{host})
+	scope := host.Metrics().Tagged(modules.HTTPTags)
+	return fcb.AddFilters(panicFilter{scope},
+		metricsFilter{scope},
+		tracingServerFilter{scope},
+		authorizationFilter{
+			scope:      scope,
+			authClient: host.AuthClient(),
+		})
 }
 
 // NewFilterChainBuilder creates an empty filterChainBuilder for setup
--- a/modules/uhttp/filterchain_builder.go
+++ b/modules/uhttp/filterchain_builder.go
@@ -54,7 +54,8 @@ type filterChainBuilder struct {
 func defaultFilterChainBuilder(host service.Host) filterChainBuilder {
 	fcb := newFilterChainBuilder(host)
 	scope := host.Metrics().Tagged(modules.HTTPTags)
-	return fcb.AddFilters(panicFilter{scope},
+	return fcb.AddFilters(
+		panicFilter{scope},
 		metricsFilter{scope},
 		tracingServerFilter{scope},
 		authorizationFilter{
--- a/modules/uhttp/filterchain_builder.go
+++ b/modules/uhttp/filterchain_builder.go
@@ -55,12 +55,12 @@ func defaultFilterChainBuilder(host service.Host) filterChainBuilder {
 	fcb := newFilterChainBuilder(host)
 	scope := host.Metrics().Tagged(modules.HTTPTags)
 	return fcb.AddFilters(
-		panicFilter{scope},
+		panicFilter{scope.Counter("panic")},
 		metricsFilter{scope},
 		tracingServerFilter{scope},
 		authorizationFilter{
-			scope:      scope,
-			authClient: host.AuthClient(),
+			authClient:  host.AuthClient(),
+			authCounter: scope.Counter("auth.fail"),
 		})
 }
 
--- a/modules/uhttp/filters.go
+++ b/modules/uhttp/filters.go
@@ -105,7 +105,7 @@ func (f panicFilter) Apply(ctx fx.Context, w http.ResponseWriter, r *http.Reques
 	defer func() {
 		if err := recover(); err != nil {
 			ctx.Logger().Error("Panic recovered serving request", "error", err, "url", r.URL)
-			f.Host.Metrics().Counter("http.panic").Inc(1)
+			f.Host.Metrics().SubScope("http").Counter("panic").Inc(1)
 			w.Header().Add(ContentType, ContentTypeText)
 			w.WriteHeader(http.StatusInternalServerError)
 			fmt.Fprintf(w, "Server error: %+v", err)
--- a/modules/uhttp/filters.go
+++ b/modules/uhttp/filters.go
@@ -104,7 +104,6 @@ func (f panicFilter) Apply(ctx fx.Context, w http.ResponseWriter, r *http.Reques
 			w.Header().Add(ContentType, ContentTypeText)
 			w.WriteHeader(http.StatusInternalServerError)
 			fmt.Fprintf(w, "Server error: %+v", err)
-			tally.DefaultSeparator
 		}
 	}()
 	next.ServeHTTP(ctx, w, r)
--- a/modules/uhttp/filters.go
+++ b/modules/uhttp/filters.go
@@ -64,10 +64,11 @@ func (f tracingServerFilter) Apply(ctx fx.Context, w http.ResponseWriter, r *htt
 	ext.HTTPUrl.Set(span, r.URL.String())
 	defer span.Finish()
 
-	ctx = ctx.(*fxcontext.Context).WithContextAwareLogger(span)
-	ctx = &fxcontext.Context{
+	yoctx = fxcontext.Context{
 		Context: opentracing.ContextWithSpan(ctx, span),
 	}
+	ctx = ctx.(*fxcontext.Context).WithContextAwareLogger(span)
+
 	r = r.WithContext(ctx)
 	next.ServeHTTP(ctx, w, r)
 }
@@ -103,6 +104,7 @@ func (f panicFilter) Apply(ctx fx.Context, w http.ResponseWriter, r *http.Reques
 			w.Header().Add(ContentType, ContentTypeText)
 			w.WriteHeader(http.StatusInternalServerError)
 			fmt.Fprintf(w, "Server error: %+v", err)
+			tally.DefaultSeparator
 		}
 	}()
 	next.ServeHTTP(ctx, w, r)
--- a/modules/uhttp/filters.go
+++ b/modules/uhttp/filters.go
@@ -28,10 +28,10 @@ import (
 	"go.uber.org/fx"
 	"go.uber.org/fx/auth"
 	"go.uber.org/fx/internal/fxcontext"
-	"go.uber.org/fx/service"
 
 	"github.com/opentracing/opentracing-go"
 	"github.com/opentracing/opentracing-go/ext"
+	"github.com/uber-go/tally"
 )
 
 // Filter applies filters on requests, request contexts or responses such as
@@ -49,11 +49,11 @@ func (f FilterFunc) Apply(ctx fx.Context, w http.ResponseWriter, r *http.Request
 }
 
 type tracingServerFilter struct {
-	service.Host
+	scope tally.Scope
 }
 
 func (f tracingServerFilter) Apply(ctx fx.Context, w http.ResponseWriter, r *http.Request, next Handler) {
-	stopWatch := f.Host.Metrics().SubScope("http").SubScope("request").Timer("tracing").Start()
+	stopWatch := f.scope.Timer("tracing").Start()
 	operationName := r.Method
 	carrier := opentracing.HTTPHeadersCarrier(r.Header)
 	spanCtx, err := opentracing.GlobalTracer().Extract(opentracing.HTTPHeaders, carrier)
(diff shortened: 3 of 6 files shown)


modules/uhttp/filters.go
```diff
@@ -48,76 +48,82 @@ func (f FilterFunc) Apply(ctx fx.Context, w http.ResponseWriter, r *http.Request
...
-			Context: ctx,
+type panicFilter struct {
+	service.Host
+}
+
+func (f panicFilter) Apply(ctx fx.Context, w http.ResponseWriter, r *http.Request, next Handler) {
+	defer func() {
+		if err := recover(); err != nil {
+			ctx.Logger().Error("Panic recovered serving request", "error", err, "url", r.URL)
+			f.Host.Metrics().Counter("http.panic").Inc(1)
```
alsamylkin: use scope?

modules/uhttp/filters.go
```diff
@@ -48,76 +48,82 @@ func (f FilterFunc) Apply(ctx fx.Context, w http.ResponseWriter, r *http.Request
...
-		}()
-
-		fxctx = fxctx.WithContextAwareLogger(span)
-		fxctx = &fxcontext.Context{
-			Context: opentracing.ContextWithSpan(fxctx, span),
-		}
-		r = r.WithContext(fxctx)
-		next.ServeHTTP(fxctx, w, r)
+func (f tracingServerFilter) Apply(ctx fx.Context, w http.ResponseWriter, r *http.Request, next Handler) {
+	stopWatch := f.Host.Metrics().SubScope("http").SubScope("request").Timer("tracing").Start()
```
alsamylkin: Should we cache the request scope in the filter chain?

alsamylkin: Should it also use a server scope?

alsamylkin: Also, should we use tags instead of scopes?

alsamylkin: Or may be in the filter itself..

modules/uhttp/filters_test.go
```diff
@@ -86,9 +86,8 @@ func TestTracingFilterWithLogs(t *testing.T) {
 func TestFilterChainFilters(t *testing.T) {
 	host := service.NopHost()
 	chain := newFilterChainBuilder(host).AddFilters(
-		contextFilter(host),
-		tracingServerFilter(host),
-		authorizationFilter(host)).Build(getNopHandler(host))
+		tracingServerFilter{host},
+		authorizationFilter{host}).Build(getNopHandler(host))
```
alsamylkin: Shouldn't we test that appropriate counters are incremented?

modules/uhttp/filters.go
anuptalwalkar: Its a bit tricky, unless we pass the scope down each filter to make it available. It doesn't belong in a context as well. Since its easily accessible through global, I think this approach is works.

anuptalwalkar: What does that benefit to metrics? Isn't all metrics are part of server? Client metrics on the other hand are usually explicitly specified.

modules/uhttp/filters.go
anuptalwalkar: I only refactored. Fixing it.

modules/uhttp/filters_test.go
anuptalwalkar: Tests are added in http_test.

modules/uhttp/filters.go
anuptalwalkar: I played around with tags, I am leaving Scope for now. Tags are getting messy. We can circle back in the future if we want to provide tags through out the framework.

modules/module_tags.go
```diff
@@ -0,0 +1,32 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package modules
+
+var (
+	moduleTag = "Module"
```
alsamylkin: I think tags are usually lowercase.

modules/uhttp/filterchain_builder.go
```diff
@@ -52,7 +53,14 @@ type filterChainBuilder struct {
 
 func defaultFilterChainBuilder(host service.Host) filterChainBuilder {
 	fcb := newFilterChainBuilder(host)
-	return fcb.AddFilters(contextFilter(host), panicFilter(host), tracingServerFilter(host), authorizationFilter(host))
+	scope := host.Metrics().Tagged(modules.HTTPTags)
+	return fcb.AddFilters(panicFilter{scope},
```
madhuravi: nit: formatting
fcb.AddFilters(
  filter 1,
  filter 2
  ..
)

modules/uhttp/filters.go
```diff
@@ -48,76 +48,83 @@ func (f FilterFunc) Apply(ctx fx.Context, w http.ResponseWriter, r *http.Request
...
+	carrier := opentracing.HTTPHeadersCarrier(r.Header)
+	spanCtx, err := opentracing.GlobalTracer().Extract(opentracing.HTTPHeaders, carrier)
+	if err != nil && err != opentracing.ErrSpanContextNotFound {
+		ctx.Logger().Info("Malformed inbound tracing context: ", "error", err.Error())
 	}
+	span := opentracing.GlobalTracer().StartSpan(operationName, ext.RPCServerOption(spanCtx))
+	ext.HTTPUrl.Set(span, r.URL.String())
+	defer func() {
+		span.Finish()
+		ctx.Logger().Debug("Span finished")
```
madhuravi: Not related but this debug statement can be removed now.

madhuravi: so you can just do
defer span.Finish()

modules/uhttp/handler.go
```diff
@@ -59,5 +59,7 @@ type handlerWrapper struct {
 // ServeHTTP calls Handler.ServeHTTP(ctx, w, r) and injects a new service context for use.
 func (h *handlerWrapper) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	ctx := fxcontext.New(context.Background(), h.host)
+	stopwatch := h.host.Metrics().SubScope("http").SubScope("handle").Timer("sla").Start()
```
madhuravi: We should definitely add the operation name into the scope. Also don't feel "sla" is a good name here. It does not really depict sla at all. Maybe latency?

modules/uhttp/filters.go
```diff
@@ -48,76 +48,83 @@ func (f FilterFunc) Apply(ctx fx.Context, w http.ResponseWriter, r *http.Request
...
-		}()
-
-		fxctx = fxctx.WithContextAwareLogger(span)
-		fxctx = &fxcontext.Context{
-			Context: opentracing.ContextWithSpan(fxctx, span),
-		}
-		r = r.WithContext(fxctx)
-		next.ServeHTTP(fxctx, w, r)
+func (f tracingServerFilter) Apply(ctx fx.Context, w http.ResponseWriter, r *http.Request, next Handler) {
+	stopWatch := f.scope.Timer("tracing").Start()
```
madhuravi: Are we convinced we want to measure latencies for each of the filters? It seems useful but more metrics add latency overhead too. We should think about which ones we need.

modules/uhttp/filters.go
```diff
@@ -48,76 +48,83 @@ func (f FilterFunc) Apply(ctx fx.Context, w http.ResponseWriter, r *http.Request
...
-			fxctx.Logger().Error(auth.ErrAuthorization, "error", err)
-			w.WriteHeader(http.StatusUnauthorized)
-			fmt.Fprintf(w, "Unauthorized access: %+v", err)
-			return
-		}
-		next.ServeHTTP(fxctx, w, r)
+func (f authorizationFilter) Apply(ctx fx.Context, w http.ResponseWriter, r *http.Request, next Handler) {
+	stopWatch := f.scope.Timer("auth").Start()
+	if err := f.authClient.Authorize(ctx); err != nil {
+		f.scope.SubScope("auth").Counter("fail").Inc(1)
```
alsamylkin: Why do we need a subscope here?

alsamylkin: BTW, you can also tag error in the counter and get distribution by error types. Same for the panic filter.

modules/uhttp/handler.go
madhuravi: Not sure if this it but we need overall latency metrics (including filters and actual execution).
Also need a metric for response status codes.

sectioneight:
To echo Madhu's comments, I don't think we want metrics on each filter just
yet

modules/uhttp/filters.go
alsamylkin: It can be dangerous though, if there are lots of different errors, we can create too much load on metrics backend.. So may be only panics then?

glibsm: subscopes should never be created in the hot loop - it's a pretty expensive operation.

You can see here that it does a lot of copying: https://github.com/uber-go/tally/blob/master/scope.go#L309

We should either cache the sub-scope on some object, or figure out a way to emit metrics without creating subscopes for each filter operation

modules/uhttp/handler.go
glibsm: do not chain subscopes: do `http.handle` in one go so we don't copy stuff around without need

modules/uhttp/filters.go
anuptalwalkar: done

modules/uhttp/handler.go
anuptalwalkar: Yes, this is the overall latency metrics. This is the point where we receive request from outside world, and respond.

modules/uhttp/filters.go
anuptalwalkar: I added for auth because even if its a library call, we would want to understand the performance over time. I think I'll remove it for rest of the filters.

madhuravi: Doesn't auth (galileo) generate metrics that we can look at instead?

modules/module_tags.go
```diff
@@ -0,0 +1,36 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package modules
+
+var (
```
sectioneight: const?

modules/uhttp/filters.go
```diff
@@ -48,76 +49,71 @@ func (f FilterFunc) Apply(ctx fx.Context, w http.ResponseWriter, r *http.Request
...
-			Context: opentracing.ContextWithSpan(fxctx, span),
-		}
-		r = r.WithContext(fxctx)
-		next.ServeHTTP(fxctx, w, r)
+func (f tracingServerFilter) Apply(ctx fx.Context, w http.ResponseWriter, r *http.Request, next Handler) {
+	operationName := r.Method
+	carrier := opentracing.HTTPHeadersCarrier(r.Header)
+	spanCtx, err := opentracing.GlobalTracer().Extract(opentracing.HTTPHeaders, carrier)
+	if err != nil && err != opentracing.ErrSpanContextNotFound {
+		ctx.Logger().Info("Malformed inbound tracing context: ", "error", err.Error())
```
sectioneight: is this really just an info? should this be warning so it goes to sentry?

modules/uhttp/filters.go
```diff
@@ -48,76 +49,71 @@ func (f FilterFunc) Apply(ctx fx.Context, w http.ResponseWriter, r *http.Request
...
-			Context: ctx,
+type panicFilter struct {
+	scope tally.Scope
+}
+
+func (f panicFilter) Apply(ctx fx.Context, w http.ResponseWriter, r *http.Request, next Handler) {
+	defer func() {
+		if err := recover(); err != nil {
+			ctx.Logger().Error("Panic recovered serving request", "error", err, "url", r.URL)
+			f.scope.Counter("panic").Inc(1)
```
sectioneight: i don't recall exactly how these work, but i thought the general guidance was to avoid instantiating counters on-demand, and instead frontload the allocation

modules/uhttp/filters.go
```diff
@@ -48,76 +49,71 @@ func (f FilterFunc) Apply(ctx fx.Context, w http.ResponseWriter, r *http.Request
...
+	next.ServeHTTP(ctx, w, r)
+}
+
+// metricsFilter adds any default metrics related to HTTP
+type metricsFilter struct {
+	scope tally.Scope
+}
+
+func (f metricsFilter) Apply(ctx fx.Context, w http.ResponseWriter, r *http.Request, next Handler) {
+	defer f.scope.Tagged(map[string]string{modules.TagStatus: w.Header().Get("Status")}).Counter("total").Inc(1)
```
sectioneight: what does status mean here? status code? is it an integer or a string?

modules/uhttp/handler.go
```diff
@@ -59,5 +59,7 @@ type handlerWrapper struct {
 // ServeHTTP calls Handler.ServeHTTP(ctx, w, r) and injects a new service context for use.
 func (h *handlerWrapper) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	ctx := fxcontext.New(context.Background(), h.host)
+	stopwatch := h.host.Metrics().Timer("http." + r.Method + ".latency").Start()
```
sectioneight: See email today about potentially using histograms

sectioneight: Also is "latency" really the right term? This isn't waiting, this is execution time

modules/uhttp/http_test.go
```diff
@@ -49,6 +50,7 @@ func TestNew_OK(t *testing.T) {
 	WithService(New(registerNothing, nil), nil, []service.Option{configOption()}, func(s service.Owner) {
 		assert.NotNil(t, s, "Should create a module")
 	})
+
```
sectioneight: remove?

modules/uhttp/filters.go
anuptalwalkar: Status is a http header with status code value parsed as a string. Here, it would be "200" or "500" etc.

modules/uhttp/filters.go
anuptalwalkar: no, counter is on internal repo. I think ideally we want to keep counters in fx, but for short time till internal auth is ready, I have it setup for both.

modules/uhttp/handler.go
anuptalwalkar: update to "time". I have also used "all" previously, but "time" makes more sense here.

anuptalwalkar: I am keeping time for now, but we can discuss offline if we want to update.

modules/uhttp/filters.go
anuptalwalkar: I moved the allocation out of request path.

anuptalwalkar:
I think I have addressed all outstanding blockers. Mind taking a look?

modules/uhttp/filters.go
```diff
@@ -48,76 +49,71 @@ func (f FilterFunc) Apply(ctx fx.Context, w http.ResponseWriter, r *http.Request
...
+	carrier := opentracing.HTTPHeadersCarrier(r.Header)
+	spanCtx, err := opentracing.GlobalTracer().Extract(opentracing.HTTPHeaders, carrier)
+	if err != nil && err != opentracing.ErrSpanContextNotFound {
+		ctx.Logger().Warn("Malformed inbound tracing context: ", "error", err.Error())
 	}
+	span := opentracing.GlobalTracer().StartSpan(operationName, ext.RPCServerOption(spanCtx))
+	ext.HTTPUrl.Set(span, r.URL.String())
+	defer span.Finish()
+
+	ctx = ctx.(*fxcontext.Context).WithContextAwareLogger(span)
```
alsamylkin: Shouldn't we do this after line 68 and avoid type assertion?

modules/uhttp/filters.go
alsamylkin: I am really curious how much does this counter costs us.. Did you try benchmarking it?

modules/uhttp/filterchain_builder.go
```diff
@@ -52,7 +53,15 @@ type filterChainBuilder struct {
...
 	fcb := newFilterChainBuilder(host)
-	return fcb.AddFilters(contextFilter(host), panicFilter(host), tracingServerFilter(host), authorizationFilter(host))
+	scope := host.Metrics().Tagged(modules.HTTPTags)
+	return fcb.AddFilters(
+		panicFilter{scope.Counter("panic")},
+		metricsFilter{scope},
+		tracingServerFilter{scope},
+		authorizationFilter{
+			authClient:  host.AuthClient(),
+			authCounter: scope.Counter("auth.fail"),
```
alsamylkin: We shouldn't use a dot as a scope separator..

anuptalwalkar: Why not? Its the default separator in Tally anyway.

glibsm: I think I'll take the fall for that, I suggested to use the dot.

The problem with doing two scopes in a row is that there is a lot of allocations and copying happening under the hood.

We should consider making it easier to create subscopes within tally, perhaps taking in a variadic list rather than a single string.

alsamylkin: I think it is called on creation only, so no need to worry about extra allocations.

Well, there is a tally.DefaultSeparator, not a '.' :)

modules/uhttp/filters.go
```diff
@@ -48,76 +49,71 @@ func (f FilterFunc) Apply(ctx fx.Context, w http.ResponseWriter, r *http.Request
...
-			fmt.Fprintf(w, "Unauthorized access: %+v", err)
-			return
-		}
-		next.ServeHTTP(fxctx, w, r)
+func (f authorizationFilter) Apply(ctx fx.Context, w http.ResponseWriter, r *http.Request, next Handler) {
+	if err := f.authClient.Authorize(ctx); err != nil {
+		f.authCounter.Inc(1)
+		ctx.Logger().Error(auth.ErrAuthorization, "error", err)
+		w.WriteHeader(http.StatusUnauthorized)
+		fmt.Fprintf(w, "Unauthorized access: %+v", err)
```
madhuravi: Remove?

anuptalwalkar: Why? This is writing response body to the user.

modules/uhttp/filterchain_builder.go
anuptalwalkar: Okey. `tally.DefaultSeparator` is a "." It would be a pretty huge breaking change for tally to switch it out to something else. 
 I don't mind having to subScope everything, or keeping in one string, but IMAO having a "." allows readability to users.

modules/uhttp/filters.go
anuptalwalkar: Good idea. I'll move it down, but needs type assertion anyway.

modules/uhttp/filterchain_builder.go
alsamylkin: One can use the prometheus as backend and the use '_' for qualified names: https://godoc.org/github.com/prometheus/client_golang/prometheus#Opts
Up to you anyway.

modules/uhttp/filters.go
anuptalwalkar: Tally has its own benchmarks - https://github.com/uber-go/tally/. The only addition here is to extract header value from http, which is a map[string] get. I don't see any value adding beanchmark here, but we should benchmark overall modules performance.

alsamylkin: It has nothing todo with tally: we allocate a map(always) and create a counter on request. Overall perf would be great to see, I think tracing will be the most expensive part anyway.

anuptalwalkar: I'll add a separate task for perf, but this is an important metric for service health.

PR #190, opened 2017-01-26
Add initial docs for async task module

madhuravi:


anuptalwalkar:
@Lupie

modules/task/README.md
```diff
@@ -0,0 +1,49 @@
...
+```
+
+The async task module is a singleton and a service can intialize only one at this time.
+Users are free to define their own "backends" and "encodings" for message passing.
+
+## Async function restrictions
+
+The function to be invoked asynchronously has the following restrictions:
+* The function has to return only one value which should be an error
+* There is no way for the caller to receive a return value from the called
```
Lupie: - There must be? Or there is, and you want to implement that way in the future?

modules/task/README.md
```diff
@@ -0,0 +1,49 @@
...
+  return nil
+}
+```
+
+The async task module is a singleton and a service can intialize only one at this time.
+Users are free to define their own "backends" and "encodings" for message passing.
+
+## Async function restrictions
+
+The function to be invoked asynchronously has the following restrictions:
```
Lupie: When I read "restrictions," I think constraints, like some criteria that need to be met before async functions can work. But I think you might mean "limitations" or "drawbacks," something that is not quite right yet.

Not sure.

modules/task/README.md
```diff
@@ -0,0 +1,49 @@
...
+* The function has to return only one value which should be an error
+* There is no way for the caller to receive a return value from the called
+function
+* The function cannot have variadic arguments at this time. Support for this
```
Lupie: - I'd put "support for this coming soon" in parens so you don't have to add periods to all the list items.
- Cut "at this time" (implied by "coming soon")

modules/task/README.md
```diff
@@ -0,0 +1,49 @@
+# Async task module
+
+The async task module presents an distributed task execution framework
+for services that want to execute a function asynchronously and in a durable
+fashion. It achieves durability with "backends" that are messaging transports.
```
madhuravi: It's tied to the code. Also though technically it should be back ends, I've observed it being used as one word very frequently in other documentation.

modules/task/README.md
```diff
@@ -0,0 +1,49 @@
...
+To use the module, initialize it at service startup and register any functions
+that will be invoked async. Simply call task.Enqueue to call a function async
```
madhuravi: haha good point about simply ;)

modules/task/README.md
madhuravi: How about requirements?

modules/task/README.md
```diff
@@ -0,0 +1,49 @@
...
+and the execution framework will take care of running it at a later time.
```
madhuravi: The backend transport. I was trying to avoid too much detail here but you're right, people will wonder that. Will clarify.

modules/task/README.md
```diff
@@ -0,0 +1,49 @@
...
+}
+```
+
+The async task module is a singleton and a service can intialize only one at this time.
+Users are free to define their own "backends" and "encodings" for message passing.
+
+## Async function restrictions
+
+The function to be invoked asynchronously has the following restrictions:
+* The function has to return only one value which should be an error
```
madhuravi: This is  by design at this point.

modules/task/README.md
madhuravi: fixed.

modules/task/doc.go
```diff
@@ -0,0 +1,81 @@
...
+//   }
+//
+//
+//   func updateCache(input string, results string) error {
+//     // update cache with the name
+//     return nil
+//   }
+//
+// The async task module is a singleton and a service can intialize only one at this time.
+// Users are free to define their own "backends" and "encodings" for message passing.
```
Lupie: These quotes are more confusing than helpful because you've already defined backends (back in line 29). If you want to introduce _backend_ as a term more formally, you can italicize it in line 29.

The quotes here make me think you're referring to some new thing. It's like if I introduce an embedded tech writer program and then say that this "program" has been working well. By putting it in quotes, I make it sound like a fake program—something that's not quite a regular program—when really I'm just trying to refer to that specific program again. Hope that makes sense.

Once you introduce something, you can drop the quotes. We'll remember that it's a messaging transport.

modules/task/README.md
```diff
@@ -0,0 +1,54 @@
...
+  return nil
+}
+```
+
+The async task module is a singleton and a service can intialize only one at this time.
+Users are free to define their own "backends" and "encodings" for message passing.
+
+## Async function requirements
+
+The function to be invoked asynchronously needs to meet the following criteria:
```
Lupie: The function isn't actually meeting these criteria. Be more precise:

> For the function to be invoked asynchronously, the following criteria must be met:

modules/task/README.md
```diff
@@ -0,0 +1,54 @@
...
+* The function must return only one value which should be an error
```
Lupie: - Add comma before "which." This first bullet point still has ambiguous phrasing. The function must return one value, and that value must be an error? Every sentence in here should leave no room for misinterpretation.
- We also need to make these either "does" or "must" throughout. Otherwise it sounds like some of these bullets are optional requirements. That'll make the first two bullets either "The function returns only one…" and "The caller does not receive…" OR "The function must return only one…" and "The caller must not receive…." Does that make sense?

modules/task/README.md
```diff
@@ -0,0 +1,54 @@
...
+* The caller does not receive a return value from the called function
+* The function cannot take variadic arguments as input. Support for this is coming soon
```
Lupie: If any bullet is two sentences, all bullets need periods. You can either put "support for this is coming soon" in parens or add periods at the end of each bullet point. I'd go with parentheses.

modules/task/README.md
madhuravi: The reason for must vs does is different here. 

must - > user action i.e. user must do this
does -> this is the state of the world, no action from the user here

I changed it a little.

modules/task/README.md
```diff
@@ -44,11 +44,11 @@ func updateCache(input string, results string) error {
...
+Users are free to define their own backends and encodings for message passing.
 
 ## Async function requirements
 
-The function to be invoked asynchronously needs to meet the following criteria:
-* The function must return only one value which should be an error
-* The caller does not receive a return value from the called function
-* The function cannot take variadic arguments as input. Support for this is coming soon
\ No newline at end of file
+For he function to be invoked asynchronously, the following criteria must be met:
```
Lupie: he --> the

modules/task/README.md
```diff
@@ -44,11 +44,11 @@ func updateCache(input string, results string) error {
...
+* The function should return only one value, which should be an error. The caller does not receive a
+return value from the called function.
+* The function should not take variadic arguments as input (Support for this is coming soon).
```
Lupie: Support --> support

PR #191, opened 2017-01-26
Change task module to accept Backend as input and add tests

madhuravi:


modules/task/execution_test.go
```diff
@@ -199,3 +200,10 @@ func Complex(car Car) error {
 	}
 	return nil
 }
+
+func TestCastToError(t *testing.T) {
+	s := make(map[string]string)
+	err := castToError(reflect.ValueOf(s))
+	assert.Error(t, err)
```
alsamylkin: require?

modules/task/execution_test.go
```diff
@@ -199,3 +200,10 @@ func Complex(car Car) error {
...
+	assert.Contains(t, err.Error(), "be error but found")
```
alsamylkin: assert.EqualError?

modules/task/task.go
```diff
@@ -98,10 +87,3 @@ type Config struct {
 	broker  string `yaml:"broker"`
 	timeout string `yaml:"timeout"`
 }
-
-type backendRegisterFn func(service.Host, Config) (Backend, error)
-
-// RegisterAsyncBackend registers the backend for the async task module
-func RegisterAsyncBackend(backendRegisterFn backendRegisterFn) {
```
alsamylkin: Could you provide more context, why are you removing it? It seems pretty standard across modules to have a registration functions?

modules/task/execution_test.go
madhuravi: We decided not to be overly strict about error string comparisons. Using EqualError would mean I need the entire string from the error. That just makes it fragile when we change the wordig etc.

modules/task/task.go
madhuravi: Discussed this with you yesterday. Decided to make it a more explicit dependency injection to avoid traps when the function has not been set.

PR #194, opened 2017-01-27
Use tally.CachedStatsReporter replacing tally.StatsReporter

anuptalwalkar:


anuptalwalkar:
https://github.com/uber-go/tally/issues/23

tracing/tracer.go
```diff
@@ -83,20 +83,20 @@ func (jl *jaegerLogger) Infof(msg string, args ...interface{}) {
...
 
 type jaegerReporter struct {
-	reporter tally.StatsReporter
+	reporter tally.CachedStatsReporter
 }
 
 // IncCounter increments metrics counter TODO: Change to use scope with tally functions to increment/update
 func (jr *jaegerReporter) IncCounter(name string, tags map[string]string, value int64) {
-	jr.reporter.ReportCounter(name, tags, value)
+	jr.reporter.AllocateCounter(name, tags).ReportCount(value)
```
Raynos: recommend leaving TODO: about caching `AllocateCounter()` result. 
The reason the CachedStatsReporter is faster is because you can pre-allocate and cache the result of `AllocateCounter()` at startup.

PR #196, opened 2017-01-30
Lonely dispatcher

alsamylkin:
Make a single global dispatcher that will be shared across multiple rpc modules.

The dispatcher is created when any YARPC module is started. The expected use case is:
1. Create all modules
2. Start them(the very first start will actually start the dispatcher)
...
4. Receive a shutdown signal and stop the dispatcher.

We can't start and stop multiple times because some of the transports (e.g. [http](https://github.com/yarpc/yarpc-go/blob/dev/transport/http/outbound.go#L158)) are idempotent, so we have to work with the YARPC modules in the same way.

If a user needs to create a rpc client they can use the yarpc.Dispatcher().

### DIFF
--- a/modules/rpc/handler.go
+++ b/modules/rpc/handler.go
@@ -46,7 +46,6 @@ func (f UnaryHandlerFunc) Handle(ctx fx.Context, body wire.Value) (thrift.Respon
 }
 
 // WrapUnary wraps the unary handler and returns a thrift.UnaryHandlerFunc for yarpc calls
-// TODO(anup): GFM-255 Remove host parameter once updated yarpc plugin is imported in the repo
 func WrapUnary(h UnaryHandlerFunc) thrift.UnaryHandler {
 	return func(ctx context.Context, body wire.Value) (thrift.Response, error) {
 		fxctx := &fxcontext.Context{
@@ -70,7 +69,6 @@ func (f OnewayHandlerFunc) HandleOneway(ctx fx.Context, body wire.Value) error {
 }
 
 // WrapOneway wraps the oneway handler and returns a thrift.OnewayHandlerFunc for yarpc calls
-// TODO(anup): GFM-255 Remove host parameter once updated yarpc plugin is imported in the repo
 func WrapOneway(h OnewayHandlerFunc) thrift.OnewayHandler {
 	return func(ctx context.Context, body wire.Value) error {
 		fxctx := &fxcontext.Context{
@@ -130,6 +128,7 @@ func authorize(ctx context.Context, host service.Host) (fx.Context, error) {
 		host.Metrics().SubScope("rpc").SubScope("auth").Counter("fail").Inc(1)
 		fxctx.Logger().Error(auth.ErrAuthorization, "error", err)
 		// TODO(anup): GFM-255 update returned error to transport.BadRequestError (user error than server error)
+		// These ^ are internal YARPC errors, should we reference it?
 		return nil, err
 	}
 	return fxctx, nil
--- a/modules/rpc/handler.go
+++ b/modules/rpc/handler.go
@@ -128,7 +128,7 @@ func authorize(ctx context.Context, host service.Host) (fx.Context, error) {
 		host.Metrics().SubScope("rpc").SubScope("auth").Counter("fail").Inc(1)
 		fxctx.Logger().Error(auth.ErrAuthorization, "error", err)
 		// TODO(anup): GFM-255 update returned error to transport.BadRequestError (user error than server error)
-		// These ^ are internal YARPC errors, should we reference it?
+		// https://github.com/yarpc/yarpc-go/issues/687
 		return nil, err
 	}
 	return fxctx, nil
--- a/modules/rpc/thrift.go
+++ b/modules/rpc/thrift.go
@@ -38,7 +38,7 @@ type CreateThriftServiceFunc func(svc service.Host) ([]transport.Procedure, erro
 // ThriftModule creates a Thrift Module from a service func
 func ThriftModule(hookup CreateThriftServiceFunc, options ...modules.Option) service.ModuleCreateFunc {
 	return func(mi service.ModuleCreateInfo) ([]service.Module, error) {
-		mod, err := newYarpcThriftModule(mi, hookup, options...)
+		mod, err := newYARPCThriftModule(mi, hookup, options...)
 		if err != nil {
 			return nil, errors.Wrap(err, "unable to instantiate Thrift module")
 		}
@@ -47,11 +47,11 @@ func ThriftModule(hookup CreateThriftServiceFunc, options ...modules.Option) ser
 	}
 }
 
-func newYarpcThriftModule(
+func newYARPCThriftModule(
 	mi service.ModuleCreateInfo,
 	createService CreateThriftServiceFunc,
 	options ...modules.Option,
-) (*YarpcModule, error) {
+) (*YARPCModule, error) {
 	registrants, err := createService(mi.Host)
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to create YARPC thrift handler")
@@ -60,7 +60,7 @@ func newYarpcThriftModule(
 	reg := func(mod *YarpcModule) {
 		_setupMu.Lock()
 		defer _setupMu.Unlock()
-		mod.rpc.Register(registrants)
+		_dispatcher.Register(registrants)
 	}
-	return newYarpcModule(mi, reg, options...)
+	return newYARPCModule(mi, reg, options...)
 }
--- a/modules/rpc/thrift.go
+++ b/modules/rpc/thrift.go
@@ -57,10 +57,11 @@ func newYARPCThriftModule(
 		return nil, errors.Wrap(err, "unable to create YARPC thrift handler")
 	}
 
-	reg := func(mod *YarpcModule) {
+	reg := func(mod *YARPCModule) {
 		_setupMu.Lock()
 		defer _setupMu.Unlock()
-		_dispatcher.Register(registrants)
+		Dispatcher().Register(registrants)
 	}
+
 	return newYARPCModule(mi, reg, options...)
 }
--- a/modules/rpc/yarpc.go
+++ b/modules/rpc/yarpc.go
@@ -52,8 +52,10 @@ var (
 	_ service.Module = &YARPCModule{}
 
 	// Represents the global dispatcher used by all started modules.
+	// The YARPC team advised it to be a 'singleton' to control the lifecycle
+	// of all of the in/out bound traffic.
 	_dispatcher   *yarpc.Dispatcher
-	_muDispatcher sync.RWMutex
+	_dispatcherMu sync.RWMutex
 
 	_configsOnce sync.Once
 	_configs     configCollection
@@ -63,32 +65,43 @@ type registerServiceFunc func(module *YARPCModule)
 
 type yarpcConfig struct {
 	modules.ModuleConfig
-	Bind                    string `yaml:"bind"`
+	Bind                    string
 	AdvertiseName           string `yaml:"advertiseName"`
 	inboundMiddleware       []middleware.UnaryInbound
 	onewayInboundMiddleware []middleware.OnewayInbound
 	inbounds                []transport.Inbound
 }
 
-// Stores a collection of all modules configs and provides
-// operations that are safe to call from multiple go routines.
+// Stores a collection of all modules configs and provides operations that are safe to call from multiple go routines.
+// All the config must share the same AdvertiseName and represent a single service.
+// Configs are referenced and stored in the modules, we store only pointers to them to easily add/remove configs.
 type configCollection struct {
 	sync.RWMutex
 	configs []*yarpcConfig
 }
 
+// Adds the config to collection, the first config sets up the AdvertiseName requirement on consequent additions.
 func (c *configCollection) addConfig(config *yarpcConfig) error {
+	if config == nil {
+		return errors.New("passed a nil config")
+	}
+
 	c.Lock()
 	defer c.Unlock()
 
 	if len(c.configs) > 0 && config.AdvertiseName != c.configs[0].AdvertiseName {
-		return errors.New("Name mismatch")
+		return errs.Errorf("name mismatch, expected: %s; actual: %s",
+			c.configs[0].AdvertiseName,
+			config.AdvertiseName,
+		)
 	}
 
 	c.configs = append(c.configs, config)
 	return nil
 }
 
+// Removes the config from collection.
+// If the config is not found, returns an error, because it means that modules were not started properly.
 func (c *configCollection) removeConfig(config *yarpcConfig) error {
 	c.Lock()
 	defer c.Unlock()
@@ -103,16 +116,20 @@ func (c *configCollection) removeConfig(config *yarpcConfig) error {
 	return errors.New("config not found")
 }
 
+// Merge all the yarpc configs in the collection: transports and middleware are going to be shared.
+// The name comes from the first config in the collection and is the same among all configs.
 func (c *configCollection) mergeConfigs() (conf yarpc.Config, err error) {
 	c.RLock()
 	defer c.RUnlock()
 
 	if len(c.configs) <= 1 {
-		err = errors.New("empty configs")
-		return
+		// Config collection will always have an additional config storing the middleware.
+		return conf, errors.New("unable to merge empty configs")
 	}
 
 	conf.Name = c.configs[0].AdvertiseName
+
+	// Collect all Inbounds and middleware from all configs
 	var inboundMiddleware []middleware.UnaryInbound
 	var onewayInboundMiddleware []middleware.OnewayInbound
 	for _, cfg := range c.configs {
@@ -121,6 +138,7 @@ func (c *configCollection) mergeConfigs() (conf yarpc.Config, err error) {
 		onewayInboundMiddleware = append(onewayInboundMiddleware, cfg.onewayInboundMiddleware...)
 	}
 
+	// Build the inbound middleware
 	conf.InboundMiddleware = yarpc.InboundMiddleware{
 		Unary:  yarpc.UnaryInboundMiddleware(inboundMiddleware...),
 		Oneway: yarpc.OnewayInboundMiddleware(onewayInboundMiddleware...),
@@ -129,6 +147,8 @@ func (c *configCollection) mergeConfigs() (conf yarpc.Config, err error) {
 	return conf, nil
 }
 
+// Creates a new yarpc module and adds a common middleware to the global config collection.
+// The first created module defines the service name.
 func newYARPCModule(
 	mi service.ModuleCreateInfo,
 	reg registerServiceFunc,
@@ -206,24 +226,14 @@ func (m *YARPCModule) Start(readyCh chan<- struct{}) <-chan error {
 
 	ret := make(chan error, 1)
 	if err := _configs.addConfig(&m.config); err != nil {
-		ret <- errors.New("can't add module config " + err.Error())
+		ret <- errs.Wrap(err, "can't add module config")
 		return ret
 	}
 
-	config, err := _configs.mergeConfigs()
-
-	if err != nil {
-		ret <- errors.New("unable to merge configs from multiple modules " + err.Error())
+	if config, err := _configs.mergeConfigs(); err != nil {
+		ret <- errs.Wrap(err, "unable to merge configs from multiple modules")
 		return ret
-	}
-
-	dispatcher, err := _dispatcherFn(m.Host(), config)
-	if err != nil {
-		ret <- err
-		return ret
-	}
-
-	if err := resetDispatcher(dispatcher); err != nil {
+	} else if err := resetDispatcher(m.Host(), config); err != nil {
 		ret <- err
 		return ret
 	}
@@ -238,7 +248,8 @@ func (m *YARPCModule) Start(readyCh chan<- struct{}) <-chan error {
 	return ret
 }
 
-// Stop shuts down a YARPC module
+// Stop shuts down a YARPC module. It will stop the current dispatcher, remove the module config from the
+// config collection and start the new dispatcher with a new merged config.
 func (m *YARPCModule) Stop() error {
 	if !m.IsRunning() {
 		panic("module is not running")
@@ -249,20 +260,9 @@ func (m *YARPCModule) Stop() error {
 
 	if err := _configs.removeConfig(&m.config); err != nil {
 		return err
-	}
-
-	config, err := _configs.mergeConfigs()
-
-	if err != nil {
-		return errors.New("unable to merge configs from multiple modules " + err.Error())
-	}
-
-	dispatcher, err := _dispatcherFn(m.Host(), config)
-	if err != nil {
-		return err
-	}
-
-	if err := resetDispatcher(dispatcher); err != nil {
+	} else if config, err := _configs.mergeConfigs(); err != nil {
+		return errs.Wrap(err, "unable to merge configs from multiple modules")
+	} else if err = resetDispatcher(m.Host(), config); err != nil {
 		return err
 	}
 
@@ -292,15 +292,21 @@ func defaultYARPCDispatcher(_ service.Host, cfg yarpc.Config) (*yarpc.Dispatcher
 // It should be called after all modules have been started, because
 // the each module start creates a new dispatcher and stops previous.
 func Dispatcher() *yarpc.Dispatcher {
-	_muDispatcher.RLock()
-	defer _muDispatcher.RUnlock()
+	_dispatcherMu.RLock()
+	defer _dispatcherMu.RUnlock()
 	return _dispatcher
 }
 
-// Stops the current dispatcher, sets the new one and starts it.
-func resetDispatcher(d *yarpc.Dispatcher) error {
-	_muDispatcher.Lock()
-	defer _muDispatcher.Unlock()
+// Stops the current dispatcher, sets the new one with the host, config and _dispatcherFn, and then starts it.
+func resetDispatcher(host service.Host, config yarpc.Config) error {
+	d, err := _dispatcherFn(host, config)
+	if err != nil {
+		return err
+	}
+
+	_dispatcherMu.Lock()
+	defer _dispatcherMu.Unlock()
+
 	if _dispatcher == nil {
 		_dispatcher = d
 		return _dispatcher.Start()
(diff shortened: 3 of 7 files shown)


sectioneight:
@breerly can you suggest reviewers from your side?

modules/rpc/handler.go
```diff
@@ -130,6 +128,7 @@ func authorize(ctx context.Context, host service.Host) (fx.Context, error) {
 		host.Metrics().SubScope("rpc").SubScope("auth").Counter("fail").Inc(1)
 		fxctx.Logger().Error(auth.ErrAuthorization, "error", err)
 		// TODO(anup): GFM-255 update returned error to transport.BadRequestError (user error than server error)
+		// These ^ are internal YARPC errors, should we reference it?
```
sectioneight: What does this mean? Can we get somebody from the YARPC team to weigh in?

modules/rpc/thrift.go
```diff
@@ -34,7 +34,7 @@ type CreateThriftServiceFunc func(svc service.Host) ([]transport.Procedure, erro
 // ThriftModule creates a Thrift Module from a service func
 func ThriftModule(hookup CreateThriftServiceFunc, options ...modules.Option) service.ModuleCreateFunc {
 	return func(mi service.ModuleCreateInfo) ([]service.Module, error) {
-		mod, err := newYarpcThriftModule(mi, hookup, options...)
+		mod, err := newYARPCThriftModule(mi, hookup, options...)
```
sectioneight: Thank you :)

modules/rpc/thrift.go
```diff
@@ -43,18 +43,18 @@ func ThriftModule(hookup CreateThriftServiceFunc, options ...modules.Option) ser
 	}
 }
 
-func newYarpcThriftModule(
+func newYARPCThriftModule(
 	mi service.ModuleCreateInfo,
 	createService CreateThriftServiceFunc,
 	options ...modules.Option,
-) (*YarpcModule, error) {
+) (*YARPCModule, error) {
```
sectioneight: and thank you one more time :)

breerly:
Yup, myself, will h, and Kris should probably look 'er over.

modules/rpc/yarpc.go
```diff
@@ -36,66 +35,138 @@ import (
...
 }
 
 var (
-	_dispatcherFn = defaultYarpcDispatcher
+	_dispatcherFn = defaultYARPCDispatcher
 
-	_ service.Module = &YarpcModule{}
+	_ service.Module = &YARPCModule{}
+
+	// Represents the global dispatcher used by all started modules.
```
sectioneight: Maybe just for fun link to explanation about why we share a dispatcher?

modules/rpc/yarpc.go
```diff
@@ -36,66 +35,138 @@ import (
...
+	_dispatcher   *yarpc.Dispatcher
+	_muDispatcher sync.RWMutex
```
sectioneight: tiniest of tiny nits: i generally see mutexes named "fooMu" not "muFoo". IOW, consider `_dispatcherMu`

modules/rpc/yarpc.go
```diff
@@ -36,66 +35,138 @@ import (
...
+
+	_configsOnce sync.Once
```
sectioneight: like you did here, the Once goes after the thing it's describing. this reads better to me, and i believe is more consistent with existing code

modules/rpc/yarpc.go
```diff
@@ -36,66 +35,138 @@ import (
...
 )
 
-type registerServiceFunc func(module *YarpcModule)
+type registerServiceFunc func(module *YARPCModule)
 
 type yarpcConfig struct {
 	modules.ModuleConfig
-	Bind          string `yaml:"bind"`
-	AdvertiseName string `yaml:"advertiseName"`
+	Bind                    string `yaml:"bind"`
```
sectioneight: do we need this yaml tag? i guess it's good to be explicit, i thought it would work to deserialize (but not serialize) the lowercase yaml into the uppercase field

modules/rpc/yarpc.go
```diff
@@ -36,66 +35,138 @@ import (
...
+	sync.RWMutex
+	configs []*yarpcConfig
+}
+
+func (c *configCollection) addConfig(config *yarpcConfig) error {
+	c.Lock()
+	defer c.Unlock()
+
+	if len(c.configs) > 0 && config.AdvertiseName != c.configs[0].AdvertiseName {
+		return errors.New("Name mismatch")
```
sectioneight: errors should start with a lowercase letter, also consider using `fmt.Errorf` (or the fancier error package) to include the values we're seeing. this could be frustrating to debug otherwise

modules/rpc/yarpc.go
```diff
@@ -36,66 +35,138 @@ import (
...
+	return nil
+}
+
+func (c *configCollection) removeConfig(config *yarpcConfig) error {
+	c.Lock()
+	defer c.Unlock()
+
+	for i := range c.configs {
+		if c.configs[i] == config {
+			c.configs = append(c.configs[:i], c.configs[i+1:]...)
```
sectioneight: can you explain what we're doing here with a comment? we're triple nested and it's not the most readable assignment

modules/rpc/yarpc.go
```diff
@@ -36,66 +35,138 @@ import (
...
+
+	return errors.New("config not found")
+}
+
+func (c *configCollection) mergeConfigs() (conf yarpc.Config, err error) {
+	c.RLock()
+	defer c.RUnlock()
+
+	if len(c.configs) <= 1 {
+		err = errors.New("empty configs")
```
sectioneight: maybe provide a bit more context, e.g. "unable to merge empty configs" (mention that this happened in the merge process)

modules/rpc/yarpc.go
```diff
@@ -36,66 +35,138 @@ import (
...
+		return
+	}
+
+	conf.Name = c.configs[0].AdvertiseName
```
sectioneight: ok, this next 15 lines  definitely needs a comment or two about what's going on

modules/rpc/yarpc.go
```diff
@@ -36,66 +35,138 @@ import (
...
-		}),
-		WithOnewayInboundMiddleware(fxContextOnewayInboundMiddleware{
-			Host: mi.Host,
-		}),
-		WithOnewayInboundMiddleware(authOnewayInboundMiddleware{
-			Host: mi.Host,
-		}),
-	}, options...)
+		config: yarpcConfig{
+			AdvertiseName: mi.Host.Name(),
```
sectioneight: we seem to have lost a `Bind: ":0"` in the config. i'm assuming that was intentional, can you explain the change?

modules/rpc/yarpc.go
```diff
@@ -36,66 +35,138 @@ import (
...
+		},
+	}
+
+	var err error
+	_configsOnce.Do(func() {
+		// All the modules are going to share the same middleware. This middleware is going to be the first:
+		err = _configs.addConfig(&yarpcConfig{
+			AdvertiseName: mi.Host.Name(),
+			inboundMiddleware: []middleware.UnaryInbound{
```
sectioneight: this is where, IMHO, the unary/inbound/oneway/outbound stuff is getting a bit out of hand. hopefully our users won't have to understand this

modules/rpc/yarpc.go
```diff
@@ -105,78 +176,105 @@ func newYarpcModule(
 		}
 	}
 
-	err := module.Host().Config().Get(fmt.Sprintf("modules.%s", module.Name())).PopulateStruct(cfg)
-	// found values, update module
-	module.config = *cfg
+	err = module.Host().Config().Scope("modules").Get(module.Name()).PopulateStruct(&module.config)
```
sectioneight: I know you're going for readability, but block-scoping `errs` should almost always be preferred. could we instead break up this crazy accessor chain a bit for clarity?

```go
mod := module.Host().Config().Scope("modules").Get(module.Name())
if err := mod.PopulateStruct(&module.conig); err != nil {
  return module, err
}
```

also what exactly is this code doing? i know it's loading some config, but the term "module" and "config" show up so many times in a single line it's really hard to tell (to me) exactly what's going on except that we're definitely fetching some configuration for a module

modules/rpc/yarpc.go
```diff
@@ -105,78 +176,105 @@ func newYarpcModule(
...
 	return module, err
 }
 
-// Start begins serving requests over YARPC
-func (m *YarpcModule) Start(readyCh chan<- struct{}) <-chan error {
+// Start begins serving requests over RPC. It first stops the current dispatcher, merges configs from the global
+// collection and then starts the new dispatcher.
+func (m *YARPCModule) Start(readyCh chan<- struct{}) <-chan error {
+	if m.IsRunning() {
+		panic("module is already running")
```
sectioneight: under what circumstances would this happen? can we mention in the method signature? I know we have internal guidelines that it should be reserved for unrecoverable startup errors only

modules/rpc/yarpc.go
```diff
@@ -105,78 +176,105 @@ func newYarpcModule(
...
-		InboundMiddleware: yarpc.InboundMiddleware{
-			Unary:  interceptor,
-			Oneway: onewayInterceptor,
-		},
-		Tracer: m.Tracer(),
-	})
+	config, err := _configs.mergeConfigs()
+
+	if err != nil {
+		ret <- errors.New("unable to merge configs from multiple modules " + err.Error())
```
sectioneight: maybe a colon after `modules` to show that the rest is the error.

also, should this be `errors.Wrap` instead?

modules/rpc/yarpc.go
```diff
@@ -105,78 +176,105 @@ func newYarpcModule(
...
+		return ret
+	}
 
+	dispatcher, err := _dispatcherFn(m.Host(), config)
 	if err != nil {
 		ret <- err
 		return ret
 	}
 
+	if err := resetDispatcher(dispatcher); err != nil {
```
sectioneight: why do we reset here? what does it mean to reset a dispatcher?

also could do with a bit of DRYing up for error returning

```go
errorOut := func(err error) chan error {
  ret <- err
  return ret
}
```

thene these all just become:

```go
if err := tryThing(); err != nil {
  return errorOut(err)
}
```

may not be worth it, just a thought

modules/rpc/yarpc.go
```diff
@@ -105,78 +176,105 @@ func newYarpcModule(
...
+	m.isRunning = true
 	readyCh <- struct{}{}
 	return ret
 }
 
 // Stop shuts down a YARPC module
-func (m *YarpcModule) Stop() error {
+func (m *YARPCModule) Stop() error {
+	if !m.IsRunning() {
+		panic("module is not running")
```
sectioneight: why panic?

modules/rpc/yarpc.go
```diff
@@ -105,78 +176,105 @@ func newYarpcModule(
...
-		err := m.rpc.Stop()
-		m.rpc = nil
+	if err := _configs.removeConfig(&m.config); err != nil {
+		return err
+	}
+
+	config, err := _configs.mergeConfigs()
+
+	if err != nil {
+		return errors.New("unable to merge configs from multiple modules " + err.Error())
```
sectioneight: errors.Wrap

modules/rpc/yarpc.go
```diff
@@ -105,78 +176,105 @@ func newYarpcModule(
...
+	}
+
+	dispatcher, err := _dispatcherFn(m.Host(), config)
```
sectioneight: seems to be an almost copy-paste of above (call _dispatcherFn, resetDispatcher).

modules/rpc/yarpc.go
alsamylkin: Yeah, this is all over our code. We need to refactor after a beta launch.

modules/rpc/yarpc.go
alsamylkin: Yes, errors are very raw in this PR, need to update.

modules/rpc/yarpc.go
alsamylkin: Remove the config, duh :)
But yeah, need more comments how do we store configs.

modules/rpc/yarpc.go
alsamylkin: I was confused about it as well: this is a default value, that is going to be overridden by the PopulateStruct, so here we simply remove a default port ":0" if it is not in the config.

I am open to discussion if we want to keep it as default though.

modules/rpc/handler.go
anuptalwalkar: This is an open issue with YARPC team, with PR open - https://github.com/yarpc/yarpc-go/issues/687

modules/rpc/yarpc.go
alsamylkin: We haven't started doing the outbound middleware yet ;)

modules/rpc/yarpc.go
alsamylkin: We read the config, the error line was too long to fit in if. Let me refactor it more.

modules/rpc/yarpc.go
alsamylkin: When a user calls it. Method is public, so it can happen. We can push errors to channels, but in this case it seems to be an initialization issue, modules are called in the wrong order/out of sync, user should check if it is running first.

modules/rpc/yarpc.go
alsamylkin: There is a comment in the reset method, and yeah, it needs to be refactored more here.

modules/rpc/yarpc.go
alsamylkin: Same as with start, why someone wants to stop something that is not running?

modules/rpc/yarpc.go
alsamylkin: Yep, I should move it to the reset.

modules/rpc/yarpc.go
```diff
@@ -36,66 +35,138 @@ import (
...
+	inboundMiddleware       []middleware.UnaryInbound
+	onewayInboundMiddleware []middleware.OnewayInbound
+	inbounds                []transport.Inbound
+}
+
+// Stores a collection of all modules configs and provides
+// operations that are safe to call from multiple go routines.
+type configCollection struct {
+	sync.RWMutex
+	configs []*yarpcConfig
```
anuptalwalkar: Why do you need collection of yarpc configs? Why not use one yarpcConfig?

alsamylkin: To be able to add/remove them in case a module is started/stopped, e.g. if they have their own middleware/transports.

modules/rpc/yarpc_options.go
```diff
@@ -31,28 +31,28 @@ const (
...
+		inboundMiddleware := inboundMiddlewareFromCreateInfo(*mci)
+		inboundMiddleware = append(inboundMiddleware, i...)
+		mci.Items[_interceptorKey] = inboundMiddleware
 
 		return nil
 	}
 }
 
-// WithOnewayInboundMiddleware adds custom YARPC inboundMiddlewares to the module
+// WithOnewayInboundMiddleware adds custom YARPC inboundMiddleware to the module
```
anuptalwalkar: nit: previous comment seemed to be inline with variadic arguments. Can we keep it, or not have variadic arguments here? I have no preference on the definition though:)

breerly:
Poor lonely dispatcher 
5b4

modules/rpc/yarpc.go
```diff
@@ -186,6 +284,38 @@ func RegisterDispatcher(dispatchFn yarpcDispatcherFn) {
...
+
+	_dispatcherMu.Lock()
+	defer _dispatcherMu.Unlock()
+
+	if _dispatcher == nil {
+		_dispatcher = d
+		return _dispatcher.Start()
+	}
+
+	if err := _dispatcher.Stop(); err != nil {
```
anuptalwalkar: I don't like the idea of stopping dispatcher and restarting new one. Can we do away with it somehow? You will be stopping the world for one module Stop() call. How about we start up new Dispatcher, free up the lock and move old dispatcher for closing with drain?

modules/rpc/yarpc.go
```diff
@@ -36,66 +35,158 @@ import (
...
+			return nil
+		}
+	}
+
+	return errors.New("config not found")
+}
+
+// Merge all the yarpc configs in the collection: transports and middleware are going to be shared.
+// The name comes from the first config in the collection and is the same among all configs.
+func (c *configCollection) mergeConfigs() (conf yarpc.Config, err error) {
```
anuptalwalkar: tests? I don't think we have tests with multiple modules using single dispatcher.

alsamylkin: yep, it is in progress

modules/rpc/yarpc.go
alsamylkin: I hope RPC team can we us suggestions on the right way to do it.

modules/rpc/yarpc.go
```diff
@@ -186,6 +284,38 @@ func RegisterDispatcher(dispatchFn yarpcDispatcherFn) {
 	_dispatcherFn = dispatchFn
 }
 
-func defaultYarpcDispatcher(_ service.Host, cfg yarpc.Config) (*yarpc.Dispatcher, error) {
+func defaultYARPCDispatcher(_ service.Host, cfg yarpc.Config) (*yarpc.Dispatcher, error) {
```
willhug: why does this function return an error?

modules/rpc/yarpc.go
willhug: Is the goal to be able to start/stop fx modules individually during the lifetime of the application?  Or is this for the shutdown/drain of the entire application?

If this is for the shutdown of the entire application, in YARPC we can update dispatcher.Stop/Start() to be idempotent so we can shutdown all modules at the same time, and they will all block on a sync.Once until the dispatcher shuts down.  (We could also wrap the dispatcher start/stop in uberfx to do the same)

We currently do this for outbounds/inbounds in YARPC using our LifecycleOnce abstraction, it can be added to the dispatcher as well.
https://github.com/yarpc/yarpc-go/blob/dev/internal/sync/lifecycle.go#L58

If this is for starting/stopping during the lifetime of the application, what is the use case we want to support?

The current implementation here will probably lead to some unknown/strange behavior since we'll be starting/stopping the inbounds dynamically which is not supported by YARPC inbounds/outbounds.

modules/rpc/yarpc.go
```diff
@@ -36,66 +35,158 @@ import (
...
-	_dispatcherFn = defaultYarpcDispatcher
+	_dispatcherFn = defaultYARPCDispatcher
 
-	_ service.Module = &YarpcModule{}
+	_ service.Module = &YARPCModule{}
+
+	// Represents the global dispatcher used by all started modules.
+	// The YARPC team advised it to be a 'singleton' to control the lifecycle
+	// of all of the in/out bound traffic.
+	_dispatcher   *yarpc.Dispatcher
```
willhug: do we have a guarantee that this will be generated before we use it?

If not, should we control access of this dispatcher through a function?

modules/rpc/yarpc.go
```diff
@@ -105,78 +196,85 @@ func newYarpcModule(
...
 	return ret
 }
 
-// Stop shuts down a YARPC module
-func (m *YarpcModule) Stop() error {
+// Stop shuts down a YARPC module. It will stop the current dispatcher, remove the module config from the
+// config collection and start the new dispatcher with a new merged config.
+func (m *YARPCModule) Stop() error {
+	if !m.IsRunning() {
+		panic("module is not running")
```
willhug: This will only happen once we're shutting down the whole application right?  If not it seems unsafe to have a panic which could damage the whole application.  We might want to consider making start/stop idempotent.

modules/rpc/yarpc.go
```diff
@@ -105,78 +196,85 @@ func newYarpcModule(
...
 	defer m.stateMu.Unlock()
 
-	if m.rpc != nil {
-		err := m.rpc.Stop()
-		m.rpc = nil
+	if err := _configs.removeConfig(&m.config); err != nil {
+		return err
+	} else if config, err := _configs.mergeConfigs(); err != nil {
+		return errs.Wrap(err, "unable to merge configs from multiple modules")
+	} else if err = resetDispatcher(m.Host(), config); err != nil {
```
willhug: if this is the path we go down, it might be clearer to make these separate if statements (since they all return anyway)
```go
if err := _configs.removeConfig(&m.config); err != nil {
    return err
}

config, err := _configs.mergeConfigs()
if err != nil {
    return errs.Wrap(err, "unable to merge configs from multiple modules")
}

if err = resetDispatcher(m.Host(), config); err != nil {
    return err
}
```

modules/rpc/yarpc.go
```diff
@@ -105,78 +196,85 @@ func newYarpcModule(
...
+	module.config.inbounds = []transport.Inbound{tchTransport.NewInbound()}
 
 	return module, err
 }
 
-// Start begins serving requests over YARPC
-func (m *YarpcModule) Start(readyCh chan<- struct{}) <-chan error {
+// Start begins serving requests over RPC. It first stops the current dispatcher, merges configs from the global
+// collection and then starts the new dispatcher.
+func (m *YARPCModule) Start(readyCh chan<- struct{}) <-chan error {
```
willhug: Start/Stop should only be called once we've defined all the Modules of the application right?  Do we see a situation where after Start is called on a YARPCModule we'd want to add another YARPCModule?  Or is the plan:

1) Configure/Create all Modules
2) Start all Modules
3) Let application run
4) Get shutdown signal, call stop on all Modules

With no dynamic module creation.

modules/rpc/yarpc.go
alsamylkin: Legacy code and also it is nicer to have an actual error than chase nils.

modules/rpc/yarpc_options.go
alsamylkin: The plural of middleware is still middleware :)

modules/rpc/yarpc.go
alsamylkin: Yes, moved it to configCollection

modules/rpc/yarpc.go
alsamylkin: Yes, thanks @willhug! We'll have to follow the same pattern here too.

modules/rpc/yarpc.go
```diff
@@ -36,67 +35,177 @@ import (
...
+			return
+		}
+
+		_dispatcherMu.Lock()
+		defer _dispatcherMu.Unlock()
+		if c.dispatcher, err = _dispatcherFn(host, cfg); err != nil {
+			return
+		}
+
+		err = c.dispatcher.Start()
```
willhug: We might want to store the error from the start.Do on the configCollection.  Currently the code will forget about the error after the first call:
```
c := configCollection(...)
c.Start() // Returns Error
c.Start() // No error
```

Instead we should store it:
```
c.start.Do(func() {...
    c.startErr := c.dispatcher.Start()
}

return c.startErr
```

modules/rpc/yarpc.go
```diff
@@ -36,67 +35,177 @@ import (
...
+
+	return err
+}
+
+// Return the result of the dispatcher Stop() on the first call.
+// No-op on subsequent calls.
+func (c *configCollection) Stop() error {
+	var err error
+	c.stop.Do(func() {
+		err = c.dispatcher.Stop()
```
willhug: ditto for storing the error

modules/rpc/yarpc.go
willhug: fair enough

modules/rpc/yarpc.go
alsamylkin: I was reluctant to do this because will need to sync errors, but I guess it is better to stay consistent :)

modules/rpc/yarpc.go
```diff
@@ -36,67 +35,192 @@ import (
...
 
-func newYarpcModule(
+// Merge all the YARPC configs in the collection: transports and middleware are going to be shared.
+// The name comes from the first config in the collection and is the same among all configs.
+func (c *configCollection) mergeConfigs() (conf yarpc.Config, err error) {
+	c.RLock()
+	defer c.RUnlock()
+
+	if len(c.configs) <= 1 {
+		// Config collection should always have an additional config with the default middleware.
```
anuptalwalkar: nit: may be comment above if? Not sure if its needed.

modules/rpc/yarpc.go
```diff
@@ -36,67 +35,192 @@ import (
...
+	AdvertiseName           string `yaml:"advertiseName"`
+	inboundMiddleware       []middleware.UnaryInbound
+	onewayInboundMiddleware []middleware.OnewayInbound
+	inbounds                []transport.Inbound
+}
+
+// Stores a collection of all modules configs with a shared dispatcher
+// that are safe to call from multiple go routines. All the configs must
+// share the same AdvertiseName and represent a single service.
+type configCollection struct {
```
anuptalwalkar: can we call this something concrete? It is doing a lot more things than just collecting yarpc configurations. `rpcController` or `dispatcherManager`. configCollection seems pretty generic to have a dispatcher in it.

modules/rpc/yarpc.go
```diff
@@ -36,67 +35,192 @@ import (
...
+	}
+
+	return nil
+}
+
+// Starts the dispatcher: wait until all modules call start, create a single dispatcher and then start it.
+// Once started the collection will not start the dispatcher again.
+func (c *configCollection) Start(host service.Host) error {
+	c.errorMu.Lock()
+	defer c.errorMu.Unlock()
```
willhug: do we need the errorMu?  The updates to the start/stop errors should already be protected by the `sync.Once` right?

modules/rpc/yarpc.go
```diff
@@ -36,67 +35,192 @@ import (
...
+	})
+
+	return c.startError
+}
+
+// Return the result of the dispatcher Stop() on the first call.
+// No-op on subsequent calls.
+func (c *configCollection) Stop() error {
+	c.errorMu.Lock()
+	defer c.errorMu.Unlock()
```
willhug: ditto, do we need `errorMu`?

modules/rpc/yarpc.go
alsamylkin: We do, otherwise we'll lose consistency:
A calls Start() and start executing Dispatcher start, but haven't finished before B calls Start().
Now B will return a nil error, while it can be not true.

willhug: afaik, sync.Once objects will block, so if both A and B call into c.start.Do, only one will execute the function, but, both will block until the function is finished, so we don't need the lock on top of that.

alsamylkin: You right, thanks!

PR #197, opened 2017-01-30
Remove Consume method from the async task Backend

madhuravi:
When the Start method is called on backend, it should kick off the listen stream and start consuming messages. A separate method is not required as part of the interface

### DIFF
--- a/modules/task/backend.go
+++ b/modules/task/backend.go
@@ -97,8 +97,10 @@ func (b *inMemBackend) Start(ready chan<- struct{}) <-chan error {
 	go func(buf chan []byte) {
 		for {
 			select {
-			case msg := <-buf:
-				errorCh <- Run(msg)
+			case msg, ok := <-buf:
+				if ok {
+					errorCh <- Run(msg)
+				}
 			case <-time.After(time.Second):
 				fmt.Println("Timed out after 1 second")
 			}
--- a/modules/task/backend.go
+++ b/modules/task/backend.go
@@ -73,11 +73,12 @@ func (b NopBackend) IsRunning() bool {
 type inMemBackend struct {
 	bufQueue  chan []byte
 	isRunning bool
+	errorCh   chan error
 }
 
 // NewInMemBackend creates a new in memory backend, designed for use in tests
 func NewInMemBackend() Backend {
-	return &inMemBackend{bufQueue: make(chan []byte, 2)}
+	return &inMemBackend{bufQueue: make(chan []byte, 2), errorCh: make(chan error, 2)}
 }
 
 // Encoder implements the Backend interface
@@ -92,23 +93,24 @@ func (b *inMemBackend) Name() string {
 
 // Start implements the Module interface
 func (b *inMemBackend) Start(ready chan<- struct{}) <-chan error {
-	b.isRunning = true
-	errorCh := make(chan error, 2)
-	go func() {
-		for {
-			select {
-			case msg, ok := <-b.bufQueue:
-				if ok {
-					errorCh <- Run(msg)
-				} else {
-					return
+	if !b.IsRunning() {
+		b.isRunning = true
+		go func() {
+			for {
+				select {
+				case msg, ok := <-b.bufQueue:
+					if ok {
+						b.errorCh <- Run(msg)
+					} else {
+						return
+					}
+				case <-time.After(time.Second):
+					ulog.Logger().Error("Timed out after 1 second")
 				}
-			case <-time.After(time.Second):
-				ulog.Logger().Error("Timed out after 1 second")
 			}
-		}
-	}()
-	return errorCh
+		}()
+	}
+	return b.errorCh
 }
 
 // Publish implements the Backend interface
@@ -123,6 +125,7 @@ func (b *inMemBackend) Publish(message []byte, userContext map[string]string) er
 func (b *inMemBackend) Stop() error {
 	b.isRunning = false
 	close(b.bufQueue)
+	close(b.errorCh)
 	return nil
 }
 
--- a/modules/task/backend.go
+++ b/modules/task/backend.go
@@ -21,6 +21,7 @@
 package task
 
 import (
+	"errors"
 	"time"
 
 	"go.uber.org/fx/service"
@@ -73,12 +74,14 @@ func (b NopBackend) IsRunning() bool {
 type inMemBackend struct {
 	bufQueue  chan []byte
 	isRunning bool
-	errorCh   chan error
+	isStopped bool
 }
 
 // NewInMemBackend creates a new in memory backend, designed for use in tests
 func NewInMemBackend() Backend {
-	return &inMemBackend{bufQueue: make(chan []byte, 2), errorCh: make(chan error, 2)}
+	return &inMemBackend{
+		bufQueue: make(chan []byte, 2),
+	}
 }
 
 // Encoder implements the Backend interface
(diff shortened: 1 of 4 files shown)


modules/task/backend.go
```diff
@@ -118,7 +93,28 @@ func (b *inMemBackend) Name() string {
...
+			select {
+			case msg, ok := <-buf:
+				if ok {
+					errorCh <- Run(msg)
+				}
+			case <-time.After(time.Second):
+				fmt.Println("Timed out after 1 second")
+			}
+		}
+	}(b.bufQueue)
```
sectioneight: why pass this as a parameter instead of just grabbing it from the closure? code clarity?

modules/task/backend.go
```diff
@@ -118,7 +93,28 @@ func (b *inMemBackend) Name() string {
...
+	errorCh := make(chan error, 2)
+	go func(buf chan []byte) {
+		for {
+			select {
+			case msg, ok := <-buf:
+				if ok {
+					errorCh <- Run(msg)
+				}
+			case <-time.After(time.Second):
+				fmt.Println("Timed out after 1 second")
```
sectioneight: Is this supposed to be an error log? Or were you just testing something?

madhuravi: Error log only

modules/task/backend.go
madhuravi: Haha no, one of the many things I tried while I got the channels to work properly. Changed.

modules/task/backend.go
```diff
@@ -118,7 +93,28 @@ func (b *inMemBackend) Name() string {
 // Start implements the Module interface
 func (b *inMemBackend) Start(ready chan<- struct{}) <-chan error {
 	b.isRunning = true
-	return make(chan error)
+	errorCh := make(chan error, 2)
```
alsamylkin: Nit, can Start() be called twice, is it ok?

modules/task/backend.go
sectioneight: Ok, is there a change forthcoming that is going to use a proper logger?

modules/task/backend.go
```diff
@@ -118,7 +93,28 @@ func (b *inMemBackend) Name() string {
...
+	go func() {
+		for {
+			select {
+			case msg, ok := <-b.bufQueue:
+				if ok {
```
alsamylkin: if not ok is not an error?

modules/task/execution.go
```diff
@@ -37,14 +37,14 @@ type fnRegister struct {
 
 var fnLookup = fnRegister{fnNameMap: make(map[string]interface{})}
 
-// fnSignature represents a function and its arguments
-type fnSignature struct {
+// FnSignature represents a function and its arguments
+type FnSignature struct {
```
alsamylkin: why does it need to become public?

modules/task/backend.go
madhuravi: What are we using for logs? ulog.Logger?

modules/task/backend.go
madhuravi: Since we control backend behavior in the task module, this is ok. We will not start more than once. It's ok if tests are calling it twice.

modules/task/backend.go
madhuravi: No, it should just stop listening if the channel is closed. Added an else return here.

modules/task/execution.go
madhuravi: Backend implementations and tests need access to it and they can be in different repos and different modules.

modules/task/backend.go
alsamylkin: What do you mean 'we control' the behavior? It is exposed to a user through NewInMemBackend.. If it is called twice, is it ok to create a new channel, because someone else can hold a reference to the previously started error channel?

modules/task/execution.go
alsamylkin: Can we expose it with a New()... function returning a Backend interface, not a exporting a struct?

modules/task/backend.go
madhuravi: That behavior will work fine but it's cleaner to have a single reference. Changed.

modules/task/execution.go
madhuravi: FnSignature and Backend are different things. I am exposing FnSignature so I can create test function structs in uberfx-private or end to end tests.

modules/task/backend.go
```diff
@@ -78,31 +73,12 @@ func (b NopBackend) IsRunning() bool {
...
-	case v, ok := <-b.bufQueue:
-		if ok {
-			return Run(v)
-		}
-		return errors.New("The bufQueue channel has been closed")
-	default:
-		// No value ready in channel, moving on
-	}
-	return nil
+	return &inMemBackend{bufQueue: make(chan []byte, 2), errorCh: make(chan error, 2)}
```
sectioneight: this is a bit hard to read, prefer one field per line:

```go
return &inMemBackend{
  bufQueue: ...,
  errorCh: ...,
}

File: modules/task/backend.go [hunk 7]
```diff
@@ -117,14 +93,39 @@ func (b *inMemBackend) Name() string {
 
 // Start implements the Module interface
 func (b *inMemBackend) Start(ready chan<- struct{}) <-chan error {
-	b.isRunning = true
-	return make(chan error)
+	if !b.IsRunning() {
+		b.isRunning = true
+		go func() {
```
sectioneight: this gets pretty deeply nested, consider extracting a private method on `*inMemBackend` that you can `go` run here

modules/task/backend.go
```diff
@@ -117,14 +93,39 @@ func (b *inMemBackend) Name() string {
...
+			for {
+				select {
+				case msg, ok := <-b.bufQueue:
+					if ok {
+						b.errorCh <- Run(msg)
+					} else {
+						return
+					}
+				case <-time.After(time.Second):
```
sectioneight: why 1 second?

madhuravi: this is pretty conservative. I can do 5-10 milliseconds instead.

modules/task/backend.go
```diff
@@ -78,31 +74,14 @@ func (b NopBackend) IsRunning() bool {
 type inMemBackend struct {
 	bufQueue  chan []byte
 	isRunning bool
+	isStopped bool
```
sectioneight: Feels weird to have both isRunning and isStopped -- shouldn't they always be the opposite of each other?

modules/task/backend.go
```diff
@@ -117,13 +96,45 @@ func (b *inMemBackend) Name() string {
 
 // Start implements the Module interface
 func (b *inMemBackend) Start(ready chan<- struct{}) <-chan error {
-	b.isRunning = true
-	return make(chan error)
+	errorCh := make(chan error, 2)
+	if b.IsRunning() {
+		errorCh <- errors.New("Cannot start when module is already running")
```
sectioneight: style: errors should start with lowercase, they're meant to be interpolated into sentences

modules/task/backend.go
madhuravi: There are 3 states,
Not started -> isRunning = false, isStopped = false
Started/Running -> isRunning = true, isStopped = false
Stopped -> isRunning = false, isStopped = true

I added this extra bool after @alsamylkin 's  comment to add validation for cases when someone calls Start after Stop. In previous code, it would panic/timeout on a channel. He was recommending a clearer error message. With isRunning, this is hard to distinguish because it's false both before the service started and after the service has been stopped.

sectioneight:
Makes sense. It's probably better represented by a single value, rather
than two different booleans. Something like:

type backendState int

const (
  invalidBackendState backendState = iota
  notStarted
  started
  stopped
)

modules/task/backend.go
glibsm: Sounds like a enum-type struct would be better, no?

Especially since `isRunning = true, isStopped = true` is an invalid state and two booleans give you four.

sectioneight: Yes, that was my comment down below, I replied via email so it didn't show up next to this :(

PR #198, opened 2017-01-31
Call the uber-licence executable from node_modules/.bin

dmcaulay:
The current bin path used in the script (`./node_modules/uber-licence/bin/licence`) did not work for me, because the file is now named [licence.js](https://github.com/uber/uber-licence/blob/master/bin/licence.js).  This updates the bin path to use the symlink in `node_modules/.bin`. This path will be less likely to change in the future.

### DIFF
--- a/.build/check_license.sh
+++ b/.build/check_license.sh
@@ -9,7 +9,7 @@ run_uber_licence() {
   # Ok, somebody hasn't run `npm install -g uber-licence`, that's cool
   if ! which "$bin" >/dev/null ; then
     npm install "$bin" >/dev/null
-    bin="./node_modules/uber-licence/bin/licence"
+    bin="./node_modules/.bin/uber-licence"
   fi
 
   readonly local output=$("$bin" --file "*.go" | sed "s/^fix //")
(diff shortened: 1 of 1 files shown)


.build/check_license.sh
```diff
@@ -9,7 +9,7 @@ run_uber_licence() {
   # Ok, somebody hasn't run `npm install -g uber-licence`, that's cool
   if ! which "$bin" >/dev/null ; then
     npm install "$bin" >/dev/null
-    bin="./node_modules/uber-licence/bin/licence"
```
glibsm: this still won't fully work, as `which uber-licence` does not puck up the local modules and will run the install every time.

```bash
$ which uber-licence
$ find . -name uber-licence
./node_modules/.bin/uber-licence
```

The proper way to fix would be check `find uber-licence` or `node_modules/uber-licence`, else install.

dmcaulay: i added a fix in the script. you can add `node_modules/.bin` to your `$PATH` if you want it to work outside the script.

the other option would be to add `-g` when you run `npm install`.

PR #199, opened 2017-01-31
Fix APP_CONFIG_DIR usage

dmcaulay:
The `APP_CONFIG_DIR` environment variable points directly to the the configuration directory. If we want to use this directory with the `FileResolver` we need to include file names without the `./config` prefix. This diff adds those files to the list of config files.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -41,7 +41,7 @@ const (
 	_environment = "_ENVIRONMENT"
 	_datacenter  = "_DATACENTER"
 	_configDir   = "_CONFIG_DIR"
-	_config      = "config"
+	_configRoot  = "./config"
 	_baseFile    = "base"
 	_secretsFile = "secrets"
 )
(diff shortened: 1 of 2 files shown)


config/config_test.go
```diff
@@ -351,3 +354,22 @@ func TestEnvProvider_Callbacks(t *testing.T) {
 	assert.NoError(t, p.RegisterChangeCallback("test", nil))
 	assert.NoError(t, p.UnregisterChangeCallback("token"))
 }
+
+func TestGetConfigFiles(t *testing.T) {
+	SetEnvironmentPrefix("TEST")
```
anuptalwalkar: You can use env.Override available in the testutils. 
``defer env.Override(t, <env key>, <temp value>()``

config/config.go
```diff
@@ -62,14 +62,18 @@ func getConfigFiles() []string {
...
 
-	var files []string
+	baseFiles := []string{_baseFile, env, _secretsFile}
 	if dc != "" && env != "" {
-		files = append(files, fmt.Sprintf("./%s/%s-%s.yaml", _config, env, dc))
+		baseFiles = append(baseFiles, fmt.Sprintf("%s-%s", env, dc))
+	}
+
+	var files []string
+	dirs := []string{".", _configRoot}
```
anuptalwalkar: Question: Why "." you will be adding ./base.yaml then ./config/base.yaml etc.. won't the lookup in _configRoot be sufficient?

dmcaulay: I attempted to explain the the PR description. Does that make sense?

anuptalwalkar: ah, sorry, I misread the description. Makes sense now!

anuptalwalkar:
lgtm. Thanks for the fix! I am merging the changes now.

PR #201, opened 2017-01-31
Use iota enums for states in InMemBackend

madhuravi:


modules/task/backend.go
```diff
@@ -30,6 +30,12 @@ import (
 
 var gobEncoding = &GobEncoding{}
 
+const (
+	_initialized = iota
```
sectioneight: hmm i think breaking the underscore rule here might be warranted. this is hard to look at

madhuravi: Either one is ok but we should be consistent across our repo. Let me know if you want to start following no underscores for constants.

glibsm: i think the convention for the first one is typically `unknown` or something like that, and the three valid states are listed below. I might be wrong though

madhuravi: @glibsm I don't see that in examples online. Let me know if you have examples. I personally don't like unknown, the name itself does not indicate what it is and is not meaningful here.

PR #202, opened 2017-01-31
Fix race condition in access of fnNameMap in execution test

madhuravi:


modules/task/execution_test.go
```diff
@@ -93,7 +93,9 @@ func TestConsumeWithoutRegister(t *testing.T) {
 	require.NoError(t, err)
 	err = Enqueue(fn, float64(1.0))
 	require.NoError(t, err)
+	fnLookup.Lock()
 	fnLookup.fnNameMap = make(map[string]interface{})
```
alsamylkin: I think we should access to the map through a method..

madhuravi: good idea, I thought this was unnecessary earlier but it's time to do this now.

PR #203, opened 2017-01-31
Fail the build when race is detected

glibsm:
Several things changes:

- Old cover profile is deleted when running tests
- If new profile is not generated, that means tests have failed
- The `die` call will make sure travis reports a failure

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,8 +42,10 @@ endif
 COVER_OUT := profile.coverprofile
 
 $(COV_REPORT): $(PKG_FILES) $(ALL_SRC)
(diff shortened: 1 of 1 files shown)


madhuravi:
Nice! Pull in latest changes and the build should succeed.

glibsm:
Yeah I was testing to make sure the build will actually fail. Will rebase

madhuravi:
Actually this is a different error - panic on closed channel. That's different from the data race issue. Might have to debug this one.

madhuravi:
Hmm looks like that was transient. We should keep an eye out if it comes up again.

PR #204, opened 2017-01-31
Fix APP_CONFIG_DIR usage

dmcaulay:
The `APP_CONFIG_DIR` environment variable points directly to the the configuration directory. If we want to use this directory with the `FileResolver` we need to include file names without the `./config` prefix. This diff adds those files to the list of config files.


dmcaulay:
@anuptalwalkar Sorry this was originally a PR against my license branch, because I couldn't get the license check to work without it.

PR #205, opened 2017-01-31
Fix number of arguments in error call

glibsm:
While creating a new example, I noticed our setup is actually not correct.
We panic at runtime with incorrect number of arguments provided to the log function.

### DIFF
--- a/examples/keyvalue/server/main.go
+++ b/examples/keyvalue/server/main.go
@@ -21,11 +21,10 @@
 package main
 
 import (
-	"log"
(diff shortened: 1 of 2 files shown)


sectioneight:
Yea, this wants me to re-prioritize getting off the sugar internally :\

anuptalwalkar:
This shouldn't be a panic, but just an error message with invalid number of arguments. But I agree that we need to get off internal sugar implementation.

PR #206, opened 2017-01-31
Add APP_ROOT environment variable to file resolver

dmcaulay:
This gives developers the ability to set the root directory of their application. This can be useful for resolving relative directories during service configuration.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -41,6 +41,7 @@ const (
 	_environment = "_ENVIRONMENT"
 	_datacenter  = "_DATACENTER"
 	_configDir   = "_CONFIG_DIR"
+	_root        = "_ROOT"
 	_configRoot  = "./config"
 	_baseFile    = "base"
 	_secretsFile = "secrets"
--- a/config/config.go
+++ b/config/config.go
@@ -38,10 +38,10 @@ const (
 )
 
 const (
+	_appRoot     = "APP_ROOT"
 	_environment = "_ENVIRONMENT"
 	_datacenter  = "_DATACENTER"
 	_configDir   = "_CONFIG_DIR"
-	_root        = "_ROOT"
 	_configRoot  = "./config"
 	_baseFile    = "base"
 	_secretsFile = "secrets"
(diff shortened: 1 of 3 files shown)


config/file_resolver.go
```diff
@@ -42,8 +42,11 @@ func NewRelativeResolver(paths ...string) FileResolver {
 
 	copy(pathList, paths)
 
-	// add the current cwd
-	if cwd, err := os.Getwd(); err == nil {
+	if appRoot := os.Getenv(EnvironmentPrefix() + _root); appRoot != "" {
```
anuptalwalkar: Note that environment prefix for uber services won't be "APP". I believe you want _root to be just "APP_ROOT" and not use EnvironmentPrefix()

config/file_resolver.go
```diff
@@ -42,8 +42,11 @@ func NewRelativeResolver(paths ...string) FileResolver {
...
+		// add the app root
+		pathList = append(pathList, appRoot)
+	} else if cwd, err := os.Getwd(); err == nil {
+		// add the current cwd
```
anuptalwalkar: No need for comment here, but may be detailed comment about why we are adding APP_ROOT on 45?

config/file_resolver.go
dmcaulay: yeah, i wasn't sure what i wanted. i can hardcode it to `APP_ROOT` if we think that's best.

glibsm: APP_ROOT sounds good, since it's one thing to rule them all, rather than environment/company specific.

config/config.go
```diff
@@ -58,6 +59,21 @@ var (
...
+// can edit this via the APP_ROOT environment variable. If the environment
+// variable is not set then it will fallback to the current working directory.
+// This is often used for resolving relative paths in your service.
+func AppRoot() string {
+	if appRoot := os.Getenv(_appRoot); appRoot != "" {
+		return appRoot
+	}
+	if cwd, err := os.Getwd(); err != nil {
+		panic(fmt.Sprintf("Unable to get the current working directory: %s", err.Error()))
+	} else {
```
glibsm: no need for this else, technically

dmcaulay: i need it for the scope, but it is a little awkward. i could split the assignment and the error check.

glibsm: oh, i see. Not a big deal

PR #208, opened 2017-02-01
Add config.ResolvePath helper for resolving relative paths

dmcaulay:
ResolvePath returns an absolute path derived from AppRoot and the relative path. If the input parameter is is already an absolute path it will be returned immediately.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -23,6 +23,8 @@ package config
 import (
 	"fmt"
 	"os"
+	"path"
+	"path/filepath"
 	"strings"
 	"sync"
 )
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -74,6 +75,19 @@ func AppRoot() string {
 	}
 }
 
+// ResolvePath returns an absolute path derived from AppRoot and the relative path.
+// If the input parameter is is already an absolute path it will be returned immediately.
```
sectioneight: nit: `is is`

config/config.go
```diff
@@ -74,6 +75,19 @@ func AppRoot() string {
 	}
 }
 
+// ResolvePath returns an absolute path derived from AppRoot and the relative path.
+// If the input parameter is already an absolute path it will be returned immediately.
+func ResolvePath(relative string) (string, error) {
```
alsamylkin: Where is it used?

dmcaulay: this is used by fx consumers. there's a diff out that uses it in the uber private repo. i'll also be using it in other projects.

dmcaulay:
i'm still not authorized to write to master. feel free to merge for me :). thanks!

PR #209, opened 2017-02-02
Use a backend create function to initialize the task module

madhuravi:


madhuravi:
@uber-go/service-framework

modules/task/README.md
```diff
@@ -23,9 +23,17 @@ import (
...
+    task.NewModule(newBackend),
   ).Build()
-  task.Register(updateCache)
+  svc.Start()
+  if err := task.Register(updateCache); err != nil {
+    // handle error
+  }
+}
+
+func newBackend(host service.Host) (task.Backend, error) {
```
anuptalwalkar: Not sure why this needs to be in README.

modules/task/task.go
```diff
@@ -36,6 +36,7 @@ var (
 	_globalBackendMu sync.RWMutex
 	_globalBackend   Backend = &NopBackend{}
 	_asyncMod        service.Module
+	_asyncModErr     error
```
anuptalwalkar: Why is this global? You can just have it locally on line 53.

madhuravi: Need it to be global so every time the function is called, it returns the module and error from the very first call.

modules/task/README.md
madhuravi: I was torn about this too. What do you suggest instead?

PR #210, opened 2017-02-02
Change config application keys to have consistent naming within fx

madhuravi:


service/core.go
```diff
@@ -126,25 +126,25 @@ func (s *serviceCore) AuthClient() auth.Client {
 }
 
 func (s *serviceCore) Name() string {
```
dmcaulay: Should `Name()` be changed to `ID()`?

madhuravi: I think Name is fine. The main reason I changed the names is so it's clear what's missing in the error messages :) when you have a service object, calling Name() makes more sense since it is the service name and id can indicate a number.

dmcaulay: It's not 100% clear that application ID maps back to the service Name. Did you consider renaming the yaml tags to service instead of renaming the config attribute names to application?

alsamylkin: Should we change the struct name too?
serviceCore -> applicationCore :)

madhuravi: There is an isolation here of how the user configures in their yaml "applicationID" versus how we use it internally.
UberFx defines a "service" and we use service extensively within our repo. I don't want to name this applicationCore when it is inside a service struct. It will add confusion within the repo.

madhuravi: You're right to point this out. In UberFx code, we usually use "service" and not "application". I believe the reason we kept the config keys "application.*" is to keep parity with existing services and avoid introducing new terms. In a lot of our repos, we use applicationID so it will easier to adopt the same terminology for config. It is translated internally to the standardConfig and the key names match the yaml but everywhere else we use service.

dmcaulay: There are a lot of repos using `service_name` and new services are actually generated with `serviceName` instead of `applicationID`.

madhuravi: Based on discussion with the team, decided to get rid of all application* and rename yaml keys to service*. It will be serviceName. Expect an update soon.

anuptalwalkar:
Looks good! update examples?

config/config.go
```diff
@@ -30,13 +30,13 @@ import (
 )
 
 const (
-	// ApplicationIDKey is the identifier of an application ID
-	ApplicationIDKey = "applicationID"
-	// ApplicationDescriptionKey is the configuration key of the application's
+	// ServiceNameKey is the config key of the service name
+	ServiceNameKey = "name"
```
glibsm: why is this `name` and not `serviceName`?

examples/simple/config/base.yaml
```diff
@@ -1,6 +1,6 @@
-applicationID: simple
-applicationDesc: A simple example service
-applicationOwner: owner@service.com
+name: simple
+description: A simple example service
+owner: owner@service.com
```
glibsm: Now all these are ambiguous. How about nesting them together, maybe?

```
service:
  name: simple
  description: foo
  lalala: bar
```

madhuravi: Thought about all the options and decided to strip away service and application for 2 reasons:
1. Making them top level constructs, it indicates they are metadata about the repo. At the top level, a name, owner and description indicate it is about the service.
2. To avoid confusion with users considering it a services vs application vs something else. At times, there are multiple services (thrift services for eg.) within a repo.

PR #211, opened 2017-02-02
Integrate richgo for better test output

sectioneight:
A few things are going on here:

First, we're upgrading overalls for a PR I merged this morning (https://github.com/go-playground/overalls/pull/18) that adds
suppor for richgo (https://github.com/kyoh86/richgo). We're currently
pointing at my fork of richgo until my PRs get merged.

Mostly we need this: https://github.com/kyoh86/richgo/pull/6

But I also fixed some stuff about TTY handling that I have an open issue
with the author for.

The second thing I changed is that the coverage data is now dumped to
coverage.txt by default. It grew large enough that it didn't fit on a
single page, or even three pages. The output of richgo actually gives
you package-level coverage and visualization so I think this is a good
tradeoff. A few new ways to run the tests, but the default is my
favorite:

```
make test TEST_FLAGS=-v
```

This shows off richgo's really neat features, but is more verbose

```
make test COVER=1
```

This returns the old behavior of showing the gigantic coverage breakdown
on stdout after each test run.

### DIFF
--- a/.build/deps.mk
+++ b/.build/deps.mk
@@ -18,6 +18,7 @@ deps: libdeps
 	$(ECHO_V)go install ./vendor/github.com/sectioneight/md-to-godoc
 	@$(call label,Installing interfacer...)
 	$(ECHO_V)go install ./vendor/github.com/mvdan/interfacer/cmd/interfacer
+	$(ECHO_V)richgo testfilter 2>/dev/null || ($(call label,Installing richgo) && go get github.com/sectioneight/richgo)
 
 GOCOV := gocov
 OVERALLS := overalls
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 *.out
 *.html
 *.coverprofile
+coverage.txt
 *.pprof
 node_modules
 /.bin
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,13 @@ else
 _FILTER_OVERALLS = grep -v "^Processing:"
 endif
 
+COV_TXT = coverage.txt
+ifeq ($(COVER),1)
+_OUTPUT_COVERAGE = cat -
+else
+_OUTPUT_COVERAGE = cat - > $(COV_TXT)
+endif
+
 # This is the default for overalls
 COVER_OUT := profile.coverprofile
 
@@ -47,6 +54,7 @@ $(COV_REPORT): $(PKG_FILES) $(ALL_SRC)
 
 	@$(call label,Running tests)
 	$(ECHO_V)$(OVERALLS) -project=$(PROJECT_ROOT) \
+		-go-binary=richgo \
 		-ignore "$(OVERALLS_IGNORE)" \
 		-covermode=atomic \
 		$(DEBUG_FLAG) -- \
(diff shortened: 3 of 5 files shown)


sectioneight:
Hmm, going to have to fix some weird dependency stuff. Fun.

sectioneight:
Ok, here's the default local output:

image

Note that it includes aggregate coverage as well as test results. If you want the full coverage like you used to have, you can do `COVER=1` when you run `make test` or you can `cat coverage.txt` afterwards:

image

Here's `make test V=1`:

image

I really, really like it. What do you all think?

If you want to run it for an individual test, you can just replace `go test` with `richgo test`, e.g.:

image

sectioneight:
cc @shawnburke who expressed interest in this, also I guess @prashantv and @peter-edge

.build/deps.mk
```diff
@@ -18,6 +18,7 @@ deps: libdeps
 	$(ECHO_V)go install ./vendor/github.com/sectioneight/md-to-godoc
 	@$(call label,Installing interfacer...)
 	$(ECHO_V)go install ./vendor/github.com/mvdan/interfacer/cmd/interfacer
+	$(ECHO_V)richgo testfilter 2>/dev/null || ($(call label,Installing richgo) && go get github.com/sectioneight/richgo)
```
sectioneight: Note: intentionally not vendoring this and just testing for the `testfilter` command because I don't want to encourage living on my fork for any longer than required

dmcaulay: this didn't work out of the box for me. i ran make and saw are issues with overalls. as a result, i manually upgraded overalls. i then saw that i was missing richgo so i manually installed it. my output is also not as pretty.

screenshot 2017-02-02 13 58 35

sectioneight:
What do you mean "didn't work out of the box"? What sort of issues with
overalls? When you manually upgraded it, to what version? How did you
manually install richgo? We're using custom versions of both tools so it's
important you have the ones provided by the repo. Weird that it didn't
install them for you... do you have error messages?

dmcaulay:
I meant that I pulled the latest, ran make and saw a overalls error. Sorry for not capturing it, but I believe it was an invalid arg issue.

I used go get to install both packages. I used the standard packages for overalls and coveralls and your branch when I installed gorich. That's probably the issue.

sectioneight:
So it's fixed now? Or you're still having issues?

dmcaulay:
Tests pass, but the output doesn't look like your image. I'll try the correct versions when I get a chance.

dmcaulay:
I ran

```
$ go install ./vendor/github.com/mattn/goveralls
$ go install ./vendor/github.com/go-playground/overalls
$ go get github.com/sectioneight/richgo
```

Tests pass, but the output looks like the screenshot above. Is this working for everyone else?

sectioneight:
Team, can you chime in here? I can't repro on my workstation but my
environment is probably in a special state given that I was patching and
testing the requisite supporting tools to make this work, and may not have
imported them into our build rules properly

alsamylkin:
I see the same screen as Dan.

sectioneight:
Blast. I'll fix this when I get back.

sectioneight:
Ok, I figured it out. I forgot to push the terminal-color-detection fix from master -> origin/master (https://github.com/sectioneight/richgo/commit/7e4cfa07f15bdb0ca918fc958f8b07c04abbf9ea)

Medium-term I'm working with the richgo author to fix the terminal color detection, at which point I'll lock in the version in glide and we'll get consistent output.

Short term run `rm $(which richgo) && make deps` and this should fix the output.

dmcaulay:
This is working for me. I also had to remove the source so `go get` would refresh it. Pulling the latest would also work.

PR #216, opened 2017-02-03
Fix formatting and redundant code in task docs

madhuravi:


modules/task/README.md
```diff
@@ -38,10 +38,8 @@ func newBackend(host service.Host) (task.Backend, error) {
 
 func runActivity(input string) error {
   // do things
-  results := "results"
-  if err := task.Enqueue(updateCache, input, results); err != nil {
-    return err
-  }
+  results := // calculate results here
```
alsamylkin: I am pretty sure this^ will fail to compile :)

madhuravi: I'm aware. This is a place holder where stuff goes.

alsamylkin: I prefer it to be compilable: 
It makes easy to test examples and for us to think hard about APIs

Minor thing anyway.

modules/task/README.md
```diff
@@ -17,18 +17,19 @@ from the backend, they will execute the function.
 package main
 
 import (
-	"go.uber.org/fx/modules/task"
+  "go.uber.org/fx/modules/task"
   "go.uber.org/fx/service"
+	"go.uber.org/fx/ulog"
```
alsamylkin: Formatting?:)

PR #217, opened 2017-02-03
Thriftsync to work with yarpc generated code

anuptalwalkar:


anuptalwalkar:
sure. @abhinav? 
The couple of problems addressed here that I was running into -
1. creating server from thriftrw generated code. (This has uber specific stuff, so I made it parameterized as YARPC server)
2. Clean return variables. Pretty much that if string is returned, don't return `nil` for compile issues.

modules/rpc/thriftrw-plugin-thriftsync/generate.go
```diff
@@ -32,27 +32,32 @@ const newHandlerTemplate = `
...
   // TODO: modify the <.Service.Name> handler with your suitable structure
 }
 // NewYARPCThriftHandler for your service
-func NewYARPCThriftHandler(<$service>.Host) ([]<$transport>.Procedure, error) {
-  handler := &YARPCHandler{}
+func NewYARPCThriftHandler(host <$service>.Host) ([]<$transport>.Procedure, error) {
+  handler := &YARPCHandler{
+		host: host,
+	}
+	<$yarpcserver := import $yarpcServerPath>
```
abhinav: unused import/variable

modules/rpc/thriftrw-plugin-thriftsync/generate.go
```diff
@@ -32,27 +32,32 @@ const newHandlerTemplate = `
...
   return <lower .Service.Name>server.New(handler), nil
 }
 
 <$name := lower .Service.Name>
 <range .Service.Functions>
 
-  func (h *YARPCHandler)<.Name>(ctx fx.Context, <range .Arguments> <lower .Name> <formatType .Type>, <end>)<if .ReturnType> (<formatType .ReturnType>, error) <else> error <end> {
+  func (h *YARPCHandler)<.Name>(ctx <$context>.Context, <range .Arguments> <lower .Name> <formatType .Type>, <end>)<if .ReturnType> (<formatType .ReturnType>, error) <else> error <end> {
```
abhinav: This will lower case the full argument name. Consider writing a little function to lower case only the first character and make it available in the template using TemplateFunc.

modules/rpc/thriftrw-plugin-thriftsync/main.go
```diff
@@ -45,6 +48,14 @@ var (
 
 var templateOptions = []plugin.TemplateOption{
 	plugin.TemplateFunc("lower", strings.ToLower),
+	plugin.TemplateFunc("isStringType", func(spec api.Type) bool {
+		simple := *spec.SimpleType
```
abhinav: nil pointer dereference if `api.Type` is not a simple type

modules/rpc/thriftrw-plugin-thriftsync/generate.go
anuptalwalkar: Oh, I actually wanted to use it in the next line. Fixing it.

modules/rpc/thriftrw-plugin-thriftsync/generate.go
anuptalwalkar: +1

modules/rpc/thriftrw-plugin-thriftsync/generate.go
```diff
@@ -32,27 +32,32 @@ const newHandlerTemplate = `
...
 }
 
 <$name := lower .Service.Name>
 <range .Service.Functions>
 
-  func (h *YARPCHandler)<.Name>(ctx fx.Context, <range .Arguments> <lower .Name> <formatType .Type>, <end>)<if .ReturnType> (<formatType .ReturnType>, error) <else> error <end> {
+  func (h *YARPCHandler)<.Name>(ctx <$context>.Context, <range .Arguments> <lowerFirst .Name> <formatType .Type>, <end>)<if .ReturnType> (<formatType .ReturnType>, error) <else> error <end> {
     // TODO: write your code here
-    return nil
+    return <if .ReturnType> <if isStringType .ReturnType> "", <else> nil, <end> <end> nil
```
abhinav: Is the intention to return zero values here by default?

That'll require a fair amount of work to support all types. It'll be easier if
you declare a var with the type and return that or use a named return value:

    func <.Name>(...) (success <formatType .ReturnType, err error) {
        // TODO: Write your code here
        return
    }
    
Although if people are going to be modifying this code, we should probably not
advise them to use named returns. Maybe the generated function should panic by
default?

    func <.Name>(...) (<formatType .ReturnType, error) {
        // TODO: Write your code here
        panic("To be implemented")
    }
    
This saves you the trouble of having to worry about zero values of all
possible types.

anuptalwalkar: My intention was to return `nil` and let user decide what they want to return. The only weird case was about string, where I can't return nil causing compilation errors. If I panic the code won't be compilable on booting process.

abhinav: What about the case where the return type is an integer or boolean?

anuptalwalkar: Ah I see what you mean. I think panic here makes sense then instead of setting all zero values since we are not changing return type on method updates as well.

anuptalwalkar:
ping?

modules/rpc/thriftrw-plugin-thriftsync/generate.go
```diff
@@ -32,27 +32,32 @@ const newHandlerTemplate = `
...
 type YARPCHandler struct{
+	host <$service>.Host
   // TODO: modify the <.Service.Name> handler with your suitable structure
 }
 // NewYARPCThriftHandler for your service
-func NewYARPCThriftHandler(<$service>.Host) ([]<$transport>.Procedure, error) {
-  handler := &YARPCHandler{}
-  return <lower .Service.Name>server.New(handler), nil
+func NewYARPCThriftHandler(host <$service>.Host) ([]<$transport>.Procedure, error) {
+  handler := &YARPCHandler{
```
abhinav: nit: inconsistent indentation here and below

modules/rpc/thriftrw-plugin-thriftsync/generate.go
```diff
@@ -32,27 +32,32 @@ const newHandlerTemplate = `
...
 }
 
 <$name := lower .Service.Name>
 <range .Service.Functions>
 
-  func (h *YARPCHandler)<.Name>(ctx fx.Context, <range .Arguments> <lower .Name> <formatType .Type>, <end>)<if .ReturnType> (<formatType .ReturnType>, error) <else> error <end> {
+  func (h *YARPCHandler)<.Name>(ctx <$context>.Context, <range .Arguments> <lowerFirst .Name> <formatType .Type>, <end>)<if .ReturnType> (<formatType .ReturnType>, error) <else> error <end> {
     // TODO: write your code here
-    return nil
+    panic("Implement return values")
```
abhinav: nit: Try "To be implemented" or similar.

modules/rpc/thriftrw-plugin-thriftsync/main.go
```diff
@@ -45,6 +49,22 @@ var (
 
 var templateOptions = []plugin.TemplateOption{
 	plugin.TemplateFunc("lower", strings.ToLower),
+	plugin.TemplateFunc("isStringType", func(spec api.Type) bool {
```
abhinav: unused function

anuptalwalkar:
Thanks, yes tested it locally. I will add some concrete testing with examples.

PR #218, opened 2017-02-04
Log stacktrace for wrapped errors

alsamylkin:
If an error contains StackTrace() we'll wrap using format:
"functionName":"{filename}:line#" with the full error message.

### DIFF
--- a/ulog/log.go
+++ b/ulog/log.go
@@ -197,9 +197,9 @@ func (l *baseLogger) fieldsConversion(keyVals ...interface{}) []zap.Field {
 				stack := value.StackTrace()
 				lines := make([]zap.Field, 0, len(stack))
 				for _, frame := range stack {
-					s := fmt.Sprintf("%s", frame)
-					l := fmt.Sprintf("line:%d", frame)
-					lines = append(lines, zap.String(s, l))
+					function := fmt.Sprintf("%n", frame)
+					source := fmt.Sprintf("%v", frame)
+					lines = append(lines, zap.String(function, source))
 				}
 
 				fields = append(fields, zap.Nest("stacktrace", lines...), zap.Error(value))
--- a/ulog/log.go
+++ b/ulog/log.go
@@ -21,7 +21,7 @@
 package ulog
 
 import (
-	err "errors"
+	stdlibErr "errors"
 	"fmt"
 	"time"
 
@@ -162,7 +162,7 @@ type stackTracer interface {
 
 func (l *baseLogger) fieldsConversion(keyVals ...interface{}) []zap.Field {
 	if len(keyVals)%2 != 0 {
-		return []zap.Field{zap.Error(err.New("expected even number of arguments"))}
+		return []zap.Field{zap.Error(stdlibErr.New("expected even number of arguments"))}
 	}
 
 	fields := make([]zap.Field, 0, len(keyVals)/2)
--- a/ulog/log.go
+++ b/ulog/log.go
@@ -21,11 +21,13 @@
 package ulog
 
 import (
+	err "errors"
 	"fmt"
 	"time"
 
 	"go.uber.org/fx/ulog/sentry"
 
+	"github.com/pkg/errors"
 	"github.com/uber-go/zap"
 )
 
@@ -153,12 +155,17 @@ func (l *baseLogger) Log(lvl zap.Level, message string, keyVals ...interface{}) 
 	}
 }
 
+type stackTracer interface {
+	error
+	StackTrace() errors.StackTrace
+}
+
 func (l *baseLogger) fieldsConversion(keyVals ...interface{}) []zap.Field {
-	fields := make([]zap.Field, 0, len(keyVals)/2)
 	if len(keyVals)%2 != 0 {
-		fields = append(fields, zap.Error(fmt.Errorf("expected even number of arguments")))
-		return fields
+		return []zap.Field{zap.Error(err.New("expected even number of arguments"))}
 	}
+
+	fields := make([]zap.Field, 0, len(keyVals)/2)
 	for idx := 0; idx < len(keyVals); idx += 2 {
 		if key, ok := keyVals[idx].(string); ok {
 			switch value := keyVals[idx+1].(type) {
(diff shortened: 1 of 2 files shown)


ulog/log.go
```diff
@@ -21,11 +21,13 @@
 package ulog
 
 import (
+	err "errors"
```
glibsm: wat

ulog/log.go
```diff
@@ -21,11 +21,13 @@
...
 	"fmt"
 	"time"
 
 	"go.uber.org/fx/ulog/sentry"
 
+	"github.com/pkg/errors"
```
glibsm: i see. ugh

ulog/log.go
```diff
@@ -153,12 +155,17 @@ func (l *baseLogger) Log(lvl zap.Level, message string, keyVals ...interface{})
 	}
 }
 
+type stackTracer interface {
```
glibsm: i thought there was already an interface for that... `Causer` or something, no?

ulog/log.go
```diff
@@ -186,6 +193,16 @@ func (l *baseLogger) fieldsConversion(keyVals ...interface{}) []zap.Field {
 				fields = append(fields, zap.Marshaler(key, value))
 			case fmt.Stringer:
 				fields = append(fields, zap.Stringer(key, value))
+			case stackTracer:
+				stack := value.StackTrace()
+				lines := make([]zap.Field, 0, len(stack))
+				for _, frame := range stack {
+					s := fmt.Sprintf("%s", frame)
+					l := fmt.Sprintf("line:%d", frame)
```
glibsm: I think `file:line` is more like go traces, rather than `file: foo, line: 123`
If that's the case, we can just do an array of strings for trace, thoughts?

ulog/log.go
glibsm: Nvm, Cause is just for the original error, if you need that sort of thing, which in here we dont.

alsamylkin: Unfortunately, no: https://github.com/pkg/errors/blob/master/errors.go#L67
stack is a private struct.

ulog/log.go
madhuravi: you're using frame in both s and l. Did you mean to read the index from the for loop?

alsamylkin: I don't think zap supports arrays: only key/value pairs. I am mostly concerned about the size of the logged string, the entire trace will be printed in one line. Logging each line separately should be better, but I am afraid we can lose error then.(In terms of context, something else can be logged at the same time)

ulog/log.go
madhuravi: can you name it stdliberr? It is a bit confusing which one is which.

ulog/log.go
alsamylkin: No, just reading source file and line number:
https://github.com/pkg/errors/blob/master/stack.go#L42

madhuravi: You can achieve this with removing "line:"
s := fmt.Sprintf("%s", frame)
l := fmt.Sprintf("%d", frame)
lines = append(lines, zap.String(s, l))

alsamylkin: `line` is intentional. Otherwise it can be confusing, what does this magic number mean?
If you think it is obvious, I can remove `line` though

ulog/log_test.go
```diff
@@ -161,3 +162,18 @@ func TestSentryHookDoesNotMutatePrevious(t *testing.T) {
...
 }
+
+func TestStackTraceLogger(t *testing.T) {
+	t.Parallel()
+	testutils.WithInMemoryLogger(t, nil, func(zapLogger zap.Logger, buf *testutils.TestBuffer) {
+		log := Builder().SetLogger(zapLogger).Build()
+		err1 := errors.New("for sure")
+		err2 := errors.Wrap(err1, "it's a trap")
+		log.Error("error message", "error", err2)
+		trace := `{"level":"error","msg":"error message","stacktrace":{"log_test.go":"line:171","in_memory_log.go":"line:60","log_test.go":"line:178"`
```
madhuravi: Can we include more information? You can get the function name too with %n from https://github.com/pkg/errors/blob/master/stack.go#L42

alsamylkin: We can, my only concern is readability: the line is going to be really big. I've already trimmed it to have only file name not a full path.
A function name is a bit useless in case if we have line numbers. We can change provided information based on a feedback.

ulog/log.go
madhuravi: Most people are used to seeing line numbers in stack traces. I think we can remove it.

madhuravi: Also from here https://dave.cheney.net/2016/06/12/stack-traces-and-the-errors-package
fmt.Printf("%+v\n", err)
will output:
not enough arguments, expected at least 3, got 0
main.parseArgs
        /home/dfc/src/github.com/pkg/errors/_examples/wrap/main.go:12
main.main
        /home/dfc/src/github.com/pkg/errors/_examples/wrap/main.go:18
runtime.main
        /home/dfc/go/src/runtime/proc.go:183
runtime.goexit
        /home/dfc/go/src/runtime/asm_amd64.s:2059

This seems like what we want. Instead of stackTraces, we can then just do:

case error:
fields = append(fields, zap.String("error", fmt.Sprintf("%v", err)))

ulog/log_test.go
madhuravi: Understand the readability concerns but function name is very useful. It helps track down the root cause much faster and the information is right there whereas with the line number you almost always have to look up the file and go to the line and trace back up to see what func it is. The func names  also give a quick sense of the trace of function calls it took to get to this error.

ulog/log.go
alsamylkin: Not really: we want to have error and stack trace to be in separate fields. This will just do a string interpolation we try to avoid.

madhuravi: That is orthogonal. We can still have them in separate fields:
case error:
line = fmt.Sprintf("%v", err)
fields = append(fields, zap.Nest("stacktrace", line), zap.Error(value))

alsamylkin: `line` will contain the error message, right?

ulog/log_test.go
alsamylkin: Make sense

anuptalwalkar:
Will this be useful once we move to zap 1.0? AFAIK, we won't require fieldConversion method, and use sugar logger directly. Do we want to add stackTrace under sugar logger then?

glibsm:
Zap also has `AddCaller` hook which attaches a stacktrace to all the logging messages. The downside is that it will re-sample the trace, i think. Might be worthwhile to see if we can contribute this to zap, since we already have a trace we just need to log it.

alsamylkin:
We want to have a stack trace of an error, not a logger :)

PR #219, opened 2017-02-04
Populate struct with non-ptr user defined types

anuptalwalkar:
1. populateStruct when variables are non pointers, and user defined types
2. Refactored convertValue to work with both primitive type Get() and PopulateStruct()

### DIFF
--- a/config/value.go
+++ b/config/value.go
@@ -497,7 +497,7 @@ func (cv Value) valueStruct(key string, target interface{}) (interface{}, error)
 			}
 			if val != nil {
 				// First try to convert primitive type values, if convertValue wasn't able
-				// to convert to primitive,try converting the value asa struct value
+				// to convert to primitive,try converting the value as a struct value
 				if ret, err := convertValue(val, fieldType); ret != nil {
 					if err != nil {
 						return nil, err
--- a/config/value.go
+++ b/config/value.go
@@ -343,8 +343,8 @@ func convertValueFromStruct(value interface{}, targetType reflect.Type, fieldTyp
 	// The fieldType is probably a custom type here. We will try and set the fieldValue by
 	// the custom type
 	// TODO: refactor switch cases into isType functions
-	kind := fieldType.Kind()
-	switch kind {
+	// TODO: refactor this whole method
+	switch fieldType.Kind() {
 	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
 		fieldValue.SetInt(int64(value.(int)))
 	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64, reflect.Uintptr:
--- a/config/value.go
+++ b/config/value.go
@@ -333,17 +333,11 @@ func derefType(t reflect.Type) reflect.Type {
 	return t
 }
 
-func convertValueFromStruct(value interface{}, targetType reflect.Type, fieldType reflect.Type, fieldValue reflect.Value) (interface{}, error) {
-	ret, err := convertValue(value, targetType)
-	if ret != nil {
-		// convertValue was a success
-		return ret, err
-	}
+func convertValueFromStruct(value interface{}, targetType reflect.Type, fieldType reflect.Type, fieldValue reflect.Value) error {
 
 	// The fieldType is probably a custom type here. We will try and set the fieldValue by
 	// the custom type
 	// TODO: refactor switch cases into isType functions
-	// TODO: refactor this whole method
 	switch fieldType.Kind() {
 	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
 		fieldValue.SetInt(int64(value.(int)))
@@ -356,10 +350,9 @@ func convertValueFromStruct(value interface{}, targetType reflect.Type, fieldTyp
 	case reflect.String:
 		fieldValue.SetString(value.(string))
 	default:
-		return nil, fmt.Errorf("can't convert %v to %v", reflect.TypeOf(value).String(), targetType)
+		return fmt.Errorf("can't convert %v to %v", reflect.TypeOf(value).String(), targetType)
 	}
-
-	return ret, nil
+	return nil
 }
 
 // this is a quick-and-dirty conversion method that only handles
@@ -371,7 +364,6 @@ func convertValue(value interface{}, targetType reflect.Type) (interface{}, erro
 	}
 
 	valueType := reflect.TypeOf(value)
-
 	if valueType.AssignableTo(targetType) {
 		return value, nil
 	} else if targetType.Name() == "string" {
@@ -387,7 +379,6 @@ func convertValue(value interface{}, targetType reflect.Type) (interface{}, erro
 			return strconv.ParseBool(v)
 		case encoding.TextUnmarshaler:
 			err := t.UnmarshalText([]byte(v))
-
 			// target should have a pointer receiver to be able to change itself based on text
 			return reflect.ValueOf(target).Elem().Interface(), err
 		}
@@ -506,13 +497,18 @@ func (cv Value) valueStruct(key string, target interface{}) (interface{}, error)
 				val = fieldInfo.DefaultValue
 			}
 			if val != nil {
-				v3, err := convertValueFromStruct(val, fieldValue.Type(), fieldType, fieldValue)
-				if err != nil {
+				ret, err := convertValue(val, fieldType)
+				// For Unmarshalled text, if convertValue fails, and not returns
+				// non nil value, we fail populateStruct call by returning the error
+				if ret != nil && err != nil {
 					return nil, err
-				}
-				if v3 != nil {
-					val = v3
-					fieldValue.Set(reflect.ValueOf(val))
+				} else if ret != nil {
+					fieldValue.Set(reflect.ValueOf(ret))
+				} else {
+					err := convertValueFromStruct(val, fieldValue.Type(), fieldType, fieldValue)
+					if err != nil {
+						return nil, err
+					}
 				}
 			}
 			continue
--- a/config/value.go
+++ b/config/value.go
@@ -334,12 +334,14 @@ func derefType(t reflect.Type) reflect.Type {
 }
 
 func convertValueFromStruct(value interface{}, targetType reflect.Type, fieldType reflect.Type, fieldValue reflect.Value) (interface{}, error) {
-	if ret, err := convertValue(value, targetType); ret != nil {
+	ret, err := convertValue(value, targetType)
+	if ret != nil {
+		// convertValue was a success
 		return ret, err
 	}
 
-	// The fieldType is probably custom type here. We will try and set the fieldValue by
-	// the type of custom type
+	// The fieldType is probably a custom type here. We will try and set the fieldValue by
+	// the custom type
 	// TODO: refactor switch cases into isType functions
 	kind := fieldType.Kind()
 	switch kind {
@@ -356,7 +358,8 @@ func convertValueFromStruct(value interface{}, targetType reflect.Type, fieldTyp
 	default:
 		return nil, fmt.Errorf("can't convert %v to %v", reflect.TypeOf(value).String(), targetType)
 	}
-	return nil, nil
+
+	return ret, nil
 }
 
 // this is a quick-and-dirty conversion method that only handles
--- a/config/value.go
+++ b/config/value.go
@@ -334,7 +334,6 @@ func derefType(t reflect.Type) reflect.Type {
 }
 
 func convertValueFromStruct(value interface{}, targetType reflect.Type, fieldType reflect.Type, fieldValue reflect.Value) error {
-
 	// The fieldType is probably a custom type here. We will try and set the fieldValue by
 	// the custom type
 	// TODO: refactor switch cases into isType functions
@@ -497,16 +496,15 @@ func (cv Value) valueStruct(key string, target interface{}) (interface{}, error)
 				val = fieldInfo.DefaultValue
 			}
 			if val != nil {
-				ret, err := convertValue(val, fieldType)
-				// For Unmarshalled text, if convertValue fails, and not returns
-				// non nil value, we fail populateStruct call by returning the error
-				if ret != nil && err != nil {
-					return nil, err
-				} else if ret != nil {
+				// First try to convert primitive type values, if convertValue wasn't able
+				// to convert to primitive,try converting the value asa struct value
+				if ret, err := convertValue(val, fieldType); ret != nil {
+					if err != nil {
+						return nil, err
+					}
 					fieldValue.Set(reflect.ValueOf(ret))
 				} else {
-					err := convertValueFromStruct(val, fieldValue.Type(), fieldType, fieldValue)
-					if err != nil {
+					if err := convertValueFromStruct(val, fieldValue.Type(), fieldType, fieldValue); err != nil {
 						return nil, err
 					}
 				}
--- a/config/value.go
+++ b/config/value.go
@@ -348,6 +348,7 @@ func convertStructValue(value interface{}, targetType reflect.Type, fieldType re
 
 	// The fieldType is probably custom type here. We will try and set the fieldValue by
 	// the type of custom type
+	// TODO: refactor switch cases into isType functions
 	kind := fieldType.Kind()
 	switch kind {
 	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
@@ -361,7 +362,7 @@ func convertStructValue(value interface{}, targetType reflect.Type, fieldType re
 	case reflect.String:
 		fieldValue.SetString(value.(string))
 	default:
-		return nil, fmt.Errorf("can't convert %v to %v", reflect.TypeOf(value), targetType)
+		return nil, fmt.Errorf("can't convert %v to %v", reflect.TypeOf(value).String(), targetType)
 	}
 	return nil, nil
 }
--- a/config/value.go
+++ b/config/value.go
@@ -336,7 +336,7 @@ func derefType(t reflect.Type) reflect.Type {
 func convertPrimitiveValue(value interface{}, targetType reflect.Type) (interface{}, error) {
 	ret, err := convertVal(value, targetType)
 	if ret == nil {
-		return nil, fmt.Errorf("can't convert %v to %v", reflect.TypeOf(value), targetType)
+		return nil, fmt.Errorf("can't convert %v to %v", reflect.TypeOf(value).String(), targetType)
 	}
 	return ret, err
 }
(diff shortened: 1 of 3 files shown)


config/value.go
```diff
@@ -333,10 +333,41 @@ func derefType(t reflect.Type) reflect.Type {
 	return t
 }
 
+func convertPrimitiveValue(value interface{}, targetType reflect.Type) (interface{}, error) {
+	ret, err := convertVal(value, targetType)
+	if ret == nil {
```
alsamylkin: We don't care about err, but ret? This behavior is very confusing.

config/value.go
```diff
@@ -333,10 +333,41 @@ func derefType(t reflect.Type) reflect.Type {
...
+		return nil, fmt.Errorf("can't convert %v to %v", reflect.TypeOf(value), targetType)
```
alsamylkin: Use types %t instead of %v? Or reflect.TypeOf(value).String()?

config/value.go
```diff
@@ -333,10 +333,41 @@ func derefType(t reflect.Type) reflect.Type {
...
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		fieldValue.SetInt(int64(value.(int)))
+	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64, reflect.Uintptr:
+		fieldValue.SetUint(uint64(value.(int)))
+	case reflect.Float32, reflect.Float64:
+		fieldValue.SetFloat(float64(value.(float64)))
+	case reflect.String:
+		fieldValue.SetString(string(value.(string)))
+	default:
+		return nil, fmt.Errorf("can't convert %v to %v", reflect.TypeOf(value), targetType)
```
alsamylkin: What about booleans, duration/(text.Unmarshaller)?

config/yaml_test_data.go
```diff
@@ -138,12 +138,24 @@ type nestedTypeUInt uint32
...
+type nestedTypeStructPtr struct {
 	TestInt   *userDefinedTypeInt   `yaml:"testInt"`
 	TestUInt  *userDefinedTypeUInt  `yaml:"testUInt"`
 	TestFloat *userDefinedTypeFloat `yaml:"testFloat"`
 }
 
+type typeStructPtr nestedTypeStructPtr
+
+type typeStructStructPtr struct {
+	TypeStruct typeStructPtr `yaml:"typeStruct"`
```
alsamylkin: What is going to happen if we'll use a pointer to struct:
```
TypeStruct *typeStructPtr
```
?

config/yaml_test.go
```diff
@@ -255,6 +255,17 @@ func TestTypeOfTypes(t *testing.T) {
 		tts := typeStructStruct{}
 		err := provider.Get(Root).PopulateStruct(&tts)
 		assert.NoError(t, err)
+		assert.Equal(t, userDefinedTypeInt(123), tts.TypeStruct.TestInt)
+		assert.Equal(t, userDefinedTypeUInt(456), tts.TypeStruct.TestUInt)
+		assert.Equal(t, userDefinedTypeFloat(123.456), tts.TypeStruct.TestFloat)
```
alsamylkin: I remember we had some issues with converting a string that was written as int, but the target type is actually a float, could you test it here too?

config/value.go
```diff
@@ -333,10 +333,41 @@ func derefType(t reflect.Type) reflect.Type {
...
+	if ret, err := convertVal(value, targetType); ret != nil {
+		return ret, err
+	}
+
+	// The fieldType is probably custom type here. We will try and set the fieldValue by
+	// the type of custom type
+	kind := fieldType.Kind()
+	switch kind {
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		fieldValue.SetInt(int64(value.(int)))
```
alsamylkin: We just assume that fieldType and value type are the same? Can value be e.g. a float here?

config/yaml_test.go
```diff
@@ -255,6 +255,17 @@ func TestTypeOfTypes(t *testing.T) {
...
+	})
+}
+
+func TestTypeOfTypesPtr(t *testing.T) {
```
alsamylkin: Where is a test for strings?

config/value.go
madhuravi: +1 safer to check for err, and it's the convention.

config/value.go
```diff
@@ -333,10 +333,41 @@ func derefType(t reflect.Type) reflect.Type {
...
+func convertPrimitiveValue(value interface{}, targetType reflect.Type) (interface{}, error) {
+	ret, err := convertVal(value, targetType)
+	if ret == nil {
+		return nil, fmt.Errorf("can't convert %v to %v", reflect.TypeOf(value), targetType)
+	}
+	return ret, err
+}
+
+func convertStructValue(value interface{}, targetType reflect.Type, fieldType reflect.Type, fieldValue reflect.Value) (interface{}, error) {
+	if ret, err := convertVal(value, targetType); ret != nil {
```
madhuravi: same as above, check for err and not ret

config/value.go
```diff
@@ -333,10 +333,41 @@ func derefType(t reflect.Type) reflect.Type {
 	return t
 }
 
+func convertPrimitiveValue(value interface{}, targetType reflect.Type) (interface{}, error) {
```
madhuravi: This method seems redundant. Why can't we call convertVal directly?

config/value.go
```diff
@@ -333,10 +333,41 @@ func derefType(t reflect.Type) reflect.Type {
...
+	// The fieldType is probably custom type here. We will try and set the fieldValue by
+	// the type of custom type
+	kind := fieldType.Kind()
+	switch kind {
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		fieldValue.SetInt(int64(value.(int)))
+	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64, reflect.Uintptr:
+		fieldValue.SetUint(uint64(value.(int)))
+	case reflect.Float32, reflect.Float64:
+		fieldValue.SetFloat(float64(value.(float64)))
```
madhuravi: why do we need all these levels? fieldValue.SetFloat(val.(float64)) doesn't work? Same for other case statements.

config/value.go
```diff
@@ -333,10 +333,41 @@ func derefType(t reflect.Type) reflect.Type {
...
+func convertStructValue(value interface{}, targetType reflect.Type, fieldType reflect.Type, fieldValue reflect.Value) (interface{}, error) {
+	if ret, err := convertVal(value, targetType); ret != nil {
+		return ret, err
+	}
+
+	// The fieldType is probably custom type here. We will try and set the fieldValue by
+	// the type of custom type
+	kind := fieldType.Kind()
+	switch kind {
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
```
madhuravi: This is in a few places now, add it as a function.

switch {
  case isInt(kind):
}

func isInt() {
  return kind in [reflect.Int, reflect.Int8 ...]
}

config/value.go
anuptalwalkar: This is refactored from previous convertValue. The error is bubbled up to decide whether TryAs* method should try additional val.(type) or return false and panic. The most I can do here is to log, but it will be just too verbose at this point since the code panics with "Can't convert to" message.

config/value.go
anuptalwalkar: updated to .String()

config/value.go
anuptalwalkar: This is conversion into fieldValue which comes from struct, vs value comes from yaml and is perceived into the struct type.

config/value.go
anuptalwalkar: Forgot about those. Fixing it and adding tests :)

config/yaml_test_data.go
anuptalwalkar: Same? Any pointer structs will call populate struct. Added tests

config/yaml_test.go
anuptalwalkar: Its covered in config_tests.

config/yaml_test.go
anuptalwalkar: Added

config/yaml_test_data.go
alsamylkin: I was curious, who is allocating memory for them and where.

config/value.go
anuptalwalkar: This is a refactor since this method was previously common between primitive type conversion and populateStruct. We don't want to do additional type conversion.

config/value.go
anuptalwalkar: Doesn't work for all cases, just float.

config/yaml_test_data.go
anuptalwalkar: If you read through populateStruct, its a recursive call for `bucketObject`

config/value.go
madhuravi: Not sure if we are talking about the same thing. I just mean:
if err != nil {
return nil, fmt.Errorf("can't convert %v to %v", reflect.TypeOf(value).String(), targetType)
}
return ret, nil

i.e. the if condition should be on err and not ret

anuptalwalkar: The err check here is not relevant here. The only concern is to check if return type is nil (as in convertible primitive type or not) Thats why we check `v != nil && err == nil` in Try* method and perform further type casting. I can check err here, but it will just be a no op, since Try* methods are taking care of the rest.

anuptalwalkar: These are the three conditions which make err check irrelevant and I can simply jump to 2nd ret != nil check. 
```
// if err is not nil, type is not convertible and we don't need additional type conversion so panic out
if err != nil {
	return ret, err
}
// err is nil, but ret is nil, so we need additional type conversion
if ret != nil {
	return ret, err
}
// just return error
```

config/yaml_test.go
alsamylkin: I don't think it covered pointers, but may be it is not needed..

config/value.go
anuptalwalkar: I am not sure thats the syntax, but if I have to return reflect.Type, I'll need to allocate and return type. having it in case is clear for reading.

config/yaml_test.go
```diff
@@ -255,9 +255,46 @@ func TestTypeOfTypes(t *testing.T) {
...
+		assert.Equal(t, userDefinedTypeString("hello"), tts.TypeStruct.TestString)
+		assert.Equal(t, 10*time.Second, tts.TypeStruct.TestDuration.Seconds)
+		assert.Equal(t, 20*time.Minute, tts.TypeStruct.TestDuration.Minutes)
+		assert.Equal(t, 30*time.Hour, tts.TypeStruct.TestDuration.Hours)
+	})
+}
+
+func TestTypeOfTypesPtr(t *testing.T) {
+	withYamlBytes(t, typeStructYaml, func(provider Provider) {
+		tts := typeStructStructPtr{}
```
alsamylkin: Could you also add a test of array's of pointers? I think we process them differently..

config/value.go
```diff
@@ -333,10 +333,43 @@ func derefType(t reflect.Type) reflect.Type {
...
+	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64, reflect.Uintptr:
+		fieldValue.SetUint(uint64(value.(int)))
+	case reflect.Float32, reflect.Float64:
+		fieldValue.SetFloat(value.(float64))
+	case reflect.Bool:
+		fieldValue.SetBool(value.(bool))
+	case reflect.String:
+		fieldValue.SetString(value.(string))
+	default:
+		return nil, fmt.Errorf("can't convert %v to %v", reflect.TypeOf(value), targetType)
```
alsamylkin: Same as above, should we print types, not values?

config/value.go
```diff
@@ -363,8 +396,7 @@ func convertValue(value interface{}, targetType reflect.Type) (interface{}, erro
 			return reflect.ValueOf(target).Elem().Interface(), err
 		}
 	}
-
-	return nil, fmt.Errorf("can't convert %v to %v", valueType, targetType)
+	return nil, nil
```
alsamylkin: Why we don't return an error anymore?

config/value.go
madhuravi: I missed that, makes sense.

config/value.go
madhuravi: Not sure what you mean. I want to avoid having hardcoded lists of int types and float types everywhere.
So we would have:
kind := fieldType.Kind()
 	switch {
 	case isInt(kind):
               fieldValue.SetInt(int64(value.(int)))
        case isUInt(kind):
		fieldValue.SetUint(uint64(value.(int)))
        etc.

config/yaml_test.go
anuptalwalkar: I am not adding test for code I didn't write here. I am happy open a separate issue for that, otherwise this conversation will go on and on untill we cover all combinatorial of test cases..

config/value.go
alsamylkin: Is it a yes, or no?:)
What will happen if a struct field value was parsed as int, but actually has a float pointer type?

config/value.go
anuptalwalkar: Did you see the refactor? I have explained this above https://github.com/uber-go/fx/pull/219#discussion_r99660210

config/value.go
anuptalwalkar: updating it, but it doesn't make any difference %v of `reflect.TypeOf(value)` is a primitive type.

config/value.go
anuptalwalkar: 1. yes, we assume that the type are same or convertable via Try* methods. If it doesn't we panic. This code is accessed on bootstrapping anyway.

anuptalwalkar: "What will happen if a struct field value was parsed as int, but actually has a float pointer type?" panic will tell you not to do that.:)

alsamylkin: So TryAsFloat() will panic, correct?

anuptalwalkar: yes

config/value.go
alsamylkin: 1. It doesn't make sense to me to communicate through returned pointers(yeah, I know, I am weird :)
2. Could you please update the comments to the function then?

anuptalwalkar:
@madhuravi @glibsm We chatted offline. The major concern on this pr is basically the structure of the populateStruct calls, which can't be fixed without heavy refactor of value.go. The change however fixes the use case where struct has user defined type objects. If you guys feel we don't need this change right now, we can circle back in later time. Thoughts?

madhuravi:
I'm for fixing the issue. Currently, it is breaking rpc port but we might have users following this pattern don't want it to break for the.

config/value.go
```diff
@@ -333,6 +333,32 @@ func derefType(t reflect.Type) reflect.Type {
...
+	case reflect.Float32, reflect.Float64:
+		fieldValue.SetFloat(value.(float64))
+	case reflect.Bool:
+		fieldValue.SetBool(value.(bool))
+	case reflect.String:
+		fieldValue.SetString(value.(string))
+	default:
+		return nil, fmt.Errorf("can't convert %v to %v", reflect.TypeOf(value).String(), targetType)
+	}
+	return nil, nil
```
madhuravi: return an error if nil. Add this behavior to the comment.

config/value.go
```diff
@@ -476,13 +503,14 @@ func (cv Value) valueStruct(key string, target interface{}) (interface{}, error)
...
 			if val != nil {
-				v3, err := convertValue(val, fieldValue.Type())
+				v3, err := convertValueFromStruct(val, fieldValue.Type(), fieldType, fieldValue)
 				if err != nil {
 					return nil, err
 				}
-
-				val = v3
-				fieldValue.Set(reflect.ValueOf(val))
+				if v3 != nil {
```
madhuravi: This if check should be redundant now.

config/value.go
```diff
@@ -333,6 +333,35 @@ func derefType(t reflect.Type) reflect.Type {
...
+	if ret != nil {
+		// convertValue was a success
+		return ret, err
+	}
+
+	// The fieldType is probably a custom type here. We will try and set the fieldValue by
+	// the custom type
+	// TODO: refactor switch cases into isType functions
+	kind := fieldType.Kind()
+	switch kind {
```
madhuravi: nit: switch fieldType.Kind()

config/value.go
```diff
@@ -333,6 +333,35 @@ func derefType(t reflect.Type) reflect.Type {
 	return t
 }
 
+func convertValueFromStruct(value interface{}, targetType reflect.Type, fieldType reflect.Type, fieldValue reflect.Value) (interface{}, error) {
+	ret, err := convertValue(value, targetType)
+	if ret != nil {
```
madhuravi: this should be err == nil now.

config/value.go
```diff
@@ -333,6 +333,28 @@ func derefType(t reflect.Type) reflect.Type {
 	return t
 }
 
+func convertValueFromStruct(value interface{}, targetType reflect.Type, fieldType reflect.Type, fieldValue reflect.Value) error {
+
```
madhuravi: nit: remove blank line

config/value.go
```diff
@@ -476,13 +497,19 @@ func (cv Value) valueStruct(key string, target interface{}) (interface{}, error)
 				val = fieldInfo.DefaultValue
 			}
 			if val != nil {
-				v3, err := convertValue(val, fieldValue.Type())
-				if err != nil {
+				ret, err := convertValue(val, fieldType)
+				// For Unmarshalled text, if convertValue fails, and not returns
+				// non nil value, we fail populateStruct call by returning the error
+				if ret != nil && err != nil {
```
madhuravi: nit: might be more readable to have
if ret != nil {
  if err != nil {
     return nil, err
  } else {
     fieldValue.set
  }
} else {
 if err := convertValueFromStruct; err != nil {
   return nil, err
 }
}

config/value.go
```diff
@@ -476,13 +497,19 @@ func (cv Value) valueStruct(key string, target interface{}) (interface{}, error)
 				val = fieldInfo.DefaultValue
 			}
 			if val != nil {
-				v3, err := convertValue(val, fieldValue.Type())
-				if err != nil {
+				ret, err := convertValue(val, fieldType)
+				// For Unmarshalled text, if convertValue fails, and not returns
```
madhuravi: comment should be something like "first try converting primitive values. If nil is returned, it wasn't able to convert to primitive, so try converting as a struct value"

config/value.go
```diff
@@ -476,13 +497,19 @@ func (cv Value) valueStruct(key string, target interface{}) (interface{}, error)
...
+				// non nil value, we fail populateStruct call by returning the error
+				if ret != nil && err != nil {
 					return nil, err
+				} else if ret != nil {
+					fieldValue.Set(reflect.ValueOf(ret))
+				} else {
+					err := convertValueFromStruct(val, fieldValue.Type(), fieldType, fieldValue)
```
madhuravi: condense 508 and 509

config/value.go
```diff
@@ -476,13 +496,18 @@ func (cv Value) valueStruct(key string, target interface{}) (interface{}, error)
 				val = fieldInfo.DefaultValue
 			}
 			if val != nil {
-				v3, err := convertValue(val, fieldValue.Type())
-				if err != nil {
-					return nil, err
+				// First try to convert primitive type values, if convertValue wasn't able
+				// to convert to primitive,try converting the value asa struct value
```
madhuravi: nit: asa -> as a

PR #220, opened 2017-02-07
Add tracing docs

madhuravi:


tracing/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Sample usage
+```go
+package main
+
+import (
+  "go.uber.org/fx/tracing"
+  "go.uber.org/fx/ulog"
+
+	"github.com/uber/jaeger-client-go/config"
```
alsamylkin: nit: formatting

tracing/README.md
```diff
@@ -0,0 +1,37 @@
...
+)
+
+func main() {
+  logger := ulog.Logger()
+  statsReporter := // initialize stats reporter
+  tracer, closer, err := tracing.InitGlobalTracer(
+    &config.Configuration{},
+    "service-name",
+    ulog.Logger(),
```
alsamylkin: pass logger?

tracing/README.md
```diff
@@ -0,0 +1,37 @@
...
+  "go.uber.org/fx/tracing"
+  "go.uber.org/fx/ulog"
+
+	"github.com/uber/jaeger-client-go/config"
+)
+
+func main() {
+  logger := ulog.Logger()
+  statsReporter := // initialize stats reporter
+  tracer, closer, err := tracing.InitGlobalTracer(
```
alsamylkin: do we need tracer? it doesn't seem to be used..

tracing/README.md
```diff
@@ -0,0 +1,37 @@
...
+import (
+  "go.uber.org/fx/tracing"
+  "go.uber.org/fx/ulog"
+
+	"github.com/uber/jaeger-client-go/config"
+)
+
+func main() {
+  logger := ulog.Logger()
+  statsReporter := // initialize stats reporter
```
alsamylkin: why logger is initialized, but stastReporter not?

tracing/README.md
```diff
@@ -0,0 +1,37 @@
+# Tracing package
+
+`package tracing` provides an API wrapper around the tracing library
+[jaeger](https://github.com/uber/jaeger-client-go) that allows you to
+instrument operations in your application.
+Jaeger tracer can be configured with an optional logger that logs errors/spans
+and a stats reporter for emitting metrics.
+Note that the tracer is initialized by default with usage of UberFx modules.
```
alsamylkin: When it is initialized, on init? How to avoid a race condition with FX and make sure FX will not override it?

tracing/README.md
```diff
@@ -0,0 +1,37 @@
+# Tracing package
```
Lupie: We decided to go H1s = Title Case, H2s = Sentence case, so this should be "Tracing Package" instead.

tracing/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+`package tracing` provides an API wrapper around the tracing library
```
Lupie: Just a note that I should come back and style package names consistently later. We decided in another doc that "Package tracing provides" and "Use package tracing for" are both valid (i.e., that we don't need monospacing or lowercase).

Instead of making this change everywhere now, I'll copyedit the whole set with all the context closer to launch. Just something to think about.

tracing/README.md
```diff
@@ -0,0 +1,37 @@
...
+[jaeger](https://github.com/uber/jaeger-client-go) that allows you to
```
Lupie: - jaeger --> Jaeger
- Which thing allows this instrumentation? Is it package tracing, the API wrapper, or Jaeger specifically? The sentence should say so more clearly.

tracing/README.md
```diff
@@ -0,0 +1,37 @@
...
+instrument operations in your application.
```
Lupie: The phrase "allows you to instrument operations" seems odd to me. Would it be better to say "record" or "trace"? Or "see instrumentation data from your application"?

tracing/README.md
```diff
@@ -0,0 +1,37 @@
...
+Jaeger tracer can be configured with an optional logger that logs errors/spans
```
Lupie: Delete "tracer"

tracing/README.md
Lupie: You could also make this more straightforward for users—something like:
"Using UberFx modules initializes Jaeger tracing by default."

tracing/README.md
```diff
@@ -0,0 +1,37 @@
...
+and a stats reporter for emitting metrics.
+Note that the tracer is initialized by default with usage of UberFx modules.
+If you decide to use it standalone, read on for an example on how to initialize
```
Lupie: This "it" could mean UberFx, a module, or Jaeger. You could rephrase for clarity. Telling people to keep reading UberFx docs if they just wanna do tracing seems odd, but maybe "the tracer" means package tracing? Need to clarify.

tracing/README.md
madhuravi: Yea similar to how logging, metrics etc. is set up. It is possible to configure the uberfx service module with your own custom tracer but that's not relevant in this package doc.

tracing/README.md
madhuravi: Creating a statsReporter is not straightforward right now and even for a Nop implementation, we have our own struct. But this should be replaced with tally's soon. I didn't want to add confusion with that here.

tracing/README.md
madhuravi: Added a comment.

tracing/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+  "go.uber.org/fx/tracing"
+  "go.uber.org/fx/ulog"
+
+  "github.com/uber/jaeger-client-go/config"
+)
+
+func main() {
+  logger := ulog.Logger()
```
alsamylkin: I think it would require context parameter soon. Just in case you don't want to send a PR twice :)

PR #222, opened 2017-02-07
Refactor config providers

alsamylkin:
Mostly renaming: 
unstutter ConfigurationChangeCallback -> ChangeCallback
Yaml ->YAML

Remove lock from staticProvider: it is read only, no reason to do any locks.
Fix Env provider in case of empty prefix.
Make scopeProvider private: all other providers are private. This also removes a need a prefix check for every lookup.

### DIFF
--- a/config/provider.go
+++ b/config/provider.go
@@ -20,8 +20,8 @@
 
 package config
 
-// ConfigurationChangeCallback is called for updates of configuration data
-type ConfigurationChangeCallback func(key string, provider string, configdata interface{})
+// ChangeCallback is called for updates of configuration data
+type ChangeCallback func(key string, provider string, data interface{})
 
 // Root marks the root node in a Provider
 const Root = ""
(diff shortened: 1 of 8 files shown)


config/provider.go
```diff
@@ -37,54 +37,54 @@ type Provider interface {
...
 	// These callbacks are nop if a dynamic provider is not configured for the service.
-	RegisterChangeCallback(key string, callback ConfigurationChangeCallback) error
+	RegisterChangeCallback(key string, callback ChangeCallback) error
 	UnregisterChangeCallback(token string) error
 }
 
-// ScopedProvider defines recursive interface of providers based on the prefix
-type ScopedProvider struct {
+// scopedProvider defines recursive interface of providers based on the prefix
+type scopedProvider struct {
```
sectioneight: Any way to know if any consumers are currently using this?

config/provider.go
```diff
@@ -37,54 +37,54 @@ type Provider interface {
...
-}
-
-func addPrefix(prefix, key string) string {
+func NewScopedProvider(prefix string, provider Provider) Provider {
 	if prefix == "" {
-		return key
+		return provider
 	}
 
+	return &scopedProvider{provider, prefix}
```
sectioneight: nit: can we use named fields here?


```go
return &scopedProvider{
  Provider: provider,
  prefix: prefix,
}
```

It may seem overly verbose in this case but relying on implicit field orderings is frowned upon and can lead to subtle breakage if people move things around or add another field. Granted, the test suite should catch this, but I'd prefer to follow best practices regardless.

config/static_provider.go
```diff
@@ -55,23 +49,19 @@ func (*staticProvider) Name() string {
 }
 
 func (s *staticProvider) Get(key string) Value {
-	s.RLock()
-	defer s.RUnlock()
-
 	if key == "" {
-		// NOTE: This returns access to the underlying map, which does not guarantee
```
sectioneight: Hmm, did I write this comment? Did we ever return map access directly?

config/static_provider.go
```diff
@@ -20,13 +20,7 @@
 
 package config
 
-import (
-	"fmt"
-	"sync"
-)
-
 type staticProvider struct {
-	sync.RWMutex
```
sectioneight: This is technically correct here, but it should be noted that we **do** still have the potential for data races as we take in a map (which is essentially a reference) that the user controls, and as such they may be modifying it themselves while we access it. I feel like it's overkill at this point to copy the map, but it should be noted that we're considering upgrading StaticProvider to a full-fledged config provider, since @shawnburke would like to use it in his project.

config/provider.go
```diff
@@ -37,54 +37,54 @@ type Provider interface {
...
 	prefix string
 }
 
 // NewScopedProvider creates a child provider given a prefix
-func NewScopedProvider(prefix string, provider Provider) *ScopedProvider {
-	return &ScopedProvider{provider, prefix}
-}
-
-func addPrefix(prefix, key string) string {
+func NewScopedProvider(prefix string, provider Provider) Provider {
```
sectioneight: Nice catch on the return type

config/static_provider.go
```diff
@@ -55,23 +49,19 @@ func (*staticProvider) Name() string {
...
 		return NewValue(s, key, s.data, true, GetType(s.data), nil)
 	}
+
 	val, found := s.data[key]
 	return NewValue(s, key, val, found, GetType(val), nil)
 }
 
 func (s *staticProvider) Scope(prefix string) Provider {
-	return newScopedStaticProvider(s, prefix)
+	return NewScopedProvider(prefix, s)
```
sectioneight: Awesome refactor 💯

config/provider.go
alsamylkin: Probably no, it is very specific to us. I want to hide it before beta.

config/static_provider.go
alsamylkin: Well, we can get value directly:
https://github.com/uber-go/fx/blob/master/config/value.go#L277
So this lock doesn't do anything and we should copy map.

config/provider.go
sectioneight: Sounds good to me :)

config/static_provider.go
sectioneight: Yes, I agree the lock is useless. Just pointing out that StaticProvider wasn't "race-free" even with the lock.

alsamylkin: Added a related comment, a normal copy would require lots of reflection tricks, which we can add later.

PR #223, opened 2017-02-07
Use string to log stacktrace

alsamylkin:
Log stacktrace zap-way:
https://github.com/uber-go/zap/blob/master/field.go#L146

Unfortunately errors package doesn't provide a granular format control for the entire stack, so I opted out for simplicity to log the full trace with full path, line numbers and function names.

### DIFF
--- a/ulog/log.go
+++ b/ulog/log.go
@@ -205,17 +205,9 @@ func (l *baseLogger) fieldsConversion(keyVals ...interface{}) []zap.Field {
 			case fmt.Stringer:
 				fields = append(fields, zap.Stringer(key, value))
 			case stackTracer:
-				stack := value.StackTrace()
-				lines := make([]zap.Field, 0, len(stack))
-				for _, frame := range stack {
-					// Trick go vet to allow use of %n as a formatter
-					format := "%n"
-					function := fmt.Sprintf(format, frame)
-					source := fmt.Sprintf("%v", frame)
-					lines = append(lines, zap.String(function, source))
-				}
-
-				fields = append(fields, zap.Nest("stacktrace", lines...), zap.Error(value))
+				fields = append(fields,
+					zap.String("stacktrace", fmt.Sprintf("%+v", value.StackTrace())),
+					zap.Error(value))
 			case error:
 				fields = append(fields, zap.Error(value))
 			default:
--- a/ulog/log_test.go
+++ b/ulog/log_test.go
@@ -23,6 +23,7 @@ package ulog
 import (
 	"fmt"
 	"net"
+	"path/filepath"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


ulog/log.go
```diff
@@ -194,17 +194,9 @@ func (l *baseLogger) fieldsConversion(keyVals ...interface{}) []zap.Field {
...
-					// Trick go vet to allow use of %n as a formatter
-					format := "%n"
-					function := fmt.Sprintf(format, frame)
-					source := fmt.Sprintf("%v", frame)
-					lines = append(lines, zap.String(function, source))
-				}
-
-				fields = append(fields, zap.Nest("stacktrace", lines...), zap.Error(value))
+				fields = append(fields,
+					zap.String("stacktrace", fmt.Sprintf("%+v", value.StackTrace())),
```
sectioneight: What's the concrete difference here between `+v` and normal `v`? Or can you show an example of what the output looks like?

alsamylkin: %v   - outputs only filename and line number, 
%+v - outputs fullpath + filename + line number + function name.

ulog/log_test.go
```diff
@@ -170,9 +170,12 @@ func TestStackTraceLogger(t *testing.T) {
 		err1 := errors.New("for sure")
 		err2 := errors.Wrap(err1, "it's a trap")
 		log.Error("error message", "error", err2)
-		trace := `{"level":"error","msg":"error message","stacktrace":{"TestStackTraceLogger.func1":"log_test.go:171","WithInMemoryLogger":"in_memory_log.go:60",`
 		line := buf.Lines()[0]
-		assert.Contains(t, line, trace, "No stack trace")
+		assert.Contains(t, line, "TestStackTraceLogger.func1", "Missing first function")
```
madhuravi: Can you test them all in one assert.Contains so we are guaranteed of the order?

ulog/log.go
madhuravi: This is so much nicer!

ulog/log_test.go
alsamylkin: Not really, they have a fullpath, but I can compare positions

madhuravi: We should verify full path too since that's our intent.

alsamylkin: I don't expect Jenkins to have the same GOPATH as my laptop :)

madhuravi: path can be relative ;) doesn't have to be absolute.

alsamylkin: Now I have to think about cross platform compatibility with windows!

ulog/log_test.go
```diff
@@ -170,9 +171,13 @@ func TestStackTraceLogger(t *testing.T) {
...
+		assert.Contains(t, line, "log_test.go:172", "Missing source")
```
glibsm: lets remove the lines from here so we don't make a brittle test again?

glibsm: better yet, bonus points for using `log_test.go:\d+` style regex

PR #226, opened 2017-02-08
Remove custom context and introduce ctx aware logger

anuptalwalkar:
Changes include
1. No more fx.Context
2. Context aware logger is accessible as fx.Logger(ctx)
3. Removed unnecessary WrapUnary, WrapOneway methods
4. Logger setup in context via contextFilter for uhttp and middleware for rpc

Not included in the PR- 
1. Doc changes
2. Make SetContextStore API private?


sectioneight:
FYI @prashantv @abhinav @breerly we had a long discussion and decided on stdlib 🎉

context.go
```diff
@@ -21,25 +21,61 @@
...
 
-type nopContext struct {
-	gcontext.Context
+// WithContextAwareLogger returns a new context with a context-aware logger
+func WithContextAwareLogger(ctx context.Context, span opentracing.Span) context.Context {
+	store := contextStore(ctx)
+	// Note that opentracing.Tracer does not expose the tracer id
+	// We only inject tracing information for jaeger.Tracer
+	if jSpanCtx, ok := span.Context().(jaeger.SpanContext); ok {
+		traceLogger := Logger(ctx).With(
```
alsamylkin: nit: store.log = is probably shorter :)

context.go
```diff
@@ -21,25 +21,61 @@
...
 
-// Context embeds Host and Go stdlib context for use
-type Context interface {
-	gcontext.Context
+type contextKey int
 
-	Logger() ulog.Log
+const _contextStore contextKey = iota
+
+type ctxStore struct {
```
alsamylkin: Why do you prefer to have a struct over storing e.g. ulog.Log directly in the context?
Each module can provide functions to work with context, which gives nice separation..

The only bad part I see is performance: access to elements in context seems to be linear:
https://golang.org/src/context/context.go?s=7222:7247#L469
But it shouldn't be a big deal, we'll have only few elements.

modules/uhttp/client/client_filters.go
```diff
@@ -69,9 +69,8 @@ func tracingFilter() FilterFunc {
 		ext.HTTPUrl.Set(span, req.URL.String())
 		defer span.Finish()
 
-		ctx = &fxcontext.Context{
-			Context: opentracing.ContextWithSpan(ctx, span),
-		}
+		ctx = opentracing.ContextWithSpan(ctx, span)
```
alsamylkin: Our code suddenly looks so much nicer!:)

context.go
alsamylkin: It seems 
```
ulog.LogError(ctx, "message")
```
looks better than 
```
Logger(ctx).LogError("message")
```

anuptalwalkar: context.Value() is a `o(n)` operation, with `n` being number of elements added `WithValue`. Keeping it in struct will save us future optimization if/when we will add more per request data like auth context. rpc team had good arguments in https://github.com/uber-go/fx/issues/126 and PR to keep the optimization in place, and I don't see any point not to.

anuptalwalkar: I would like to punt the discussion on changing the method signature on logger since we won't be having Log inerface in the future with zap 1.0 :).

context.go
```diff
@@ -21,25 +21,61 @@
...
 	"go.uber.org/fx/ulog"
 )
 
-// Context embeds Host and Go stdlib context for use
-type Context interface {
-	gcontext.Context
+type contextKey int
 
-	Logger() ulog.Log
+const _contextStore contextKey = iota
```
madhuravi: nit: The name contextStore is very generic. I would prefer contextHost or something along those lines that indicate this has logging, auth etc.

context.go
alsamylkin: When n = 1, it doesn't matter :)

I am still not convinced about the struct, can we used Host directly here?

examples/keyvalue/gen.go
```diff
@@ -20,4 +20,4 @@
 
 package keyvalue
 
-//go:generate thriftrw "--plugin=yarpc --context-import-path=go.uber.org/fx --unary-handler-wrapper=go.uber.org/fx/modules/rpc.WrapUnary --oneway-handler-wrapper=go.uber.org/fx/modules/rpc.WrapOneway" kv.thrift
+//go:generate thriftrw "--plugin=yarpc" kv.thrift
```
glibsm: 🎉

context.go
```diff
@@ -21,25 +21,61 @@
...
 
-type nopContext struct {
-	gcontext.Context
+// WithContextAwareLogger returns a new context with a context-aware logger
+func WithContextAwareLogger(ctx context.Context, span opentracing.Span) context.Context {
+	store := contextStore(ctx)
+	// Note that opentracing.Tracer does not expose the tracer id
+	// We only inject tracing information for jaeger.Tracer
+	if jSpanCtx, ok := span.Context().(jaeger.SpanContext); ok {
+		traceLogger := store.log.With(
```
alsamylkin: I meant to replace traceLogger too :)

context_test.go
```diff
@@ -33,24 +33,28 @@ import (
...
-	ctx := New(gctx, service.NopHost())
-	assert.Equal(t, "val", ctx.Value("key"))
-
-	gctx1 := gcontext.WithValue(gcontext.Background(), "key1", "val1")
-	ctx = &Context{
-		Context: gctx1,
-	}
-	assert.Equal(t, nil, ctx.Value("key"))
-	assert.Equal(t, "val1", ctx.Value("key1"))
+func TestContext_WithLogger(t *testing.T) {
```
glibsm: what is this testing, exactly?

context.go
anuptalwalkar: oh you are right. Actually this makes more sense here -
```
		store.log = Logger(ctx).With(
			"traceID", jSpanCtx.TraceID(), "spanID", jSpanCtx.SpanID(),
		)
```

context.go
anuptalwalkar: changing *store to *host

context_test.go
anuptalwalkar: hmm. I just ported the internal tests out. This was testing if we get the keys out of context. I'll remove it.

context.go
anuptalwalkar: We had Host in the context earlier, but the object was too big and we decided not to keep it as a whole in the context.

context.go
```diff
@@ -21,25 +21,60 @@
...
+		c = ctxHost{}
+		ctx = context.WithValue(ctx, _contextHost, c)
+	}
+	return c.(ctxHost)
 }
 
-// NopContext is defined for testing purpose
-var NopContext = nopContext{gcontext.Background()}
+// SetContextHost sets the context with context aware logger
+func SetContextHost(ctx context.Context, host service.Host) context.Context {
```
dmcaulay: You could name this `NewContext`.

https://golang.org/pkg/context/#Context

It's slightly different than the example, because you don't expose the struct you're storing in the context, but I thought I'd point out the standard name.

context.go
```diff
@@ -21,25 +21,60 @@
...
+}
 
-type nopContext struct {
-	gcontext.Context
+// WithContextAwareLogger returns a new context with a context-aware logger
+func WithContextAwareLogger(ctx context.Context, span opentracing.Span) context.Context {
+	store := contextHost(ctx)
+	// Note that opentracing.Tracer does not expose the tracer id
+	// We only inject tracing information for jaeger.Tracer
+	if jSpanCtx, ok := span.Context().(jaeger.SpanContext); ok {
```
dmcaulay: can you use `opentracing.SpanFromContext(ctx)` instead of passing the span into this method?

context.go
```diff
@@ -21,25 +21,60 @@
...
 }
 
-func (nopContext) Logger() ulog.Log {
-	return ulog.NopLogger
+// Logger returns a context aware logger. If logger is absent from the context store,
+// the function updates the context with a new context based logger
+func Logger(ctx context.Context) ulog.Log {
+	store := contextHost(ctx)
+	if store.log == nil {
+		store.log = ulog.Logger()
```
dmcaulay: you could return here, since this assignment doesn't change the underlying store in the context.

context.go
anuptalwalkar: Thanks Dan, I already merged, but I will follow up the comments in incoming PR.

context.go
anuptalwalkar: The span is passed from `tracingServerFilter` which is extracted from headers. Moving the headers here won't be appropriate.

examples/keyvalue/gen.go
breerly: Nice, this should simplify usage a bit.

PR #227, opened 2017-02-08
Upgrade tally to 1.1.0 to get memory improvements

madhuravi:


glide.yaml
```diff
@@ -4,7 +4,7 @@ import:
   # TODO(ai): Pin to semver range after 1.0 is released
   version: master
 - package: github.com/uber-go/tally
-  version: ^1
+  version: ^1.1.0
```
alsamylkin: Why do we care about version if there is no code change?

madhuravi: Existing tally version is causing memory leaks in a service that's using fx. The new version has dedups and memory fixes.

PR #228, opened 2017-02-08
Update docs to cleanup fx.Context traces

anuptalwalkar:
1. Updated docs and readme with context.Context
2. Fixed follow up from #226

### DIFF
--- a/context.go
+++ b/context.go
@@ -39,12 +39,14 @@ type ctxHost struct {
 
 func contextHost(ctx context.Context) ctxHost {
 	c := ctx.Value(_contextHost)
-	if c == nil {
-		c = ctxHost{
-			log: ulog.Logger(),
-		}
-		ctx = context.WithValue(ctx, _contextHost, c)
+	if c != nil {
+		return c.(ctxHost)
 	}
+
+	c = ctxHost{
+		log: ulog.Logger(),
+	}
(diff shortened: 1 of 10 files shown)


anuptalwalkar:
@dmcaulay

context.go
```diff
@@ -40,14 +40,16 @@ type ctxHost struct {
 func contextHost(ctx context.Context) ctxHost {
 	c := ctx.Value(_contextHost)
 	if c == nil {
-		c = ctxHost{}
+		c = ctxHost{
```
sectioneight: nit: prefer "early returns"

```go
if c != nil {
  return c.(ctxHost)
}

c = ctxHost{
...
```

modules/rpc/thriftrw-plugin-thriftsync/README.md
```diff
@@ -10,146 +10,150 @@ Following examples show how thriftsync syncs handler code with the updated thrif
...
 package main
 
 import (
-	"go.uber.org/fx"
-	"testservice/testservice/testserviceserver"
-	"go.uber.org/fx/service"
-	"go.uber.org/yarpc/api/transport"
+  "context"
+
+  "testservice/testservice/testserviceserver"
```
sectioneight: style: this is a "project" import so should be separate from 3rd-party (go.uber.org) imports

PR #229, opened 2017-02-08
Added useful metrics for fx modules

anuptalwalkar:
Moved metrics to static initialization to save on allocs


modules/rpc/handler.go
```diff
@@ -34,6 +35,8 @@ type contextInboundMiddleware struct {
 }
 
 func (f contextInboundMiddleware) Handle(ctx context.Context, req *transport.Request, resw transport.ResponseWriter, handler transport.UnaryHandler) error {
+	stopwatch := stats.RPCHandleTimer.Timer("rpc.handler." + req.Procedure + ".time").Start()
```
alsamylkin: Prometheus backend is using "_" as separators..
https://prometheus.io/docs/practices/naming/

modules/stats/metrics.go
```diff
@@ -0,0 +1,115 @@
...
+	// TaskExecutionTime is a turnaround time for execution
+	TaskExecutionTime tally.Timer
+	// TaskPublishTime is a publish time for tasks
+	TaskPublishTime tally.Timer
+)
+
+// SetupHTTPMetrics allocates counters for necessary setup
+func SetupHTTPMetrics(host service.Host) {
+	HTTPPanicCounter = host.Metrics().Tagged(HTTPTags).Counter("panic")
+	HTTPAuthFailCounter = host.Metrics().Tagged(HTTPTags).Counter("http.auth.fail")
```
alsamylkin: Same as above with separators..

alsamylkin: Also nit, shouldn't we tag once, not 3 times?

modules/stats/metrics.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// SetupRPCMetrics allocates counters for necessary setup
+func SetupRPCMetrics(host service.Host) {
+	RPCAuthFailCounter = host.Metrics().Tagged(RPCTags).Counter("rpc.auth.fail")
+	RPCHandleTimer = host.Metrics().Tagged(RPCTags)
+}
+
+// SetupTaskMetrics allocates counters for necessary setup
+func SetupTaskMetrics(host service.Host) {
```
alsamylkin: Why do we setup metrics for all modules in the metrics module? If someone wants do delete or update metrics they need to make change in their own local module, not in 2 places..

modules/stats/metrics.go
```diff
@@ -0,0 +1,115 @@
...
+	TaskExecutionCount = host.Metrics().Tagged(TaskTags).Counter("task.execution.count")
```
alsamylkin: Same as above, lots of Tagged(TaskTags) repetition..

modules/task/execution.go
```diff
@@ -81,20 +84,26 @@ func Enqueue(fn interface{}, args ...interface{}) error {
 	fnName := getFunctionName(fn)
 	_, ok := fnLookup.getFn(fnName)
 	if !ok {
+		stats.TaskPublishFail.Inc(1)
```
alsamylkin: I don't see tests for that..

modules/stats/metrics.go
```diff
@@ -0,0 +1,115 @@
...
+const (
+	//TagModule is module tag for metrics
+	TagModule = "module"
+	// TagType is either request or response
+	TagType = "type"
+	// TagStatus is status of the request
+	TagStatus = "status"
+)
+
+var httpMethods = []string{"GET", "HEAD", "PUT", "POST", "DELETE", "CONNECT", "OPTIONS", "TRACE", "PATCH"}
```
alsamylkin: Please, use standard constants: https://golang.org/pkg/net/http/#pkg-constants

modules/stats/metrics.go
```diff
@@ -0,0 +1,115 @@
...
+
+// HTTPTags creates metrics scope with defined tags
+var (
+	HTTPTags = map[string]string{
+		TagModule: "http",
+		TagType:   "request",
```
alsamylkin: No response tags??

modules/uhttp/handler.go
```diff
@@ -58,7 +59,7 @@ type handlerWrapper struct {
 // ServeHTTP calls Handler.ServeHTTP(ctx, w, r) and injects a new service context for use.
 func (h *handlerWrapper) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	ctx := fx.NewContext(context.Background(), h.host)
-	stopwatch := h.host.Metrics().Timer("http." + r.Method + ".time").Start()
+	stopwatch := stats.HTTPMethodTimer[r.Method].Start()
```
alsamylkin: I am really curious, how thread safe is it..
Theoretically it should be fine, but I'd prefer to test it.

modules/rpc/handler.go
anuptalwalkar: I am following muttleys naming schemes. Having "rpc.handler.procedure" should be fine for query. We are not using prometheus right now. We should fix how we define scope however, in other news...

modules/stats/metrics.go
anuptalwalkar: We don't have response level metrics, yet.

modules/stats/metrics.go
```diff
@@ -0,0 +1,115 @@
...
+	HTTPMethodTimer = make(map[string]tally.Timer)
+	for _, method := range httpMethods {
+		HTTPMethodTimer[method] = host.Metrics().Tagged(HTTPTags).Timer("http." + method + ".time")
+	}
+	HTTPStatusCountScope = host.Metrics().Tagged(HTTPTags)
+}
+
+// SetupRPCMetrics allocates counters for necessary setup
+func SetupRPCMetrics(host service.Host) {
+	RPCAuthFailCounter = host.Metrics().Tagged(RPCTags).Counter("rpc.auth.fail")
```
dmcaulay: A couple thoughts:

1. You could also add a tag for auth. In our services we use `middleware: auth` so we can pick up all the middleware stats for a handler in one query.

2. Why do we have the full path in the name? Is that for metric clients that don't support tags?

sectioneight:
@anuptalwalkar can you sync with @breerly and co to make sure we're not emitting overlapping metrics? I know YARPC takes care of some stuff, we should validate that we're not double-measuring things.

modules/rpc/handler.go
dmcaulay: should we use tags for the procedure?

modules/stats/metrics.go
alsamylkin: I am a bit confused, should server record a request latency or response?

anuptalwalkar: Server should only record request latency and turnaround time. "response" metrics are usually defined on client side. for example rtt for client calling a service and getting response back comes under response time.

modules/task/execution.go
anuptalwalkar: added

modules/uhttp/handler.go
anuptalwalkar: Not done yet. Need to add test for this.

modules/stats/metrics.go
anuptalwalkar: Added the tag. Also yes, and it also provides more readability without affecting perf.

modules/uhttp/handler.go
anuptalwalkar: Actually, you are right. Im removing this since underlying library timers can be not thread safe.

modules/rpc/handler.go
```diff
@@ -34,6 +35,11 @@ type contextInboundMiddleware struct {
 }
 
 func (f contextInboundMiddleware) Handle(ctx context.Context, req *transport.Request, resw transport.ResponseWriter, handler transport.UnaryHandler) error {
+	stopwatch := stats.RPCHandleTimer.
+		Tagged(map[string]string{stats.TagProcedure: req.Procedure}).
+		Timer("rpc.handler." + req.Procedure + ".time").
```
glibsm: i think `.time` is extraneous, no? we already know it's a timer, but I'm a bit behind on the M3

modules/rpc/handler.go
```diff
@@ -34,6 +35,11 @@ type contextInboundMiddleware struct {
...
+		Start()
+	defer stopwatch.Stop()
 	ctx = fx.NewContext(ctx, f.Host)
```
glibsm: newline after defer

modules/task/execution_test.go
```diff
@@ -135,6 +142,7 @@ func TestEnqueueSimpleFn(t *testing.T) {
 	err = <-_errorCh
 	require.Error(t, err)
 	assert.Contains(t, err.Error(), "Simple error")
+	verifyMetrics(t, _testScope)
```
glibsm: pulling this out into a function is not totally helpful, since it's a one-and-done function that can't be easily reused

anuptalwalkar: +1 moved back in the test

modules/rpc/stats/metrics.go
```diff
@@ -18,19 +18,39 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-package modules
+package stats
+
+import "github.com/uber-go/tally"
 
 const (
 	//TagModule is module tag for metrics
```
alsamylkin: nit formatting

modules/rpc/handler.go
```diff
@@ -34,6 +35,12 @@ type contextInboundMiddleware struct {
 }
 
 func (f contextInboundMiddleware) Handle(ctx context.Context, req *transport.Request, resw transport.ResponseWriter, handler transport.UnaryHandler) error {
+	stopwatch := stats.RPCHandleTimer.
+		Tagged(map[string]string{stats.TagProcedure: req.Procedure}).
+		Timer("rpc.handler." + req.Procedure).
```
alsamylkin: It looks like all these things should be in tags.

modules/rpc/stats/metrics.go
```diff
@@ -18,19 +18,39 @@
...
+	// RPCAuthFailCounter counts auth failures
+	RPCAuthFailCounter tally.Counter
+	// RPCHandleTimer is a turnaround time for rpc handler
+	RPCHandleTimer tally.Scope
+)
+
+// SetupRPCMetrics allocates counters for necessary setup
+func SetupRPCMetrics(scope tally.Scope) {
+	rpcTagsScope := scope.Tagged(RPCTags)
+	RPCAuthFailCounter = rpcTagsScope.Tagged(map[string]string{TagMiddleware: "auth"}).Counter("rpc.auth.fail")
```
alsamylkin: The counter already has tags rpc and auth, no need to include it in the name

modules/task/execution_test.go
```diff
@@ -135,6 +142,17 @@ func TestEnqueueSimpleFn(t *testing.T) {
...
 	require.Error(t, err)
 	assert.Contains(t, err.Error(), "Simple error")
+
+	snapshot := _testScope.(tally.TestScope).Snapshot()
+	timers := snapshot.Timers()
+	counters := snapshot.Counters()
+
+	assert.NotNil(t, timers["task.publish.time"].Values())
+	assert.NotNil(t, counters["task.publish.count"].Value())
+	assert.NotNil(t, counters["task.publish.fail"].Value())
```
alsamylkin: I think we should check that is is not 0..

modules/rpc/stats/metrics.go
```diff
@@ -18,19 +18,39 @@
...
 )
 
 // HTTPTags creates metrics scope with defined tags
-var HTTPTags = map[string]string{
-	TagModule: "http",
-	TagType:   "request",
+var (
+
+	// RPCTags creates metrics scope with defined tags
+	RPCTags = map[string]string{
```
alsamylkin: Why all this variables are public, did you want to use an internal package?

modules/rpc/stats/metrics.go
dmcaulay: you lose the group/alias/apply by node functions when you use m3ql. if you have the tags you can aggregate by tags instead.

modules/rpc/stats/metrics.go
anuptalwalkar: oh makes sense. I probably missed this comment.

blampe:
There's definitely some overlap here. In particular we're going to need to turn these timers into histograms, cc @akshayjshah

modules/task/execution_test.go
```diff
@@ -29,15 +29,20 @@ import (
...
+	_testScope tally.Scope
+	_errorCh   <-chan error
+	_ctx       = context.Background()
 )
 
 func init() {
-	_globalBackend = NewInMemBackend()
+	host := service.NopHost()
+	_testScope = host.Metrics()
+	_globalBackend = NewInMemBackend(host)
```
akshayjshah: The *tests* need global state? I don't see you resetting state, so it looks like you can't make substantive assertions about these metrics. Again, these would be much less awkward if they weren't globals.

modules/task/internal/stats/metrics.go
```diff
@@ -0,0 +1,63 @@
...
+	// TaskPublishCount counts number of tasks enqueued
+	TaskPublishCount tally.Counter
+	// TaskExecuteFail counts number of tasks failed to execute
+	TaskExecuteFail tally.Counter
+	// TaskPublishFail counts number of tasks failed to enqueue
+	TaskPublishFail tally.Counter
+	// TaskExecutionTime is a turnaround time for execution
+	TaskExecutionTime tally.Timer
+	// TaskPublishTime is a publish time for tasks
+	TaskPublishTime tally.Timer
```
akshayjshah: Missing the always-crucial time spent in queue, which will require you to define an envelope for your tasks.

If you haven't read something similar already, consider reading [this](https://www.vividcortex.com/resources/queueing-theory) for a quick-and-dirty introduction to queueing theory.

modules/task/execution.go
```diff
@@ -130,13 +137,17 @@ func Register(fn interface{}) error {
...
 // Run decodes the message and executes as a task
 func Run(ctx context.Context, message []byte) error {
+	stopwatch := stats.TaskExecutionTime.Start()
+	defer stopwatch.Stop()
+
 	var s fnSignature
 	if err := GlobalBackend().Encoder().Unmarshal(message, &s); err != nil {
 		return errors.Wrap(err, "unable to decode the message")
 	}
 	// TODO (madhu): Do we need a timeout here?
```
akshayjshah: The context already carries a deadline - why not use that?

modules/task/execution_test.go
```diff
@@ -29,15 +29,20 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"github.com/uber-go/tally"
+	"go.uber.org/fx/service"
```
akshayjshah: nit: import grouping.

modules/uhttp/filterchain_builder.go
```diff
@@ -53,15 +52,13 @@ type filterChainBuilder struct {
...
-		panicFilter{scope.Counter("panic")},
-		metricsFilter{scope},
-		tracingServerFilter{scope},
+		panicFilter{},
+		metricsFilter{},
+		tracingServerFilter{},
 		authorizationFilter{
-			authClient:  host.AuthClient(),
-			authCounter: scope.Counter("auth.fail"),
+			authClient: host.AuthClient(),
```
akshayjshah: This seems like a step backwards. Why replace local state with global state?

modules/uhttp/filters.go
```diff
@@ -83,13 +81,12 @@ func (f tracingServerFilter) Apply(ctx context.Context, w http.ResponseWriter, r
...
 type authorizationFilter struct {
-	authClient  auth.Client
-	authCounter tally.Counter
+	authClient auth.Client
 }
 
 func (f authorizationFilter) Apply(ctx context.Context, w http.ResponseWriter, r *http.Request, next Handler) {
 	if err := f.authClient.Authorize(ctx); err != nil {
-		f.authCounter.Inc(1)
+		stats.HTTPAuthFailCounter.Inc(1)
```
akshayjshah: It'd be nice if the error returned from any authn/authz checks included tags for this counter: who's the caller, what scope of permissions did they fail to get, etc.

modules/uhttp/handler.go
```diff
@@ -58,7 +59,8 @@ type handlerWrapper struct {
 // ServeHTTP calls Handler.ServeHTTP(ctx, w, r) and injects a new service context for use.
 func (h *handlerWrapper) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	ctx := fx.NewContext(context.Background(), h.host)
-	stopwatch := h.host.Metrics().Timer("http." + r.Method + ".time").Start()
+	stopwatch := stats.HTTPMethodTimer.Timer(r.Method).Start()
```
akshayjshah: Consider whether you really want "getUser" to be the name of this metric. Would "latency" be a better name, with method as a dimension?

PR #230, opened 2017-02-09
Enforce first argument in async task function to be context.Context

madhuravi:


modules/task/README.md
```diff
@@ -37,13 +39,13 @@ func newBackend(host service.Host) (task.Backend, error) {
   return b, nil
 }
 
-func runActivity(input string) error {
+func runActivity(ctx context.Context, input string) error {
   // do things and calculate results
   results := "results"
-  return task.Enqueue(updateCache, input, results)
+  return task.Enqueue(updateCache, ctx, input, results)
```
sectioneight: open question: should `ctx` be first as per convention?

modules/task/execution.go
```diff
@@ -88,9 +89,14 @@ func Enqueue(fn interface{}, args ...interface{}) error {
 	if err := validateFnAgainstArgs(fnType, args); err != nil {
 		return err
 	}
+	// Register function types for encoding
+	for _, arg := range args {
```
alsamylkin: Do we need registration for decoder?

madhuravi:
NOTE: I am working on testing decode when register happens on a different host. Will report back once I do.

modules/task/README.md
madhuravi: I prefer second here because I want the enqueue arguments to map to the actual function. It's more intuitive to see
task.Enqueue(fn, arg1 arg2)
with
func fn(arg1 context.Context, arg2 string)

modules/task/execution.go
madhuravi: We do. See updated change.

sectioneight:
Ya, i think second feels natural here just wanted to have the discussion
since the de facto standard is that it's always first.

modules/task/task.go
```diff
@@ -75,6 +79,10 @@ func newAsyncModule(
 	_globalBackendMu.Lock()
 	_globalBackend = backend
 	_globalBackendMu.Unlock()
+	// Register context implementation with the encoder so it can be used in async functions
+	if err := _globalBackend.Encoder().Register(context.Background()); err != nil {
```
alsamylkin: Context.Background is of type emptyCtx. I've counted at least 3 more context types in the source code: https://golang.org/src/context/context.go?s=11899:11974#L311

I think it is really bad idea to depend on implementation on standard packages..

modules/task/execution.go
```diff
@@ -111,11 +111,15 @@ func Register(fn interface{}) error {
...
 	}
-	// Register function types for encoding
 	for i := 0; i < fnType.NumIn(); i++ {
-		arg := reflect.Zero(fnType.In(i)).Interface()
-		if err := GlobalBackend().Encoder().Register(arg); err != nil {
-			return errors.Wrap(err, "unable to register the message for encoding")
+		argType := fnType.In(i)
+		// Interfaces cannot be registered, their implementations should be
+		// https://golang.org/pkg/encoding/gob/#Register
+		if argType.Kind() != reflect.Interface {
```
madhuravi: First argument is context. This is only calling gob register if the type is not an interface. It does not do anything beyond that.

modules/task/execution.go
```diff
@@ -159,6 +163,15 @@ func validateFnFormat(fnType reflect.Type) error {
...
 		return fmt.Errorf("expected a func as input but was %s", fnType.Kind())
 	}
+	if fnType.NumIn() < 1 {
+		return fmt.Errorf(
+			"expected atleast one argument of type context.Context in function, found %d input arguments",
+			fnType.NumIn(),
+		)
+	}
+	if !isContext(fnType.In(0)) {
+		return fmt.Errorf("expected first argument to be context.Context but found %s", fnType.In(0))
```
madhuravi: this already prints the type as expected.

modules/task/task.go
madhuravi: agreed, I don't like this either. I think the right solution would be something like extracting the values from context and passing it through backend and constructing it back into context on the receiving side. Planning to discuss this with them today.

modules/task/backend.go
```diff
@@ -114,22 +113,13 @@ func (b *inMemBackend) Start(ready chan<- struct{}) <-chan error {
...
-			if ok {
-				errorCh <- Run(msg)
-			} else {
-				return
-			}
-		case <-time.After(time.Millisecond):
-			ulog.Logger().Error("Timed out after 1 ms")
-		}
+	for msg := range b.bufQueue {
+		errorCh <- Run(context.Background(), msg)
```
alsamylkin: This is ok for now, but we need to have a common serialization between backends that don't support context propagation.

PR #231, opened 2017-02-09
Add HTTP transport to YARPC module

alsamylkin:
We can have only one tchannel inbound right now.
This PR:
1. Adds ability to have multiple tchannel and http inbounds.
2. Customizes dispatcher start, e.g. users want to attach some metrics on start
3. Few minor refactors along the way


alsamylkin:
@willhug @breerly just in case I am courting YARPC with bad manners

modules/rpc/thrift_test.go
```diff
@@ -27,18 +27,56 @@ import (
...
-	mci := mch()
-	goofy, err := chip(mch())
+	type Modules struct {
+		RPC yarpcConfig
+	}
+	type Global struct {
+		Modules Modules
+	}
+
+	buf, err := yaml.Marshal(
```
sectioneight: can we separate out this yaml stuff into a helper to make the test clearer

modules/rpc/thrift_test.go
```diff
@@ -27,18 +27,56 @@ import (
...
+	return h.config
+}
+
 func TestThriftModule_OK(t *testing.T) {
 	chip := ThriftModule(okCreate, modules.WithRoles("rescue"))
 	dale := ThriftModule(okCreate, modules.WithRoles("ranges"))
 
-	mci := mch()
-	goofy, err := chip(mch())
+	type Modules struct {
```
sectioneight: Nit: prefer a type name that makes it clear it's a test. To avoid shadowing other Modules in this package (I think?)

modules/rpc/thrift_test.go
```diff
@@ -27,18 +27,56 @@ import (
...
+					Inbounds: []Inbound{
+						{TChannel: &Port{0}, HTTP: &Port{0}},
+					},
+				},
+			}})
+
+	require.NoError(t, err)
+
+	mci := service.ModuleCreateInfo{
+		Items: make(map[string]interface{}),
```
sectioneight: why can't we pass nil here? this seems like a wart for test ability, then again the ctor has some business logic we need? I really believe in th Go-ism that the zero value of the state should be valid

modules/rpc/yarpc.go
```diff
@@ -32,6 +32,10 @@ import (
 	"go.uber.org/yarpc"
 	"go.uber.org/yarpc/api/middleware"
 	"go.uber.org/yarpc/api/transport"
+
+	"fmt"
```
sectioneight: wrong place

modules/rpc/yarpc.go
```diff
@@ -48,8 +52,9 @@ type YARPCModule struct {
 }
 
 var (
-	_dispatcherFn = defaultYARPCDispatcher
 	_dispatcherMu sync.Mutex
+	_dispatcherFn = defaultYARPCDispatcher
```
sectioneight: maybe we should start commenting these? we've got a lot of globals to keep track of :(

modules/rpc/yarpc.go
```diff
@@ -62,13 +67,30 @@ var (
...
+
+// Inbound is a union that configures how to configure a single inbound.
+type Inbound struct {
+	TChannel *Port
+	HTTP     *Port
+}
+
+// Port is a struct that have a required port for tchannel/http transports.
+// TODO(alsam) make it optional
+type Port struct {
```
sectioneight: why not just 'type Port int' which you could also attach methods to

modules/rpc/thrift_test.go
alsamylkin: This is actually a trick I have to do because we don't have PopulateSlice/Static provider is not recursive..

modules/rpc/thrift_test.go
alsamylkin: Yeah, we actually don't need it, thanks!

modules/rpc/yarpc.go
alsamylkin: It is not a type that I am after, but structure: to show port for both tchannel and http:
```
Inbounds:
- http
   8080
- tchannel
   8181
```

modules/rpc/thrift_test.go
alsamylkin: Probably, byte array is more cleaner..

modules/rpc/yarpc.go
```diff
@@ -62,13 +70,30 @@ var (
...
+}
+
+// Inbound is a union that configures how to configure a single inbound.
+type Inbound struct {
+	TChannel *Address
+	HTTP     *Address
+}
+
+// Address is a struct that have a required port for tchannel/http transports.
+// TODO(alsam) make it optional
```
sectioneight: Optional how? It's already a pointer for `Inbound` so it can be ommited, right?

modules/rpc/yarpc.go
```diff
@@ -264,9 +294,7 @@ func (m *YARPCModule) Start(readyCh chan<- struct{}) <-chan error {
 	}
 
 	m.register(m)
-	m.log.Info("Module started",
-		"name", m.config.AdvertiseName,
-		"port", m.config.Bind)
+	m.log.Info("Module started")
```
sectioneight: Why remove the metadata?

modules/rpc/yarpc.go
```diff
@@ -306,6 +334,19 @@ func defaultYARPCDispatcher(_ service.Host, cfg yarpc.Config) (*yarpc.Dispatcher
...
+type yarpcStarterFn func(dispatcher *yarpc.Dispatcher) error
+
+// RegisterStarter allows you to override the YARPC dispatcher start, e.g. attach some metrics with start.
+func RegisterStarter(starterFN yarpcStarterFn) {
+	_dispatcherMu.Lock()
+	defer _dispatcherMu.Unlock()
+	_starterFn = starterFN
+}
+
+func defaultYARPCStarter(dispatcher *yarpc.Dispatcher) error {
```
sectioneight: This looks like it's only used in tests. Can we move it to a `_test.go` file so it's clear it's not used in production code?

modules/rpc/yarpc.go
```diff
@@ -32,6 +33,8 @@ import (
 	"go.uber.org/yarpc"
 	"go.uber.org/yarpc/api/middleware"
 	"go.uber.org/yarpc/api/transport"
+
```
glibsm: why the empty line?

modules/rpc/yarpc_test.go
```diff
@@ -36,30 +35,36 @@ func TestRegisterDispatcher_OK(t *testing.T) {
 	RegisterDispatcher(defaultYARPCDispatcher)
 }
 
+func TestRegisterStarter_OK(t *testing.T) {
+	t.Parallel()
+	RegisterStarter(defaultYARPCStarter)
```
glibsm: interesting test

modules/rpc/yarpc_test.go
```diff
@@ -36,30 +35,36 @@ func TestRegisterDispatcher_OK(t *testing.T) {
...
-		AdvertiseName: "fx",
-		inbounds:      []transport.Inbound{http.NewTransport().NewInbound("")},
+		transports: transports{
+			inbounds: []transport.Inbound{http.NewTransport().NewInbound("")},
+		},
 	}
 
-	require.NoError(t, c.addConfig(cfg))
+	c.addConfig(cfg)
 	assert.Error(t, c.Start(service.NopHost()))
```
glibsm: can we validate error why, i.e. check part of the error message?

modules/rpc/yarpc_test.go
alsamylkin: it is actually effective once you'll start checking for races.
I am very sad we don't use t.Parallel everywhere.

modules/rpc/yarpc.go
alsamylkin: E.g. we want to have a transport on a default port.

modules/rpc/yarpc.go
alsamylkin: It is a _starterFn and used by default in production

modules/rpc/yarpc.go
alsamylkin: Don't have an access to it on start() anymore and log it on module creation.

modules/rpc/yarpc_test.go
glibsm: That's not really my point.

My point is that if there is another parallel test with a custom starter, you have just created a potential for a flappy test, depending on which one starts first, right?

modules/rpc/thrift_test.go
```diff
@@ -27,18 +27,43 @@ import (
...
+}
+
 func TestThriftModule_OK(t *testing.T) {
 	chip := ThriftModule(okCreate, modules.WithRoles("rescue"))
 	dale := ThriftModule(okCreate, modules.WithRoles("ranges"))
+	cfg := []byte(`modules:
+  rpc:
+    inbounds:
+     - tchannel:
+       port: 0
```
breerly: indent 2

modules/rpc/thrift_test.go
```diff
@@ -27,18 +27,43 @@ import (
...
+     - http:
+       port: 0
```
breerly: indent 2

modules/rpc/yarpc_test.go
alsamylkin: You are right about this possibility and that's why I have only a defaultStarter.

modules/rpc/thrift_test.go
alsamylkin: * It should be similar to inbounds, I am still wiring it together lots of different types, decided to make this first and add incrementally outbounds.
* Yes, right now there is no config format for the YARPC, it accepts a struct, that's why we need something very similar.

modules/rpc/yarpc_test.go
alsamylkin: True way is to compare error messages, but here we expect a system error, can't open a default port.

Changed port to be -1, when 80 is open, all I want is to trigger transports error.

breerly:
I'm writing a doc with @kriskowal to propose a plan on sharing the config object, let's try to build some consensus in there. Will get @alsamylkin's eyes on the rough draft tomorrow.

modules/rpc/thrift_test.go
```diff
@@ -27,18 +27,43 @@ import (
...
 	"github.com/stretchr/testify/require"
 
+	"go.uber.org/fx/config"
 	"go.uber.org/fx/modules"
 	"go.uber.org/fx/service"
 	"go.uber.org/yarpc/api/transport"
 	"go.uber.org/yarpc/encoding/thrift"
 )
 
+type testHost struct {
```
madhuravi: instead add a NopHostWithConfig or something similar in service/host_mock.go. This will be useful for other tests too.

alsamylkin: I don't like to share :) Plus making it public assumes some responsibility, which I am trying to avoid.

alsamylkin: I'll made it public if there will be 3 cases where similar pattern is used.

madhuravi: that's fair. just that it leads to people now knowing that something like this exists elsewhere in code and end up duplicating it :)

modules/rpc/yarpc.go
```diff
@@ -177,13 +184,13 @@ func (c *dispatcherController) mergeConfigs() (conf yarpc.Config, err error) {
 		return conf, errors.New("unable to merge empty configs")
 	}
 
-	conf.Name = c.configs[0].AdvertiseName
```
madhuravi: We want to discourage services using a different advertiseName but this removes the ability to do so. I think services might want this for eg. when a service is re-written, migration paths might involve advertising as a different service name.

sectioneight:
Ya, we def need to support custom advertise names for special snowflakes
and certain test scenarios

modules/rpc/yarpc.go
```diff
@@ -212,39 +219,61 @@ func newYARPCModule(
...
+
+	module.config.transports.inbounds = transportsIn
 	module.config.inboundMiddleware = inboundMiddlewareFromCreateInfo(mi)
 	module.config.onewayInboundMiddleware = onewayInboundMiddlewareFromCreateInfo(mi)
 
-	tchTransport, err := tch.NewChannelTransport(
-		tch.ServiceName(module.config.AdvertiseName),
-		tch.ListenAddr(module.config.Bind),
-	)
+	_controller.addConfig(module.config)
```
madhuravi: Won't this cause duplicates since addDefaultMiddleware has already added the yarpcConfig to c.configs?

modules/rpc/yarpc.go
```diff
@@ -212,39 +219,61 @@ func newYARPCModule(
...
+				return nil, errs.Wrap(err, "can't create tchannel transport")
+			}
 
-	err = _controller.addConfig(module.config)
-	return module, err
+			transportsIn = append(transportsIn, chn.NewInbound())
+		}
+	}
+
+	return transportsIn, err
```
madhuravi: return transportsIn, nil

modules/rpc/yarpc.go
```diff
@@ -306,6 +333,19 @@ func defaultYARPCDispatcher(_ service.Host, cfg yarpc.Config) (*yarpc.Dispatcher
 	return yarpc.NewDispatcher(cfg), nil
 }
 
+type yarpcStarterFn func(dispatcher *yarpc.Dispatcher) error
+
+// RegisterStarter allows you to override the YARPC dispatcher start, e.g. attach some metrics with start.
+func RegisterStarter(starterFN yarpcStarterFn) {
```
madhuravi: nit: same format here too as startedFn

modules/rpc/thrift_test.go
```diff
@@ -27,18 +27,43 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	"go.uber.org/fx/config"
```
anuptalwalkar: fix imports. stdlib imports, pkg imports, then external pkg imports

modules/rpc/thrift_test.go
```diff
@@ -27,18 +27,43 @@ import (
...
+}
+
+func (h testHost) Config() config.Provider {
+	return h.config
+}
+
 func TestThriftModule_OK(t *testing.T) {
 	chip := ThriftModule(okCreate, modules.WithRoles("rescue"))
 	dale := ThriftModule(okCreate, modules.WithRoles("ranges"))
+	cfg := []byte(`modules:
```
anuptalwalkar: nit: newline after `

modules/rpc/yarpc.go
```diff
@@ -62,13 +69,30 @@ var (
...
+
+// Inbound is a union that configures how to configure a single inbound.
+type Inbound struct {
+	TChannel *Address
+	HTTP     *Address
+}
+
+// Address is a struct that have a required port for tchannel/http transports.
+// TODO(alsam) make it optional
+type Address struct {
```
anuptalwalkar: can this not be just a type `Port`? or is there anything else going to be in Address?
```
type Port int
type Inbound struct {
	TChannel *Port
	HTTP     *Port
}
```

modules/rpc/yarpc.go
madhuravi: Do we even need _startedFn at all? Why can't we call c.dispatcher.Start() in line 152?

modules/rpc/yarpc_test.go
madhuravi: let's test that message then?

modules/rpc/thrift_test.go
```diff
@@ -27,18 +27,43 @@ import (
...
+  rpc:
+    inbounds:
```
madhuravi: Test with only one inbound and one with 2 inbounds of the same type.

alsamylkin: Isn't it a superset, 2 and nothing? Or you think there is something that is not covered?

modules/rpc/yarpc.go
alsamylkin: Yes, http can have additional configs(like keep alive, tracer) that we can add later on demand, I know it is a bit weird, but keeps our options open.

modules/rpc/yarpc.go
alsamylkin: This is not in our power, but YARPC: dispatcher allows us to have only one service name.

Snowflakes can create an extra dispatcher and we expose enough functionality: create and start dispatcher functions, for users to synchronize with the second dispatcher if they'd need it. Anyway, I prefer RPC team to tinker these puzzles.

modules/rpc/thrift_test.go
alsamylkin: Believe me, I'll see it ;)

modules/rpc/thrift_test.go
alsamylkin: The other problem is a single dispatcher: we can start it only once. Transports can be started only once: you can't really create another e.g. http transport, test it, stop it and start it again.

modules/rpc/yarpc.go
alsamylkin: Default middleware are private functions, so it is hard to add duplicates, unless users really want it.
Am I missing something?

modules/rpc/yarpc.go
alsamylkin: There is a comment above RegisterStarter: in case a user wants to have e.g. a boot counter, or RPC team wants to do some other naughty things with the dispatcher in private :)

modules/rpc/yarpc_test.go
alsamylkin: I am ok with testing errors that we produce, but testing error messages that produces OS is a bit too much.

modules/rpc/yarpc.go
alsamylkin: Added a todo

PR #232, opened 2017-02-09
Add testing logger and sink

glibsm:
GFM-47

`WithInMemoryLogger` gives access to the logged bytes. This sink adds access to the underlying zap tuples and logged message, in the same way that zap itself is tested.

Very useful for testing that traceID was included with a particular value, for instance, without having to construct part of that JSON to validate against the logged bytes.

### DIFF
--- a/ulog/log_test.go
+++ b/ulog/log_test.go
@@ -34,6 +34,16 @@ import (
 	"github.com/uber-go/zap"
 )
 
+func TestZapSink(t *testing.T) {
+	l, s := TestingLogger()
+	l.Info("This is my message", "And this is", "My field")
+	assert.Equal(t, 1, len(s.Logs()))
+
+	e := s.Logs()[0]
+	assert.Equal(t, "This is my message", e.Msg)
+	assert.Contains(t, e.Fields, zap.String("And this is", "My field"))
+}
+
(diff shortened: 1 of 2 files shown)


ulog/log_test.go
```diff
@@ -34,6 +34,16 @@ import (
...
 )
 
+func TestZapSink(t *testing.T) {
+	l, s := TestingLogger()
+	l.Info("This is my message", "And this is", "My field")
+	assert.Equal(t, 1, len(s.Logs()))
+
+	e := s.Logs()[0]
+	assert.Equal(t, "This is my message", e.Msg)
+	assert.Contains(t, e.Fields, zap.String("And this is", "My field"))
```
alsamylkin: Do you want other tests to use your fancy technique?

glibsm: Certainly some tests I have in mind in the future, we can see if some tests can be improved by this approach.
I think it's not worth it to go change all the tests that use `WithInMemoryLogger` to this.

ulog/log_test.go
```diff
@@ -34,6 +34,16 @@ import (
 	"github.com/uber-go/zap"
 )
 
+func TestZapSink(t *testing.T) {
```
alsamylkin: Lets make it parallel?

Issue #237, opened 2017-02-10
follow std http package Handler interface


yutongp:
Just want open a discussion about:
```
 // Handler is a context-aware extension of http.Handler.
type Handler interface {
	ServeHTTP(ctx context.Context, w http.ResponseWriter, r *http.Request)
}
```
vs
```
type Handler interface {
        ServeHTTP(ResponseWriter, *Request)
}
```
I am playing with std http with context recently and I found myself adopt easily. Here are reasons why I think following std defined interface is better:
1. context in one place. you always get context out of a request every time you use it and put context back every time you update it. if you look at uhttp/filters.go, 
```
func (f contextFilter) Apply(ctx context.Context, w http.ResponseWriter, r *http.Request, next Handler) {
	ctx = fx.NewContext(ctx, f.host)
	next.ServeHTTP(ctx, w, r)
}
```
contextFilter failed to set context in request.
and at https://github.com/uber-go/fx/blob/master/modules/uhttp/filters.go#L80, tracingFilter overwrite request context directly instead wrapping original request context.
Since context is already in request, there is no need to pass context again and have the risk to make these two context out of sync.
Since `request` has `Context` and `WithContext`, user and std will use `context` in `request`. If we need to pick one source of truth, I think we should go with `context` in `request`.

2. no std http.Handler wrapper. 100% compatible with any http related package out there including std right away and always. I think this is important as well. When I select package, I will try to use ones follow std since std API appear to be stable and rarely change.

3. no "golang.org/x/net/context" incompatible issue. a handler implemented: `ServeHTTP(ctx context.Context, w http.ResponseWriter, r *http.Request)` can not be used as a
```
type Handler interface {
		ServeHTTP(ctx golang.org/x/net/context.Context, w http.ResponseWriter, r *http.Request)
}
```
context.Context can be passed as golnag.org/x/next/context.Context, but these two handler interface are not compatible. go1.8 has `go fix` which aiming to "solve" this issue, but following std interface will let us move a way from this mess.

Those are why I think we should use std http handler interface. Let's discuss :)

yutongp:
BTW I do agree context should be the first param in all business logic layers passed HTTP layer. So the user define HTTP handler probably looks like:
```
func (h Handler) ServeHTTP(w ResponseWriter, r *Request) {
  ctx := r.Context() 
  businessReq := toBusinessModel(r)
  businessResp := handleBusiness(ctx , businessReq)
  ....
}
```

yutongp:
[...]

PR #239, opened 2017-02-11
use http.Request context

yutongp:
refer #237 

use http.Request instead of explicitly passing context, so we can follow std API instead of define our own.

### DIFF
--- a/modules/uhttp/filters.go
+++ b/modules/uhttp/filters.go
@@ -52,14 +52,11 @@ type contextFilter struct {
 }
 
 func (f contextFilter) Apply(w http.ResponseWriter, r *http.Request, next http.Handler) {
-	ctx := r.Context()
-	ctx = fx.NewContext(ctx, f.host)
-	r = r.WithContext(ctx)
-	next.ServeHTTP(w, r)
+	ctx := fx.NewContext(r.Context(), f.host)
+	next.ServeHTTP(w, r.WithContext(ctx))
 }
 
-type tracingServerFilter struct {
-}
+type tracingServerFilter struct{}
 
 func (f tracingServerFilter) Apply(w http.ResponseWriter, r *http.Request, next http.Handler) {
 	ctx := r.Context()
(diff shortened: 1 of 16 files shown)


yutongp:
@sectioneight I think `TestTracingFilterWithLogs` in filter_test.go should cover OpenTracing integration

sectioneight:
Right, I know there's unit test coverage, I just want to make sure my team verifies that it works in production :)

modules/uhttp/filters.go
```diff
@@ -34,33 +33,36 @@ import (
...
 }
 
 type contextFilter struct {
 	host service.Host
 }
 
-func (f contextFilter) Apply(ctx context.Context, w http.ResponseWriter, r *http.Request, next Handler) {
+func (f contextFilter) Apply(w http.ResponseWriter, r *http.Request, next http.Handler) {
+	ctx := r.Context()
 	ctx = fx.NewContext(ctx, f.host)
```
anuptalwalkar: this can be single line now `fx.NewContext(r.Context(), f.host)`

modules/uhttp/handler.go
```diff
@@ -21,46 +21,31 @@
...
 	host    service.Host
-	handler Handler
+	handler http.Handler
 }
 
-// ServeHTTP calls Handler.ServeHTTP(ctx, w, r) and injects a new service context for use.
-func (h *handlerWrapper) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	ctx := fx.NewContext(context.Background(), h.host)
+// ServeHTTP calls Handler.ServeHTTP( w, r) and injects a new service context for use.
+func (h *handlerWithHost) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
anuptalwalkar: hmmm. Not sure if this whole wrapping is useful now. Perhaps we can just remove it? The whole point of wrapping is to embed context for service owners who write ServeHTTP to have access to the context, which will now be available via `r.Context()`

yutongp: I agree. We can remove the wrap host part now. request timing can be another filter. I will put an separate diff to remove this wrapper and add a timing filter.

sectioneight:
Thanks! This is much nicer :)

PR #240, opened 2017-02-12
add rpc panic recovery middleware

yutongp:
refer to #240 
I talked with @abhinav. He said the panic recovery middleware does not exist yet. If they add one, that one will be in rpcinit with go-common integration.
In that case, I think fx need one as well for sentry and metrics integrations.


modules/rpc/handler.go
```diff
@@ -88,3 +90,27 @@ func authorize(ctx context.Context, host service.Host) (context.Context, error)
...
+
+func (p panicOnewayInboundMiddleware) HandleOneway(ctx context.Context, req *transport.Request, handler transport.OnewayHandler) error {
+	defer panicRecovery(ctx)
+	return handler.HandleOneway(ctx, req)
+}
+
+func panicRecovery(ctx context.Context) {
+	if err := recover(); err != nil {
+		stats.RPCPanicCounter.Inc(1)
+		fx.Logger(ctx).Error("Panic recovered serving request", "error", err)
```
yutongp: should I also log `debug.Stack()` here as well? raven does send stack trace to sentry, but even users are not running with sentry, they will miss the stack trace.

modules/rpc/handler.go
```diff
@@ -30,6 +30,8 @@ import (
 	"go.uber.org/yarpc/api/transport"
 )
 
+const panicRespnose = "Server Error"
```
madhuravi: typo: panicResponse

modules/rpc/handler.go
madhuravi: We have an interceptor for error logs, and print the stack trace by default. @alsamylkin to confirm that it works in this case as well.

modules/rpc/handler.go
```diff
@@ -88,3 +90,27 @@ func authorize(ctx context.Context, host service.Host) (context.Context, error)
...
+	return handler.HandleOneway(ctx, req)
+}
+
+func panicRecovery(ctx context.Context) {
+	if err := recover(); err != nil {
+		stats.RPCPanicCounter.Inc(1)
+		fx.Logger(ctx).Error("Panic recovered serving request", "error", err)
+		// rethrow panic back to yarpc
+		// before https://github.com/yarpc/yarpc-go/issues/734 fixed, throw a generic error.
+		panic(panicRespnose)
```
madhuravi: it should be something along the lines of panic(err.Error()). server error is too generic and does not convey much.

modules/rpc/handler.go
madhuravi: We should also print some request specific information here.

modules/rpc/handler.go
```diff
@@ -88,3 +90,27 @@ func authorize(ctx context.Context, host service.Host) (context.Context, error)
...
+	defer panicRecovery(ctx)
+	return handler.HandleOneway(ctx, req)
+}
+
+func panicRecovery(ctx context.Context) {
+	if err := recover(); err != nil {
+		stats.RPCPanicCounter.Inc(1)
+		fx.Logger(ctx).Error("Panic recovered serving request", "error", err)
+		// rethrow panic back to yarpc
+		// before https://github.com/yarpc/yarpc-go/issues/734 fixed, throw a generic error.
```
madhuravi: From the issue, it looks like yarpc already implements a panic handler. Why does UberFx need one?

modules/rpc/handler.go
yutongp: I think we should return general error message to client (since client can be something out of DC) so we don't leak something like: server use golang, server has nil pointer etc. Meaningful info can be present through introspector and sentry to service owners .

modules/rpc/handler.go
yutongp: Yarpc only log to stderr, this is more for log to all hooks in fx like sentry. Of course, if filebeat is almost ready, it will be not that useful for us. But should be still useful for open source when they hook with sentry

modules/rpc/handler.go
alsamylkin: It will not log a stacktrace here, because err can technically not implement error interface :)
The easiest way to get the stacktrace, is to wrap err with errors.Wrap and log it, then you'll get a stacktrace for free.

modules/rpc/handler.go
```diff
@@ -88,3 +90,27 @@ func authorize(ctx context.Context, host service.Host) (context.Context, error)
...
+		panic(panicResponse)
```
alsamylkin: Why do we rethrow panic here? I thought the point of this filter was to not cause panics, like https://github.com/uber-go/fx/blob/master/modules/uhttp/filters.go#L102

modules/rpc/handler.go
```diff
@@ -88,3 +90,27 @@ func authorize(ctx context.Context, host service.Host) (context.Context, error)
 	}
 	return ctx, nil
 }
+
+type panicInboundMiddleware struct{}
```
glibsm: unary

modules/rpc/handler.go
```diff
@@ -30,6 +30,8 @@ import (
 	"go.uber.org/yarpc/api/transport"
 )
 
+const panicResponse = "Server Error"
```
glibsm: two things:

1. This var looks familiar from another diff
2. Underscore the const variables, please

modules/rpc/handler.go
glibsm: Im not sure what the correct behavior should be here, but I don't think we should just "swallow" the panic. We do want to let the caller know something went wrong

Even if we re-throw in this case, we got a chance to record a metric and log an error. Any other processing that we like can also be added easily

modules/rpc/handler.go
yutongp: 1. ya, I want to through a same general error msg for panic. do we need to share this const between `rpc` and `uhttp`
2. sure

modules/rpc/handler.go
yutongp: this follows the naming convention with `contextInboundMiddleware` and `authInboundMiddleware` right now
maybe change all of them in another diff?

modules/rpc/handler.go
alsamylkin: Services can return errors, that's what we should do in this case. Panic here is an inconsistent behavior between different modules.

modules/rpc/handler.go
yutongp: what kind of request specific information we are thinking about to log here?

modules/rpc/handler.go
yutongp: the reason I rethrow a panic is: here in https://github.com/yarpc/yarpc-go/blob/61fa69b4a04faa5d1dc57b5a2ce44f93c16e6485/api/transport/handler.go#L100
 yarpc does have recover from panic and log the panic to std output. but I don't like how they return the panic msg back to client, so I rethrow a more generic msg...

alsamylkin: Ok, since they documented recovery, its fine with me to panic here.

modules/rpc/handler.go
alsamylkin: Entire request?

yutongp: Entire request may have PII concern. sentry can mask password but kafka probably will not.

alsamylkin: Then we should have another filter? Panic shouldn't handle the PII part..

glibsm: The point is that we re-throw panic with a new error that we fully control. We're not obscuring PII or doing anything else crazy, just masking whatever actually went front and returning a generic `500 service error` so to speak

alsamylkin: @glibsm We are talking about logging additional information about panic: request, how can we stripe the PII information from it and safely log.

glibsm: I see, I think there are too many threads in here that got mixed up

PII concern applies to *all* logging and sentry, lets not try to solve it only for YARPC

PII in response on failure is what we should be talking about here, and returning a custom error seems to satisfy the need.

yutongp: so what's the conclusion we have here? keep what I have right now?
I agree obscuring PII is a seperate issue, before we have a story about that, not logging request make sense to me.

alsamylkin: Why not logging the request makes sense?

alsamylkin: synced offline, it does make sense to not log request information, because it is not safe in general case.

modules/rpc/yarpc.go
```diff
@@ -126,10 +126,12 @@ func (c *dispatcherController) addDefaultMiddleware(host service.Host) {
 	cfg := yarpcConfig{
 		inboundMiddleware: []middleware.UnaryInbound{
 			contextInboundMiddleware{host},
+			panicInboundMiddleware{},
```
alsamylkin: I think panic middleware should be the very first in the slice, to catch all possible issues.

modules/rpc/handler_test.go
```diff
@@ -116,3 +116,32 @@ func (f fakeOnewayHandler) HandleOneway(ctx context.Context, p *transport.Reques
 	assert.NotNil(f.t, ctx)
 	return errors.New("oneway handle")
 }
+
+func TestInboundMiddleware_panic(t *testing.T) {
+	defer testPanicHandler(t)
```
alsamylkin: nit: use t.Parallel()?

modules/rpc/handler_test.go
```diff
@@ -116,3 +116,32 @@ func (f fakeOnewayHandler) HandleOneway(ctx context.Context, p *transport.Reques
...
+	oneway := panicOnewayInboundMiddleware{}
+	oneway.HandleOneway(context.Background(), &transport.Request{}, &panicOnewayHandler{})
+}
+
+func testPanicHandler(t *testing.T) {
+	r := recover()
+	assert.EqualValues(t, r, _panicResponse)
+}
+
+type panicUnaryHandler struct{}
```
alsamylkin: nit: name it alwaysPanicUnary?

modules/rpc/handler.go
```diff
@@ -88,3 +90,27 @@ func authorize(ctx context.Context, host service.Host) (context.Context, error)
...
+type panicOnewayInboundMiddleware struct{}
+
+func (p panicOnewayInboundMiddleware) HandleOneway(ctx context.Context, req *transport.Request, handler transport.OnewayHandler) error {
+	defer panicRecovery(ctx)
+	return handler.HandleOneway(ctx, req)
+}
+
+func panicRecovery(ctx context.Context) {
+	if err := recover(); err != nil {
+		stats.RPCPanicCounter.Inc(1)
```
alsamylkin: I don't see where this counter update is tested..

modules/rpc/yarpc.go
yutongp: but then we don't have the logger with host in context

alsamylkin: oh, ok

modules/rpc/handler.go
yutongp: is there way to test metrics count/timer is called?  I would like to test in https://github.com/uber-go/fx/pull/251 as well. Maybe you can point me somewhere metrics is tested.

alsamylkin: Here is an example:
https://github.com/uber-go/fx/blob/master/modules/task/execution_test.go#L172

modules/rpc/handler_test.go
yutongp: ya, then maybe we should apply parallel for all tests in this file. Let me pull another PR to address that. I do want to do some refactor as well.

modules/rpc/handler.go
```diff
@@ -88,3 +92,27 @@ func authorize(ctx context.Context, host service.Host) (context.Context, error)
...
+		fx.Logger(ctx).Error("Panic recovered serving request", "error", errors.Errorf("panic: %+v", err))
```
yutongp: @alsamylkin I cannot use `errors.Wrap` since it takes in an `error` type. let me know if I am doing it wrong. This is to get the stack trace.

modules/rpc/handler.go
```diff
@@ -28,8 +28,12 @@ import (
 	"go.uber.org/fx/modules/rpc/internal/stats"
 	"go.uber.org/fx/service"
 	"go.uber.org/yarpc/api/transport"
+
+	"github.com/pkg/errors"
```
sectioneight: nit: this should be grouped with yarpc since it's a "3rd-party" import

modules/rpc/handler.go
sectioneight:
```go
fx.Loger(ctx).Error("Panic recovered serving request", "error", errors.Wrap(err, "panic in handler")
```

yutongp: `errors.Wrap` asked for an `error` interface, `recover` only returns `interface{}`, so we cannot use `errors.Wrap` here. I will change the text

PR #241, opened 2017-02-12
uhttp stop return panic detail to client

yutongp:
I opened a similar issue to yarpc [hide panic detail to caller](https://github.com/yarpc/yarpc-go/issues/734)

I think we should hide server panic detail from client to avoid some information leaking. the panic detail can be found in server's log and maybe sentry.

### DIFF
--- a/modules/uhttp/filters.go
+++ b/modules/uhttp/filters.go
@@ -33,6 +33,8 @@ import (
 	"github.com/opentracing/opentracing-go/ext"
 )
 
+const panicRespnose = "Server Error"
+
 // Filter applies filters on requests or responses such as
 // adding tracing to the context
 type Filter interface {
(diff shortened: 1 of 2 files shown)


sectioneight:
We actually have an open ticket to expose debug info to employees in a consistent way but provide generic errors user-facing

yutongp:
@sectioneight maybe we can start with returning general msg to client? what kind of channel you think we can use to return debug info to employees? or maybe a debug flag during build?

modules/uhttp/filters.go
```diff
@@ -34,6 +34,8 @@ import (
 	"github.com/opentracing/opentracing-go/ext"
 )
 
+const panicRespnose = "Server Error"
```
madhuravi: typo: panicResponse

yutongp:
@alsamylkin  I think it should not be different in windows: https://golang.org/src/fmt/print.go?s=7002:7065#L1127 `doPrintln` only add `\ n`.

PR #243, opened 2017-02-12
Update zap to latest commit on dev branch

peter-edge:
I was working on writing a new Module for another project that also uses zap, and I noticed that FX currently uses zap on master, which is way behind the latest zap development on dev. I think this was on purpose (ie I think the whole purpose of the master branch on zap is so that FX could develop against a stable zap API), but as you approach 1.0 for FX, I think getting the latest updates would be a really good idea, so that users of FX are using zap as it will be going forward.

This also does a global glide update, but I think this is the only breaking change. It would be a good idea to audit the rest of the dependencies for updates before FX goes to 1.0 as well. If this is merged, it would be nice to just merge zap dev into master, and then have this depend on master again.

A few notes:

- This is a half hack job. I am just learning the workings of FX, and this should be a drop-in replacement, but I worked to get this just passing tests. This could use a fine-tooth comb by one of the FX maintainers, and I left some TODOs in the code as it is now.
- I think LogBuilder could use some general work afterwards, but I know you're all pushing for 1.0 so this is for later.
- Instead of using LogConfiguration, it might be nice to directly use zap's new configuration struct.
- A general wish: I wish FX wasn't so tightly tied to zap, I wish we could have other logging libraries as well (and a lot of users will want this). It would be non-trivial but not too difficult to separate the tight coupling to zap.
- This PR was a pain in the butt, and if you want to merge it, we should go over it NOW and get it in before this gets into rebase world. I'm happy to do a video chat to get this done quickly, or if someone else wants to take it over, go for it.

@sectioneight @akshayjshah @glibsm @anuptalwalkar

### DIFF
--- a/testutils/in_memory_log.go
+++ b/testutils/in_memory_log.go
@@ -21,41 +21,35 @@
 package testutils
 
 import (
-	"bytes"
-	"strings"
 	"testing"
 
-	"github.com/uber-go/zap"
+	"go.uber.org/zap"
+	"go.uber.org/zap/testutils"
+	"go.uber.org/zap/zapcore"
 )
 
-// TestBuffer is a buffer used to test the zap logger
-type TestBuffer struct {
-	bytes.Buffer
-}
-
-// Sync is a nop to conform to zap.WriteSyncer interface
-func (b *TestBuffer) Sync() error {
-	return nil
-}
-
-// Lines returns buffer as array of strings
-func (b *TestBuffer) Lines() []string {
-	output := strings.Split(b.String(), "\n")
-	return output[:len(output)-1]
-}
-
-// Stripped returns buffer as a string without the newline
-func (b *TestBuffer) Stripped() string {
-	return strings.TrimRight(b.String(), "\n")
-}
-
 // WithInMemoryLogger creates an in-memory zap logger that can be used in tests
-func WithInMemoryLogger(t *testing.T, opts []zap.Option, f func(zap.Logger, *TestBuffer)) {
-	sink := &TestBuffer{}
-	errSink := &TestBuffer{}
-
-	allOpts := make([]zap.Option, 0, 3+len(opts))
-	allOpts = append(allOpts, zap.DebugLevel, zap.Output(sink), zap.ErrorOutput(errSink))
-	allOpts = append(allOpts, opts...)
-	f(zap.New(zap.NewJSONEncoder(zap.NoTime()), allOpts...), sink)
+func WithInMemoryLogger(t *testing.T, opts []zap.Option, f func(*zap.Logger, *testutils.Buffer)) {
+	buffer := &testutils.Buffer{}
+	f(
+		zap.New(
+			zapcore.WriterFacility(
+				zapcore.NewJSONEncoder(
+					zapcore.EncoderConfig{
+						MessageKey:    "msg",
+						LevelKey:      "level",
+						CallerKey:     "caller",
+						StacktraceKey: "stacktrace",
+						//EncodeTime:     zapcore.EpochTimeEncoder,
+						EncodeDuration: zapcore.SecondsDurationEncoder,
+						EncodeLevel:    zapcore.LowercaseLevelEncoder,
+					},
+				),
+				buffer,
+				zapcore.DebugLevel,
+			),
+			opts...,
+		),
+		buffer,
+	)
 }
--- a/ulog/builder.go
+++ b/ulog/builder.go
@@ -27,7 +27,6 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/uber-go/tally"
-	"go.uber.org/fx/config"
 	"go.uber.org/fx/ulog/metrics"
 	"go.uber.org/fx/ulog/sentry"
 	"go.uber.org/zap"
@@ -120,13 +119,7 @@ func (lb *LogBuilder) Build() (Log, error) {
 		}, nil
 	}
 
-	var log *zap.Logger
-	var err error
-	if config.IsDevelopmentEnv() {
-		log, err = lb.devLogger()
-	} else {
-		log, err = lb.Configure()
-	}
+	log, err := lb.Configure()
 	if err != nil {
 		return nil, err
 	}
@@ -151,10 +144,6 @@ func (lb *LogBuilder) Build() (Log, error) {
 		log: log,
 		sh:  lb.sentryHook,
 	}, nil
-}
-
-func (lb *LogBuilder) devLogger() (*zap.Logger, error) {
-	return zap.NewDevelopment()
 }
 
 // Configure Log object with the provided log.Configuration
--- a/ulog/builder.go
+++ b/ulog/builder.go
@@ -25,13 +25,26 @@ import (
 	"os"
 	"path"
 
+	"github.com/pkg/errors"
+	"github.com/uber-go/tally"
 	"go.uber.org/fx/config"
 	"go.uber.org/fx/ulog/metrics"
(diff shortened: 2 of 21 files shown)


akshayjshah:
I'm happy to take this over and manage the rebasing. I'd much prefer to wait 1-3 days and pull in the RC of zap 1.0.

peter-edge:
Please do, thank you :) I'd like to file a "friend of the court" brief and
note I'd love this merged sooner than later, the project I'm working on
will be a lot easier if I don't have to rewrite the logging parts of FX
integration twice (it's not that bad to have to change a few things in this
project twice, but just annoying)

sectioneight:
This sounds good, but keep in mind that we're releasing UberFx beta a week from tomorrow, so if you want this change merged we'll need it before EOD Thursday so we can end-to-end test before release.

peter-edge:
It's "ready" to merge but Aisha's should give it a once over for sure.

peter-edge:
*Akshay (iOS spell correct)

peter-edge:
@akshayjshah when can we have that? :)

I don't think you need to worry about SugaredLogger for now. We can do two things:

1. Wait to merge. In that case, this PR is going to be in rebase hell and it's just going to be annoying work for someone to do :)
2. Just merge now, merge zap dev into master, continue going along as is (where FX depends on zap master)

Obviously I'm biased :)

testutils/in_memory_log.go
```diff
@@ -21,41 +21,35 @@
...
+	f(
+		zap.New(
+			zapcore.WriterFacility(
+				zapcore.NewJSONEncoder(
+					zapcore.EncoderConfig{
+						MessageKey:    "msg",
+						LevelKey:      "level",
+						CallerKey:     "caller",
+						StacktraceKey: "stacktrace",
+						//EncodeTime:     zapcore.EpochTimeEncoder,
```
sectioneight: Why is this here / commented-out?

testutils/in_memory_log.go
```diff
@@ -21,41 +21,35 @@
...
-	allOpts = append(allOpts, zap.DebugLevel, zap.Output(sink), zap.ErrorOutput(errSink))
-	allOpts = append(allOpts, opts...)
-	f(zap.New(zap.NewJSONEncoder(zap.NoTime()), allOpts...), sink)
+func WithInMemoryLogger(t *testing.T, opts []zap.Option, f func(*zap.Logger, *testutils.Buffer)) {
+	buffer := &testutils.Buffer{}
+	f(
+		zap.New(
+			zapcore.WriterFacility(
+				zapcore.NewJSONEncoder(
+					zapcore.EncoderConfig{
```
sectioneight: Holy nesting, batman. Can we pull this out to a `var` to make this more readable?

ulog/builder.go
```diff
@@ -99,108 +112,104 @@ func (lb *LogBuilder) WithSentryHook(hook *sentry.Hook) *LogBuilder {
...
 	// TODO(glib): document that yaml configuration takes precedence or
 	// make the situation better so yaml overrides only the DSN
 	if lb.logConfig.Sentry != nil {
 		if len(lb.logConfig.Sentry.DSN) > 0 {
 			hook, err := sentry.Configure(*lb.logConfig.Sentry)
-			if err == nil {
-				lb.sentryHook = hook
-			} else {
+			if err != nil {
+				// TODO(pedge): it would be nicer to return the error
```
sectioneight: Interesting. Why don't we return an error here?

ulog/builder.go
```diff
@@ -99,108 +112,104 @@ func (lb *LogBuilder) WithSentryHook(hook *sentry.Hook) *LogBuilder {
...
-func (lb *LogBuilder) Configure() zap.Logger {
-	lb.log = lb.defaultLogger()
-	options := lb.zapOptions()
-
-	if lb.logConfig.TextFormatter != nil && *lb.logConfig.TextFormatter {
-		return zap.New(zap.NewTextEncoder(), options...)
+func (lb *LogBuilder) Configure() (*zap.Logger, error) {
+	levelEnabler := zap.DebugLevel
+	if !lb.logConfig.Verbose {
+		// TODO(pedge): not set anymore
```
sectioneight: ?

ulog/builder.go
```diff
@@ -99,108 +112,104 @@ func (lb *LogBuilder) WithSentryHook(hook *sentry.Hook) *LogBuilder {
...
+		//lb.log.Info("Setting log level", zap.String("level", lb.logConfig.Level))
```
sectioneight: ?

ulog/builder_test.go
```diff
@@ -165,6 +170,8 @@ func TestMetricsHook(t *testing.T) {
 	assert.Equal(t, 1, len(r.Counters))
 }
 
+// TODO(pedge): it's non-trivial to test this now with how zap sets up hooks
+/*
```
sectioneight: wait, was this already commented-out?

ulog/doc.go
```diff
@@ -21,7 +21,7 @@
 // Package ulog is the Logging package.
 //
 // package ulog provides an API wrapper around the logging library
-// zap (https://github.com/uber-go/zap). package ulog uses the builder pattern
+// zap (https://go.uber.org/zap). package ulog uses the builder pattern
```
sectioneight: I like the previous link better, since godoc does'nt link back to github but github links to godoc

sectioneight: Also, these files are auto generated (`make gendoc`) from README so if you do want to change it, change it there first

sectioneight:
note to @alsamylkin -- this appears to remove `IsDevelopmentEnv()` which is on your task list. We should ensure that we're still using a text outputter in dev, but your work may already be done :)

peter-edge:
As this is right now, it just has you set the logging config separately for development and production. In development.yaml, I have:

```yaml
logging:
  level: debug
  stdout: true
  textFormatter: true
```

In production.yaml, I have:
```yaml
logging:
  level: info
```

I'm not sure if you want to assume the formatter based on environment, in fact I'd argue against this, and against some more auto-magical stuff happening with the logging configuation..I'll write something up.

Also as I just mentioned, zap lets do you:

```
go run main.go | jq .
```

So you don't always want a text formatter in development anyways.

sectioneight:
Having textFormatter: true seems like the right solution to this.

peter-edge:
Hey, I'm starting to agree with @anuptalwalkar on this - zap is still changing a lot, and this is probably still a decent base, but even now it's getting into rebase land. @akshayjshah let me know what you want to do, I really want to get this in pre-beta but it might need refactoring again, you probably have more context

akshayjshah:
Zap has tagged a release candidate for a stable 1.0 release, so I don't anticipate many more changes (if any). I'll take this over today and get it ready for merge.

Fair warning, though - this will involve backwards-incompatible changes.

peter-edge:
Thank you so much! @akshayjshah I would note that I also wanted to get in a refactor of the logging configuration in fx, to match zap more closely, but we might try to do that later?

akshayjshah:
I'd prefer to get it in before UberFX starts their open beta, which sounds like it's early next week. The changes will be visible to users, and I'd prefer to not break them right after they start using a new framework.

sectioneight:
NOTE: I had to pin zap today in #264 to unblock service creation -- right now, master is not usable for us

anuptalwalkar:
We should close this PR in favor of https://github.com/uber-go/fx/pull/290 I will hold off by tomorrow if you disagree.

peter-edge:
sgtm

PR #244, opened 2017-02-12
Generic module package

peter-edge:
I was working on writing a new Module, and I realized that there's a lot of duplicated code between the existing Module implementations in FX, and that most implementations of Module just need a couple things:

- Access to the service/config data
- Start
- Stop

It would be nice if there was a simplification of this. This is what I have come up with for now. I'm not crazy about the Controller interface, and the call to Initialize, but this should make writing Modules a lot easier. I may have missed something or done something incorrectly or not per FX idioms, so let me know. I'll add more testing if this is something you'd want, but didn't want to go further until I got some feedback.

### DIFF
--- a/modules/generic/generic.go
+++ b/modules/generic/generic.go
@@ -63,7 +63,7 @@ type Module interface {
 	Stop() error
 }
 
-// NewModule returns a ModuleCreateFunc for the given GenericModule.
+// NewModule returns a ModuleCreateFunc for the given Module.
 func NewModule(
 	moduleName string,
 	module Module,
--- a/modules/generic/generic.go
+++ b/modules/generic/generic.go
@@ -23,36 +23,47 @@ package generic
 import (
 	"sync"
 
+	"github.com/opentracing/opentracing-go"
+
 	"go.uber.org/fx/modules"
 	"go.uber.org/fx/service"
 	"go.uber.org/fx/ulog"
 )
 
-// Module is a simpliciation of the Module interface
-// that can be wrapped with a Module for easier implemenation.
+// Controller holds data for a Module, and provides functionality
+// for the Module to notify the wrapping service.Module of state changes.
+// TODO(pedge): not the best name
+// TODO(pedge): split up into read-only and read-write interfaces?
+type Controller interface {
+	Host() service.Host
+	Roles() []string
+	Name() string
+	Tracer() opentracing.Tracer
+	Log() ulog.Log
+	// NotifyStopped will result in the wrapping service.Module's
+	// IsRunning() method to return false.
+	NotifyStopped()
+}
+
+// Module is a simpliciation of the service.Module interface
+// that can be wrapped with a service.Module for easier implemenation.
 type Module interface {
-	// This will be called after ModuleBase, log, and config are populated.
+	// This will be called after the Controller and config are populated.
 	// If the module wishes, these can be stored for use in the module.
-	// Start and stop should not be called before this is called.
-	Initialize(moduleBase modules.ModuleBase, log ulog.Log, config interface{}) error
-	// Start the module, on return the module is expected to be started
-	// If there is an error, the module is not expected to be started.
+	//
+	// This should only be called by the generic package, and unless there are
+	// other calls, it can be safely assumed that this will be called exactly once,
+	// and called before Start() and Stop() are ever called.
+	Initialize(controller Controller, config interface{}) error
+	// Start the module.
+	// On return,  the module is expected to be started, unless there is an error.
 	Start() error
-	// Stop the module, on return the module is expected to be stopped
-	// If there is an error, the module is still expected to be stopped.
+	// Stop the module.
+	// On return the module is expected to be stopped, regardless of if there is an error.
 	Stop() error
 }
 
 // NewModule returns a ModuleCreateFunc for the given GenericModule.
-//
-// config should be a struct pointer to a configuration struct.
-//
-//   type FooConfig struct {
-//       modules.ModuleConfig
-//       Foo string
-//   }
-//
-//   NewModule("foo", module, &FooConfig{})
 func NewModule(
 	moduleName string,
 	module Module,
@@ -69,6 +80,8 @@ func NewModule(
 }
 
 type wrapperModule struct {
+	modules.ModuleBase
+	log        ulog.Log
 	moduleName string
 	module     Module
 	lock       sync.RWMutex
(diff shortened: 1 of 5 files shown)


sectioneight:
Two main comments:

1) It looks like we're re-defining interfaces that look very similar to others. Is this to avoid import cycles?
2) This is essentially "dangling" code right now. Is there any way we can refactor our existing modules to use this pattern? Otherwise I'm worried that a) the implementation will break/drift and b) by not dog-fooding we may rely on internal accessors that may make non-core modules more difficult to implement

peter-edge:
1. The Controller interface isn't quite a redefinition, it just models the struct ModuleBase along with an additional Log() function (plus the NotifyStopped() function). If anything, it would be good to have as an interface that ModuleBase just implements.

2. We could just refactor the main Module implementation, I just figured it was the way it is on purpose. However, making services.Module what Generic.Module is now might not be a bad idea. I can send up a PR to this effect.

modules/generic/generic.go
```diff
@@ -0,0 +1,165 @@
...
+		return nil, err
+	}
+	return wrapperModule, nil
+}
+
+func (m *wrapperModule) Name() string {
+	return m.moduleName
+}
+
+func (m *wrapperModule) Start(readyC chan<- struct{}) <-chan error {
```
dmcaulay: What are your thoughts on making this a synchronous method? I think module authors will appreciate the simpler interface. UberFx initialization code can always spin up goroutines to parallelize if we think it's necessary.

modules/generic/generic.go
```diff
@@ -0,0 +1,165 @@
...
+	m.lock.Lock()
+	defer m.lock.Unlock()
```
dmcaulay: Can we get rid of locking as a default? It seems like overkill to me.

modules/generic/generic.go
```diff
@@ -0,0 +1,165 @@
...
+		moduleName = moduleCreateInfo.Name
+	}
+	moduleBase := *modules.NewModuleBase(
+		moduleName,
+		moduleCreateInfo.Host,
+		moduleCreateInfo.Roles,
+	)
+	wrapperModule := &wrapperModule{
+		ModuleBase: moduleBase,
+		log:        ulog.Logger().With("moduleName", moduleName),
```
dmcaulay: Scoping the logger is interesting. It seems like we could also scope metrics and config too. When I was integrating with fx I found myself wanting a `Scope` method on the `Host` interface so I could scope the host before passing it into my constructors.

service/host_mock.go
```diff
@@ -32,7 +32,12 @@ import (
 
 // NopHost is to be used in tests
 func NopHost() Host {
-	return NopHostConfigured(auth.NopClient, ulog.NopLogger, opentracing.NoopTracer{})
+	return NopHostWithConfig(nil)
+}
+
+// NopHostWithConfig is to be used in tests and allows setting of config.
+func NopHostWithConfig(configProvider config.Provider) Host {
```
dmcaulay: 👍

modules/generic/generic.go
```diff
@@ -0,0 +1,165 @@
...
+// PopulateStruct populates the given configuration struct using the given Controller.
+func PopulateStruct(controller Controller, config interface{}) error {
+	return controller.Host().Config().Scope("modules").Get(controller.Name()).PopulateStruct(config)
+}
+
+// Controller holds data for a Module, and provides functionality
+// for the Module to notify the wrapping service.Module of state changes.
+// TODO(pedge): not the best name
+// TODO(pedge): split up into read-only and read-write interfaces?
+type Controller interface {
```
dmcaulay: Is `Controller` supposed to represent the global backends (RPC Dispatcher, Kafka client, Cherami client) that most modules depend on?

modules/generic/generic.go
peter-edge: That's definitely worth considering

modules/generic/generic.go
peter-edge: Yes, I wrote this PR on the thinking that Module was locked down, we should just make it a synchronous method.

modules/generic/generic.go
peter-edge: I don't think it has much cost, and it is proper. I don't think of locking vs not as "overkill", its about what is correct - without it, wrapperModule isn't thread-safe

modules/generic/generic.go
peter-edge: It's basically a hacked way to do DI, along with providing a mechanism to give the Module control over saying if events have happened (like the module stopped without Stop being called).

modules/generic/generic.go
```diff
@@ -0,0 +1,165 @@
...
+	Tracer() opentracing.Tracer
+	Log() ulog.Log
+	// NotifyStopped will result in the wrapping service.Module's
+	// IsRunning() method to return false.
+	NotifyStopped()
+}
+
+// Module is a simpliciation of the service.Module interface
+// that can be wrapped with a service.Module for easier implemenation.
+type Module interface {
```
alsamylkin: Name is the same as service.Module interface, which would lead to many confusions, maybe something like SyncModule?

modules/generic/generic.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+// NewModule returns a ModuleCreateFunc for the given Module.
+func NewModule(
+	moduleName string,
+	module Module,
+	options ...modules.Option,
+) service.ModuleCreateFunc {
+	return func(moduleCreateInfo service.ModuleCreateInfo) ([]service.Module, error) {
+		module, err := newWrapperModule(moduleCreateInfo, moduleName, module, options...)
```
alsamylkin: this is a very confusing line: we have on module being declared using an external variable with the same name.

modules/generic/generic_test.go
```diff
@@ -0,0 +1,153 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"go.uber.org/fx/config"
+	"go.uber.org/fx/modules"
+	"go.uber.org/fx/service"
+)
+
+func TestConfig(t *testing.T) {
+	testConfig := &testConfig{Test: "foo"}
```
alsamylkin: nit: can we run tests in parallel?

service/host_mock.go
alsamylkin: Could you replace a similar struct: https://github.com/uber-go/fx/blob/master/modules/rpc/thrift_test.go#L37 ?

modules/generic/generic.go
```diff
@@ -0,0 +1,165 @@
...
+// Module is a simpliciation of the service.Module interface
+// that can be wrapped with a service.Module for easier implemenation.
+type Module interface {
+	// This will be called after the Controller is populated.
+	// If the module wishes, this can be stored for use in the module.
+	//
+	// This should only be called by the generic package, and unless there are
+	// other calls, it can be safely assumed that this will be called exactly once,
+	// and called before Start() and Stop() are ever called.
+	Initialize(controller Controller) error
```
peter-edge: Ya, I could protect it better, and Start might be a better place. This PR was really just a first pass, I'll address the rest of these comments too.

modules/generic/generic.go
peter-edge: That's fine per golang idioms - the initial intent would be to name it generic.GenericModule, but this fails golang linters because you shouldnt have foo.FooWhatever.

modules/generic/generic.go
```diff
@@ -0,0 +1,165 @@
...
+// THE SOFTWARE.
+
+package generic
+
+import (
+	"sync"
+
+	"github.com/opentracing/opentracing-go"
+
+	"go.uber.org/fx/modules"
```
peter-edge: I think it's the other way around usually

peter-edge:
The logging PR is higher priority to me, so I'm going to work on that first, and see if I can make a pass on this tonight or tomorrow morning.

peter-edge:
Closing this for https://github.com/uber-go/fx/pull/275 which hopefully I will finish today.

PR #247, opened 2017-02-13
Make register function types public

alsamylkin:
Right now they are public and it looks weird using them in public functions.


glibsm:
Pro tip: include `Fixes #issue_num` to auto-close the issue after PR lands

PR #248, opened 2017-02-13
Restrict ulog.Logger to always require context

anuptalwalkar:
1. users should only be able to access `ulog.Logger()` when a context is provided.

Not included in the PR - 
1. Logger without context is still available at host level as `host.Logger()`. Follow up PR to remove that functionality.


examples/keyvalue/server/main.go
```diff
@@ -39,7 +39,7 @@ func main() {
 	).Build()
 
 	if err != nil {
-		ulog.Logger().Fatal("Unable to initialize service", "error", err)
+		ulog.Logger(nil).Fatal("Unable to initialize service", "error", err)
```
alsamylkin: context.Background()?

ulog/log.go
```diff
@@ -99,14 +107,59 @@ func TestingLogger() (Log, *spy.Sink) {
 	}, sink
 }
 
-// Logger returns the package-level logger configured for the service
-// TODO:(at) Remove Logger() call, _std and defaultLogger() access in ulog
-func Logger() Log {
+func logger() Log {
+	_setupMu.Lock()
```
alsamylkin: RLock?

ulog/log.go
```diff
@@ -99,14 +107,59 @@ func TestingLogger() (Log, *spy.Sink) {
...
+	defer _setupMu.Unlock()
+
 	return &baseLogger{
 		log: _std.(*baseLogger).log,
```
alsamylkin: Probably we should have _std to be baseLogger type, to catch a panic during SetLogger..

ulog/log.go
```diff
@@ -99,14 +107,59 @@ func TestingLogger() (Log, *spy.Sink) {
...
+func NewLogContext(ctx context.Context) context.Context {
+	if ctx == nil {
+		ctx = context.Background()
+	}
+
+	return context.WithValue(ctx, internal.ContextLogger, logger())
+}
+
+// WithContextAwareLogger returns a new context with a context-aware logger
+func WithContextAwareLogger(ctx context.Context, span opentracing.Span) context.Context {
```
alsamylkin: Should it be called WithTracingAware?

modules/uhttp/filters.go
```diff
@@ -65,15 +65,15 @@ func (f tracingServerFilter) Apply(ctx context.Context, w http.ResponseWriter, r
 	carrier := opentracing.HTTPHeadersCarrier(r.Header)
 	spanCtx, err := opentracing.GlobalTracer().Extract(opentracing.HTTPHeaders, carrier)
 	if err != nil && err != opentracing.ErrSpanContextNotFound {
-		fx.Logger(ctx).Warn("Malformed inbound tracing context: ", "error", err.Error())
+		ulog.Logger(ctx).Warn("Malformed inbound tracing context: ", "error", err.Error())
```
anuptalwalkar: I'll have a follow up PR when host.Logger() is removed, since this PR is only refactoring current setup.

ulog/log.go
```diff
@@ -99,12 +107,57 @@ func TestingLogger() (Log, *spy.Sink) {
...
-		log: _std.(*baseLogger).log,
+		log: _std.log,
+	}
+}
+
+// SetLogger sets configured logger at the start of the service
+func SetLogger(log Log) {
+	_setupMu.Lock()
+	defer _setupMu.Unlock()
+	_std = log.(*baseLogger)
```
alsamylkin: This is very non-intuitive, could you add comments, why do we want `log` variable to be an internal type?

examples/simple/main.go
```diff
@@ -32,7 +32,7 @@ func main() {
 	).Build()
 
 	if err != nil {
-		ulog.Logger().Fatal("Unable to initialize service", "error", err)
+		ulog.Logger(nil).Fatal("Unable to initialize service", "error", err)
```
alsamylkin: should we use Fatal() from standard log package here?

modules/uhttp/client/client_filters.go
```diff
@@ -98,13 +98,13 @@ func authenticationFilter(info auth.CreateAuthInfo) FilterFunc {
...
 		if err != nil {
-			fx.Logger(ctx).Error(auth.ErrAuthentication, "error", err)
+			ulog.Logger(ctx).Error(auth.ErrAuthentication, "error", err)
 			return nil, err
 		}
 
 		span := opentracing.SpanFromContext(authCtx)
 		if err := injectSpanIntoHeaders(req.Header, span); err != nil {
-			fx.Logger(ctx).Error("Error injecting auth context", "error", err)
+			ulog.Logger(ctx).Error("Error injecting auth context", "error", err)
```
alsamylkin: should we use authCtx here?

ulog/log.go
```diff
@@ -99,12 +107,57 @@ func TestingLogger() (Log, *spy.Sink) {
...
+}
+
+// Logger is the context based logger
+func Logger(ctx context.Context) Log {
+	// TODO(anup): Do we want users to access Logger(nil)?
```
alsamylkin: I would prefer to force them(and us) to use a background context.

ulog/log.go
```diff
@@ -99,12 +107,57 @@ func TestingLogger() (Log, *spy.Sink) {
...
+	if ctx == nil {
+		return logger()
```
alsamylkin: Looks like we have an allocation, each time we call Log()..

Could you add a TODO to improve it?

ulog/log_test.go
```diff
@@ -46,6 +49,38 @@ func TestZapSink(t *testing.T) {
 	assert.Contains(t, e.Fields, zap.String("And this is", "My field"))
 }
 
+func TestLogger_SetLogger(t *testing.T) {
+	SetLogger(logger())
```
alsamylkin: I think I am the only person who is using t.Parallel() :(

ulog/log.go
yutongp: what if type assertion failed?

ulog/log.go
```diff
@@ -99,12 +107,57 @@ func TestingLogger() (Log, *spy.Sink) {
...
+	}
+	log := ctx.Value(internal.ContextLogger)
+	if log != nil {
+		return log.(Log)
+	}
+	log = logger()
+	ctx = context.WithValue(ctx, internal.ContextLogger, log)
```
yutongp: this new context isn't used or passed to anywhere?

examples/keyvalue/server/main.go
yutongp: +1

ulog/log.go
sectioneight: +1

examples/simple/main.go
sectioneight: +1

ulog/log.go
anuptalwalkar: cool.. I'll add panic on nil

ulog/log.go
anuptalwalkar: added panic on nil from previous comment

ulog/log.go
anuptalwalkar: Its ok to panic here since SetLogger() is meant to be used in service setup only.

ulog/log.go
anuptalwalkar: removed

modules/uhttp/client/client_filters.go
anuptalwalkar: +1

PR #250, opened 2017-02-14
Fix config readme spelling mistake

sectioneight:


sectioneight:
@anuptalwalkar @Lupie

PR #251, opened 2017-02-14
remove host wrapper handler and add http response time metrics to metrics filter

yutongp:
remove host wrapper handler since host is in request.Context after context filter. And add http response time metrics to metrics filter


yutongp:
will add same test after #241 landed since #241 change some test functions. I don't want to deal with the rebase now..

yutongp:
test added

PR #252, opened 2017-02-14
use http.Error to return authFilter error

yutongp:
`http.Error` sets some common header by default

### DIFF
--- a/modules/uhttp/filters_test.go
+++ b/modules/uhttp/filters_test.go
@@ -106,7 +106,7 @@ func TestFilterChainFilters_AuthFailure(t *testing.T) {
 			authClient: host.AuthClient(),
 		}).Build(getNopHandler(host))
 	response := testServeHTTP(chain, host)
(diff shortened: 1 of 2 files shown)


modules/uhttp/filters_test.go
```diff
@@ -106,7 +106,7 @@ func TestFilterChainFilters_AuthFailure(t *testing.T) {
 			authClient: host.AuthClient(),
 		}).Build(getNopHandler(host))
 	response := testServeHTTP(chain, host)
-	assert.Contains(t, "Unauthorized access: Error authorizing the service", response.Body.String())
+	assert.Contains(t, response.Body.String(), "Unauthorized access: Error authorizing the service")
```
yutongp: this is because the right order is `str, sub`

alsamylkin: Hmm, since it is commutative, should we replace Contains with Equal?

PR #255, opened 2017-02-14
Use vendored richgo

sectioneight:
Use a stable version rather than go-getting from my private fork now that changes have been upstreamed. Verified locally.


sectioneight:
Self-merging to unblock contributors

PR #259, opened 2017-02-15
Add an observer that will trigger after service starts

alsamylkin:
This observer is useful if we want to call some function after service starts, e.g. yarpc client.

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -152,8 +152,8 @@ type niladicStart func()
 func (n niladicStart) OnInit(service Host) error      { return nil }
 func (n niladicStart) OnShutdown(reason Exit)         {}
 func (n niladicStart) OnCriticalError(err error) bool { return true }
-func (n niladicStart) OnStateChange(old State, new State) {
-	if old == Starting && new == Running {
+func (n niladicStart) OnStateChange(old State, curr State) {
+	if old == Starting && curr == Running {
 		n()
 	}
 }
--- a/service/service.go
+++ b/service/service.go
@@ -146,3 +146,19 @@ func New(options ...Option) (Owner, error) {
 
 	return svc, nil
 }
+
+type niladicStart func()
+
+func (n niladicStart) OnInit(service Host) error      { return nil }
+func (n niladicStart) OnShutdown(reason Exit)         {}
+func (n niladicStart) OnCriticalError(err error) bool { return true }
+func (n niladicStart) OnStateChange(old State, new State) {
+	if old == Starting && new == Running {
+		n()
+	}
+}
+
+// AfterStart will create an observer that will execute f() immediately after service starts.
+func AfterStart(f func()) Observer {
+	return niladicStart(f)
+}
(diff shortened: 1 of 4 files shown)


service/service.go
```diff
@@ -146,3 +146,19 @@ func New(options ...Option) (Owner, error) {
...
 	return svc, nil
 }
+
+type niladicStart func()
+
+func (n niladicStart) OnInit(service Host) error      { return nil }
+func (n niladicStart) OnShutdown(reason Exit)         {}
+func (n niladicStart) OnCriticalError(err error) bool { return true }
+func (n niladicStart) OnStateChange(old State, new State) {
+	if old == Starting && new == Running {
```
sectioneight: Don't we transition through Starting and then running? E.g. won't `n` be called twice? Also, don't shadow `new`

alsamylkin: If n called twice, the test should panic..

service/service_test.go
```diff
@@ -137,3 +139,22 @@ var validServiceConfig = map[string]interface{}{
...
+			loggingCore: loggingCore{
+				log: ulog.NopLogger,
+			},
+		},
+		observer: AfterStart(func() {
+			wg.Done()
+		}),
+	}
+
+	h.StartAsync()
```
madhuravi: Add tests to trigger all functions.

service/service.go
```diff
@@ -146,3 +146,19 @@ func New(options ...Option) (Owner, error) {
 
 	return svc, nil
 }
+
+type niladicStart func()
```
madhuravi: Is niladic a well known word? Slight preference for a simpler word for better readability.

alsamylkin: Got it from https://golang.org/pkg/sync/#Once.Do
Niladic - zero arguments,
Monadic - one argument,
Dyadic...

service/service_test.go
alsamylkin: This is just silly :)
I don't think we should test empty functions or functions always returning nils/true..

service/service.go
alsamylkin: We do transition from [Starting](https://github.com/uber-go/fx/blob/master/service/host.go#L235) to [Running](https://github.com/uber-go/fx/blob/master/service/host.go#L297) in start

service/service.go
```diff
@@ -146,3 +146,19 @@ func New(options ...Option) (Owner, error) {
...
+
+func (n niladicStart) OnInit(service Host) error      { return nil }
+func (n niladicStart) OnShutdown(reason Exit)         {}
+func (n niladicStart) OnCriticalError(err error) bool { return true }
+func (n niladicStart) OnStateChange(old State, curr State) {
+	if old == Starting && curr == Running {
```
sectioneight: Oooh, I read this as `||` for some reason which is why I was wondering about the double-trigger. 🤦‍♂️

service/service.go
glibsm: we tend to call things like this `nop`, nope?

sectioneight: It's not a noop though. A noop is a stub, this is a function with zero arguments that **does** get called.

PR #260, opened 2017-02-15
Remove host Logger() access in favor of ctx logger

anuptalwalkar:


sectioneight:
Do we lose any fidelity here? Is there context on the host logger not
present on ctx?

anuptalwalkar:
Nothing to lose here. We add module name to the logger today, which is now passed and available in the Logger(ctx), other than that additional fields should come from zap logger which is initialized during service setup and accessed on ulog.Logger(ctx)

service/host.go
```diff
@@ -278,9 +278,9 @@ func (s *host) start() Control {
 
 				s.shutdownMu.Unlock()
 				if _, err := s.shutdown(e, "", nil); err != nil {
-					s.Logger().Error("Unable to shut down modules", "initialError", e, "shutdownError", err)
+					s.log.Error("Unable to shut down modules", "initialError", e, "shutdownError", err)
```
alsamylkin: Looks like instead of logging here, we should wrap an error.

service/service_setup.go
```diff
@@ -45,7 +45,7 @@ func (svc *serviceCore) setupLogging() {
 		svc.log = logBuilder.WithConfiguration(svc.logConfig).Build()
 		ulog.SetLogger(svc.log)
 	} else {
-		svc.log.Debug("Using custom log provider due to service.WithLogger option")
+		ulog.Logger(context.Background()).Debug("Using custom log provider due to service.WithLogger option")
```
alsamylkin: What does this log message even mean?

modules/uhttp/filters_test.go
```diff
@@ -42,6 +41,7 @@ import (
 	"github.com/uber/jaeger-client-go/config"
 )
 
+<<<<<<< HEAD
```
anuptalwalkar: rebasing/cleaning up

service/service_setup.go
anuptalwalkar: We have a service.WithLogger option during the service setup (similar to WithModule, etc). If that is set, we don't update the service.Logger

anuptalwalkar: I can see this is a confusing debug message though, will update.

service/host.go
anuptalwalkar: This is a separate error on shutdown though, it needs to be logged?

ulog/log.go
```diff
@@ -140,11 +140,13 @@ func Logger(ctx context.Context) Log {
 }
 
 // NewLogContext sets the context with the context aware logger
-func NewLogContext(ctx context.Context) context.Context {
+func NewLogContext(ctx context.Context, log Log) context.Context {
```
madhuravi: why pass this in explicitly? just use ulog.Logger() here

service/host.go
```diff
@@ -95,7 +95,7 @@ func (s *host) IsRunning() bool {
 func (s *host) OnCriticalError(err error) {
 	shutdown := true
 	if s.observer == nil {
-		s.Logger().Warn(
+		s.log.Warn(
```
madhuravi: If we decide on using the ulog.Logger, maybe we should be consistent and use it everywhere in our code. We don't have access to context here but so do service owners in some cases and we still expect them to inject a context.

anuptalwalkar: I was initially thinking to use s.log internally, only for setup purpose, but you are right that we should be consistent. I'll remove log object on the host itself.

ulog/log.go
anuptalwalkar: This is used when middleware sets up Log object on the context with module specific fields.

madhuravi: mild preference for method be to ContextWithLogger

modules/uhttp/http.go
```diff
@@ -114,20 +115,22 @@ func newModule(
...
 	handlers := addHealth(getHandlers(mi.Host))
 
+	log := ulog.Logger(context.Background()).With("moduleName", mi.Name)
+
 	// TODO (madhu): Add other middleware - logging, metrics.
 	module := &Module{
 		ModuleBase: *modules.NewModuleBase(mi.Name, mi.Host, []string{}),
 		handlers:   handlers,
-		fcb:        defaultFilterChainBuilder(mi.Host),
+		fcb:        defaultFilterChainBuilder(log, mi.Host.AuthClient()),
```
madhuravi: why not set ulog.SetLogger and access ulog.Logger from contextFilter?

ulog/log.go
```diff
@@ -140,11 +140,13 @@ func Logger(ctx context.Context) Log {
...
 	if ctx == nil {
 		ctx = context.Background()
 	}
-
+	if log != nil {
+		return context.WithValue(ctx, internal.ContextLogger, log)
+	}
 	return context.WithValue(ctx, internal.ContextLogger, logger())
 }
```
madhuravi: likewise, WithTracingAware would be ContextWithTraceLogger, along those lines.

ulog/log.go
glibsm: more than mind here

modules/uhttp/http.go
anuptalwalkar: `SetLogger` updates the global logger, we don't want to set it to be uhttp specific.

modules/uhttp/http.go
```diff
@@ -114,20 +115,22 @@ func newModule(
...
 
 	err := module.Host().Config().Get(getConfigKey(mi.Name)).PopulateStruct(cfg)
 	if err != nil {
-		module.Host().Logger().Error("Error loading http module configuration", "error", err)
+		log.Error("Error loading http module configuration", "error", err)
 	}
 	module.config = *cfg
 
-	module.log = module.Host().Logger().With("moduleName", mi.Name)
+	module.log = log
```
madhuravi: Is this used anywhere?

service/service_setup.go
```diff
@@ -34,19 +34,16 @@ import (
 	"github.com/pkg/errors"
 )
 
-func (svc *serviceCore) setupLogging() {
-	if svc.log == nil {
-		err := svc.configProvider.Get("logging").PopulateStruct(&svc.logConfig)
-		if err != nil {
-			ulog.Logger(context.Background()).Info("Logging configuration is not provided, setting to default logger", "error", err)
-		}
+var _simpleCtx = context.Background()
```
alsamylkin: nit, it is actually empty context, probably should name _emptyCtx?

alsamylkin: Although, since it is used only for logger, can we have an _svcLogger = ulog.Logger(_simpleCtx)?

modules/uhttp/http.go
anuptalwalkar: yes, logging on Start/Stop module

service/service_setup.go
anuptalwalkar: I think its better that we use `ulog.Logger(_simpleCtx)` in service. If we have context based objects in the future, we will have use of _simpleContext.

ulog/log.go
anuptalwalkar: updated both.

PR #262, opened 2017-02-15
use sub tests for uhttp/filter_test

yutongp:
- run tests which use same host config in parallel
- tear down in filter tests to make sure tests are independent. tests should not depend on one or other to set up global state (here is the http metrics)

### DIFF
--- a/modules/uhttp/filters_test.go
+++ b/modules/uhttp/filters_test.go
@@ -42,14 +42,57 @@ import (
 	"github.com/uber/jaeger-client-go/config"
 )
 
-func TestFilterChain(t *testing.T) {
+func TestDefaultFiltersWithNopHost(t *testing.T) {
+	// setup
 	host := service.NopHost()
+	stats.SetupHTTPMetrics(host.Metrics())
+
+	t.Run("run parallel", func(t *testing.T) {
+		t.Run("testFilterChain", func(t *testing.T) {
+			testFilterChain(t, host)
+		})
+		t.Run("testFilterChainFilters", func(t *testing.T) {
+			testFilterChainFilters(t, host)
+		})
+		t.Run("testPanicFilter", func(t *testing.T) {
+			testPanicFilter(t, host)
+		})
+	})
+
+	// teardown
+	httpMetricsTeardown()
+}
+
+func TestDefaultFiltersWithNopHostAuthFailure(t *testing.T) {
+	// setup
+	host := service.NopHostAuthFailure()
+	stats.SetupHTTPMetrics(host.Metrics())
+
+	t.Run("run parallel", func(t *testing.T) {
+		t.Run("testFilterChainFilters_AuthFailure", func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


modules/uhttp/filters_test.go
```diff
@@ -42,14 +42,58 @@ import (
 	"github.com/uber/jaeger-client-go/config"
 )
 
-func TestFilterChain(t *testing.T) {
+func TestDefaultFiltersWithNopHost(t *testing.T) {
+	// setup
 	host := service.NopHost()
+	stats.SetupHTTPMetrics(host.Metrics())
+
+	t.Run("run parallel", func(t *testing.T) {
```
alsamylkin: How is it parallel? It looks like 3 sequential runs, I think t.Run is sync..
May be I am missing something, but it looks like you've forgot to set t.Parallel()?

yutongp: right, I should add `t.Parallel()` in closure

alsamylkin: Could you set t.Parallel() in the beginning(line 46)?
And testFilter... functions can be easily combined in a map and iterated over:
```
tests := map[string]func(*testing.T, service.Host){
"testFilterChain": testFilterChain,...}
for _, test := range tests {
    test.Run(t, host)
}
```

modules/uhttp/filters_test.go
```diff
@@ -42,14 +42,67 @@ import (
...
+
+	// teardown
+	httpMetricsTeardown()
+}
+
+func TestDefaultFiltersWithNopHostConfigured(t *testing.T) {
+	// this test's sub tests cannot run parallel
+	// and they need to build host by theirselves
+	t.Run("testTracingFilterWithLogs", func(t *testing.T) {
+		t.Parallel()
```
alsamylkin: There is no point of making it a subtest and marking for parallel run
Same as above..

modules/uhttp/filters_test.go
```diff
@@ -42,14 +42,67 @@ import (
...
+			testPanicFilter(t, host)
+		})
+		t.Run("testMetricsFilter", func(t *testing.T) {
+			t.Parallel()
+			testMetricsFilter(t, host)
+		})
+	})
+
+	// teardown
+	httpMetricsTeardown()
```
glibsm: defer?

modules/uhttp/filters_test.go
yutongp: we cannot set t.Parallel at line 46, since `stats.SetupHTTPMetrics(host.Metrics())` set global, no test should run parallel when they are going to touch this value. so `TestDefaultFiltersWithNopHost` `TestDefaultFiltersWithNopHostAuthFailure` and `TestDefaultFiltersWithNopHostConfigured` have to run one after another

modules/uhttp/filters_test.go
yutongp: I put it here so when follow devs add more tests, they will follow the style and add t.Parallel in their sub tests

PR #263, opened 2017-02-15
remove []Filter in uhttp.New param and add uhttp.WithFilters func

yutongp:
follow `rpc` packge, make `WithFilters` returns `modules.Option`, so we don't need to pass `[]Filter` as an additional param in `uhttp.New`.
No more `uhttp.New(hander, nil)` :D

### DIFF
--- a/modules/uhttp/http_options.go
+++ b/modules/uhttp/http_options.go
@@ -1,0 +1,52 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package uhttp
+
+import (
+	"go.uber.org/fx/modules"
+	"go.uber.org/fx/service"
+)
+
+const _filterKey = "uhttpFilter"
+
+// WithFilter adds Filter to uhttp Module that will be applied to all incoming http requests.
+func WithFilter(fs ...Filter) modules.Option {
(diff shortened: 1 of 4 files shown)


modules/uhttp/http_options.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"go.uber.org/fx/modules"
+	"go.uber.org/fx/service"
+)
+
+const _filterKey = "uhttpFilter"
+
+// WithFilter adds Filter to uhttp Module that will be applied to all incoming http requests.
+func WithFilter(fs ...Filter) modules.Option {
```
sectioneight: should this be `WithFilters` if it accepts multiple filters?

yutongp: I 100% agree. However: https://godoc.org/go.uber.org/yarpc#UnaryInboundMiddleware, https://godoc.org/go.uber.org/fx/modules/rpc#WithInboundMiddleware are all singular...
So I thought we may follow the same style here

sectioneight:
:shrug: we have WithFilters() and WithOptions and WithModules, I think it's clearer, but you're correct that it's inconsistent

modules/uhttp/http_options.go
```diff
@@ -0,0 +1,52 @@
...
+			mci.Items = make(map[string]interface{})
+		}
+		mci.Items[_filterKey] = filters
+
+		return nil
+	}
+}
+
+func filtersFromCreateInfo(mci service.ModuleCreateInfo) []Filter {
+	items, ok := mci.Items[_filterKey]
```
alsamylkin:
```golang
if items, ok := mci.Items[_filterKey]; ok {
    return items.([]Filter)
}

return nil
```
?

sectioneight: Yes, this is more idiomatic.

modules/uhttp/http_options.go
```diff
@@ -0,0 +1,52 @@
...
+
+// WithFilters adds Filters to uhttp Module that will be applied to all incoming http requests.
+func WithFilters(fs ...Filter) modules.Option {
+	return func(mci *service.ModuleCreateInfo) error {
+		filters := filtersFromCreateInfo(*mci)
+		filters = append(filters, fs...)
+		if mci.Items == nil {
+			mci.Items = make(map[string]interface{})
+		}
+		mci.Items[_filterKey] = filters
```
sectioneight: Hmm, this looks like it doesn't respect calling `WithFilters` multiple times and keeping old results. Is that intentional / am I misreading this? IMHO you should be able to apply the same option multiple times and if the contents are additive, they should add, rather than override.

yutongp: not sure I got it. currently I first got old filters from ModuleCreateInfo and add the new ones in:
```
filters := filtersFromCreateInfo(*mci)
filters = append(filters, fs...)
```
I added a new test `TestWithFiltersTwice`. hopefully that what you are concerning.

modules/uhttp/http_options.go
yutongp: i am slightly prefer "error return early, normal return later". easier for me to quick go through the code. no big preference though.

modules/uhttp/http_options.go
sectioneight: Derp, just me being dumb.

modules/uhttp/http_options.go
sectioneight: Either way is fine. Let's not bikeshed on this.

alsamylkin: This is not an error, but "comma ok": https://golang.org/doc/effective_go.html#maps

sectioneight: I'm going to merge as-is. Alex, if you'd like to comma-OK it, that would be my preference as well, but this is not a big enough deal to spend any more PR cycles on.

yutongp: Thanks for pointing out. comma-OK makes sense. Compare to the API changes, that is small internal change, I can just pull another quick PR

alsamylkin: It is not a big deal :)

PR #264, opened 2017-02-15
Pin zap to current lock SHA to unblock

sectioneight:
Currently zap is underoing a heavy refator, making master unstable and
not compatible with our current usage.


sectioneight:
We're still using spy, this pins zap to a version that still uses spy

akshayjshah:
Sure, sounds good.

PR #268, opened 2017-02-16
make uhttp Config Debug use bool instead of *bool

yutongp:
We don't need to use pointer on basic type when we don't have to.
use  ```Debug   bool          `yaml:"debug" default:"true"` ```


modules/uhttp/http.go
```diff
@@ -78,7 +78,7 @@ type Config struct {
 	modules.ModuleConfig
 	Port    int           `yaml:"port"`
 	Timeout time.Duration `yaml:"timeout"`
-	Debug   *bool         `yaml:"debug"`
+	Debug   bool          `yaml:"debug"`
```
yutongp: I can change it to Debug bool `yaml:"debug" default:"true"` if we prefer set default value to be true

sectioneight:
Where is the code that reads `default`? If that's working, then this is definitely a positive change

modules/uhttp/http.go
glibsm: FYI, there are three distinct states for a bool when serializing from yaml: `true`, `false` and `did not specify`, that's why we use `*bool` as opposed to just bool.

yutongp: Thanks for the context. I believe there are use cases for that. But when we are not using `did not specify` has different behavior with `false`, let's just use basic `bool` instead of `*bool`.

yutongp:
@sectioneight 
http.go line 124:
```
 	err := module.Host().Config().Get(getConfigKey(mi.Name)).PopulateStruct(cfg)
```
If Host().Config() is a YAMLProvider(), when the value is not set, we will use the default we have.

modules/uhttp/http.go
```diff
@@ -154,7 +154,7 @@ func (m *Module) Start(ready chan<- struct{}) <-chan error {
 		router.Handle(h.Path, m.fcb.Build(h.Handler))
 	}
 
-	if m.config.Debug == nil || *m.config.Debug {
+	if m.config.Debug {
```
alsamylkin: I prefer to keep pprof true by default, unless users really know what they are doing.

yutongp: changed to` default: "true"`

modules/uhttp/http.go
```diff
@@ -78,7 +78,7 @@ type Config struct {
 	modules.ModuleConfig
 	Port    int           `yaml:"port"`
 	Timeout time.Duration `yaml:"timeout"`
-	Debug   *bool         `yaml:"debug"`
+	Debug   bool          `yaml:"debug" default:"true"`
```
alsamylkin: This default is going to work iff config comes from yaml. Otherwise, I think, it will be false.

alsamylkin: Speaking of which, I think we should remove yaml tags above anyway. Nit, feel free to ignore.

yutongp: #269 i can remove that `if !cv.HasValue{...}` so default tag can be populate when value is empty

madhuravi: That seems dangerous if default values can differ with config providers. We should stick to one.

alsamylkin: Actually, I was wrong: it will work on all providers, because we iterate over a struct to get values and take defaults from the object itself.

Issue #171, opened 2017-01-18
Rename interceptor/filter terminology


breerly:
Before going 1.0, YARPC renamed the terms interceptor and filter to:

```
inbound middleware
outbound middleware
```

Why? We found that in introducing the interceptor/filter concept, that we were having to say "inbound/outbound middleware" anyways. This is because, without doing that, people were confused and didn't understand what these things did right away.

Of course, inbound and outbound middleware is much more verbose, which is why we started out with interceptor/filter in the first place, and because these are relatively common terms. In the end, we are happy we made the tradeoff, because the verbosity only occurs while setting up the framework, and not during request time.

The result has been great - users "get it" right away because the term is self describing and well understood. No need to define the term every time it's being used.

What do you say about using these terms in UberFx as well, so that both projects share terminology? I imagine this can only help our customers, where as having separate terms would only conflate the space.

cc @abhinav @kriskowal @willhug @bombela @blampe @prashantv @akshayjshah

glibsm:
linkerino #170

breerly:
One more note - filters and interceptors *are* pretty common in other frameworks, except that we found that in practice their direction and semantics are not as well understood as middleware. Some camps of developers get it (we certainly did), but we found that our customers at large had a much easier time with the "middleware" terminology.

glibsm:
@breerly we've discussed this internally as a team and agree with your findings. We'll take it on to move away from interceptor/filter and clean up the API to be consistent with yarpc.

breerly:
Oh nice - this is good stuff 👍

breerly:
Looks like this is resolved now?

glibsm:
@breerly Negative. Our http module is still called a Filter, just haven't gotten to it yet.

yutongp:
close #271

glibsm:
thank you @yutongp

PR #271, opened 2017-02-16
bye filter and hello middleware

yutongp:
refer #171 

~~I didn't end up with adding `Inbound` and `Outbound` since HTTP client is in a seperate package. Adding `Inbound` and `Outbound` doesn't seem to add much for distinguish one from another.~~

But I agree if we want to share the same terminology we may add them as well. Let me know if `Inbound` and `Outbound` are good to have.
It is not hard to make that change with GoRename.

@breerly @glibsm 

update:
use `Inbound` `Outbound`

### DIFF
--- a/examples/simple/handlers.go
+++ b/examples/simple/handlers.go
@@ -47,9 +47,9 @@ func registerHTTPers(service service.Host) []uhttp.RouteHandler {
 	}
 }
 
-type simpleMiddleware struct{}
+type simpleInboundMiddleware struct{}
 
-func (simpleMiddleware) Handle(w http.ResponseWriter, r *http.Request, next http.Handler) {
-	io.WriteString(w, "Going through simpleMiddleware")
+func (simpleInboundMiddleware) Handle(w http.ResponseWriter, r *http.Request, next http.Handler) {
+	io.WriteString(w, "Going through simpleInboundMiddleware")
 	next.ServeHTTP(w, r)
 }
--- a/examples/simple/handlers.go
+++ b/examples/simple/handlers.go
@@ -47,9 +47,9 @@ func registerHTTPers(service service.Host) []uhttp.RouteHandler {
 	}
 }
 
-type simpleFilter struct{}
+type simpleMiddleware struct{}
 
-func (simpleFilter) Apply(w http.ResponseWriter, r *http.Request, next http.Handler) {
-	io.WriteString(w, "Going through simpleFilter")
+func (simpleMiddleware) Handle(w http.ResponseWriter, r *http.Request, next http.Handler) {
+	io.WriteString(w, "Going through simpleMiddleware")
 	next.ServeHTTP(w, r)
 }
--- a/examples/simple/main.go
+++ b/examples/simple/main.go
@@ -29,7 +29,7 @@ import (
 
 func main() {
 	svc, err := service.WithModules(
-		uhttp.New(registerHTTPers, uhttp.WithMiddlewares(simpleMiddleware{})),
+		uhttp.New(registerHTTPers, uhttp.WithInboundMiddlewares(simpleInboundMiddleware{})),
 	).Build()
 	if err != nil {
 		log.Fatal("Unable to initialize service", "error", err)
--- a/examples/simple/main.go
+++ b/examples/simple/main.go
@@ -29,7 +29,7 @@ import (
 
 func main() {
 	svc, err := service.WithModules(
-		uhttp.New(registerHTTPers, uhttp.WithInboundMiddlewares(simpleInboundMiddleware{})),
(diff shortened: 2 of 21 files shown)


examples/simple/main.go
```diff
@@ -29,7 +29,7 @@ import (
 
 func main() {
 	svc, err := service.WithModules(
-		uhttp.New(registerHTTPers, []uhttp.Filter{simpleFilter{}}),
+		uhttp.New(registerHTTPers),
```
yutongp: I broke this example in #263 ...

examples/simple/handlers.go
```diff
@@ -47,9 +47,9 @@ func registerHTTPers(service service.Host) []uhttp.RouteHandler {
 	}
 }
 
-type simpleFilter struct{}
+type simpleMiddleware struct{}
 
-func (simpleFilter) Apply(w http.ResponseWriter, r *http.Request, next http.Handler) {
-	io.WriteString(w, "Going through simpleFilter")
+func (simpleMiddleware) Apply(w http.ResponseWriter, r *http.Request, next http.Handler) {
```
sectioneight: what's the point of this being in the example if it's not used? shouldn't we change simple/main.go to use this as an option?

yutongp: #273

examples/simple/main.go
yutongp: Not a good idea to change this issue in this PR. opened another one: #273

yutongp:
@sectioneight that sounds good to me. I care more on settle down some thing we have consensus before v1 compare to which one we actually pick.

sectioneight:
cc @breerly because I'm pretty certain he's spent a lot of time thinking about this problem :)

yutongp:
Thought about `Inbound` and `Outbound` again. Since our main goal is share terminology, I feel it make sense to have them.

modules/uhttp/client/client_middlewares_benchmark_test.go
```diff
@@ -34,26 +34,26 @@ import (
 	jconfig "github.com/uber/jaeger-client-go/config"
 )
 
-// BenchmarkClientFilters/empty-8         	  500000	      3517 ns/op	     256 B/op	       2 allocs/op
-// BenchmarkClientFilters/tracing-8       	   20000	     64421 ns/op	    1729 B/op	      29 allocs/op
-// BenchmarkClientFilters/auth-8          	   50000	     36574 ns/op	     728 B/op	      16 allocs/op
-// BenchmarkClientFilters/default-8       	   10000	    104374 ns/op	    2275 B/op	      43 allocs/op
-func BenchmarkClientFilters(b *testing.B) {
+// BenchmarkClientMiddlewares/empty-8          100000000           10.8 ns/op         0 B/op          0 allocs/op
```
alsamylkin: How renaming reduced allocations? Did we forget to update it last time?

modules/uhttp/README.md
```diff
@@ -39,15 +39,15 @@ func registerHTTP(service service.Host) []uhttp.RouteHandler {
 }
 ```
 
-HTTP handlers are set up with filters that inject tracing, authentication information etc. into the
+HTTP handlers are set up with inbound middlewares that inject tracing, authentication information etc. into the
```
alsamylkin: Nit: plural of middleware is still middleware..

modules/uhttp/middlewarechain_builder.go
```diff
@@ -26,58 +26,58 @@ import (
...
+	return inboundMiddlewareChainBuilder{
 		Host: host,
 	}
 }
 
-func (f filterChainBuilder) AddFilters(filters ...Filter) filterChainBuilder {
-	for _, filter := range filters {
-		f.filters = append(f.filters, filter)
+func (m inboundMiddlewareChainBuilder) AddMiddlewares(middlewares ...InboundMiddleware) inboundMiddlewareChainBuilder {
+	for _, middleware := range middlewares {
```
alsamylkin: Nit, just:
```golang
m.middleware = append(m.middleware, middleware...)
```

modules/uhttp/client/client_middlewares_benchmark_test.go
yutongp: #239 reduced allocations and I forgot to update here. The same benchmark result is written is 3 places like: README.MD, doc.go and client_middlewares_benchmark_test.go

modules/uhttp/README.md
yutongp: I see! that is why `rpc.WithInboundMiddleware` is singular! Should I change it? it will effect `inboundMiddlewareChainBuilder.AddMiddlewares` and `uhttp.WithInboundMiddlewares`

alsamylkin: That would be awesome!

yutongp:
just did a long rebase...

Issue #269, opened 2017-02-16
config: populate default when value is empty


yutongp:
```
func (cv Value) PopulateStruct(target interface{}) error {
	if !cv.HasValue() {
		return nil
	}

	_, err := cv.valueStruct(cv.key, target)

	return err
}
```
right now when Value has no value, it is a noop. I think we should populate value in `default` tag instead.

alsamylkin:
The change was merged.

PR #272, opened 2017-02-16
set value to default tag when config is empty

yutongp:
refer #269 

When config Value is nil, default tag can be populate to object.

I know @alsamylkin has some thoughts on this. Plz take a look

### DIFF
--- a/config/value.go
+++ b/config/value.go
@@ -393,12 +393,7 @@ func convertValue(value interface{}, targetType reflect.Type) (interface{}, erro
 
 // PopulateStruct fills in a struct from configuration
 func (cv Value) PopulateStruct(target interface{}) error {
-	if !cv.HasValue() {
-		return nil
-	}
-
 	_, err := cv.valueStruct(cv.key, target)
-
 	return err
 }
 
@@ -519,18 +514,16 @@ func (cv Value) valueStruct(key string, target interface{}) (interface{}, error)
 		case bucketObject:
 			ntt := derefType(fieldType)
 			newTarget := reflect.New(ntt)
-			if v2 := global.Get(childKey); v2.HasValue() {
-
-				if err := v2.PopulateStruct(newTarget.Interface()); err != nil {
-					return nil, errors.Wrap(err, "unable to populate struct of object target")
-				}
-				// if the target is not a pointer, deref the value
-				// for copy semantics
-				if fieldType.Kind() != reflect.Ptr {
-					newTarget = newTarget.Elem()
-				}
-				fieldValue.Set(newTarget)
+			v2 := global.Get(childKey)
+			if err := v2.PopulateStruct(newTarget.Interface()); err != nil {
+				return nil, errors.Wrap(err, "unable to populate struct of object target")
 			}
+			// if the target is not a pointer, deref the value
+			// for copy semantics
+			if fieldType.Kind() != reflect.Ptr {
+				newTarget = newTarget.Elem()
+			}
+			fieldValue.Set(newTarget)
 		case bucketArray:
 			// TODO(alsam) fix array type DRI-12.
 		case bucketSlice:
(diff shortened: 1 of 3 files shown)


yutongp:
I totally agree "unset" should be supported. The case I am thinking about is like this:
say if the yaml is defined like:
```
config1:
  sub1: "string"
```
user config struct is:
```
type Config struct {
   Config1 `yaml: "config1"`
   Config2 Inner `yaml: "config2" default`
}

type Inner struct {
  MyString string `yaml: "my_string" default: "a_empty_string"`
}
```

if user do:
```
var config2 Inner
NewYAMLProvider.Get("Config2").PopulateStruct(&config2)
```
right now, `config2.MyString` will have a value of `""` after populated which I think is very weird since it have a default tag: `default: "a_empty_string"`.

This PR is intend to make default tag work more consistent. Unset will and should be continually supported

config/value.go
```diff
@@ -387,12 +387,7 @@ func convertValue(value interface{}, targetType reflect.Type) (interface{}, erro
 
 // PopulateStruct fills in a struct from configuration
 func (cv Value) PopulateStruct(target interface{}) error {
-	if !cv.HasValue() {
-		return nil
-	}
-
 	_, err := cv.valueStruct(cv.key, target)
```
anuptalwalkar: Not sure if we need `cv.valueStruct()` method now since first parameter returned on valueStruct is never used.  We can just have this method signature for PopulateStruct on line 401 below that would just return `error`.

config/yaml_test.go
```diff
@@ -351,3 +351,27 @@ func TestMerge(t *testing.T) {
 		})
 	}
 }
+
+func TestNilYAMLProviderSetDefaultTagValue(t *testing.T) {
+	type Inner struct {
+		Set bool `yaml:"set" default:"true"`
```
anuptalwalkar: wait. don't we want default `false`?

yutongp: I think I am confused with `default` tag. My thought is `default` tag is for user to set a custom default value user like to an entity and auto apply it when that entity is not present.
The default golang behavior is set bool to `false` if entity is not present, then we don't need to have a `default` tag at all.
I think I may missed some context.

config/yaml_test.go
```diff
@@ -351,3 +351,27 @@ func TestMerge(t *testing.T) {
...
+	}
+	data := struct {
+		ID0 int             `yaml:"id0" default:"10"`
+		ID1 string          `yaml:"id1" default:"string"`
+		ID2 Inner           `yaml:"id2"`
+		ID3 []Inner         `yaml:"id3"`
+		ID4 map[Inner]Inner `yaml:"id4"`
+		ID5 *Inner          `yaml:"id5"`
```
alsamylkin: Could you add a test e.g. 
```golang
ID6 [6]Inner `yaml:"id6"`
```
?

yutongp: sure. and I already know that will break before I even add it... Thanks for pointing that out.
Will add `[6]Inner` and `[7]*Inner`

config/yaml_test.go
```diff
@@ -351,3 +351,27 @@ func TestMerge(t *testing.T) {
...
+	}{}
+
+	p := NewYAMLProviderFromBytes(nil)
```
alsamylkin: If we have two providers/yamls, how defaults will affect merge top default override a bottom explicit?

yutongp: I think there will be only one place default is defined where is on the struct. And default will only apply on empty non-ptr entities.

say we have base.yaml:
```
abc:
  str: "base"
  int: 1
```
production.yaml:
```
abc: 
  str: "prod"
```
if the reciver struct defined like this:
```golang
type Config1 struct {
  Str string `yaml:"str" default:"nope"`
  Int int `yaml:"int" default: "0"`
  Bool bool `yaml:"bool" default:"true"`
  BoolPtr *bool `yaml:"bool_ptr"`
}

type Config struct {
  ABC Config1 `yaml:"abc"`
}

var abc Config1
NewYAMLProvider("base.yaml","production.yaml").Get("abc").Populate(&abc)
```
the end result will be:
``` golang
{
  Str: "prod",
  Int: 1,
  Bool: true,
  BoolPtr: nil
}
```

alsamylkin: Lets add a test for this case.

config/value.go
yutongp: agree. will put a seperate PR to make that clear.

yutongp:
any update for this PR?

PR #273, opened 2017-02-16
fix broken simple example

yutongp:


sectioneight:
https://jira.uberinternal.com/browse/GFM-370

PR #274, opened 2017-02-16
Test examples as part of CI run

sectioneight:
Technically this should probably be `/...` but seems like overkill for now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -142,6 +142,7 @@ examples:
 	@$(call label,Building examples)
 	@echo
 	$(ECHO_V)$(MAKE) -C examples/keyvalue ECHO_V=$(ECHO_V)
+	$(ECHO_V)go test ./examples/simple
 
 .PHONY: vendor
 vendor:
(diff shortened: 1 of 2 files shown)


examples/simple/main.go
```diff
@@ -29,9 +29,8 @@ import (
 
 func main() {
 	svc, err := service.WithModules(
-		uhttp.New(registerHTTPers, []uhttp.Filter{simpleFilter{}}),
+		uhttp.New(registerHTTPers),
```
yutongp: is this change intended?

sectioneight: Sorta, I meant to include it to get the build passing but now I merged your real fix and rebased.

Makefile
```diff
@@ -142,6 +142,7 @@ examples:
 	@$(call label,Building examples)
 	@echo
 	$(ECHO_V)$(MAKE) -C examples/keyvalue ECHO_V=$(ECHO_V)
+	$(ECHO_V)go test ./examples/simple
```
madhuravi: why not follow the same format as keyvalue? shouldn't we run the same commands to test both?

sectioneight:
A makefile seems overkill to run a single test command. The make rules in
key value are much more complex. I can move this command to a separate
Makefile, but that feels like overkill to me right now.

sectioneight:
All tests passed after re-triggering in Travis

PR #275, opened 2017-02-16
Module rewrite

peter-edge:


modules/rpc/yarpc_options.go
```diff
@@ -32,28 +32,29 @@ const (
...
-func WithInboundMiddleware(i ...middleware.UnaryInbound) modules.Option {
-	return func(mci *service.ModuleCreateInfo) error {
-		inboundMiddleware := inboundMiddlewareFromCreateInfo(*mci)
-		inboundMiddleware = append(inboundMiddleware, i...)
-		mci.Items[_interceptorKey] = inboundMiddleware
-
-		return nil
-	}
+func WithInboundMiddleware(i ...middleware.UnaryInbound) service.ModuleOption {
+	return service.WithModuleItem(_interceptorKey, func(existing interface{}) interface{} {
```
yutongp: I like this wrapping on write to module item.
maybe a similar wrapping for read as well?

service/module.go
```diff
@@ -20,27 +20,187 @@
...
+		return nil
+	}
+}
+
+// WithModuleItem adds the value to the Module. If there is an existing value,
+// it will be passed as the argument to the population function.
+func WithModuleItem(key string, f func(interface{})interface{}) ModuleOption {
+	return func(o *moduleOption) error {
+		value, ok := o.items[key]
+		if ok {
```
yutongp: nit:
```
if value, ok := o.items[key]; ok {
  o.items[key] = f(value)
  return nil
}
o.items[key] = f(nil)
retur nil
```

service/host.go
```diff
@@ -323,29 +304,34 @@ func (s *host) Stop(reason string, exitCode int) error {
 	return err
 }
 
-func (s *host) startModules() map[Module]error {
-	results := map[Module]error{}
+func (s *host) startModules() []error {
```
dmcaulay: so this only happens once, right? how much time are we saving be starting these in parallel?

service/host_mock.go
```diff
@@ -45,9 +50,16 @@ func NopHostAuthFailure() Host {
...
+func nopHostConfigured(client auth.Client, logger ulog.Log, tracer opentracing.Tracer, configProvider config.Provider) Host {
+	if configProvider == nil {
+		configProvider = config.NewStaticProvider(nil)
+	}
 	return &serviceCore{
 		authClient:     client,
-		configProvider: config.NewStaticProvider(nil),
+		configProvider: configProvider,
 		standardConfig: serviceConfig{
 			Name:        "dummy",
```
dmcaulay: can we read the name from our config provider? i need this for some integration tests.

service/module.go
```diff
@@ -20,27 +20,187 @@
...
+		name = moduleOptions.Name
+	}
+	return &moduleWrapper{module: module, moduleInfo: moduleInfo, name: name}, nil
+}
+
+func (m *moduleWrapper) Name() string {
+	return m.name
+}
+
+func (m *moduleWrapper) Start() error {
```
dmcaulay: can we remove locking if we single thread start/stop?

service/module.go
```diff
@@ -20,27 +20,187 @@
...
+	return &moduleInfo{host, roles, items}
+}
+
+// TODO(pedge): what about the Host's roles?
+func (mi *moduleInfo) Roles() []string {
+	return mi.roles
+}
+
+func (mi *moduleInfo) Metrics() tally.Scope {
+	return mi.Host.Metrics().SubScope("modules").SubScope(mi.name)
```
dmcaulay: is it more efficient to memoize this scope?

service/module.go
```diff
@@ -20,27 +20,187 @@
...
-	Roles []string
-	Items map[string]interface{}
-	Host  Host
+// ModuleInfo is the information passed to a Module on creation.
+// This can be stored inside the module for use.
+type ModuleInfo interface {
+	Host
+	Items() map[string]interface{}
+	// TODO(pedge)
+	ConfigValue() config.Value
```
peter-edge: Ya I'm going to remove this for now

service/module.go
peter-edge: ya whoops, sorry

service/host.go
peter-edge: No idea, but this PR is just addressing the external interface, I think we can iterate on implementation some other time.

service/module.go
peter-edge: Ya we can, but I really think the locking is "proper".

service/module.go
peter-edge: Yes, I think I have a TODO somewhere in this file to do that

modules/rpc/yarpc.go
```diff
@@ -65,7 +60,8 @@ var (
 	// that are stored together to create a shared dispatcher.
 	// The YARPC team advised it to be a 'singleton' to control
 	// the lifecycle of all of the in/out bound traffic.
-	_controller dispatcherController
+	_controller        dispatcherController
+	_controllerRunning bool
```
alsamylkin: This global var is definitely not needed

modules/rpc/thrift.go
```diff
@@ -33,24 +32,23 @@ import (
 var _setupMu sync.Mutex
 
 // CreateThriftServiceFunc creates a Thrift service from a service host
+// TODO(pedge): pass all module info?
 type CreateThriftServiceFunc func(svc service.Host) ([]transport.Procedure, error)
 
 // ThriftModule creates a Thrift Module from a service func
-func ThriftModule(hookup CreateThriftServiceFunc, options ...modules.Option) service.ModuleCreateFunc {
```
alsamylkin: Why are you removing options?

modules/task/backend.go
```diff
@@ -101,22 +93,20 @@ func (b *inMemBackend) Name() string {
...
+func (b *inMemBackend) Start() error {
+	go b.consumeFromQueue()
+	return nil
 }
 
-func (b *inMemBackend) consumeFromQueue(errorCh chan error) {
+func (b *inMemBackend) consumeFromQueue() {
 	for msg := range b.bufQueue {
-		errorCh <- Run(context.Background(), msg)
+		// TODO(pedge): this was effectively not being handled and was a bug
```
alsamylkin: It is not a bug, but a feature :) Seriously, this is a way to communicate with a sender about errors in function executions.

service/host.go
dmcaulay: sg

service/module.go
dmcaulay: okay. fine with it.

modules/task/backend.go
peter-edge: If you look into how this channel was read, this error would have never been read with the current service code

modules/rpc/thrift.go
peter-edge: They are now on the service builder only

modules/rpc/yarpc.go
peter-edge: Removed

service/host_mock.go
peter-edge: We can, but probably out of scope for this PR

modules/rpc/yarpc_options.go
peter-edge: What do you mean exactly? Wrap `Items()` access?

peter-edge:
I've updated for the latest changes. Before continuing, if everyone could give a quick lookover of what has happened so far, that would be great - the next step is going to be a lot of test refactoring and documentation updates, and before doing that, it would be nice to have a general "OK" from what I have so far.

@sectioneight @dmcaulay @alsamylkin @madhuravi @yutongp @glibsm

service/builder.go
```diff
@@ -22,22 +22,26 @@ package service
 
 import "github.com/pkg/errors"
 
-// WithModules is a helper to create a service without any options
-func WithModules(modules ...ModuleCreateFunc) *Builder {
```
madhuravi: Why? I would prefer keeping WithModules.
1.
service.WithModule(
  rpc,
  uhttp,
  task
)

2.
service.WithModule(uhttp)
  .WithModule(rpc)
  .WithModule(task)

1 (the existing way) seems simpler and nicer than 2.

peter-edge:
The reason is because now WithModule takes options instead of the
individual derivative functions, there are other ways to handle this but
this seemed the easiest.

modules/task/backend.go
madhuravi: We do want to propagate these errors up so nothing to change here. Need to not drop them on the floor in module Start where this gets called.

modules/task/doc.go
```diff
@@ -48,7 +48,8 @@
 //   )
```
madhuravi: This file shouldn't be included?

service/host.go
```diff
@@ -447,3 +436,7 @@ func loadInstanceConfig(cfg config.Provider, key string, instance interface{}) b
 	}
 	return false
 }
+
+func logger() ulog.Log {
```
madhuravi: Make this a var instead? That's how we handle other such globals.

service/host.go
```diff
@@ -184,32 +174,24 @@ func (s *manager) shutdown(err error, reason string, exitCode *int) (bool, error
...
-			ulog.Logger(_simpleCtx).Info(
-				"module will not be added due to selected roles",
-				"roles", mi.Roles,
-			)
-		}
-
-		for _, mod := range mods {
-			err = s.addModule(mod)
-		}
+func (s *manager) addModule(name string, module ModuleCreateFunc, options ...ModuleOption) error {
```
madhuravi: why add name here? Makes module initialization a bit bloated. There are some big changes in this PR, can you add reasoning behind the major changes in PR description?

service/builder.go
peter-edge: You can't do this anymore, `WithModule` takes more than one parameter for every module.

service/host.go
peter-edge: Having worked through this, it doesn't make sense to have default names for modules - it completely messes with scoping. Part of the point of this change is to eventually be able to scope all relevant `service.Host` methods by module (if we fix https://github.com/uber-go/fx/issues/306), and you need a name to do so. The default names got to be completely problematic, and presented a chicken-or-egg problem with creating a ModuleInfo vs creating a Module.

peter-edge:
Tentatively ready for review, tests are passing. Will add more context in a bit.

service/host.go
```diff
@@ -184,32 +174,24 @@ func (s *manager) shutdown(err error, reason string, exitCode *int) (bool, error
...
+	if s.locked {
+		return fmt.Errorf("can't add module: service already started")
```
alsamylkin: Why do you want to remove ability to hot plug modules?

modules/rpc/thrift.go
```diff
@@ -36,27 +35,21 @@ var _setupMu sync.Mutex
...
```
alsamylkin: Each module have a specific set of options and they can overlap, passing everything to a builder is very confusing..

service/host.go
peter-edge: This PR does nothing to affect this, this was a previous PR, the location was just moved.

modules/rpc/thrift.go
peter-edge: I don't see the issue, what would you propose?

peter-edge:
I'm not really entirely sure how to split this into separate PRs to be honest. It's a big change, and every cascading change came out of semi-necessity by simplifying modules. I wanted to get this done sooner than later as I'm already a day back from working on what I intended to this week - I'd love if we could do line by line over video on this one and get it done.

modules/rpc/thrift.go
peter-edge: How the code is in this PR barely changes usage. If anything, I like the idea of being able to say:

```
var FooThriftModuleFunc = rpc.ThriftModule(FooThriftServiceFunc)

WithModule("myFoo", FooThriftModuleFunc, mySpecificOptionsForThisRegistration...)
```

service/host.go
peter-edge: Wait, this is weird, I didn't add this bool in, why does it show that I did? Or am I going crazy?

modules/task/doc.go
peter-edge: ? It changed from `make gendoc`.

service/host.go
alsamylkin: it was there before, you just moved it :)

service/host.go
madhuravi: Second part of my question  - Can you update PR description with a list of main changes and why.

alsamylkin:
Sure, feel free to set up a meeting, it feels it is going to be a long one :)

peter-edge:
Code coverage is back up

modules/rpc/yarpc_options.go
peter-edge: Done

sectioneight:
👏 🎉

PR #276, opened 2017-02-16
service: remove `NewBuilder` and `Owner.AddModules` and `New`

yichengq:
To create builder, it is more clear to use `WithModules`
and `WithOptions`.

For `Owner.AddModules`, the modules should be added at the
build stage. There is no need to add modules after Owner
is created.

`New` is removed, and it is encouraged to always use builder.

for https://jira.uberinternal.com/browse/DRI-11

### DIFF
--- a/service/builder.go
+++ b/service/builder.go
@@ -22,16 +22,16 @@ package service
 
 import "github.com/pkg/errors"
 
+// WithModules is a helper to create a service without any options
+func WithModules(modules ...ModuleCreateFunc) *Builder {
+	b := &Builder{}
+	return b.WithModules(modules...)
+}
+
 // A Builder is a helper to create a service
 type Builder struct {
 	options []Option
 	modules []ModuleCreateFunc
-}
-
-// NewBuilder returns a new Builder for instantiating services
-func NewBuilder(options ...Option) *Builder {
-	b := &Builder{}
-	return b.WithOptions(options...)
 }
 
 // WithModules adds the given modules to the service
@@ -39,12 +39,6 @@ func (b *Builder) WithModules(modules ...ModuleCreateFunc) *Builder {
 	b.modules = append(b.modules, modules...)
 
 	return b
-}
-
-// WithModules is a helper to create a service without any options
-func WithModules(modules ...ModuleCreateFunc) *Builder {
-	b := NewBuilder()
-	return b.WithModules(modules...)
 }
 
 // WithOptions adds service Options to the builder
@@ -55,14 +49,9 @@ func (b *Builder) WithOptions(options ...Option) *Builder {
 
 // Build returns the service, or any errors encountered during build phase.
 func (b *Builder) Build() (Owner, error) {
-	svc, err := New(b.options...)
+	svc, err := NewOwner(b.modules, b.options...)
 	if err != nil {
-		return nil, errors.Wrap(err, "service instantiation failed due to options")
+		return nil, errors.Wrap(err, "service instantiation failed")
 	}
-
-	if err := svc.AddModules(b.modules...); err != nil {
-		return nil, errors.Wrap(err, "service modules failed to initialize")
-	}
-
 	return svc, nil
 }
--- a/service/builder.go
+++ b/service/builder.go
@@ -49,7 +49,7 @@ func (b *Builder) WithOptions(options ...Option) *Builder {
 
 // Build returns the service, or any errors encountered during build phase.
 func (b *Builder) Build() (Owner, error) {
-	svc, err := NewOwner(b.modules, b.options...)
+	svc, err := newOwner(b.modules, b.options...)
 	if err != nil {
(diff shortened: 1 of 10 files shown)


service/service.go
```diff
@@ -83,9 +81,8 @@ type serviceConfig struct {
 	Roles       []string `yaml:"roles"`
 }
 
-// New creates a service owner from a set of service instances and options
-// TODO(glib): Something is fishy here... `service.New` returns a service.Owner -_-
-func New(options ...Option) (Owner, error) {
+// NewOwner creates a service owner from a set of module creation functions and options.
+func NewOwner(modules []ModuleCreateFunc, options ...Option) (Owner, error) {
```
yichengq: an even more aggressive thought is to make this function private, and we always encourage people to use `Builder` to build the owner.

alsamylkin: Lets keep modules in options..

service/builder.go
```diff
@@ -22,31 +22,25 @@ package service
 
 import "github.com/pkg/errors"
 
+// WithModules is a helper to create a service without any options
+func WithModules(modules ...ModuleCreateFunc) *Builder {
```
alsamylkin: nit: that looks a bit weird from a user perspective: `service.WithModules` returns a builder, not a service.

service/host.go
```diff
@@ -183,8 +183,8 @@ func (s *host) shutdown(err error, reason string, exitCode *int) (bool, error) {
 	return true, err
 }
 
-// AddModules adds the given modules to a service host
-func (s *host) AddModules(modules ...ModuleCreateFunc) error {
+// addModules adds the given modules to a service host
+func (s *host) addModules(modules ...ModuleCreateFunc) error {
```
alsamylkin: Thanks!

service/service.go
madhuravi: +1 we can make this private.

service/host.go
```diff
@@ -183,8 +183,8 @@ func (s *host) shutdown(err error, reason string, exitCode *int) (bool, error) {
 	return true, err
 }
 
-// AddModules adds the given modules to a service host
-func (s *host) AddModules(modules ...ModuleCreateFunc) error {
+// addModules adds the given modules to a service host
```
anuptalwalkar: no need for comment here

service/service.go
```diff
@@ -48,8 +48,6 @@ const (
 type Owner interface {
 	Host
 
-	AddModules(modules ...ModuleCreateFunc) error
```
anuptalwalkar: Can you update the README as well? It explicitly mentions use of AddModules

service/builder.go
anuptalwalkar: Well, builder pattern! This might be more suitable in builder package, but then we'll dive into circular dependency issues.

service/host.go
madhuravi: I prefer to err on the side of having comments. Let's not remove comments because they are not required by linting :)

service/service.go
yichengq: Let me do this and check how it goes

yichengq: Plan to make this private, and builder uses optional arguments for modules.

service/builder.go
yichengq: @alsamylkin The concern here is that building service needs `modules` and `options`, and people like to to optional arguments to pass them in. This results in split them into 2 functions anyway.

@anuptalwalkar Plan to make this standard building method. Another possibility here is to use a config struct. This is beyond the scope of this diff though.

service/service.go
yichengq: sure

yichengq:
Addressed comments and rebased.

service/host.go
anuptalwalkar: true, but method name is pretty self explanatory :)

service/builder.go
sectioneight: Agree that it's confusing, but we will give people (at uber) a templated app that calls .Build() so hopefully they won't even notice, they just get a `(Service, error)`

service/options_test.go
```diff
@@ -32,27 +32,26 @@ import (
...
+		newOwner([]ModuleCreateFunc{successModuleCreate}, WithMetrics(tally.NoopScope, metrics.NopCachedStatsReporter))
 	})
 }
 
-func TestWithTracing_OK(t *testing.T) {
+func TestNewOwner_WithTracingOK(t *testing.T) {
 	tracer := &opentracing.NoopTracer{}
 	assert.NotPanics(t, func() {
-		New(WithTracer(tracer))
+		newOwner([]ModuleCreateFunc{successModuleCreate}, WithTracer(tracer))
```
sectioneight: can we pass `nil` instead of the success? these tests just got a lot harder to read. wondering if we can pass an empty slice of module create funcs, or whether the framework disallows that

sectioneight:
Thanks @yichengq, this is great stuff

PR #277, opened 2017-02-16
Initial DIG implementation

glibsm:
Several things still to tidy up, but seems to be working to review.

`README.md` is a good place to start and `examples/dig/*` to see how one might use it in a "real" application.

TODO list:

- [x] Thread safety
- [x] Rebrand `Inject` to `Register`

Other TODOs are now filed as issues.

#282
#283
#284
#285


dig/README.md
```diff
@@ -0,0 +1,78 @@
...
+`*Object` that requires `*Dep` to be present in the graph
+
+```go
+type Dep struct{}
+
+type Object struct{
+  Dep
+}
+
+func NewObject(d *Dep) *Object {
```
dmcaulay: can constructors return an error? it seems reasonable since resolve returns an error.

dig/README.md
```diff
@@ -0,0 +1,78 @@
...
+There are two sides of dig: `Inject` and `Resolve`
+
+## Inject
+
+Adding an object, or a constructor of an object to the graph.
+
+There are two ways of providing an object:
+
+1. Injecting a pointer to an existing object
+1. Injecting a "constructor function" that returns one pointer (or interface)
```
breerly: Guessing this should be `2.`

dig/README.md
```diff
@@ -0,0 +1,78 @@
...
+### Injecting an object
+
+Injecting an object means it has no dependencies, and will be used as a
+**shared** singleton instance for all resolutions within the graph
+
+```go
+type Fake struct {
+    Name string
+}
+err := g.Inject(&Fake{Name: "I am an injected thing"})
```
breerly: Suggest renaming this to `g.Provide`, `g.Register`, or `g.Set` - 

Using `Inject` terminology here is confusing because dependency injection uses that term to refer to instantiating an object with its dependencies (either through constructor, property, or setter injection).

In this case, we aren't "injecting" the object into the graph, we are "providing" the object so that  it can be *injected* into other objects.

dig/README.md
glibsm: That is a good point. perhaps a tuple of `obj, error` is in order here to tie in the rest of the error stuff here
Alternative could be to check for zero value

dig/README.md
```diff
@@ -0,0 +1,78 @@
+# Dependency Injection Graph
+
+This package provides a fairly opinionated way of resolving object dependencies.
```
Lupie: Can you start this with "Package dig provides" just to make it super clear that "Dependency Injection Graph" and "this package" and "dig" all mean the same thing?

dig/README.md
```diff
@@ -0,0 +1,78 @@
...
+There are two sides of dig: `Inject` and `Resolve`
```
Lupie: Period at the end

dig/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Inject
+
+Adding an object, or a constructor of an object to the graph.
```
Lupie: `Inject` adds an object, or a constructor of an object, to the graph.

dig/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+There are two ways of providing an object:
```
Lupie: of providing --> to provide

glibsm:
@dmcaulay absolutely. thread safety, panic wraps and a few more tests are definitely in order. I'll add to the description list

dig/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+1. Injecting a pointer to an existing object
```
Lupie: Injecting --> Inject

dig/README.md
Lupie: - Injecting --> Inject
- Why isn't this "2."?

dig/README.md
```diff
@@ -0,0 +1,78 @@
...
+1. Injecting a "constructor function" that returns one pointer (or interface)
+
+### Injecting an object
```
Lupie: Nit: Injecting --> Inject

Verb+ing as a heading can be good for people skimming, but "Inject" is more direct. Plus, you have "Injecting an object" in the paragraph that follows for anyone searching for those terms instead.

dig/README.md
Lupie: Would it be better to say "ways to use dig" instead of "sides of dig"?

dig/dig.go
```diff
@@ -0,0 +1,288 @@
...
+	for _, node := range g.nodes {
+		for _, dep := range node.dependencies() {
+			node, ok := g.nodes[dep]
+			if ok {
+				l.PushFront(node)
+			}
+		}
+	}
+
+	return cv.Call(args)[0], nil
```
dmcaulay: couldn't you grab the constructor error from the `Call` result?

dig/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+type Object struct{
+  Dep
+}
+
+func NewObject(d *Dep) *Object {
+  return &Object{Dep: d}
+}
+
+err := dig.Inject(NewObject)
```
breerly: How does the user get access to `Object` here? How can `NewObject` be called if `Dep` hasn't first been provided to the graph?

dig/dig.go
dmcaulay: i thought returning it here might make constructors with errors work.

glibsm: If we force a `ptr,error` tuple for constructors, yes error would be `[1]` in the returns

dmcaulay: 👍  i think forcing ptr, error is reasonable

dig/README.md
```diff
@@ -0,0 +1,78 @@
...
+Adding an object, or a constructor of an object to the graph.
+
+There are two ways of providing an object:
+
+1. Injecting a pointer to an existing object
+1. Injecting a "constructor function" that returns one pointer (or interface)
+
+### Injecting an object
+
+Injecting an object means it has no dependencies, and will be used as a
```
Lupie: "means it has no dependencies" could mean:
- it's a requirement for the object to have no dependencies (i.e., you can only inject objects without dependencies)
- removes its dependencies (i.e., when you inject an object, that object drops any dependencies)
- doesn't add any dependencies (i.e., just because you injected the object into this graph doesn't mean it's useful yet—you still need to tie it to other things)

dig/README.md
```diff
@@ -0,0 +1,78 @@
...
+**shared** singleton instance for all resolutions within the graph
```
Lupie: Period at the end

dig/README.md
```diff
@@ -0,0 +1,78 @@
...
+require.NoError(t, err)
+
+var f1 *Fake
+err = g.Resolve(&f1)
+require.NoError(t, err)
+
+// f1 is ready to use here...
+```
+
+### Injecting a constructor
```
Lupie: Nit: Injecting --> Inject
If you do this, also change "This is a more interesting…" to "Injecting a constructor is a more interesting…".

dig/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+This is a more interesting and widely used scenario. Constructor is defined as a
+function that returns exactly one pointer (or interface) and takes 0-N number of
+arguments. Each one of the arguments is automatically registered as a
+**dependency** and must also be an interface or a pointer.
+
+The following example illustrates injecting a constructor function for type
```
Lupie: It'd be sweet to have a real example in here, too. Could we add one once more people are using it?

dig/README.md
glibsm: if you like tracking 1-2-3-4 numbers etc, sure. Markdown properly numbers the list if they are all 1 :)

dig/README.md
```diff
@@ -0,0 +1,78 @@
...
+func NewObject(d *Dep) *Object {
+  return &Object{Dep: d}
+}
+
+err := dig.Inject(NewObject)
+```
+
+## Resolve
+
+Mechanism of retrieving objects from the graph by type.
```
Lupie: Mechanism of retrieving --> `Resolve` retrieves

dig/README.md
breerly: I did not know that.

dig/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+In the future, there are plans to do named retrievals to support multiple
```
Lupie: There are future plans to…

dig/dig.go
```diff
@@ -0,0 +1,288 @@
...
+	}
+}
+
+// Resolve all of the dependencies of the provided class
+//
+// Provided object must be a pointer
+// Any dependencies of the object will receive constructor calls, or be initialized (once)
+// Constructor with return value *object will be called
+//
+// TODO(glib): catch any and all panics from this method, as there is a lot of reflect going on
```
Lupie: We cool with publishing this?

examples/dig/README.md
```diff
@@ -0,0 +1,20 @@
+# Dependency Injection Graph example
```
Lupie: example --> Example

examples/dig/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+This example illustrates how `package go.uber.org/fx/dig` can be used to inject
+a dependency.
+
+`struct HelloHandler` exploses a dependency on `hello.Sayer` but does not
```
Lupie: exploses --> exposes

examples/dig/README.md
```diff
@@ -0,0 +1,20 @@
...
+provide any guidance of when or where to get it. On the other side package sayer
```
Lupie: "On the other side package sayer" is awkward. How about "However, package sayer injects…"? I'm expecting to see something about package sayer providing guidance in contrast to HelloHandler not providing any guidance. Not sure what the point is here.

examples/dig/README.md
```diff
@@ -0,0 +1,20 @@
...
+injects an implementation of `hello.Sayer` interface into the graph
```
Lupie: Period at the end

examples/dig/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+When the application is initialized, `dig` is asked to resolve the dependencies
+of the `HelloHandler` and create an instance.
+
+## Running the example
```
Lupie: Running --> Run

examples/dig/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+```
+$ go build
+$ ./dig
+$ curl localhost:8080
+Well hello there DIG!. How are you?
```
Lupie: Haha this is awesome, but I wish that "!." weren't in the message.

examples/dig/hello/sayer.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+func init() {
+	err := dig.Inject(NewPoliteSayer)
+	if err != nil {
+		log.Fatalf("Failed to inject a new polite sayer: %v", err)
+	}
+}
+
+// Sayer returns a string that say hello to the person
```
Lupie: say --> says

examples/dig/hello/sayer.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+type politeSayer struct{}
+
+func (ps *politeSayer) SayHello(person string) string {
+	return fmt.Sprintf("Well hello there %v. How are you?", person)
+}
+
+// DIG injection function that provides a polite sayer
+// Function takes no parameters, i.e. it has no dependencies
```
Lupie: parameters, i.e. it --> parameters (i.e., it)

glibsm:
@breerly `TestBasicResolve` does that, right? Unless im misunderstanding something from your example. It injects `kGrandchild1 -> Child1 -> Parent1`

examples/dig/main.go
```diff
@@ -0,0 +1,55 @@
...
+
+	if err != nil {
+		log.Fatal("Unable to initialize service", "error", err)
+	}
+
+	svc.Start()
+}
+
+func router(_ service.Host) []uhttp.RouteHandler {
+	// when service calls us to provide a handler, we simply grab one from DIG
```
Lupie: Let's keep dig lowercase or uppercase throughout. I vote lowercase. I think it's just here and in the "Well hello there DIG.!" that it's capped.

dig/README.md
glibsm: API started out with `Register` instead of `Inject`, but wouldn't that compromise the `dig` package name? :(

dig/README.md
glibsm: it's the same way to use it, one gives, one takes. If you have better ideas of how to phrase it let me know

dig/README.md
glibsm: markdown works properly with these, i promise. it's so lists can be easily rearranged without fixing the numbers

dig/README.md
breerly: I was just a bit confused that `Inject` actually instantiated the thing, when in reality it was just providing it to the graph so that when `Resolve` was called, it had everything it needed.

In the context of a regular app, I like this terminology:

1. A DIG graph is created to manage the application graph
2. Structs and constructors are provided to the graph
3. The graph is resolved, resulting in the application graph being fully instantiated

dig/README.md
Lupie: "two sides" works.

dig/README.md
Lupie: Learning

glibsm:
@breerly Yes! I was just talking about taking config values/keys as params to a constructor with Maxim. That seems like a YUUUUGE win, lets see if we can get there.

examples/dig/main.go
glibsm: oKaY

dig/dig.go
glibsm: that TODO is in scope for this PR to fix

dmcaulay:
I've been thinking about DB integrations for Fx and I think this would be very helpful. I have two questions:

1. Will each host have its own graph?
This is nice for tests because I create a new host for each test and I don't want tests to share the same DB client. Ideally, each test gets a clean DB.

2. Can the graph implement the `io.Closer` interface?
This is also good for tests, because I want an easy way to cleanup everything in the graph. Unfortunately, objects created by constructors would probably need to be closed by the consumer. This is a little awkward.

glibsm:
@dmcaulay 

1. Yes, I think the plan was to attach a graph to each host, though I think the global graph (based on default) is going to be pretty useful for most people

2. There is a `Reset()` method right now on the default graph, which also needs to make it's way to the main interface for `Graph`. `Close()` seems a little silly, as it implies that you have to re-create from scratch. We'd just want to clear out the existing nodes, right?

dmcaulay:
> There is a Reset() method right now on the default graph, which also needs to make it's way to the main interface for Graph. Close() seems a little silly, as it implies that you have to re-create from scratch. We'd just want to clear out the existing nodes, right?

I'd also like it to `Close` nodes if they contain closers. Do you think that's too much?

dig/dig.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+
+	// check that all the dependencies have nodes present in the graph
+	// doesn't mean everything will go smoothly during resolve, but it
+	// drastically increases the chances that we're not missing something
+	for _, node := range g.nodes {
+		for _, dep := range node.dependencies() {
+			// check that the dependency has been nodes
+			_, ok := g.nodes[dep]
+			if !ok {
```
alsamylkin:
```golang
if _, ok := g.nodes[objElemType]; ok {
```

dig/default.go
```diff
@@ -0,0 +1,45 @@
...
+	return defaultGraph.Inject(i)
+}
+
+// InjectAll into the default graph
+func InjectAll(is ...interface{}) error {
+	return defaultGraph.InjectAll(is...)
+}
+
+// Resolve an object through the default graph
+func Resolve(i interface{}) error {
```
dmcaulay: What do you think about a `ResolveAll` function?

dig/dig.go
```diff
@@ -0,0 +1,286 @@
...
+	// check if the type has been nodes
+	n, ok := g.nodes[objElemType]
+	if !ok {
+		return fmt.Errorf("type %v is not registered", objType)
+	}
+
+	// check that all the dependencies have nodes present in the graph
+	// doesn't mean everything will go smoothly during resolve, but it
+	// drastically increases the chances that we're not missing something
+	for _, node := range g.nodes {
```
alsamylkin: Why we don't resolve all dependencies recursively?

dig/dig.go
```diff
@@ -0,0 +1,286 @@
...
+func newGraph() *graph {
+	return &graph{
+		nodes: make(map[interface{}]object),
+	}
+}
+
+func (g graph) String() string {
+	var b bytes.Buffer
+	for key, reg := range g.nodes {
+		b.WriteString(fmt.Sprintf("%v -> %v\n", key, reg))
```
alsamylkin:
```golang
fmt.Fprintf(b, "%v -> %v\n", key, reg)
```
Or
```golang
fmt.Fprintln(b, key, "->", reg)
```

dig/dig.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+	return fmt.Sprintf("{nodes:\n%v}", b.String())
```
alsamylkin: %s?

dig/dig.go
alsamylkin: Print extra tabs before each pair?:)

service/core.go
```diff
@@ -45,6 +45,7 @@ type Host interface {
 	RuntimeMetricsCollector() *metrics.RuntimeCollector
 	Observer() Observer
 	Config() config.Provider
+	// TODO: delete me!
```
alsamylkin: ?

dig/README.md
anuptalwalkar: This could be a 3 step process. I like the term `Register` then `Resolve` graph, and `Inject`. In think `Inject` is used when object is ready to be consumed, and gets 'Injected' in the code. That way may be we can hide the Resolve call from users?
 
1. Graph is registered with all the dependency objects
2. DIG resolves the dependency graph
3. Objects are Injected in the code

dig/dig.go
```diff
@@ -0,0 +1,286 @@
...
+	if !ok {
+		return fmt.Errorf("type %v is not registered", objType)
+	}
+
+	// check that all the dependencies have nodes present in the graph
+	// doesn't mean everything will go smoothly during resolve, but it
+	// drastically increases the chances that we're not missing something
+	for _, node := range g.nodes {
+		for _, dep := range node.dependencies() {
+			// check that the dependency has been nodes
```
anuptalwalkar: what does this mean?

dig/dig.go
```diff
@@ -0,0 +1,286 @@
...
+	// Resolve the dependencies of the object and populate the pointer value
+	Resolve(interface{}) error
+}
+
+// Inject an object in the dependency graph
+//
+// Provided argument must be a function that returns exactly one pointer argument
+// All arguments to the function must be pointers
+func (g *graph) Inject(c interface{}) error {
+	ctype := reflect.TypeOf(c)
```
alsamylkin: Cycle detection in this PR?

dig/dig.go
```diff
@@ -0,0 +1,286 @@
...
+			objType: objType,
+		},
+	}
+	for i := 0; i < argc; i++ {
+		arg := ctype.In(i)
+		if arg.Kind() != reflect.Ptr && arg.Kind() != reflect.Interface {
+			return errArgKind
+		}
+
+		n.deps[i] = arg
```
alsamylkin: Should we inject this dependency first?

dig/README.md
madhuravi: It doesn't look like Inject is instantiating it. It takes in an instantiated object and adds it to the graph.
But to Grayson 's point, typically the injection part refers to populating the target object.

examples/dig/main.go
```diff
@@ -0,0 +1,55 @@
...
+	var h *handlers.HelloHandler
+	err := dig.Resolve(&h)
```
glibsm: I don't really understand what you mean. Grab me in person and lets chat, or else this thread could get large

dig/dig.go
```diff
@@ -0,0 +1,286 @@
...
+	n := funcNode{
+		deps:        make([]interface{}, argc),
+		constructor: c,
+		node: node{
+			objType: objType,
+		},
+	}
+	for i := 0; i < argc; i++ {
+		arg := ctype.In(i)
+		if arg.Kind() != reflect.Ptr && arg.Kind() != reflect.Interface {
```
glibsm: No constants can and should be part of the graph, else how do you match the arguments to the constructors?

dig/dig.go
```diff
@@ -0,0 +1,286 @@
...
+	return fmt.Sprintf(
+		"(function) deps: %v, type: %v, constructor: %v, cached: %v",
+		n.deps,
+		n.objType,
+		n.constructor,
+		n.cached,
+	)
+}
+
+func (g *graph) injectObject(o interface{}) error {
```
glibsm: so we don't panic if someone provides a non-injectable object, like a struct value

dig/dig.go
glibsm: we do, just not in this method

dig/README.md
```diff
@@ -0,0 +1,78 @@
...
+}
+
+err := dig.Inject(NewObject)
+```
+
+## Resolve
+
+`Resolve` retrieves objects from the graph by type.
+
+There are future plans to do named retrievals to support multiple
```
madhuravi: What about a constructor and object for the same type? Would named retrievals be used to solve that too?

dig/default.go
glibsm: Great idea. 

Same reason `InjectAll` exists I suppose.

dig/dig.go
glibsm: Depends how much bigger this PR is going to get, might have to push it out for the sanity of the reviewers

dig/README.md
glibsm: point taken, will revert back to `Provide/Register` :(
might flip a coin again or something

dig/README.md
glibsm: This would actually be an error as Dep is not provided. Good catch, Dep needs to be provided first

dig/README.md
glibsm: There is a disclaimer in the readme, everything at the moment is one-of-a-kind and a singleton that gets shared.

Support for multiple types would have to be done through the names or something like that, yes.

dig/dig.go
glibsm: I don't understand what you mean. This is just someone providing a constructor with several parameters, and each parameter type becomes a dependency of the return type.

Injecting provides an object, dependency is just a type.

dig/dig.go
```diff
@@ -0,0 +1,292 @@
...
+	// Inject into the dependency graph
+	// Parameter must be a pointer or a constructor function returning exactly one pointer
+	Inject(interface{}) error
+
+	// Same as calling .Inject for each provided vararg
+	// Returns the first error encountered
+	InjectAll(...interface{}) error
+
+	// Resolve the dependencies of the object and populate the pointer value
+	Resolve(interface{}) error
```
madhuravi: When will instantiation errors be caught? Say a constructor returns an error, when is this surfaced? Resolve time seems a bit late. Is there a way we can instantiate all on startup and generate errors right there for objects/constructors that failed to resolve? This can get tricky with constructors etc but worth thinking about.

glibsm: Error is bubbled up as soon as it's encountered.

You can't resolve on startup, because dependencies are not added to the graph in a logical order. This assumes that, for example, you have to provide the Redis constructor before Handler constructor is provided. You simply can't enforce that order.

Using reflect comes with a lot of runtime implications. The proper way to validate DI is to do codegen, but that comes with it's own set of limitations and problems. Perhaps `dig v2` will do codegen instead

service/core.go
glibsm: dig is here to replace resources

dig/dig.go
```diff
@@ -0,0 +1,292 @@
...
+func (g *graph) Inject(c interface{}) error {
+	ctype := reflect.TypeOf(c)
+
+	switch ctype.Kind() {
+	case reflect.Func:
+		if ctype.NumOut() != 1 {
+			return errReturnCount
+		}
+
+		objType := ctype.Out(0)
```
madhuravi: This will panic if there is no return argument. First return err if ctype.NumOut() != 1

madhuravi: never mind, Im blind.

dig/dig.go
```diff
@@ -0,0 +1,292 @@
...
+	objType := reflect.TypeOf(obj)
+	if objType.Kind() != reflect.Ptr {
+		return fmt.Errorf("can not resolve non-pointer object of type %v", objType)
+	}
+
+	objElemType := reflect.TypeOf(obj).Elem()
+	objVal := reflect.ValueOf(obj)
+
+	// check if the type is a registered node
+	n, ok := g.nodes[objElemType]
```
madhuravi: For future reference, this is what I ended up doing for type safety.

dig/dig.go
```diff
@@ -0,0 +1,292 @@
...
+	if !ok {
+		return fmt.Errorf("type %v is not registered", objType)
+	}
+
+	// check that all the dependencies have nodes present in the graph
+	// doesn't mean everything will go smoothly during resolve, but it
+	// drastically increases the chances that we're not missing something
+	for _, node := range g.nodes {
```
madhuravi: Why are we checking for all of g.nodes here? Shouldn't it be just for n? Checking for all g.nodes can go in ResolveAll.

dig/dig.go
```diff
@@ -0,0 +1,292 @@
...
+		for _, dep := range node.dependencies() {
+			// check that the dependency is a registered node
+			if _, ok := g.nodes[dep]; !ok {
```
madhuravi: Can the dependency register checks go inside the n.value function itself? Makes sense to check this every time value is called. Along with that, you can remove the dependencies function from the object interface since value function is the only one that needs access to it.

dig/dig.go
alsamylkin: Do we need to inject constructor arguments in the graph as well?

glibsm: That is what the `Provide` function does, else the graph is incomplete therefore we can't resolve

dig/dig.go
```diff
@@ -0,0 +1,292 @@
...
+				return fmt.Errorf("%v dependency of type %v is not registered", objElemType, dep)
+			}
+		}
+	}
+
+	v, err := n.value(g)
+	if err != nil {
+		return errors.Wrapf(err, "unable to resolve %v", objType)
```
madhuravi: nit: resolve is indicative of the overall function. better to say "unable to determine value for.. "

dig/dig.go
```diff
@@ -0,0 +1,292 @@
...
+	}
+
+	// set the pointer value of the provided object to the instance pointer
+	objVal.Elem().Set(v)
+
+	return nil
+}
+
+// Reset the graph by removing all the registered nodes
+func (g *graph) Reset() {
```
madhuravi: When would this be called?

dig/dig.go
alsamylkin: Well, you can't/don't have to resolve non pointers..

What I meant is we don't have to restrict argument types and store them with a function itself.

dig/dig.go
alsamylkin: Users have to do this manually??

dig/dig.go
madhuravi: nit: how about colon instead of arrows? that's more indicative of maps in go.

dig/dig.go
```diff
@@ -0,0 +1,292 @@
...
+
+func (g graph) String() string {
+	var b bytes.Buffer
+	for key, reg := range g.nodes {
+		b.WriteString(fmt.Sprintf("%v -> %v\n", key, reg))
+	}
+	return fmt.Sprintf("{nodes:\n%v}", b.String())
+}
+
+type node struct {
```
madhuravi: Is this an objNode? Better to name it as such.

dig/dig.go
```diff
@@ -0,0 +1,292 @@
...
+	fmt.Stringer
+
+	obj         interface{}
+	objType     reflect.Type
+	cached      bool
+	cachedValue reflect.Value
```
madhuravi: unused?

glibsm: it's definitely used

dig/dig.go
```diff
@@ -0,0 +1,292 @@
...
+// Call the function and return the result
+func (n *funcNode) value(g *graph) (reflect.Value, error) {
+	if n.cached {
+		return n.cachedValue, nil
+	}
+
+	cv := reflect.ValueOf(n.constructor)
+	ct := reflect.TypeOf(n.constructor)
+
+	args := make([]reflect.Value, ct.NumIn(), ct.NumIn())
```
madhuravi: should be make([]reflect.Value, 0, ct.NumIn())

madhuravi: nm, you're not appending.

dig/dig.go
```diff
@@ -0,0 +1,292 @@
...
+	for idx := range args {
+		arg := ct.In(idx)
+		node, ok := g.nodes[arg]
```
madhuravi: condense the two lines.
if node, ok := g.nodes[arg]; ok {
}

dig/dig.go
```diff
@@ -0,0 +1,292 @@
...
+			v, err := node.value(g)
+			if err != nil {
+				return reflect.Zero(n.objType), errors.Wrap(err, "dependency resolution failed")
+			}
+			args[idx] = v
+		}
+	}
+
+	v := cv.Call(args)[0]
+	n.cached = true
```
madhuravi: will this be configurable in future?

dig/dig.go
```diff
@@ -0,0 +1,292 @@
...
+	return n.deps
+}
+
+func (n funcNode) String() string {
+	return fmt.Sprintf(
+		"(function) deps: %v, type: %v, constructor: %v, cached: %v",
+		n.deps,
+		n.objType,
+		n.constructor,
+		n.cached,
```
madhuravi: cachedValue if it exists?

dig/dig.go
madhuravi: I think he means should we check here if that argument type has already been registered. Does it have to be in that order?

dig/dig.go
madhuravi: I think he means should we check here if that argument type has already been registered. Does it have to be in that order?

glibsm: You can't check registration of arguments when constructor gets injected, because you don't control the order of when and how things get registered. We can talk about this more tomorrow, clearly this thread is already getting large

dig/README.md
```diff
@@ -0,0 +1,78 @@
...
+1. Inject a "constructor function" that returns one pointer (or interface)
+
+### Inject an object
+
+Injecting an object means it has no dependencies, and will be used as a
+**shared** singleton instance for all resolutions within the graph.
+
+```go
+type Fake struct {
+    Name string
```
sectioneight: nit: can we settle on 2 or 4 spaces? i feel like we've been using 2.

dig/README.md
```diff
@@ -0,0 +1,78 @@
...
+}
+err := g.Register(&Fake{Name: "I am an injected thing"})
+require.NoError(t, err)
```
sectioneight: instead of shadowing error, just inline:

```go
require.NoError(t, g.Register(....))
```

dig/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+var f1 *Fake
+err = g.Resolve(&f1)
+require.NoError(t, err)
```
sectioneight: samesies

dig/README.md
```diff
@@ -0,0 +1,78 @@
...
+## Resolve
+
+`Resolve` retrieves objects from the graph by type.
+
+There are future plans to do named retrievals to support multiple
+objects of the same type in the graph.
+
+```go
+var o *Object
+err := dig.Resolve(&o) // notice the pointer to a pointer as param type
```
sectioneight: Idioms:

```go
if err : = ...; err == nil {

}
```

dig/README.md
```diff
@@ -0,0 +1,78 @@
...
+if err == nil {
+    // o is ready to use
+}
+
+type Do interface{}
+var d Do
+err := dig.Resolve(&d) // notice pointer to an interface
+if err == nil {
```
sectioneight: same

sectioneight: same

examples/dig/README.md
```diff
@@ -0,0 +1,20 @@
...
+injects an implementation of `hello.Sayer` interface into the graph.
+
+When the application is initialized, `dig` is asked to resolve the dependencies
+of the `HelloHandler` and create an instance.
+
+## Run the example
+
+```
+$ go build
+$ ./dig
```
sectioneight: How does this relate to `u`, go-build, and thriftsync?

examples/dig/README.md
```diff
@@ -0,0 +1,20 @@
+# Dependency Injection Graph Example
+
+This example illustrates how `package go.uber.org/fx/dig` can be used to inject
+a dependency.
+
+`struct HelloHandler` exposes a dependency on `hello.Sayer` but does not
```
sectioneight: nit: `type HelloHandler struct` is the syntactically-correct version

examples/dig/README.md
glibsm: it's just a binary name for the example :)

dig/README.md
glibsm: nice!

glibsm:
@breerly Agreed. I deferred most TODOs to other PRs, will only add thread safety here.

dig/README.md
glibsm: filed a ticket to take a pass at the readme to fix all this up

dig/dig.go
glibsm: yes, it's in the readme

dig/dig.go
glibsm: its moved now

dig/dig.go
glibsm: reset a global graph

dig/dig.go
```diff
@@ -0,0 +1,308 @@
...
+	otype := reflect.TypeOf(o)
+
+	n := node{
+		obj:     o,
+		objType: otype,
+		cached:  true,
+	}
+
+	g.nodes[otype] = &n
+	return nil
```
glibsm: rolling api changes

dig/README.md
```diff
@@ -0,0 +1,78 @@
+# Dependency Injection Graph
+
+Package dig provides a fairly opinionated way of resolving object dependencies.
+There are two sides of dig: `Inject` and `Resolve`.
```
kriskowal: Inject and resolve are of course colloquially "dig" and "dug".

digdug

dig/dig.go
kriskowal: Looks like this TODO got merged.

dig/default.go
yutongp: Just curious, can we have `Registe(...interface{}) error` and `Resolve(...interface{}) error`, so we can provide same functionalities with less APIs?

dig/default.go
```diff
@@ -0,0 +1,49 @@
...
+// Default graph is used for all the top-level inject calls
+var defaultGraph = newGraph()
+
+// Register into the default graph
+func Register(i interface{}) error {
+	return defaultGraph.Register(i)
+}
+
+// RegisterAll into the default graph
+func RegisterAll(is ...interface{}) error {
```
glibsm: It's a valid question. Yes, at the moment this can be combined, but I'm actually saving room for `Register(interface{}, ...Option)` and `Resolve(interface{}, ...Option)`

A google doc will be shared shortly with some design thoughts and goals. Essentially, it will allow us to have named imports, private instances, etc.

I think @kriskowal had the same question, so a courtesy tag.

dig/dig.go
```diff
@@ -0,0 +1,321 @@
...
+	return reflect.ValueOf(n.obj), nil
+}
+
+func (n objNode) dependencies() []interface{} {
+	return nil
+}
+
+func (n objNode) String() string {
+	return fmt.Sprintf(
+		"(object) obj: %v, deps: nil, cached: %v, cachedValue: %v",
```
glibsm: ah, good point

dig/dig.go
```diff
@@ -0,0 +1,321 @@
...
+	return nil
+}
+
+func (g *graph) registerConstructor(c interface{}) error {
+	ctype := reflect.TypeOf(c)
+	objType := ctype.Out(0)
+
+	argc := ctype.NumIn()
+	n := funcNode{
+		deps:        make([]interface{}, argc),
```
glibsm: Probably forgot to change it, i'm pretty sure it used to be `reflect.Type`, then it was `graphNode`.

Good thing it's not externally visible :) I'll make a mental note to change to value

dig/dig.go
```diff
@@ -0,0 +1,321 @@
...
+	return &graph{
+		nodes: make(map[interface{}]object),
+	}
+}
+
+func (g *graph) String() string {
+	b := &bytes.Buffer{}
+	fmt.Fprintln(b, "{nodes:")
+	for key, reg := range g.nodes {
+		fmt.Fprintln(b, key, " -> ", reg)
```
glibsm: Good idea. Im not sure that preserving the order of dependencies from the constructor in the right order is that important for debugging, but if it gets to that I will extend it as such.

PR #279, opened 2017-02-17
add test to ensure config does not panic when populate for Array type

yutongp:
Array and Slice are different but right now we treat them the same in config package:
```
case bucketArray:
            destSlice := reflect.MakeSlice(fieldType, 0, 4)
```
this will cause panic on Array type like `[6]int`:
```
--- FAIL: TestArrayTypeNoPanic (0.00s)
panic: reflect.MakeSlice of non-slice type [recovered]
    panic: reflect.MakeSlice of non-slice type
```
Add a test to ensure we don’t panic when populate config for Array type
And add TODO for @alsamylkin

### DIFF
--- a/config/value.go
+++ b/config/value.go
@@ -292,6 +292,7 @@ const (
 	bucketArray     = 1
 	bucketObject    = 2
 	bucketMap       = 3
+	bucketSlice     = 4
 )
 
 func getBucket(t reflect.Type) int {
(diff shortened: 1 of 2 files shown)


config/value.go
```diff
@@ -526,6 +527,8 @@ func (cv Value) valueStruct(key string, target interface{}) (interface{}, error)
 				fieldValue.Set(newTarget)
 			}
 		case bucketArray:
+			// TODO(alsam) fix array type.
```
sectioneight: this should be your name, not alex, right?

sectioneight: can we task this up?

yutongp: synced with @alsamylkin before this PR. He claimed it. Will add a task to track

alsamylkin: I've asked Young to add this line. This is our bug, unrelated to PR

PR #280, opened 2017-02-17
add String() func on config.Value to solve fmt print panic

### LINKED ISSUE #278: fmt.Printf("%v", value) causes stack overflow
Understand that the common way to print the actual value of a "Value" is to call Value.Value(), however, printing a raw object with "%v" causes stack overflow is something worth notice.
```
root := config.Load()
cfg := root.Get("test_sec")
fmt.Printf("%v", cfg)
```
ends up with:
> runtime: goroutine stack exceeds 1000000000-byte limit
fatal error: stack overflow

yutongp:
Fixes #278 

I think the cause of this issue is: `Value` has a `Provider` and `yamlConfigProvider` has a `map[string]Value`. This cause finding a `String()` function recursing indefinitely.
Add `String()` function to `Value` can solve this issue.

However, we don't want user use this function, so this solution may not be ideal. Open to discuss.

### DIFF
--- a/config/static_provider_test.go
+++ b/config/static_provider_test.go
@@ -77,7 +77,7 @@ func TestStaticProviderFmtPrintOnValueNoPanic(t *testing.T) {
 	val := p.Get("something")
 
 	f := func() {
-		fmt.Sprintf("%v", val)
+		assert.Contains(t, fmt.Sprintf("%v", val), "")
 	}
 	assert.NotPanics(t, f)
 }
--- a/config/static_provider_test.go
+++ b/config/static_provider_test.go
@@ -21,6 +21,7 @@
 package config
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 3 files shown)


config/static_provider_test.go
```diff
@@ -70,3 +71,13 @@ func TestStaticProvider_Callbacks(t *testing.T) {
...
 }
+
+func TestStaticProviderFmtPrintOnValueNoPanic(t *testing.T) {
+	p := NewStaticProvider(nil)
+	val := p.Get("something")
+
+	f := func() {
+		assert.Contains(t, fmt.Sprintf("%v", val), "")
+	}
+	assert.NotPanics(t, f)
```
anuptalwalkar: nit: can be single line - 
```
assert.NotPanics(t, func() {
  assert.Contains(t, fmt.Sprintf("%v", val), "")
})
```

sectioneight: Or could just do:

```go
assert.Equal(t, fmt.Sprintf("%v", val, ""))
```

sectioneight: because a panic in the test will fail the test

PR #281, opened 2017-02-17
Allow both .yaml and .yml config file extensions

peter-edge:
This might be desirable or not - a lot of people use .yml instead of .yaml - .yaml is more "correct", but it would be nice to support both. This just checks for both. As it is now, .yml overrides .yaml, so maybe we want to reverse the order.

Just as another data point, docker-compose accepts both https://docs.docker.com/compose/compose-file

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -102,6 +102,7 @@ func getConfigFiles() []string {
 	for _, dir := range dirs {
 		for _, baseFile := range baseFiles {
 			files = append(files, fmt.Sprintf("%s/%s.yaml", dir, baseFile))
+			files = append(files, fmt.Sprintf("%s/%s.yml", dir, baseFile))
 		}
 	}
 


config/config.go
```diff
@@ -102,6 +102,7 @@ func getConfigFiles() []string {
 	for _, dir := range dirs {
 		for _, baseFile := range baseFiles {
 			files = append(files, fmt.Sprintf("%s/%s.yaml", dir, baseFile))
+			files = append(files, fmt.Sprintf("%s/%s.yml", dir, baseFile))
```
alsamylkin: Tests?:)

yutongp:
so the `yml` file also use struct tag `yaml`? like:
```
type Config struct {
  Value string `yaml:"value"`
}
```

dmcaulay:
people really use `yml`?

glibsm:
I vote we stick with `yaml` in light of the `base.yaml -> base.yml -> wtf.yaml`

sectioneight:
Agreed, what's the value-add here? We're taking a pretty strong stance on conventions, and I don't think anybody's personal preference for .yml versus .yaml is important enough to warrant two different naming schemes for config.

peter-edge:
I can buy that, this is just a suggestion - a lot of people (unfortunately)
use yml too (including us haha, travis.yml) and I was thinking of FX in a
pure OSS context - some might see yaml provider and expect it to pick up
yml as well.

sectioneight:
Let's add it if we get a request for it. Right now, every service at uber
uses .yaml for config.

peter-edge:
sgtm

madhuravi:
Closing this @peter-edge since we decided against it.

PR #286, opened 2017-02-18
Owner -> Manager

glibsm:


breerly:
Hmmm... a bit hesitant to overload the term Controller, which has already experienced quite a bit of semantic diffusion...

alsamylkin:
A Runner!
Or a Lifecycle.. <- my favorite.

service/service.go
```diff
@@ -81,8 +81,8 @@ type serviceConfig struct {
 	Roles       []string `yaml:"roles"`
 }
 
-// newOwner creates a service owner from a set of module creation functions and options.
-func newOwner(modules []ModuleCreateFunc, options ...Option) (Owner, error) {
```
madhuravi: Change to newController

madhuravi:
@breerly Controller was the name most of us were happy with. We do acknowledge it's a bit loaded but it captures this interface the best. Feel free to suggest alternatives if you think of any.

breerly:
Glue/MVCS uses this term already to mean a different thing - if for perhaps just this reason it's probably worth thinking of something else, since we will be presenting the user with 2 concepts using the same name.

breerly:
Of course, up to you guys.

glibsm:
Correct, that's our worry as well.

rails/expressjs background was always my concern. Most frameworks use `controller` to mean `handler`.

glibsm:
However, `Owner` is also not ideal :)

dmcaulay:
`Lifecycle`?

glibsm:
`type ThisTooReallyDifficultToNameSorry interface`

anuptalwalkar:
It's not a `Lifecycle` We need a verb. `service.Manager` is nice.

madhuravi:
@breerly good point about the conflict with glue. We want to avoid that.
We decided to rename to Manager.

service/host.go
```diff
@@ -57,9 +57,9 @@ type host struct {
 	started        bool
 }
 
-// TODO(glib): host is both an Owner and a Host?
+// TODO(glib): host is both an Manager and a Host?
 var _ Host = &host{}
-var _ Owner = &host{}
+var _ Manager = &host{}
```
madhuravi: Rename private struct to manager too.

alsamylkin:
service.Interface!

sectioneight:
service.Thing
service.WTF
service.TODO
service.ShedForBikes
service.Uber
service.Bag

PR #287, opened 2017-02-18
Add trace/span logging to YARPC

madhuravi:


modules/rpc/handler.go
```diff
@@ -38,31 +39,49 @@ type contextInboundMiddleware struct {
...
+	handler transport.UnaryHandler,
+) error {
 	stopwatch := stats.RPCHandleTimer.
 		Tagged(map[string]string{stats.TagProcedure: req.Procedure}).
 		Timer(req.Procedure).
 		Start()
 	defer stopwatch.Stop()
 
-	ctx = ulog.ContextWithLogger(ctx, ulog.Logger(ctx))
+	if span := opentracing.SpanFromContext(ctx); span != nil {
```
alsamylkin: Why we don't start the span?

tracing/tracer.go
```diff
@@ -100,3 +103,17 @@ func (jr *jaegerReporter) UpdateGauge(name string, tags map[string]string, value
 func (jr *jaegerReporter) RecordTimer(name string, tags map[string]string, d time.Duration) {
 	jr.reporter.AllocateTimer(name, tags).ReportTimer(d)
 }
+
+// WithSpan is used for generating a span to be used in testing
+func WithSpan(t *testing.T, log ulog.Log, f func(opentracing.Span)) {
+	tracer, closer, err := InitGlobalTracer(
```
alsamylkin: This is not safe for use in multiple go routines.

tracing/tracer.go
```diff
@@ -100,3 +103,17 @@ func (jr *jaegerReporter) UpdateGauge(name string, tags map[string]string, value
 func (jr *jaegerReporter) RecordTimer(name string, tags map[string]string, d time.Duration) {
 	jr.reporter.AllocateTimer(name, tags).ReportTimer(d)
 }
+
+// WithSpan is used for generating a span to be used in testing
+func WithSpan(t *testing.T, log ulog.Log, f func(opentracing.Span)) {
```
alsamylkin: Why is it in this file, not testutils package?

tracing/tracer.go
```diff
@@ -100,3 +103,17 @@ func (jr *jaegerReporter) UpdateGauge(name string, tags map[string]string, value
...
+	tracer, closer, err := InitGlobalTracer(
+		nil, "serviceName", log, metrics.NopCachedStatsReporter,
+	)
+	assert.NoError(t, err)
```
alsamylkin: require?

modules/rpc/handler_test.go
```diff
@@ -50,21 +56,46 @@ func (f fakeEnveloper) ToWire() (wire.Value, error) {
...
+			require.Equal(t, "oneway handle", err.Error())
+			checkLogForTrace(t, buf)
+		})
+	})
+}
+
+func checkLogForTrace(t *testing.T, buf *testutils.TestBuffer) {
+	require.True(t, len(buf.Lines()) > 0)
+	for _, line := range buf.Lines() {
+		assert.Contains(t, line, "traceID")
```
alsamylkin: Do we need to check the ID format?

modules/rpc/handler_test.go
```diff
@@ -50,21 +56,46 @@ func (f fakeEnveloper) ToWire() (wire.Value, error) {
...
 	}
-	err := oneway.HandleOneway(context.Background(), &transport.Request{}, &fakeOneway{t: t})
-	assert.Equal(t, "oneway handle", err.Error())
+	testutils.WithInMemoryLogger(t, nil, func(zapLogger zap.Logger, buf *testutils.TestBuffer) {
+		loggerWithZap := ulog.Builder().SetLogger(zapLogger).Build()
+		tracing.WithSpan(t, loggerWithZap, func(span opentracing.Span) {
+			ctx := ulog.ContextWithLogger(context.Background(), loggerWithZap)
+			ctx = opentracing.ContextWithSpan(ctx, span)
+			err := oneway.HandleOneway(ctx, &transport.Request{}, &fakeOneway{t: t})
+			require.Equal(t, "oneway handle", err.Error())
```
alsamylkin: require.EqualError

tracing/tracer.go
yutongp: +1. I feel like we probably need to avoid having testing stuffs expose through API (right now we have many of them) testutils sounds a right place to fit in

tracing/tracer.go
```diff
@@ -100,3 +103,17 @@ func (jr *jaegerReporter) UpdateGauge(name string, tags map[string]string, value
...
+	defer closer.Close()
+	opentracing.InitGlobalTracer(tracer)
+	defer opentracing.InitGlobalTracer(opentracing.NoopTracer{})
+	span := opentracing.GlobalTracer().StartSpan("test")
+	defer span.Finish()
```
yutongp: this is part tracing confuse me. `span.Finish` will report to tracer's reporter through a channel, but tracer is closed first and close the channel as well. However, this will still work somehow...

tracing/tracer.go
madhuravi: Fixed.

tracing/tracer.go
madhuravi: We decided as a team to minimize testutils package usage since it causes import cycles often.

modules/rpc/handler.go
madhuravi: YARPC does.

tracing/tracer.go
anuptalwalkar: internal/utils might be better place since I dont think there will be a usecase for developers for this API?

tracing/tracer.go
madhuravi: deferred calls are last in first out.

modules/rpc/handler_test.go
madhuravi: No, that does not belong in a unit test for handlers.

modules/rpc/handler.go
alsamylkin: Is it a feature, not a bug? If it is, I think we should put a reference to it.

tracing/tracer_testutils.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+// WithSpan is used for generating a span to be used in testing
+func WithSpan(t *testing.T, log ulog.Log, f func(opentracing.Span)) {
+	tracer, closer, err := createTracer(
+		nil, "serviceName", log, metrics.NopCachedStatsReporter,
+	)
+	require.NoError(t, err)
+	defer closer.Close()
```
alsamylkin: Looks like you are ignoring errors here..

tracing/tracer.go
yutongp: I don't know. I will be very confused if godoc said `WithSpan` takes in a `testing.T`, but the comment explains though. Still something I would try to avoid show up in API.

tracing/tracer.go
yutongp: I forgot that trick. But I tried called `span.Finish()` first, it still worked. ¯\_(ツ)_/¯

tracing/tracer_testutils.go
```diff
@@ -0,0 +1,43 @@
...
+package tracing
+
+import (
+	"testing"
+
+	"go.uber.org/fx/metrics"
+	"go.uber.org/fx/ulog"
+
+	"github.com/opentracing/opentracing-go"
+	"github.com/stretchr/testify/require"
```
yutongp: Oh, by having this function in a non testing file, we will need to have "github.com/stretchr/testify/require" in production import as well

tracing/tracer.go
yutongp: How about `tracing/testutils`? is that kind of overkill?

tracing/tracer_testutils.go
madhuravi: In certain cases defer is preferable. Not much we can do about the error here.

modules/rpc/handler.go
madhuravi: Ah, thought this was a well known thing. Added a comment in case it throws off people.

tracing/tracer_testutils.go
madhuravi: That's true of testutils package as well :)

yutongp: If my service is using `fx` which uses `fx/tracing` but not use `fx/tracing/testutils` in prod code, my prod dependencies will not `github.com/stretchr/testify/require` and `testing`, right?

yutongp:
yarpc has test packages like https://godoc.org/go.uber.org/yarpc/api/transport/transporttest and https://godoc.org/go.uber.org/yarpc/api/peer/peertest etc
Maybe we could reference from them? To have `fx/tracing/tracingtest`?
I feel this can help us achieve:
1. Avoid expose unusable APIs to our users
2. Avoid introduce testing dependencies in prod build
3. Avoid one big `testutils` import cycles
4. Make our code more organized

madhuravi:
@yutongp Good point, moved it under testutils/tracing since that's our convention so far.

glibsm:
Awesome!

PR #288, opened 2017-02-18
make dig example build

yutongp:


sectioneight:
@glibsm should we update the `simple` example to use `dig`?

glibsm:
@sectioneight Perhaps. I created an entirely new sample to illustrate dependency injection though. Perhaps we can name it something better than `dig`. 

I'd prefer to keep the simple example, well, simple....

PR #289, opened 2017-02-18
move uhttp/client into uhttp

yutongp:
After the middleware renaming, I feel it is quite nature to move `uhttp/client` into `uhttp`.
some of my thoughts:
1. All uhttp APIs are in one package like std http and xhttp. it will be more easier to read through the APIs.

2. `client` is a package name you may want to avoid to use for public facing. Because it is a common package name and even common variable name. an example is our doc example:
    ```golang
    client := client.New(svc)
    // client package cannot be used in this scope anymore!
    ```
   It is also not specific about what client it is. So when our users use it, they probably gonna rename it to something like `uhttpclient`.

3. `Filter`s have not naming overlap now. So it is more nature to have both `InboundMiddleware` and `OutboundMiddleware` live in the same package.

Overall, only two public APIs changed name in this PR:
`New` -> `NewClient` and `Executor` -> `OutboundExecutor`

Let's discuss about this change. Maybe there are some good reasons to keep them in `client` package.


yutongp:
#93 moved it out. May need more context from @madhuravi for why it has been moved out.

peter-edge:
This is probably painfully broken now, you might want to start over heh.

madhuravi:
Based on https://github.com/uber-go/fx/issues/298, this is not required anymore. Closing.

PR #290, opened 2017-02-18
Upgrade to zap 1.0 release candidate

akshayjshah:
This PR upgrades UberFX to use the release candidate of zap 1.0. Along the way, it migrates FX to use the base zap logger (what you were calling the typed logger). It includes a few potentially controversial changes, which I'll address head-on here. I'm happy to chat about any of this in person on Monday.

(Clearly, this can't get merged until after your code freeze. That's cool, I'd just like to get review done before I'm constantly rebasing.)

## No ulog.Log interface

Interfaces are good for a few things:

- depending on a subset of a type's methods
- allowing callers to pass third-party types without depending on those external packages
- abstracting away implementation, which lets you swap implementations later

However, the former `ulog.Log` interface doesn't accomplish any of these goals. It would have included the full `zap.SugaredLogger` API, and it doesn't prevent UberFX (and ulog) from taking a concrete dependency on zap. Most importantly, because we need users to be able to recover the strongly-typed `*zap.Logger`, it references tons of concrete zap types (`*zap.Logger`, which then references `Field` and `*CheckedEntry`). It's hard for me to imagine a reimplementation of that API outside zap.

That might still be okay, but interfaces are also harder to extend - adding even a single method is a breaking change. Given all that, it seems wiser to me to use `*zap.Logger` directly and eliminate a bunch of copypasta. Since we own both libraries, I don't see this as a significant problem. - users can and should declare narrow interfaces for just the portions of the logger that they're using.

## No Sentry integration

This is purely a time issue. If you can't wait for the Filebeat plugin we've discussed, I'm happy to resurrect direct Sentry integration in a follow-on PR. The latest zap should allow us to do that without so many map copies and with zero impact on non-error log performance.

## No loggers in contexts

I was a little confused by the existing code. It looks to me like we're largely working with a single global logger, which starts off with some default configuration and is reconfigured on service startup. We're supplying that logger to users by putting it into the request context and providing `ulog.Logger(ctx context.Context) ulog.Logger` as a getter. Where possible, we add tracing information to the logger before putting it into the context.

In all that, I'm not sure whether the logger *needs* to be in the context at all. We can keep `ulog.Logger`, and add just add the context's tracing information to the logger when it's retrieved. This is quite inexpensive, and it's much simpler.

## No log builder

I don't think that this is required any more either; we can just use zap's config shapes instead. They already come with sensible default values, and we can modify the service scaffold to comply with our internal logging schemas.

## Documentation

I didn't touch the ulog README much - I'll leave that for someone on your team.


akshayjshah:
Argh, there's already a massive conflict with Anup's latest logging refactoring. I'll resolve all the conflicts later this weekend.

auth/uauth.go
```diff
@@ -48,7 +48,7 @@ var (
 // CreateAuthInfo interface provides necessary data
 type CreateAuthInfo interface {
 	Config() config.Provider
-	Logger() ulog.Log
+	Logger() *zap.Logger
```
sectioneight: I think this is going to be the most contentious part cc @shawnburke who deals with the same issue in Catalyst. I'm sure your 1.x series will maintain API compatibility but we've been trying to introduce as many shim layers as possible (caveat: you'll see nearby we're exposing tally directly, which we may decide not to) to provide **you** the flexibility to change your APIs without worrying about breaking UberFx consumers.

examples/keyvalue/server/observer.go
```diff
@@ -32,7 +32,7 @@ type Observer struct {
 
 // OnInit is called during service init process. Returning an error halts the init?
 func (o *Observer) OnInit(svc service.Host) error {
-	svc.Logger().Info(
+	svc.Logger().Sugar().Infow(
```
sectioneight: `Sugar()` is low enough overhead/memoized that we don't need a `svc.SugaredLogger()`? Cool!

glide.yaml
```diff
@@ -1,8 +1,7 @@
 package: go.uber.org/fx
 import:
-- package: github.com/uber-go/zap
-  # Pinned temporarily while zap reaches 1.0 (https://github.com/uber-go/fx/pull/243)
-  version: c064b5c44b285a7e2fd5c9b26e8c38228ce2bccb
+- package: go.uber.org/zap
+  version: v1.0.0-rc.1
```
sectioneight: Any release timeline on 1.0? We had an incredibly painful amount of churn with yarpc 1.0 RC candidates, both for my team and our consumers.

modules/rpc/handler.go
```diff
@@ -85,7 +89,7 @@ func (a authOnewayInboundMiddleware) HandleOneway(ctx context.Context, req *tran
 func authorize(ctx context.Context, host service.Host) (context.Context, error) {
 	if err := host.AuthClient().Authorize(ctx); err != nil {
 		stats.RPCAuthFailCounter.Inc(1)
-		ulog.Logger(ctx).Error(auth.ErrAuthorization, "error", err)
+		ulog.Logger(ctx).Error(auth.ErrAuthorization, zap.Error(err))
```
sectioneight: Oh man, @akshayjshah, you really went above and beyond here. We have an action item to convert to unsugared usage internally. Really appreciate this!

modules/rpc/handler.go
```diff
@@ -110,7 +114,7 @@ func (p panicOnewayInboundMiddleware) HandleOneway(ctx context.Context, req *tra
 func panicRecovery(ctx context.Context) {
 	if err := recover(); err != nil {
 		stats.RPCPanicCounter.Inc(1)
-		ulog.Logger(ctx).Error("Panic recovered serving request", "error", errors.Errorf("panic in handler: %+v", err))
+		ulog.Logger(ctx).Error("Panic recovered serving request.", zap.Any("error", err))
```
sectioneight: Is this a thing? Periods after messages? If so, we should document somewhere and communicate with my team.

modules/uhttp/client/client_filters_test.go
```diff
@@ -74,7 +74,7 @@ func TestExecutionChainFiltersError(t *testing.T) {
 }
 
 func withOpentracingSetup(t *testing.T, registerFunc auth.RegisterFunc, fn func(tracer opentracing.Tracer)) {
-	tracer, closer, err := tracing.InitGlobalTracer(&jconfig.Configuration{}, "Test", ulog.NopLogger, metrics.NopCachedStatsReporter)
+	tracer, closer, err := tracing.InitGlobalTracer(&jconfig.Configuration{}, "Test", zap.New(nil), metrics.NopCachedStatsReporter)
```
sectioneight: all these `zap.New(nil)` calls are starting to stand out as a smell. Any way we can rework it so that a ZapCore is a variadic option and you scan for it first? Just a general comment any time I see `nil` as a required param repeatedly.

modules/uhttp/filters.go
```diff
@@ -55,8 +55,7 @@ type contextFilter struct {
 }
 
 func (f contextFilter) Apply(w http.ResponseWriter, r *http.Request, next http.Handler) {
-	ctx := ulog.NewLogContext(r.Context())
-	next.ServeHTTP(w, r.WithContext(ctx))
+	next.ServeHTTP(w, r)
```
sectioneight: Need @anuptalwalkar to explain why we no longer need to wrap the request with the incoming context. If we don't, then this `contextFilter` seems like a noop and should be removed, right?

modules/uhttp/filters_test.go
```diff
@@ -160,10 +160,13 @@ func testTracingFilterWithLogs(t *testing.T) {
 		var tracecount = 0
 		var spancount = 0
 		for _, line := range buf.Lines() {
-			if strings.Contains(line, "traceID") {
+			// TODO: This violates encapsulation; tests outside the ulog
```
sectioneight: I'll let you work this out with @madhuravi

service/host.go
```diff
@@ -199,7 +206,7 @@ func (s *host) AddModules(modules ...ModuleCreateFunc) error {
 		if !s.supportsRole(mi.Roles...) {
 			s.Logger().Info(
 				"module will not be added due to selected roles",
-				"roles", mi.Roles,
+				zap.Any("roles", mi.Roles),
```
sectioneight: no `zap.SliceAny`? Is there an overhead when we have to later check that it's a slice or are you going to just %v it?

service/host.go
```diff
@@ -158,15 +165,15 @@ func (s *host) shutdown(err error, reason string, exitCode *int) (bool, error) {
...
+			s.Logger().Error("Failure to close metrics", zap.Error(err))
 		}
 	}
 
 	// Flush tracing buffers
 	if s.tracerCloser != nil {
 		s.Logger().Debug("Closing tracer")
 		if err = s.tracerCloser.Close(); err != nil {
-			s.Logger().Error("Failure to close tracer", "error", err)
+			s.Logger().Error("Failure to close tracer", zap.Error(err))
```
sectioneight: TODO add period here?

service/host.go
```diff
@@ -199,7 +206,7 @@ func (s *host) AddModules(modules ...ModuleCreateFunc) error {
 		if !s.supportsRole(mi.Roles...) {
 			s.Logger().Info(
 				"module will not be added due to selected roles",
```
sectioneight: TODO(my team) capitalize "Module"

service/host.go
```diff
@@ -278,9 +285,12 @@ func (s *host) start() Control {
 
 				s.shutdownMu.Unlock()
 				if _, err := s.shutdown(e, "", nil); err != nil {
-					s.Logger().Error("Unable to shut down modules", "initialError", e, "shutdownError", err)
+					s.Logger().Error("Unable to shut down modules",
```
sectioneight: No period? Perhaps I'm misunderstanding the guideline here.

service/host.go
```diff
@@ -278,9 +285,12 @@ func (s *host) start() Control {
...
+						zap.String("initialError", e.Error()),
+						zap.String("shutdownError", err.Error()),
```
sectioneight: We're going to lose the cause/stacktrace of these errors by turning them into Strings, right?

service/host.go
```diff
@@ -400,7 +413,11 @@ func (s *host) transitionState(to State) {
 
 	// TODO(ai) this isn't used yet
 	if to < s.state {
-		s.Logger().Fatal("Can't down from state", "from", s.state, "to", to, "service", s.Name())
+		s.Logger().Fatal("Can't down from state",
```
sectioneight: Again, same question about periods. Seems like maybe we should just leave them out and back out the changes towards the top to remain consistent

auth/uauth.go
akshayjshah: Let's discuss this in person. I Have a lot of reservations about top-level frameworks own interfaces for logging and metrics; it forces components far down the stack to depend on the framework, instead of the reverse.

examples/keyvalue/server/observer.go
akshayjshah: Yep, its one allocation. Not free, but not a big deal if you're using the sugared API. If UberFx gets to a point where it's counting single allocations, memorizing this may make sense.

sectioneight:
Ya, this is definitely not black-and-white, definitely should hammer out in
person. My primary concern is stability, not ownership. If we can keep
pushing new versions of Fx without *ever* breaking people using our logging
mechanism, then my vote is for whatever mechanism makes that easiest.

examples/keyvalue/server/observer.go
sectioneight: How do you feel about `svc.TypedLogger()` and `svc.Logger()` being sugared? I'm on the fence as to how we want to expose this, but my gut (and my discussions outside of uber) is that most people find the verbosity of `zap.String()` etc too high, and will opt for typed only when perf is a concern.

glide.yaml
akshayjshah: Typical for an RC - depends on how many things I got wrong :)

I'm hoping to update some other internal projects this week and cut a 1.0 final before the end of the month. The APIs for the loggers themselves are very, very unlikely to change; more likely that some portion of the constructors changes. 

I misspelled a JSON key, so there's and RC.2 coming Monday. Shouldn't affect you guys.

auth/uauth.go
sectioneight: Meant to post to this comment, but replied to the email instead so it's at the bottom. Copy-paste:

Ya, this is definitely not black-and-white, definitely should hammer out in
person. My primary concern is stability, not ownership. If we can keep
pushing new versions of Fx without **ever** breaking people using our logging
mechanism, then my vote is for whatever mechanism makes that easiest.

modules/rpc/handler.go
akshayjshah: :+1:

sectioneight:
With regards to your comment:

```
No loggers in contexts

I was a little confused by the existing code. It looks to me like we're largely working with a single global logger, which starts off with some default configuration and is reconfigured on service startup. We're supplying that logger to users by putting it into the request context and providing ulog.Logger(ctx context.Context) ulog.Logger as a getter. Where possible, we add tracing information to the logger before putting it into the context.

In all that, I'm not sure whether the logger needs to be in the context at all. We can keep ulog.Logger, and add just add the context's tracing information to the logger when it's retrieved. This is quite inexpensive, and it's much simpler.
```

It's always seemed weird to me that the logger object was stored as a value on the context. If we can just extract the tracing when we need it, it seems much cleaner to not add additional metadata. I tried to work this out with @anuptalwalkar and @alsamylkin, but they had some technical reasons why it was necessary to store on the context.

glide.yaml
sectioneight: Cool. Again though, this makes me think we should keep a ulog.Logger even if it's currently just the exact same interface as a zap.Logger so that we can upgrade to v1 without breaking users.

auth/uauth.go
akshayjshah: Agreed.

examples/keyvalue/server/observer.go
akshayjshah: Let's defer pending the interface discussion. If we end up not putting a facade in front of logging, we can do away with this API altogether - it's just a roundabout way of giving people access to `zap.L` and `zap.S`, which FX startup configures.

modules/rpc/handler.go
akshayjshah: Sorry, broke with your convention out of habit. My strong preference is for comments and log messages to be complete sentences - initial capitals, proper spelling and grammar, and terminal punctuation. I think I'm out of the mainstream on this though.

modules/uhttp/client/client_filters_test.go
akshayjshah: Sure, makes sense.

modules/uhttp/filters.go
akshayjshah: Probably. If it's okay with you, I'd rather keep this PR restricted to the logging. I'm going to be rebasing endlessly as it is.

service/host.go
akshayjshah: Not needed - `Any` handles all those cases. The only overhead is allocating to cast the slice to `interface{}`. They're all serialized properly, no %v needed. 

If you're okay with the one extra interface allocation, just using zap.Any for every field is a big perf win over the SugaredLogger. It's exactly what sugaring does under the hood - it's zipping the keys and values into Any.

modules/uhttp/filters.go
sectioneight: Makes sense. We'll task up to remove this, since it's pretty clearly useless now.

service/host.go
akshayjshah: Possibly. I'll take another pass to make sure we don't lose any info. 

I'm not going to take a dependency on Daves error package in zap (it's pre-1.0), but I should still be able to support these errors with `Error`.

akshayjshah:
I'm out on Monday, but will try to get the remaining nits ironed out by Tuesday. Mind scheduling something for us to chat Tuesday morning?

sectioneight:
Sure. What's the scope? Fascade interfaces in general, review of this PR,
scheduling/timing/ownership?

sectioneight:
cc @shawnburke we should (maybe?) take a consistent stance on fascade interfaces between UberFx and Catalyst, otherwise we could end up with some painful conversion functions. Would you like to join the meeting, or have a separate discussion w.r.t. UberFx+Catalyst once we've figured out our strategy on zap/tally/etc?

sectioneight:
Also lolin at the merge conflicts. For such a small codebase, it's pretty amazing how bad this conflicted:

image

modules/uhttp/client/client_filters_test.go
glibsm: Consider `zap.Nop` instead of `zap.New(nil)` for NopCore. 
I actually had to pull up zap to see what it meant to pass in nil

modules/rpc/handler.go
yutongp: Golang comments expect a period in the end: https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences
I have seen we missed period in a few places.

akshayjshah:
> Sure. What's the scope? Fascade interfaces in general, review of this PR,
scheduling/timing/ownership?

I'd definitely like to get through this PR, and ideally also figure out the facades & interface ownership aspects.

modules/uhttp/client/client_filters_test.go
akshayjshah: Yep, fair point. `New` supports nil purely for safety (since I can't stop people from passing nil interfaces, this seemed like the best way to not add an error return). uber-go/zap#326 adds an explicit `NewNop`.

modules/uhttp/filters.go
anuptalwalkar: I want to keep this. We had this filter to keep the logger inside context and making it available for users. Primary reason is to keep adding fields that are not provided by zap. Logger(ctx) provides default fields from zap configuration (host, time, etc), but some of the fields come from fx. Right now, you have `module: "http"`, service name, traceID, spanID  with the logger available in the user handler. 

To answer the other question on why Logger is on the context and not fields. I see that you have added Logger(ctx) to pull all fields from context instead of returning configured logger. This seems efficient right now, but it will be O(n) operation n being number of fields in the context to make the logger rich enough for our needs. There is this slight performance benefit of having Logger in the context since most of the codebase is literally doing 'ulog.Logger(ctx)' on every logging step. What do you think? We can comeback to perf later since either way will not break any APIs.

ulog/log.go
```diff
@@ -20,254 +20,24 @@
...
+	cfg := zap.NewProductionConfig()
+	cfg.OutputPaths = []string{"stdout"}
+	return cfg
+}
+
+// SetLogger uses the provided logger to replace zap's global loggers and
+// hijack output from the standard library's "log" package. It returns a
+// function to undo these changes.
+func SetLogger(log *zap.Logger) func() {
+	undoGlobals := zap.ReplaceGlobals(log)
```
anuptalwalkar: This is sweet!

auth/uauth.go
anuptalwalkar: I started email thread before this PR, and Im curious too. I have been asked this question multiple times that why fx is so tightly integrated with zap.Logger, and can it be used with different logger implementation.

modules/rpc/handler.go
sectioneight: Comments !== log messages

peter-edge:
I really really really like this PR. This is going to break like, everything, ever, but I don't mind at all. Totally worth it.

akshayjshah:
I suspect that, at the end of the day, a bunch of the migration pain can be fixed with `gofmt -r` rules. I'll play with it a bit and write a little script for users.

auth/uauth.go
peter-edge: Ack ya I am starting to really like zap, and am super impressed with it, but...I have to agree, I hate when frameworks/libraries tightly link a logging library. I wish we could even document the `Typed()` function to return nil if zap is not being used, or something a little less unsafe.

modules/rpc/handler.go
peter-edge: I think I'm agreeing with Aiden 2-0 so far in the Aiden/Akshay "Battle of the PR" here. I actually don't even like sentences, I like "Events", and wish they were somehow documented. Think `request_panic_recovered`.

glide.yaml
akshayjshah: Any breaking change in the RC will also break your interface, so I'm not sure what this buys you.

I've had this discussion about interfaces with Peter and Josh before, and it's nuanced enough that we should take it out of PR comments. I'll send you a short proposal, and we can discuss in person. I'm certainly not blocking you from shipping what you want in FX, but I don't think that this interface will be helpful.

modules/rpc/handler.go
akshayjshah: Sure. Don't think that this is a place we need consensus.

modules/uhttp/filters.go
akshayjshah: In general, I'm unsure why UberFX is now treating loggers differently from other dependencies. To me, the most reasonable thing to do is to use DIG to inject loggers into modules; *that's* the efficient place to add static context like the service name and module name. If you're doing that on the request path, you've already lost the performance game - saving 300ns by storing the logger on the context is penny-wise and pound-foolish.

Separately, I understand that you want to make it easy (or possibly even required) for developers to include span and trace IDs in their logs. I'm all for this - as a simple field type. This is easy because it's *consistent* - tracing information is just like any other field, and doesn't require me to use some special API. It also doesn't require cooperation from my application's middleware, so `ulog` is actually independent of FX's HTTP server.

```
logger.Debugw("failed calling populous", ulog.Trace(ctx))
```

akshayjshah:
Rebased on master, addressed some nits. This go-round leaves all log messages exactly as-is :)

ulog/context.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func (t trace) MarshalLogObject(enc zapcore.ObjectEncoder) error {
+	span := opentracing.SpanFromContext(t.ctx)
+	if span == nil {
+		return nil
+	}
+	j, ok := span.Context().(jaeger.SpanContext)
+	if !ok {
+		return nil
```
alsamylkin: This looks weird..

ulog/metrics.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/uber-go/tally"
+	"go.uber.org/zap/zapcore"
+)
+
+// Metrics returns a function that counts the number of logs emitted by level.
+//
+// To register it with a zap.Logger, use zap.Hooks.
+func Metrics(s tally.Scope) func(zapcore.Entry) error {
+	// Avoid allocating strings and maps in the request path.
+	debugC := s.Tagged(map[string]string{"level": "debug"}).Counter("logs")
```
alsamylkin: can we simplify it even further and put counters in a map, so returning function will probably have only one if?

ulog/context.go
akshayjshah: Can you elaborate a bit? If there's no Jaeger span available, there's nothing to do; it's not an error, it's an expected case.

ulog/metrics.go
akshayjshah: This is faster - why take on a hash map lookup when we can use a short `switch` instead? For straightforward, well-encapsulated, often-executed code, I don't think that the brevity gains are worth it.

Of course, y'all own this code - I'll change if you really want me to.

alsamylkin: Well, this is because Zap has weird constants :) otherwise we can use arrays and be on par.
I did a simple benchmark and maps are 3 times slower, so it is good to keep it.

ulog/context.go
alsamylkin: I am confused, why is it an expected case?

alsamylkin:
My stamp is not binding to merge, I agree with the changes, but we need to figure out sentry(and fix lint errors) to merge.

tracing/tracer_test.go
```diff
@@ -67,34 +62,33 @@ func TestInitGlobalTracer_Disabled(t *testing.T) {
...
 func TestLoadAppConfig_JaegerConfigWithLogger(t *testing.T) {
-	jConfig := loadAppConfig(_jaegerConfigWithLogger, getLogger())
+	jConfig := loadAppConfig(_jaegerConfigWithLogger, zap.L())
 	assert.NotNil(t, jConfig)
 	assert.Equal(t, jaeger.NullLogger, jConfig.Logger)
 }
 
 func TestLoadAppConfig_NilJaegerConfig(t *testing.T) {
-	jConfig := loadAppConfig(nil, getLogger())
+	jConfig := loadAppConfig(nil, zap.L())
```
anuptalwalkar: zap.L and zap.S are a bit confusing. Not blocking this PR, but we may need some verbose method here.

akshayjshah: Yep, these are extremely short by design. Many logging packages have top-level `Error`, `Infof`, etc. functions; I really, really don't want these in zap, which means that we need a minimally noisy way to get the global logger. `zap.GlobalLogger().Info("something")` seems like a lot; `zap.L().Info("something")` is much less intrusive.

This is definitely a matter of taste, though - file an issue and we can discuss pros and cons there.

anuptalwalkar:
Merging the PR, since conflicts are resolved here and offline.

sectioneight:
:heart:

glibsm:
leerooooooooooyyyyyyyyy

PR #291, opened 2017-02-18
Expose deafult DI graph

alsamylkin:
Exposing the default graph helps us with testing(especially parallel) external packages, it allows us to store the graph reference in the object itself and e.g. override it with options.

### DIFF
--- a/dig/default.go
+++ b/dig/default.go
@@ -20,7 +20,7 @@
 
 package dig
 
-// defaultGraph is used for all the top-level inject calls
+// Default graph is used for all the top-level inject calls
 var defaultGraph = newGraph()
 
 // DefaultGraph returns the graph used by top-level inject calls
--- a/dig/default.go
+++ b/dig/default.go
@@ -20,30 +20,30 @@
 
 package dig
 
-// Default graph is used for all the top-level inject calls
-var defaultGraph = newGraph()
+// DefaultGraph is used for all the top-level inject calls
+var DefaultGraph Graph = newGraph()
 
 // Register into the default graph
 func Register(i interface{}) error {
-	return defaultGraph.Register(i)
+	return DefaultGraph.Register(i)
 }
 
 // RegisterAll into the default graph
 func RegisterAll(is ...interface{}) error {
-	return defaultGraph.RegisterAll(is...)
+	return DefaultGraph.RegisterAll(is...)
 }
 
 // Resolve an object through the default graph
 func Resolve(i interface{}) error {
-	return defaultGraph.Resolve(i)
+	return DefaultGraph.Resolve(i)
 }
 
 // ResolveAll the passed in pointers through the dependency graph
 func ResolveAll(is ...interface{}) error {
-	return defaultGraph.ResolveAll(is...)
+	return DefaultGraph.ResolveAll(is...)
 }
 
 // Reset the default graph
 func Reset() {
-	defaultGraph.Reset()
+	DefaultGraph.Reset()
 }
--- a/dig/default.go
+++ b/dig/default.go
@@ -20,30 +20,35 @@
 
 package dig
 
-// DefaultGraph is used for all the top-level inject calls
-var DefaultGraph Graph = newGraph()
+// defaultGraph is used for all the top-level inject calls
+var defaultGraph = newGraph()
+
+// DefaultGraph returns the graph used by top-level inject calls
+func DefaultGraph() Graph {
+	return defaultGraph
+}
 
 // Register into the default graph
 func Register(i interface{}) error {
-	return DefaultGraph.Register(i)
+	return defaultGraph.Register(i)
 }
 
 // RegisterAll into the default graph
 func RegisterAll(is ...interface{}) error {
-	return DefaultGraph.RegisterAll(is...)
+	return defaultGraph.RegisterAll(is...)
 }
 
 // Resolve an object through the default graph
 func Resolve(i interface{}) error {
-	return DefaultGraph.Resolve(i)
+	return defaultGraph.Resolve(i)
 }
 
 // ResolveAll the passed in pointers through the dependency graph
 func ResolveAll(is ...interface{}) error {
-	return DefaultGraph.ResolveAll(is...)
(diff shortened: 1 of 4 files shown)


dig/default.go
```diff
@@ -20,30 +20,30 @@
 
 package dig
 
-// Default graph is used for all the top-level inject calls
-var defaultGraph = newGraph()
+// DefaultGraph is used for all the top-level inject calls
+var DefaultGraph Graph = newGraph()
```
yutongp: I am little bit not sure about this change. Allowing other package has the ability to swap/overwrite global singleton, won't it introduce more chance of race? Like the one we saw in Parallel tests with global tracer. I am also worried some users don't know what they are doing gonna overwrite it by mistake and get into random runtime error.

I may need some help to understand why expose this variable make Parallel test easier as well...

dig/dig.go
```diff
@@ -157,7 +157,7 @@ func (g *graph) Reset() {
 	g.Lock()
 	defer g.Unlock()
 
-	defaultGraph.nodes = make(map[interface{}]object)
+	g.nodes = make(map[interface{}]object)
```
yutongp: this bug fix may deserve it own PR

dig/default.go
yutongp: aren't the five global functions(Register, Resolve etc) already exposed all methods defaultGraph interface can use?

alsamylkin: I envisioned testing other packages: constructors can accept an option to use a different graph to resolve objects, but what is the default graph for them?
For example, for uhttp.Client, we can load auth/tracer from a default graph, so users will not need to worry about parameters. But how would we test for auth errors? By injecting the injector(with options)..
An example, where a default global is used is http.DefaultClient.

dig/dig.go
alsamylkin: I am just fixing them along the way. We have too many bugs to open a PR for each bug :)

alsamylkin:
@sectioneight that's why I've added him as a reviewer. Why you don't trust me?:)

dig/default.go
alsamylkin: As per overwriting a global singleton: I can create function that will return a defaultGraph..

sectioneight:
Yea, I do trust you. I was trying to remove myself as an approver but it wasn't showing up in the UI. Now it is.

dig/default.go
yutongp: how about something like this:
``` golang
func New(info auth.CreateAuthInfo, middleware ...OutboundMiddleware) *http.Client
func new(info auth.CreateAuthInfo, digGraph dig.Graph, middleware ...OutboundMiddleware) *http.Client
```
We use the defaultGraph in public API, when we test it we use the private construct function to hijack dependencies.

`http.DefaultClient` is a good point, I never thought that before. I guess things like `http.DefaultClient` and  `http. DefaultTransport` probably are not be changed in most use cases. If golang has immutable data structure, I vote them to be immutable.
But `dig.Graph` cannot be immutable unless it works like `context.Context` which returns a new copy every time after mutation. That may not be a terrible idea.

yutongp: The example I gave won't work.
need to be something like
``` golang
type resolveFunc func(interface{}) error
func New(info auth.CreateAuthInfo, middleware ...OutboundMiddleware) *http.Client {
     return new(info, dig.Resolve, middleware...)
}

func new(info auth.CreateAuthInfo, resolveFn resolveFunc, middleware ...OutboundMiddleware) *http.Client {...}
```

I hope dig can expose a `Resolver` interface or only return copy

dig/dig.go
glibsm: interesting.

nice catch

glibsm:
I like the change to expose a function rather than making the variable global. It still means people could mess with the pointer and all that, but it's significantly harder

dig/default.go
alsamylkin: I was hoping to make something like:
```golang
const _middlewareKey = "_httpClientMiddleware"
const _graphKey = "_httpClientGraph"

func WithOutbound(middleware... OutboundMiddleware) modules.Option{
	return func(info *service.ModuleCreateInfo) error {
		info.Items[_middlewareKey] = append(info.Items[_middlewareKey].([]OutboundMiddleware, middleware...)
		return nil
	}
}

func WithGraph(graph dig.Graph) modules.Option {
	return func(info *service.ModuleCreateInfo) error {
		info.Items[_graphKey] = graph
		return nil
	}
}

func New(options... modules.Option) *http.Client {
	var info service.ModuleCreateInfo
	for _, opt := range options {
		opt(info)
	}

	graph := dig.DefaultGraph()
	if g, ok := info.Items[_graphKey]; ok {
		graph = g.(dig.Graph)
	}

	middleware := make([]OutboundMiddleware, 0, len(options) + 2)

	var tracer opentracing.Tracer
	if err := graph.Resolve(&tracer); err != nil {

	}

	middleware = append(middleware, tracingOutbound(*tracer), authenticationOutbound(info))
}
```

This way a default client will be created without any extra parameters that users need to pass, but can override almost anything they want. `uhttp.NewClient()` is better than `uhttp.NewClient(defaultAuth, graph)`

yutongp: that's pretty nit. I like that

yutongp:
@glibsm how about expose a `Resolver` interface like
```golang
type Resolver interface {
	// Resolve the dependencies of the object and populate the pointer value
	Resolve(interface{}) error

	// ResolveAll the passed in pointers through the original dependency graph
	// Returns the first error encountered
	ResolveAll(...interface{}) error
}
```
So when caller can pass in a `Resolver` only, caller don't need to worry about downstream mess up the Graph.

Or how about make `Graph` only return copy after `Register` like `context.Context`? So downstream can still call `Register` for their need, but caller can safely use it's own `Graph`.

glibsm:
@yutongp "forking" the graph on each `Register` will defeat the point of using a shared injection graph, because then like context you will have to pass it around everywhere with "new values" rather than holding on to the original reference

yutongp:
@glibsm That makes sense. Thank you!

PR #292, opened 2017-02-18
Prepare documentation for beta launch

sectioneight:


sectioneight:
Obviously, don't merge this until Monday

CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
+# Changelog
+
+## v1.0.0-beta1 (20 Feb 2017)
```
madhuravi: Why beta1? v1.0.0-beta instead?
Another option is to go with your strategy of tracking master and we can release a version when we have our 1.0.0 release.
I know this version is for users to play with and send feedback but some of them might start building services on this and pin it to v1.0.0-beta.
I have a slight preference for keeping it at master for now until our 1.0.0 release.

sectioneight: Because I expect we will have a beta2 release before we reach 1.0 :)

madhuravi: What do you think about the second part of my comment? master tracking until 1.0.0

sectioneight: After recent issues (e.g. me trying to fix cherami yesterday) I think having a tag to point at (instead of "master", which glide seems to sometimes give you an old version) will be much more useful. Otherwise I worry that people will say `version: master` but really be on a pre-beta commit. Essentially this is to reduce support burden, so we can ensure everybody's on the beta we expect them to be using.

madhuravi: sounds good.

PR #294, opened 2017-02-19
dig internal cleanup

glibsm:
Wanted to PR this on it's own, to reduce the noise from the upcoming improvements to resolve some issues.

`dig.go` has grown rather large, so i split up some functionality out into their respective files. There is changes in functionality, purely moving things around.

I also recognize that dig might be a little difficult to work with internally, I will consider a more significant refactor (no externally visible changes) post-beta


dig/graph.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	return nil
+}
+
+// Reset the graph by removing all the registered nodes
+func (g *graph) Reset() {
+	g.Lock()
+	defer g.Unlock()
+
+	defaultGraph.nodes = make(map[interface{}]graphNode)
```
alsamylkin: inherited the bug..

dig/node.go
```diff
@@ -0,0 +1,131 @@
...
+	deps        []interface{}
+}
+
+// Call the function and return the result
+func (n *funcNode) value(g *graph) (reflect.Value, error) {
+	if n.cached {
+		return n.cachedValue, nil
+	}
+
+	cv := reflect.ValueOf(n.constructor)
```
alsamylkin: Doesn't seem to be used for a long time, till line 111

dig/graph.go
glibsm: oh man this is bad :(

dig/node.go
glibsm: yes, but any time you write reflect related function, you almost always need type and value, one or multiple times. So i got into a habit of doing that at the top, i don't see much harm in that

PR #295, opened 2017-02-19
Prevent dig cycles

### LINKED ISSUE #282: dig: prevent cycles
At the moment, it's possible to `Register` a cyclical graph.
This should be defected, perhaps at registration time.

glibsm:
This is a must for beta, because it has the ability to overflow panic the stack if types are not registered correctly.

Fixes #282

### DIFF
--- a/dig/dig_test.go
+++ b/dig/dig_test.go
@@ -188,15 +188,18 @@ func TestConcurrentAccess(t *testing.T) {
 func TestCycles(t *testing.T) {
 	type Type1 interface{}
 	type Type2 interface{}
+	type Type3 interface{}
 	c1 := func(t2 Type2) Type1 { return nil }
-	c2 := func(t1 Type1) Type2 { return nil }
+	c2 := func(t3 Type3) Type2 { return nil }
+	c3 := func(t1 Type1) Type3 { return nil }
 
 	g := testGraph()
 
 	require.NoError(t, g.Register(c1))
+	require.NoError(t, g.Register(c2))
 
-	err := g.Register(c2)
-	require.Contains(t, err.Error(), "unable to register dig.Type2")
+	err := g.Register(c3)
+	require.Contains(t, err.Error(), "unable to register dig.Type3")
 	require.Contains(t, err.Error(), "cycle")
 }
 
--- a/dig/dig_test.go
+++ b/dig/dig_test.go
@@ -194,10 +194,10 @@ func TestCycles(t *testing.T) {
 	g := testGraph()
 
 	require.NoError(t, g.Register(c1))
-	require.NoError(t, g.Register(c2))
 
-	var r Type1
-	require.NoError(t, g.Resolve(&r))
+	err := g.Register(c2)
+	require.Contains(t, err.Error(), "unable to register dig.Type2")
+	require.Contains(t, err.Error(), "cycle")
 }
 
 func TestResolveAll(t *testing.T) {
--- a/dig/dig_test.go
+++ b/dig/dig_test.go
@@ -185,6 +185,21 @@ func TestConcurrentAccess(t *testing.T) {
 	}
 }
 
+func TestCycles(t *testing.T) {
+	type Type1 interface{}
+	type Type2 interface{}
+	c1 := func(t2 Type2) Type1 { return nil }
+	c2 := func(t1 Type1) Type2 { return nil }
+
+	g := testGraph()
+
+	require.NoError(t, g.Register(c1))
+	require.NoError(t, g.Register(c2))
+
+	var r Type1
+	require.NoError(t, g.Resolve(&r))
+}
+
 func TestResolveAll(t *testing.T) {
 	t.Parallel()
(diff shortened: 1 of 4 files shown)


dig/graph.go
```diff
@@ -183,5 +183,38 @@ func (g *graph) registerConstructor(c interface{}) error {
...
+func (g *graph) recursiveDetectCycles(n graphNode, visited map[string]bool) error {
+	if visited[n.id()] {
+		return errCycle
+	}
+
+	visited[n.id()] = true
+
+	for _, dep := range n.dependencies() {
+		if node, ok := g.nodes[dep]; ok {
+			if err := g.recursiveDetectCycles(node, visited); err != nil {
```
alsamylkin: Can we dump the full loop in the error with errors.Wrap?

dig/graph.go
```diff
@@ -183,5 +183,38 @@ func (g *graph) registerConstructor(c interface{}) error {
 	}
 
 	g.nodes[objType] = &n
+
+	// object needs to be part of the graph to properly detect cycles
+	if cycleErr := g.detectCycles(&n); cycleErr != nil {
+		// if the cycle was detected delete from the graph
+		delete(g.nodes, objType)
```
madhuravi: Register can be called at runtime though. If we stick to our philosophy of not calling panic outside of service startup, this should return error.

madhuravi: Adding and then deleting is a bit inefficient. Why not detect cycles first and move line 185 to the end of the function?

dig/dig_test.go
```diff
@@ -185,6 +185,21 @@ func TestConcurrentAccess(t *testing.T) {
 	}
 }
 
+func TestCycles(t *testing.T) {
```
madhuravi: Can you add a test with one more level in between? Make sure it can resolve a loop with more elements in it.

dig/graph.go
madhuravi: +1 that would be cool

dig/graph.go
anuptalwalkar: Not saying that its not the right thing, but whats the outcome of returning an error? This error will be returned, and you have a `nil pointer dereference` at the location where the resolved object is accessed. npd's are pretty vague to debug and doesn't help in understanding the cause. I think Register should be only accessed on initialization, may be not via exported method.

dig/graph.go
```diff
@@ -183,5 +183,38 @@ func (g *graph) registerConstructor(c interface{}) error {
...
+}
+
+// When a new constructor is being inserted, detect any present cycles
+func (g *graph) detectCycles(n *funcNode) error {
+	visited := make(map[string]bool)
+	return g.recursiveDetectCycles(n, visited)
+}
+
+// DFS and tracking if same node is visited twice
+func (g *graph) recursiveDetectCycles(n graphNode, visited map[string]bool) error {
```
madhuravi: When used as a set and you don't really care about the value (just existence of element in map), empty structs are more efficient. Zero size. map[string]struct{}

dig/graph.go
madhuravi: Yea I see your point as well. I like the approach of keeping Register on initialization. Worth further discussion.

glibsm: dig doesn't panic, period. this forces service owners to make their own decisions about what to do if register/resolve failed

glibsm: there is a comment above, object has to be part of the graph to detect cycles. you can't detect a cycle if you can't trace back to this object in the graph

dig/graph.go
glibsm: noted

dig/graph.go
glibsm: Yes, it's doable. I will leave a TODO and come back to it though so this can be merged. Better error is not as big of a value add as having this merged in the first place

dig/dig_test.go
glibsm: done

dig/graph.go
madhuravi: Got it, that does introduce a risk when this is executed in parallel. You might have to lock cycle detection function to only one execution at a time. One alternative to getting around that is treat the root node specially, on line 212 check g.nodes[dep] or root, check for cycles.

glibsm: This code does not execute in parallel, there is a lock on entirety of any register or resolve call

glibsm:
It's unclear to me if there are any blocking issues or not. @madhuravi if you have any concerns, please let me know. Otherwise I plan to merge this in a few

dig/graph.go
madhuravi: Ah is that for the time being or we're always planning to lock register? Can't it be executed in parallel if access to the graph map is locked?

glibsm:
a-ha well there is my answer

dig/graph.go
glibsm: I think the entirety of the register call should always be locked, I wasn't just being lazy with a big lock :)

PR #297, opened 2017-02-19
[WIP]Use DIG to create uhttp client

alsamylkin:
Simplify client construction and use a DI graph to resolve auth and tracing.
If they are not present, an error is returned.

Constructor now accepts 2 options: di graph and middleware,
which can be used to override auth/tracing and add extra middleware.

TODO: Put tracer and auth in the default graph.

### DIFF
--- a/modules/uhttp/client/client.go
+++ b/modules/uhttp/client/client.go
@@ -26,22 +26,52 @@ import (
 
 	"github.com/opentracing/opentracing-go"
 	"go.uber.org/fx/auth"
+	"go.uber.org/fx/dig"
+
+	"go.uber.org/fx/modules"
+	"go.uber.org/fx/service"
 )
 
-type Option func() OutboundMiddleware
-
-func WithOutbound(middleware OutboundMiddleware) Option{
-	return func () OutboundMiddleware{ return middleware}
+type Option struct {
+	graph dig.Graph
 }
 
-// New creates an http.Client that includes 2 extra outbound middleware: tracing and auth
-// they are going to be applied in following order: tracing, auth, remaining outbound middleware
-// and only if all of them passed the request is going to be send.
-func New(options... Option) *http.Client {
+const _middlewareKey = "_httpClientMiddleware"
+const _graphKey = "_httpClientGraph"
+
+func WithOutbound(middleware... OutboundMiddleware) modules.Option{
+	return func(info *service.ModuleCreateInfo) error {
+		info.Items[_middlewareKey] = append(info.Items[_middlewareKey].([]OutboundMiddleware, middleware...)
+		return nil
+	}
+}
+
+func WithGraph(graph dig.Graph) modules.Option {
+	return func(info *service.ModuleCreateInfo) error {
+		info.Items[_graphKey] = graph
+		return nil
+	}
+}
+
+func New(options... modules.Option) *http.Client {
+	var info service.ModuleCreateInfo
+	for _, opt := range options {
+		opt(info)
+	}
+
+	graph := dig.DefaultGraph()
+	if g, ok := info.Items[_graphKey]; ok {
+		graph = g.(dig.Graph)
+	}
+
 	middleware := make([]OutboundMiddleware, 0, len(options) + 2)
 
-	var trace
-	middleware = append(middleware, tracingOutbound(tracer), authenticationOutbound(info))
+	var tracer opentracing.Tracer
+	if err := graph.Resolve(&tracer); err != nil {
+
+	}
+
+	middleware = append(middleware, tracingOutbound(*tracer), authenticationOutbound(info))
 	for _, x := range options {
 		middleware = append(middleware, x())
 	}
--- a/modules/uhttp/client/client.go
+++ b/modules/uhttp/client/client.go
@@ -32,20 +32,25 @@ import (
 	"go.uber.org/fx/service"
 )
 
-type Option struct {
-	graph dig.Graph
-}
+const (
+	_middlewareKey = "httpClientMiddleware"
+	_graphKey      = "httpClientGraph"
+)
 
-const _middlewareKey = "_httpClientMiddleware"
-const _graphKey = "_httpClientGraph"
-
-func WithOutbound(middleware... OutboundMiddleware) modules.Option{
+// WithOutbound lets you add custom middleware to the client
+func WithOutbound(middleware ...OutboundMiddleware) modules.Option {
 	return func(info *service.ModuleCreateInfo) error {
-		info.Items[_middlewareKey] = append(info.Items[_middlewareKey].([]OutboundMiddleware, middleware...)
+		items := info.Items
+		if m, ok := items[_middlewareKey]; ok {
+			items[_middlewareKey] = append(m.([]OutboundMiddleware), middleware...)
+		} else {
+			items[_middlewareKey] = middleware
+		}
 		return nil
 	}
 }
 
+// WithGraph allows you to a custom dependency injection graph to resolve Tracer and AuthInfo
 func WithGraph(graph dig.Graph) modules.Option {
 	return func(info *service.ModuleCreateInfo) error {
 		info.Items[_graphKey] = graph
@@ -53,10 +58,16 @@ func WithGraph(graph dig.Graph) modules.Option {
 	}
 }
 
-func New(options... modules.Option) *http.Client {
-	var info service.ModuleCreateInfo
+// New creates an http.Client that includes 2 extra outbound middleware: tracing and auth
+// they are going to be applied in following order: tracing, auth, remaining outbound middleware
+// and only if all of them passed the request is going to be send.
+func New(options ...modules.Option) (*http.Client, error) {
+	info := service.ModuleCreateInfo{
+		Items: make(map[string]interface{}),
+	}
+
 	for _, opt := range options {
-		opt(info)
+		opt(&info)
 	}
 
 	graph := dig.DefaultGraph()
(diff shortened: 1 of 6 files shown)


modules/uhttp/client/client.go
```diff
@@ -25,19 +25,83 @@ import (
...
+	"github.com/opentracing/opentracing-go"
+)
+
+const (
+	_middlewareKey = "httpClientMiddleware"
+	_graphKey      = "httpClientGraph"
 )
 
+// WithOutbound lets you to add a custom middleware to the client
+func WithOutbound(middleware ...OutboundMiddleware) modules.Option {
```
anuptalwalkar: rename `WithOutboundMiddleware` to keep it consistent with rpc

modules/uhttp/client/client.go
```diff
@@ -25,19 +25,83 @@ import (
...
+	return func(info *service.ModuleCreateInfo) error {
+		items := info.Items
+		if m, ok := items[_middlewareKey]; ok {
+			items[_middlewareKey] = append(m.([]OutboundMiddleware), middleware...)
```
anuptalwalkar: should check for type assertion?

modules/uhttp/client/client.go
```diff
@@ -25,19 +25,83 @@ import (
...
+	}
+
+	for _, opt := range options {
+		if err := opt(&info); err != nil {
+			return nil, err
+		}
+	}
+
+	// Resolve auth and tracer.
+	graph := dig.DefaultGraph()
```
anuptalwalkar: I don't like this approach of Resolving dependency in New. If you have 10 vars, this method would linearly stretch.  May be refactor it into a method on the client for resolve. Or, have the variables be Resolved in the outbound filters?

modules/uhttp/client/client.go
```diff
@@ -25,19 +25,83 @@ import (
...
+	}
+
+	var auth *auth.CreateAuthInfo
+	if err := graph.Resolve(&auth); err != nil {
+		return nil, err
+	}
+
+	// Make middleware.
+	middleware := make([]OutboundMiddleware, 0, len(options)+2)
+	middleware = append(middleware, tracingOutbound(*tracer), authenticationOutbound(*auth))
```
anuptalwalkar: Can this be wrapped into some defeultOutboundMiddleware?

service/host.go
```diff
@@ -164,6 +164,7 @@ func (s *manager) shutdown(err error, reason string, exitCode *int) (bool, error
 	}
 
 	// Flush tracing buffers
+	// GFM-378 uhttp.Client is not safe to send requests anymore
```
anuptalwalkar: It this an open task, or why?

modules/uhttp/client/client.go
```diff
@@ -25,19 +25,83 @@ import (
...
 // they are going to be applied in following order: tracing, auth, remaining outbound middleware
 // and only if all of them passed the request is going to be send.
-// Client is safe to use by multiple go routines, if global tracer is not changed.
-func New(info auth.CreateAuthInfo, middleware ...OutboundMiddleware) *http.Client {
-	defaultMiddleware := []OutboundMiddleware{tracingOutbound(), authenticationOutbound(info)}
-	defaultMiddleware = append(defaultMiddleware, middleware...)
+func New(options ...modules.Option) (*http.Client, error) {
+
+	// Apply options.
+	info := service.ModuleCreateInfo{
```
anuptalwalkar: I am confused about info creation here. Why is it created with an empty map?

modules/uhttp/client/client.go
alsamylkin: I am consistent with current usage, I think we never check these: e.g. inboundMiddlewareFromCreateInfo, because this really means something went really wrong

modules/uhttp/client/client.go
alsamylkin: So we can add elements to it later in options..

modules/uhttp/client/client.go
alsamylkin: If we'll have 10 default(!) arguments, I'll refactor, but just for, it is an overkill

service/host.go
alsamylkin: I opened it when starting working on moving tracer out, we close a jaeger reporter here and tracer will panic if any requests will be made by http client

madhuravi: This is not unique to http. If Stop is called, nothing is safe to call. Doesn't this happen to YARPC client too?

alsamylkin:
Clients for YARPC are created with a dispatcher and we stop it in modules
before, so it should be ok(error I assume).

service/host.go
madhuravi: From Alex - "Clients for YARPC are created with a dispatcher and we stop it in modules
before, so it should be ok(error I assume)."

Yes but it's not a module. Clients are not necessarily aware that dispatcher comes from initializing the YARPC module. It is fetched separately from module initialization.

modules/uhttp/client/client.go
```diff
@@ -25,19 +25,83 @@ import (
...
 	"go.uber.org/fx/auth"
+	"go.uber.org/fx/dig"
+	"go.uber.org/fx/modules"
+	"go.uber.org/fx/service"
+
+	"github.com/opentracing/opentracing-go"
+)
+
+const (
+	_middlewareKey = "httpClientMiddleware"
```
madhuravi: If it's outboundMiddleware otherwise, the key should match that.

modules/uhttp/client/client.go
```diff
@@ -25,19 +25,83 @@ import (
...
+			items[_middlewareKey] = append(m.([]OutboundMiddleware), middleware...)
+		} else {
+			items[_middlewareKey] = middleware
+		}
+		return nil
+	}
+}
+
+// WithGraph allows you to use a custom dependency injection graph to resolve Tracer and AuthInfo
+func WithGraph(graph dig.Graph) modules.Option {
```
madhuravi: Why is this a public function? I think it's too much for clients to think about.
client.New(WithGraph())

They shouldn't have to know that a client is initialized with a graph or set it up etc.

Is our goal to have separate graphs per module? Why doesn't using the global graph work?
Maybe we should figure out the strategy for how to use dig before we start using it everywhere.

modules/uhttp/client/client.go
```diff
@@ -25,19 +25,83 @@ import (
...
+		if err := opt(&info); err != nil {
+			return nil, err
+		}
+	}
+
+	// Resolve auth and tracer.
+	graph := dig.DefaultGraph()
+	if g, ok := info.Items[_graphKey]; ok {
+		graph = g.(dig.Graph)
+	}
```
madhuravi: why not 
else {
  graph = dig.DefaultGraph()
}
instead of always assigning and then over-writing. Avoids the overhead of DefaultGraph if provided.

modules/uhttp/client/client_middleware_test.go
```diff
@@ -115,38 +115,35 @@ func TestExecutionChainOutboundMiddleware_AuthContextPropagationFailure(t *testi
 }
 
 func TestOutboundMiddlewareWithTracerErrors(t *testing.T) {
+	t.Parallel()
+	tr := &shadowTracer{
+		opentracing.NoopTracer{},
+		func(sm opentracing.SpanContext, format interface{}, carrier interface{}) error {
```
madhuravi: nit: sm?

modules/uhttp/client/client_middleware_test.go
```diff
@@ -115,38 +115,35 @@ func TestExecutionChainOutboundMiddleware_AuthContextPropagationFailure(t *testi
...
+			return errors.New("Very bad _tracer")
```
madhuravi: nit: I would skip the underscore. The shadowTracer is not aware of the _tracer.

modules/uhttp/client/client_test.go
```diff
@@ -33,96 +38,134 @@ var (
 	_testYaml = []byte(`
 name: test
 `)
-	_testClient = New(fakeAuthInfo{yaml: _testYaml})
+	_tracer   = opentracing.NoopTracer{}
```
madhuravi: call this noopTracer, makes it more readable

service/host.go
alsamylkin: If clients get dispatcher with yarpc.Dispatcher(), but don't initialize the module, then it will be nil and users will not be able to create client..

modules/uhttp/client/client.go
alsamylkin: And they will not have to know, a regular `client.New()` will return a client, that is initialized with a default graph.

They may need to override a tracer/auth client, then they can provide their DI graph, that's why it is public. No one forces them to use it.

modules/uhttp/client/client.go
alsamylkin: In most cases we expect people to not provide it, that's why I've initialized graph that way.

modules/uhttp/client/client_middleware_test.go
alsamylkin: This is gogland's `refactoring`.. :(

modules/uhttp/client/client.go
alsamylkin: Sure, it just becomes so much SimpleBeanFactoryAwareAspectInstanceFactory..

modules/uhttp/client/client.go
alsamylkin: Yes, it is not very concise, let me rephrase it..

service/host.go
madhuravi: yes but they are not usually aware of this at all. They don't *know* that this is how it works and they don't have. UberFx's thing is all this functionality works only if you initialize and start the service.

alsamylkin:
Looks like it is easier to start over and pass all the parameters to the constructor explicitly.

Issue #285, opened 2017-02-17
dig: README examples not idiomatic


glibsm:
Need to take a pass. Includes things like:

```go
err := dig.Resolve(&o) // notice the pointer to a pointer as param type
if err == nil {
    // o is ready to use
}
```

```
err = g.Resolve(&f1)
require.NoError(t, err)
```

dmcaulay:
I'd like to see some guidelines for usage in the docs. I started to integrate dig into one of my projects and found myself trying to solve the wrong problems with dig. Here are a couple rules I've started to form:

* Dependency injection should only be used to resolve dependencies at service startup.
* Application code should receive dependencies in their constructors and not use dig.Resolve directly.
* Tests should create dependencies manually and pass them into service constructors.

I'd also like to see recommendations on where dig.Register code should live.

Dependency injection is very cool, but without better guidelines it might start leading to magical and messy code that's difficult to manage.

Feel free to strongly disagree with any of this :smiley:.

anuptalwalkar:
`Application code should receive dependencies in their constructors and not use dig.Resolve directly.
`
This gets tricky when you have a list of dependency to be consumed. May be worthwhile to investigate intelligent way to perform `dig.Resolve()`.

glibsm:
moved to https://github.com/uber-go/dig/issues/1

PR #304, opened 2017-02-21
Register dispatcher in the default dig graph

alsamylkin:
Use DIG to store the YAPRC dispatcher.

Right now people are confused about rpc.Dispatcher(), which returns you a dispatcher only if service is started and testing handlers becomes much harder. Proposed solution is to inject it in the default dig graph.

### DIFF
--- a/modules/rpc/thrift_test.go
+++ b/modules/rpc/thrift_test.go
@@ -80,8 +80,43 @@ modules:
 	// Dispatcher must be resolved in the default graph
 	var dispatcher *yarpc.Dispatcher
 	assert.NoError(t, dig.Resolve(&dispatcher))
-	assert.NotEmpty(t, dispatcher)
+	assert.Equal(t, 2, len(dispatcher.Inbounds()))
+}
 
+func TestThriftModuleSeparateGraph_OK(t *testing.T) {
+	t.Parallel()
+	di := dig.New()
+	snowflake := ThriftModule(okCreate, modules.WithRoles("rescue"), withGraph(di))
+
+	cfg := []byte(`
+modules:
+  rpc:
+    inbounds:
+     - tchannel:
+         port: 0
+     - http:
+         port: 0
+`)
+
+	mci := service.ModuleCreateInfo{
+		Name: "RPC",
+		Host: testHost{
+			Host:   service.NopHost(),
+			config: config.NewYAMLProviderFromBytes(cfg),
+		},
+		Items: make(map[string]interface{}),
+	}
+
+	special, err := snowflake(mci)
+	require.NoError(t, err)
+	assert.NotEmpty(t, special)
+
+	testInitRunModule(t, special[0], mci)
+
+	// Dispatcher must be resolved in the default graph
+	var dispatcher *yarpc.Dispatcher
+	assert.NoError(t, di.Resolve(&dispatcher))
+	assert.Equal(t, 2, len(dispatcher.Inbounds()))
 }
 
 func TestThriftModule_BadOptions(t *testing.T) {
@@ -113,7 +148,8 @@ func testInitRunModule(t *testing.T, mod service.Module, mci service.ModuleCreat
 
 func mch() service.ModuleCreateInfo {
 	return service.ModuleCreateInfo{
-		Host: service.NopHost(),
+		Host:  service.NopHost(),
+		Items: make(map[string]interface{}),
 	}
 }
 
--- a/modules/rpc/thrift_test.go
+++ b/modules/rpc/thrift_test.go
@@ -25,12 +25,12 @@ import (
 	"testing"
 
 	"go.uber.org/fx/config"
+	"go.uber.org/fx/dig"
 	"go.uber.org/fx/modules"
 	"go.uber.org/fx/service"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"go.uber.org/fx/dig"
 	"go.uber.org/yarpc"
 	"go.uber.org/yarpc/api/transport"
 	"go.uber.org/yarpc/encoding/thrift"
--- a/modules/rpc/thrift_test.go
+++ b/modules/rpc/thrift_test.go
@@ -83,40 +83,6 @@ modules:
 	assert.Equal(t, 2, len(dispatcher.Inbounds()))
 }
 
-func TestThriftModuleSeparateGraph_OK(t *testing.T) {
-	t.Parallel()
-	di := dig.New()
-	snowflake := ThriftModule(okCreate, modules.WithRoles("rescue"), withGraph(di))
-
-	cfg := []byte(`
-modules:
-  rpc:
-    inbounds:
-     - tchannel:
-         port: 0
-`)
-
-	mci := service.ModuleCreateInfo{
-		Name: "RPC",
-		Host: testHost{
-			Host:   service.NopHost(),
-			config: config.NewYAMLProviderFromBytes(cfg),
-		},
-		Items: make(map[string]interface{}),
-	}
-
-	special, err := snowflake(mci)
-	require.NoError(t, err)
-	assert.NotEmpty(t, special)
-
-	testInitRunModule(t, special[0], mci)
-
-	// Dispatcher must be resolved in the new graph
-	var dispatcher *yarpc.Dispatcher
-	assert.NoError(t, di.Resolve(&dispatcher))
-	assert.Equal(t, 1, len(dispatcher.Inbounds()))
-}
-
 func TestThriftModule_BadOptions(t *testing.T) {
 	modCreate := ThriftModule(okCreate, errorOption)
 	_, err := modCreate(mch())
--- a/modules/rpc/thrift_test.go
+++ b/modules/rpc/thrift_test.go
@@ -30,10 +30,10 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"go.uber.org/fx/dig"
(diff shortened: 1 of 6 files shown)


doc.go
```diff
@@ -27,19 +27,28 @@
...
 // UberFx builds the following into your service:
-// * Configuration, logging, metrics, and tracing setup by default
-// * Configuration provider that seamlessly merges static and dynamic config
-// * Application-level as well as runtime metrics for effective monitoring
-// * Context-aware logging for easy debugging
-// * RPC module with Thrift interfaces for microservices
-// * HTTP module with intelligent defaults for web applications
-// * Task module for executing async tasks durably
 //
+// • Logging backed by the zap logger
```
glibsm: wat

modules/rpc/yarpc.go
```diff
@@ -49,6 +50,7 @@ type YARPCModule struct {
 	log       ulog.Log
 	stateMu   sync.RWMutex
 	isRunning bool
+	di        dig.Graph
```
glibsm: I'm sure you had your reasons, but I don't know if it's the right thing to attach di object to the YARPC module. Can we think of alternative approaches? Is this purely to make testing easier?

Wouldn't it make testing the handlers difficult, as you'd have to either:

a. Rely on the assumption that we fall back onto the default graph
b. Pass around the reference to the module

doc.go
glibsm: image

Looks like file diff shows the wrong symbol, but the conversation view is fine :/

dig/doc.go
```diff
@@ -20,7 +20,7 @@
 
 // Package dig is the Dependency Injection Graph.
 //
-// Package dig provides a fairly opinionated way of resolving object dependencies.
+// package dig provides an opinionated way of resolving object dependencies.
```
madhuravi: Rant alert :P I think we overuse the word opinionated, I don't see dig as opinionated particularly. Each type of DI provides a different way of resolving dependencies that's all.

Why not "Dig resolves dependencies and provides users with an instantiated object when requested"

dig/doc.go
```diff
@@ -20,7 +20,7 @@
```
madhuravi: Missing README files?

doc.go
```diff
@@ -27,19 +27,28 @@
 //
 // Status
 //
-// Alpha. Expect minor API changes and bug fixes. (Beta release coming soon)
+// Beta. Expect minor API changes and bug fixes. See our changelog (CHANGELOG.md)for more.
```
madhuravi: This should not show up as a change, pull latest changes.

dig/doc.go
alsamylkin: This is was picked by me randomly, here is the Glib's fix:
https://github.com/uber-go/fx/pull/305/files

doc.go
madhuravi: Likewise, there are no changes to this file from this PR. It should go away when you pull changes.

doc.go
alsamylkin: I did :)

modules/rpc/yarpc.go
madhuravi: Why not just use one global graph for UberFx? Makes things much simpler.

dig/doc.go
madhuravi: Ah, my bad. We have been forgetting doc.go in commits :)

modules/rpc/yarpc_options.go
```diff
@@ -71,3 +73,20 @@ func onewayInboundMiddlewareFromCreateInfo(mci service.ModuleCreateInfo) []middl
...
+}
+
+func graphFromCreateInfo(mci service.ModuleCreateInfo) dig.Graph {
+	g, ok := mci.Items[_graphInterceptorKey]
+	if !ok {
+		return dig.DefaultGraph()
+	}
+
+	// Intentionally panic if someone adds non-graph to Items.
+	return g.(dig.Graph)
```
dmcaulay: Should you call panic explicitly here? It seems kinda random to force a type assertion panic.

alsamylkin: Hmm, how an explicit panic will be better than a regular one? It will provide all the types needed:
`panic: interface conversion: interface is *dig.graph, not string`

dmcaulay: I misread this earlier. Ignore the previous comment.

modules/rpc/yarpc.go
alsamylkin: Short answer because of snowflakes and tests: it is much easier to write tests with custom graphs, especially when/if yarpc team will add an in memory transport.

Also, I was able to remove di from the struct. Wrote a long rant, why it is hard and then hit an eureka moment..

modules/rpc/yarpc_options.go
```diff
@@ -71,3 +73,20 @@ func onewayInboundMiddlewareFromCreateInfo(mci service.ModuleCreateInfo) []middl
 	// Intentionally panic if programmer adds non-middleware slice to the data
 	return items.([]middleware.OnewayInbound)
 }
+
+func withGraph(graph dig.Graph) modules.Option {
```
anuptalwalkar: wait a minute. Why is the graph passed as an object? I thought the whole point of having DI is to perform dependency injection and not pass around as a dependency baggage. These objects should just be access d as 'dig.Resolve()'. Otherwise Im having hard time understanding the use of graph vs passing a struct or an interface map.

breerly: This.

We need to be *really* careful about how users are meant to use this.

If we set users up so that they inject the graph into their objects, then we've messed up - the graph isn't supposed to leave main.

anuptalwalkar: @glibsm  
Totally agree and we shouldn't use it as an object in `fx` as well. I believe this is a situation where global graph is required and should be restricted to be used only within DIG. That way we can have sane dig.Resolve, dig.Register APIs and restrict baggage passing use cases.

dmcaulay: I've added some thoughts to the dig issue on idiomatic usage. https://github.com/uber-go/fx/issues/285

I think it might be better to start out with more constrained usage patterns until we get more familiar with dig.

alsamylkin: This is an a private option to a constructor for testing.

What you are proposing @anup is to have a graph to be a global dump of all objects, so we essentially left with the same globals as before(actually worse: we added an extra dependency). This is worse for testing, hard to have multiple objects of the same type, etc

anuptalwalkar: I don't understand how it is worst for testing. It might be easier to test here by passing graph, but then how do you test DI outside fx? There may be more work involved in setting up DI for test, but it will at least be transparent. 
The graph is a dump of object in either case, but right now we are exposing it to test passing it around. IMHO dig users should not be aware of the graph.

alsamylkin: Because you can't run tests in parallel, global graph...
Not only this, but we can have multiple dispatchers if we really want and "know what we are doing" :)

I am confused about "test DI outside fx", why do we care in this module about it? We take it as an interface..

Dig users should be aware of graph, otherwise it can be misused.

alsamylkin: Synced offline and decided to use a global dig graph and add ability to run tests in parallel through dig, not a local field.

modules/rpc/thrift_test.go
```diff
@@ -74,6 +76,11 @@ modules:
 
 	testInitRunModule(t, goofy[0], mci)
 	testInitRunModule(t, gopher[0], mci)
+
+	// Dispatcher must be resolved in the default graph
```
breerly: Why does FX need a default graph at all? Why not just instantiate 1 graph when FX is instantiated?

modules/rpc/yarpc.go
```diff
@@ -249,7 +269,24 @@ func newYARPCModule(
 	module.config.inboundMiddleware = inboundMiddlewareFromCreateInfo(mi)
 	module.config.onewayInboundMiddleware = onewayInboundMiddlewareFromCreateInfo(mi)
 
-	_controller.addConfig(module.config)
+	// Try to resolve a controller first
+	// TODO(alsam) use dig options when available, because we can overwrite the controller in case of multiple
+	// modules registering a controller.
+	if err := dig.Resolve(&module.controller); err != nil {
```
breerly: I'm confused by the term controller here... what exactly is it's responsibility?

alsamylkin: There can be multiple modules that can can use YARPC, dispatcherController combines their configs in one and creates a single dispatcher, that a service will use.

anuptalwalkar: Should we log this err here?

modules/rpc/yarpc_test.go
```diff
@@ -25,6 +25,7 @@ import (
 
 	"go.uber.org/fx/service"
 
+	"fmt"
```
anuptalwalkar: `fmt` should be with stdlib imports

modules/rpc/thrift_test.go
```diff
@@ -30,6 +30,8 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"go.uber.org/fx/dig"
+	"go.uber.org/yarpc"
```
anuptalwalkar: nit: arrange imports.

modules/rpc/yarpc.go
alsamylkin: No, because it can be a valid error: no one registered a controller, and this is what we are going to do in the next line.

PR #308, opened 2017-02-22
Allow arbitrary order for Register funcitons

glibsm:
Fixes a bug where cyclical deps were not correctly identified
due to the fact that all functions returned an empty string
for Name() method. Using String() instead solve the problem
for concrete types.

The modified test covers arbitrary order scenario where
dependencies are registered after the constructor that needs them.

### DIFF
--- a/dig/node.go
+++ b/dig/node.go
@@ -35,6 +35,10 @@ type graphNode interface {
 	dependencies() []interface{}
 
 	// unique identification per node
+	//
+	// TODO(glib): GFM-396
+	// consider using a custom type to identify objects, rather than a string
+	// type id struct { reflect.Type, string name, } or something of the sort
 	id() string
 }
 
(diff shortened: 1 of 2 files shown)


glibsm:
@alsamylkin probably the test called "TestCycles"?

alsamylkin:
It passed, so should we add more, not modify ResolveAll?

dig/node.go
```diff
@@ -47,7 +47,7 @@ type node struct {
 func (n node) id() string {
 	// in the future, more than just the type of node is going to be required
 	// for instance, when multiple types are allowed with different names
-	return n.objType.Name()
+	return n.objType.String()
```
alsamylkin: Please, add a todo, to use Type for comparison.

glibsm: todo added, ticket filed. good idea on the custom key

PR #309, opened 2017-02-22
Better panic information for config merge

alsamylkin:
We don't show enough information to user why merge for 2 configs could fail.

This PR solves it by printing actual values we are trying to merge and printing the correct type of destination.

### DIFF
--- a/config/yaml_test.go
+++ b/config/yaml_test.go
@@ -353,6 +353,28 @@ func TestMerge(t *testing.T) {
 	}
 }
 
+func TestMergePanics(t *testing.T) {
+	t.Parallel()
+	src := []byte(`
+map:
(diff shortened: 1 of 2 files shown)


config/yaml_test.go
```diff
@@ -353,6 +353,28 @@ func TestMerge(t *testing.T) {
 	}
 }
 
+func TestMergePanics(t *testing.T) {
+	t.Parallel()
+	src := []byte(`
+map:
```
madhuravi: nit: is the format ok here? same below.

alsamylkin: Yep:
https://github.com/uber-go/fx/blob/master/config/yaml_test_data.go#L270

PR #310, opened 2017-02-22
Close dig objects on Reset()

glibsm:
GFM-377

### DIFF
--- a/dig/graph.go
+++ b/dig/graph.go
@@ -129,7 +129,6 @@ func (g *graph) RegisterAll(cs ...interface{}) error {
 	return nil
 }
 
-// Reset the graph by removing all the registered nodes
 func (g *graph) Reset() {
 	g.Lock()
 	defer g.Unlock()
--- a/dig/graph.go
+++ b/dig/graph.go
@@ -23,6 +23,7 @@ package dig
 import (
 	"bytes"
 	"fmt"
+	"io"
 	"reflect"
 	"sync"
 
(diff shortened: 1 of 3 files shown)


dig/graph.go
```diff
@@ -133,6 +134,16 @@ func (g *graph) Reset() {
 	g.Lock()
 	defer g.Unlock()
 
+	for _, node := range g.nodes {
+		if v, err := node.value(g); err == nil {
+			if nc, ok := v.Interface().(io.Closer); ok {
+				// Error is ignored. What else can be done here, should Reset() return an error?
+				// Assignment is to satisfy errcheck :(
+				err = nc.Close()
```
sectioneight: log it?

dig/graph.go
```diff
@@ -128,11 +129,20 @@ func (g *graph) RegisterAll(cs ...interface{}) error {
 	return nil
 }
 
-// Reset the graph by removing all the registered nodes
 func (g *graph) Reset() {
 	g.Lock()
 	defer g.Unlock()
 
+	for _, node := range g.nodes {
+		if v, err := node.value(g); err == nil {
```
alsamylkin: v can be nil, right?

alsamylkin: Oh, it can't

dig/graph.go
```diff
@@ -128,11 +129,20 @@ func (g *graph) RegisterAll(cs ...interface{}) error {
 	return nil
 }
 
-// Reset the graph by removing all the registered nodes
```
alsamylkin: update, not remove?

glibsm: These need to be removed, because they appear in there different places. And as abg pointed out, these don't make it into GoDoc

glibsm:
going to table this for now

PR #311, opened 2017-02-23
Remove extraneous fmt.Stringer pointers

glibsm:
These pointers are point...less


sectioneight:
Commit message game on point

PR #316, opened 2017-02-23
Remove global stats variables for rpc

peter-edge:


modules/rpc/internal/stats/metrics.go
```diff
@@ -33,27 +33,47 @@ const (
...
-	// RPCTags creates metrics scope with defined tags
-	RPCTags = map[string]string{
+	rpcTags = map[string]string{
 		TagModule: "rpc",
 		TagType:   "request",
 	}
+)
 
+// Client is a client for stats.
+type Client interface {
```
alsamylkin: Why is it an interface?

modules/rpc/internal/stats/metrics.go
```diff
@@ -33,27 +33,47 @@ const (
...
-	RPCPanicCounter tally.Counter
-)
+	RPCPanicCounter() tally.Counter
+}
+
+// NewClient returns a new Client for the given tally.Scope.
+func NewClient(scope tally.Scope) Client {
+	rpcTagsScope := scope.Tagged(rpcTags)
+	return &client{
+		rpcTagsScope.Tagged(map[string]string{TagMiddleware: "auth"}).Counter("fail"),
```
alsamylkin: Please, use field name for initialization..

peter-edge: I would love to talk about this - I leave out the field names on purpose. This is so that if a new field is added or a field is deleted later, I want there to be a compile-time error here. By adding the tags, I've had a lot more bugs.

modules/rpc/internal/stats/metrics.go
peter-edge: I really like putting things behind interfaces in general. In the future, `stats.Client` might be something more complicated than what amounts to a struct of counters, and I always default to using interfaces over structs for actual logic so that I can swap them out in testing.

modules/rpc/stats.go
```diff
@@ -33,27 +33,39 @@ const (
 	TagMiddleware = "middleware"
```
alsamylkin: Based on Youtong's comment in other PR, this should be private.

peter-edge: Ughhhh haha ok I'll change them all

alsamylkin: Thanks!

PR #317, opened 2017-02-23
	Remove global stats variables for task

peter-edge:


modules/task/task.go
```diff
@@ -23,31 +23,21 @@ package task
...
-
 var (
-	_globalBackendMu sync.RWMutex
-	_globalBackend   Backend = &NopBackend{}
-	_asyncMod        service.Module
-	_asyncModErr     error
-	_once            sync.Once
+	_globalBackendMu          sync.RWMutex
+	_globalBackend            Backend = &NopBackend{}
+	_globalBackendStatsClient         = newStatsClient(tally.NoopScope)
```
glibsm: what's going on with this alignment here?

peter-edge: I don't know, it was gofmt that did this, so it must be "right".

madhuravi: We seem to be using the NoopScope throughout, are we ever using the scope initialized by UberFx?

modules/task/task.go
```diff
@@ -56,6 +46,12 @@ func GlobalBackend() Backend {
 	return _globalBackend
 }
 
+func globalBackendStatsClient() *statsClient {
```
madhuravi: We are just replacing one global with the other? What's the goal?

peter-edge: To factor this out when we get rid of the globals in task altogether hopefully, but getting stats down to just this is the start

modules/task/task.go
peter-edge: Whoops! Missed setting this again, fixed. Thanks for the close review.

modules/task/task.go
peter-edge: The goal would be that statsClient would be per module.

peter-edge:
Fixed!

modules/task/task.go
madhuravi: this was @anuptalwalkar 's catch :)

modules/task/task.go
madhuravi: Shouldn't it be a member of the module somehow then?

peter-edge: Ya it should be eventually, but to make the modules/task package not use globals, there needs to be a bit more separation. Really what you need is something more like `Frontend` and `Backend`, and `Encoder` needs to be split up into `Encoder` and `Decoder`. Right now though, there's no such concept, so stats for what would be on the `Frontend` are accessed in the same way as the `Backend`, ie the `Enqueue` function. If we get this PR in, when we look to get rid of the globals in modules/task, this will make it easier.

peter-edge: @akshayjshah

madhuravi: Not sure how Frontend/Backend separation helps with metrics propagation, all of task module code needs access to metrics anyway. But cool with checking this in and iterating on it.

peter-edge:
It doesn't really, saying that you'll need a global stats client for the
"Frontend" operations anyways (ie Enqueue). I could have a stats client per
Backend somehow I guess, but really what you want there is more the model
of rpc with middleware...ie ya we should iterate on this

peter-edge:
Can we merge? I'm not sure what the policy is here

madhuravi:
@peter-edge Apologies, I forget you don't have permissions.

PR #318, opened 2017-02-23
Remove global stats variables for uhttp

peter-edge:


modules/uhttp/internal/stats/metrics.go
```diff
@@ -33,31 +33,48 @@ const (
...
-	HTTPMethodTimer tally.Scope
-	// HTTPStatusCountScope is a scope for http status
-	HTTPStatusCountScope tally.Scope
 )
 
-// SetupHTTPMetrics allocates counters for necessary setup
-func SetupHTTPMetrics(scope tally.Scope) {
-	httpScope := scope.Tagged(HTTPTags)
-	HTTPPanicCounter = httpScope.Counter("panic")
+// Client is a client for stats.
```
alsamylkin: Nit: I think metrics were moved to internal package, because of the globals, now they can become private structs in the uhttp package.

peter-edge: Ugh haha ok, I'll do that

modules/uhttp/stats.go
```diff
@@ -33,31 +33,46 @@ const (
 	TagMiddleware = "middleware"
```
yutongp: Shall these tag names be private?

modules/uhttp/stats.go
```diff
@@ -18,7 +18,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-package stats
+package uhttp
```
yutongp: since we are moving `stats` back to `uhttp`, how about `client`? ;P

peter-edge: Another PR I think :)

PR #319, opened 2017-02-23
Move internal.ContextLogger to be private in ulog

peter-edge:
I think we may actually want some better key here, I think this is just setting the logger at key 0, which isn't very unique. Regardless, this should probably be private inside ulog now.


ulog/log.go
```diff
@@ -37,6 +36,10 @@ import (
 	"github.com/uber/jaeger-client-go"
 )
 
+const _contextLogger contextKey = iota
```
yutongp: I thought context key better to be unique. If anyone use 0 or used 0, value will be overwrote.

peter-edge: Fixed

sectioneight:
So, ulog is gone now, do we close this? @akshayjshah

akshayjshah:
Yep, I think so.

PR #322, opened 2017-02-23
Change service.Option to use the manager struct

peter-edge:
This is generally how i've seen options done - the package that exposes them defines them in terms of a private struct. Unless Host is editable, an option on the Host interface doesn't make sense, and there was just casting happening here anyways. I think it would be best to limit Options to be either defined in the service package, or defined in terms of options in the service package.

### DIFF
--- a/service/options.go
+++ b/service/options.go
@@ -27,44 +27,39 @@ import (
 	"github.com/uber-go/tally"
 )
 
-// A Option configures a service host
-type Option func(Host) error
+// A Option configures a manager
+type Option func(*manager) error
 
-// WithConfiguration adds configuration to a service host
+// WithConfiguration adds configuration to a manager
 func WithConfiguration(config config.Provider) Option {
-	return func(svc Host) error {
-		// TODO(ai) verify type assertion is correct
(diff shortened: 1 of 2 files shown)


sectioneight:
The problem with this is that now people can't write their own Options in their own repos, which I guess isn't a huge deal since they couldn't do the private cast before, but I'd like to have a story for custom options.

peter-edge:
Custom options I think should be done similar to how I have them (right now, but this is actually going to change) in https://github.com/uber-go/fx/pull/275 - you implement them on top of the functions provided by the service package. Think:

```go
func WithSomeSpecialConfigThing(foo Foo) service.Option {
... some other logic
  return service.WithConfiguration(configProvider)
}
```

Host (as I see it) is a read-only interface, ie it just has accessors, not setters. If this is true, it shouldn't be possible to modify a Host outside of the service package.

peter-edge:
Also while I'm on a gRPC rant heh https://github.com/grpc/grpc-go/blob/20633fa172ac711ac6a77fd573ed13f23ec56dcb/server.go#L124

PR #323, opened 2017-02-23
[WIP] Remove most metrics global variables

peter-edge:
I think the usage of the metrics package may not be optimal right now, and we may want to clean this up (the only place this is used is basically in one spot in service_setup.go, and there wasn't a need for it anyway as RegisterRootScope is never called), but this is literally just wrapping Freeze, RegisterRootScope, and RootScope in a Client interface, almost nothing else. I argue this is strictly improving FX's situation heh :)


peter-edge:
This actually is half-baked - we need a way to register a root scope on a Manager I think.

service/core.go
```diff
@@ -99,6 +99,7 @@ type serviceCore struct {
 	metricsCore
 	tracerCore
 	authClient     auth.Client
+	metricsClient  metrics.Client
```
anuptalwalkar: Why do we need this on service core? I don't see us using metrics client being used anywhere except for initialization.

service/options.go
```diff
@@ -47,6 +48,17 @@ func WithMetrics(scope tally.Scope, reporter tally.CachedStatsReporter) Option {
 	}
 }
 
+// WithRootScope registers the root scope for all the manager metrics
+func WithRootScope(scopeFunc metrics.ScopeFunc) Option {
```
anuptalwalkar: Do we need this? Whats the benefit of having RootScope available as an option on service?

peter-edge:
Closing this for now, will revisit in a bit

PR #324, opened 2017-02-23
Add a Dockerfile and Docker Make targets

peter-edge:
See https://github.com/uber-go/sally/pull/25 for a good explanation of why this is useful :)


Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM golang:1.8.0
+
+WORKDIR /go/src/go.uber.org/fx
+ADD .build/deps.mk /go/src/go.uber.org/fx/.build/
+ADD glide.yaml glide.lock /go/src/go.uber.org/fx/
+RUN make -f .build/deps.mk deps
+ADD . /go/src/go.uber.org/fx/
+RUN make clean
```
glibsm: is this right? i think it should be `make` or `make bins`

peter-edge: This is right - I can't easily individually ignore all the outputted files that are generated for code coverage (and as packages are harder, it would be unlikely to always be in sync with .dockerignore). `make` or `make bins` is the testing step that you do separately. If this were to have a running binary associated with it, usually you would do `CMD ["make", "run"]` or something along those lines as the default command to launch to binary.

glibsm: got it.

PR #325, opened 2017-02-23
Rename RPC module to YARPC

### LINKED ISSUE #267: yarpc module is called `modules/rpc`
IIRC, yarpc is just one type of RPC system which could be supported as modules. Similarly, gRPC may be supported too in the future.

sectioneight:
Fixes #267


peter-edge:
image

sectioneight:
So, I think I'm going to hold off on merging this since @peter-edge and @akshayjshah both have some gigantic PRs in-flight that they'd like to land soon, and this would cause merge conflicts. Once those are merged, I'll fix the conflicts and merge this.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## v1.0.0-beta2 (unreleased)
+
+* [Breaking] Rename `modules/rpc` to `modules/yarpc`
```
glibsm: I think we should leave a note about the yaml change that needs to happen. I think that's the way YARPC has been writing changelog notes - with an action that needs to be taken, or more detailed information of what the rename actually entails.

alsamylkin: What yaml change?

modules/yarpc/yarpc.go
```diff
@@ -230,17 +230,17 @@ func (c *dispatcherController) mergeConfigs(name string) (conf yarpc.Config, err
...
 // Creates a new YARPC module and adds a common middleware to the global config collection.
 // The first created module defines the service name.
-func newYARPCModule(
+func newModule(
 	mi service.ModuleCreateInfo,
 	reg registerServiceFunc,
 	options ...modules.Option,
-) (*YARPCModule, error) {
+) (*Module, error) {
 	name := "yarpc"
```
peter-edge: So on my "no default names" thing - note that for *newModule`, it's defaulting to `yarpc`, while in `ThriftModule` in thrift.go it's defaulting to `rpc` @madhuravi

CHANGELOG.md
peter-edge: The key should be changed to yarpc. @sectioneight you need to just remove the default naming in the `ThriftModule` function (where it sets the name to "rpc"), see my comment on yarpc.go

glibsm: yes, the yaml key also changes from "rpc" to "yarpc". It is not immediately obvious from this changelog entry that service owners should update their config

sectioneight:
This PR doesn't change the YAML behavior though, right? I can add a code
change that also changes the module config name, but it's not implicit
based on the package name currently, right?

peter-edge:
This PR *should* change the yaml behavior per #267 though, and per a convo
we had earlier (where package names should match default module names)

sectioneight:
Makes sense. Will update.

modules/yarpc/yarpc.go
madhuravi: Thanks for catching that Peter. Ot should always be "yarpc"

CHANGELOG.md
alsamylkin: I was hoping it was left on purpose..

modules/yarpc/yarpc.go
madhuravi: I looked up current code and don't see a separate name being set for thrift. Maybe I'm missing something? Which file are you looking at Peter?

akshayjshah:
> So, I think I'm going to hold off on merging this since @peter-edge and @akshayjshah both have some gigantic PRs in-flight that they'd like to land soon, and this would cause merge conflicts. Once those are merged, I'll fix the conflicts and merge this.

🙏  Your mercy is appreciated.

sectioneight:
@anuptalwalkar yea, I have an internal diff open that I've been waiting to land until this change is merged.

modules/yarpc/yarpc.go
```diff
@@ -259,14 +259,14 @@ func newYARPCModule(
...
 	for _, option := range options {
 		if err := option(moduleOptions); err != nil {
 			return nil, err
 		}
 	}
-	module := &YARPCModule{
+	module := &Module{
 		host:        host,
 		statsClient: newStatsClient(host.Metrics()),
 		log:         ulog.Logger(context.Background()).With(zap.String("module", host.Name())),
```
alsamylkin: Something fishy is going on the line 274: We are using host.Name to get yarpc config..

sectioneight: I think host.Name() is now scoped post-peter's-refactor to the module name? @peter-edge to confirm

sectioneight:
`gimme tip` appears to be broken, unrelated to this PR. Build passes on 1.7 and 1.8. Merging with the team's blessing.

PR #330, opened 2017-02-24
Remove unused config.IsDevelopmentEnv func

sectioneight:
We recognized early on that this was an anti-pattern, and with Akshay's
log refactor we no longer have any references to it.

This is not a breaking change because nobody is using this according to
our internal code search tool.


akshayjshah:
:+1:

PR #333, opened 2017-02-24
Delete kafka package

peter-edge:
The kakfa package has basically nothing in it, and it's still there as a placeholder. This has just been bugging me for a while, but feel free to close this if there is a reason to keep the package.


akshayjshah:
Sounds good to me too.

PR #334, opened 2017-02-24
[Do not merge]Consume latest tally

alsamylkin:
Tally introduced histograms to their interfaces and changed constructor for a scope, which requires us to implement it for NopReported and update some tests.

Will not merge, until migrate internal tools to the new tally version.

### DIFF
--- a/metrics/nop_reporter.go
+++ b/metrics/nop_reporter.go
@@ -46,13 +46,13 @@ func (c *capabilities) Tagging() bool {
 	return c.tagging
 }
 
-// NopCachedStatsReporter is an implementatin of tally.CachedStatsReporter than simply does nothing.
+// NopCachedStatsReporter is an implementation of tally.CachedStatsReporter that simply does nothing
+// and should be used for testing purposes only.
 // TODO:(anup) This should exist in tally. https://github.com/uber-go/tally/issues/23
-// Remove and replace metrics.NopCachedStatsReporter with tally.NopCachedStatsRepor once issue is resolved
+// Remove and replace metrics.NopCachedStatsReporter with tally.NopCachedStatsReporter once issue is resolved
 var NopCachedStatsReporter tally.CachedStatsReporter = nopCachedStatsReporter{}
 
-type nopCachedStatsReporter struct {
-}
+type nopCachedStatsReporter struct{}
 
 func (nopCachedStatsReporter) AllocateCounter(name string, tags map[string]string) tally.CachedCount {
 	return NopCachedCount
@@ -74,10 +74,10 @@ func (r nopCachedStatsReporter) Capabilities() tally.Capabilities {
 	return capabilitiesReportingNoTagging
 }
 
-func (r nopCachedStatsReporter) Flush() {
-}
+func (r nopCachedStatsReporter) Flush() {}
 
-// NopCachedCount is an implementation of tally.CachedCount
+// NopCachedCount is an implementation of tally.CachedCount and same as other Nop types don't do anything and
+// should be used for testing purposes only.
 var NopCachedCount tally.CachedCount = nopCachedCount{}
 
 type nopCachedCount struct {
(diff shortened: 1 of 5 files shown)


metrics/nop_reporter.go
```diff
@@ -99,3 +103,30 @@ type nopCachedTimer struct {
 
 func (nopCachedTimer) ReportTimer(interval time.Duration) {
 }
+
+// NopCachedHistogram is an implementation of tally.CachedHistogram
```
sectioneight: maybe mention this is exported for testing?

metrics/nop_reporter.go
```diff
@@ -99,3 +103,30 @@ type nopCachedTimer struct {
...
+var NopCachedHistogram tally.CachedHistogram = nopCachedHistogram{}
+
+type nopCachedHistogram struct {
```
sectioneight: style nit: empty struct on a single line generally, right? or are we doing differerently:

```go
type nopCachedHistogram struct{}
```

metrics/nop_reporter.go
```diff
@@ -99,3 +103,30 @@ type nopCachedTimer struct {
...
+}
+
+func (nopCachedHistogram) ValueBucket(
+	bucketLowerBound, bucketUpperBound float64,
```
sectioneight: style nit: inconsistent oneline versus multiline. we may need a style guide :|

i thought we either did:

```go
func foo(lotsAndLots, ofArgumentsHere ofThisType) {
```

or
```go
func foo(
  lotsAndLots,
  ofArguments ofThisType,
)
```

but maybe i'm misremembering

metrics/nop_reporter.go
```diff
@@ -62,6 +62,10 @@ func (nopCachedStatsReporter) AllocateGauge(name string, tags map[string]string)
 	return NopCachedGauge
 }
 
+func (nopCachedStatsReporter) AllocateHistogram(name string, tags map[string]string, buckets tally.Buckets) tally.CachedHistogram {
```
sectioneight: how come this one doesn't get godoc comments but the other one you added does?

service/service_test.go
```diff
@@ -38,7 +38,16 @@ import (
 func TestServiceCreation(t *testing.T) {
 	r := metrics.NewTestStatsReporter()
 	r.CountersWG.Add(1)
-	scope, closer := tally.NewRootScope("", nil, r, 50*time.Millisecond, tally.DefaultSeparator)
+	opts := tally.ScopeOptions{
+		Tags:           nil,
```
sectioneight: why the explicit nil/empty string stuff here? i think i actually like it, but generally see it elided as:

```go
opts := tally.ScopeOptions {
  Reporter: r,
  Separator: ...,
  DefaultBuckets: ...,
}
```

for brevity. thoughts?

testutils/metrics/metrics.go
```diff
@@ -84,6 +114,15 @@ func NewTestStatsReporter() *TestStatsReporter {
...
+	opts := tally.ScopeOptions{
+		Tags:           nil,
+		Prefix:         "",
+		Reporter:       r,
+		CachedReporter: nil,
+		Separator:      tally.DefaultSeparator,
+		DefaultBuckets: tally.DefaultBuckets,
+	}
+
+	scope, _ := tally.NewRootScope(opts, 100*time.Millisecond)
```
sectioneight: unrelated to this change, but what is this magic number for?

metrics/nop_reporter.go
alsamylkin: Well, I was following the leader(couple lines above) :)

metrics/nop_reporter.go
alsamylkin: They have the same type, but sure.

metrics/nop_reporter.go
alsamylkin: Should we do this everywhere in this file? I'd prefer that, but not sure, if it will confuse reviewers.

service/service_test.go
alsamylkin: Well, I was afraid linter would complain, because, we can extend a struct and forget some mandatory fields. But it didn't. Let's get dangerous.

metrics/nop_reporter.go
alsamylkin: nopCachedStatsReporter is the internal type.

testutils/metrics/metrics.go
alsamylkin: It is the reporting interval, how often we are going to sample metrics.

testutils/metrics/metrics.go
```diff
@@ -35,9 +35,11 @@ type TestStatsReporter struct {
 	GaugesWG   sync.WaitGroup
 	TimersWG   sync.WaitGroup
 
-	Counters map[string]int64
-	Gauges   map[string]float64
-	Timers   map[string]time.Duration
+	Counters           map[string]int64
+	Gauges             map[string]float64
+	Timers             map[string]time.Duration
+	HistogramSamples   map[string]tally.Buckets
```
anuptalwalkar: Has this been implemented internally? If not we cannot use it until the internal implementation support these APIs.

alsamylkin: Why? Nop implementations doesn't have to do anything.

alsamylkin: This is needed only to implement interfaces.

The way service is using it is in service_setup.go, where options became a separate struct.

anuptalwalkar: I meant for Tally 2.0 `CachedStatsReporter`. We don't have its implementation here, but we rely on internal implementation from our private repo.

metrics/nop_reporter.go
```diff
@@ -86,47 +86,45 @@ type nopCachedCount struct {
...
+// should be used for testing purposes only.
 var NopCachedHistogram tally.CachedHistogram = nopCachedHistogram{}
 
-type nopCachedHistogram struct {
-}
+type nopCachedHistogram struct{}
 
 func (nopCachedHistogram) ValueBucket(
-	bucketLowerBound, bucketUpperBound float64,
+	bucketLowerBound float64,
```
sectioneight: note: i wasn't suggesting repeating the type, just putting each param on its own line or having them all on one line, not some mix-and-match

alsamylkin: I know, personally, it is easier to read. Really wish go fmt will help with these things :)

glibsm:
Looks good. What Anup said about internal implementation changes applies. As soon as this lands, m3 backend will be broken.

anuptalwalkar:
merging since we have all the necessary components ready.

PR #337, opened 2017-02-27
Refactor PopulateStruct

alsamylkin:
config.PopulateStruct is not really recursive now: primitive types are available in full only for structs, but not arrays; maps can't have arrays or structs as values, etc. There is also some code duplication: the dispatch for populating values for an array, map, struct fields should be the same.

This PR splits valueStruct method in several smaller methods, that are called recursively. Which solves all the problems above.

- [x] Arrays of fixed length(they are different types than regular slices)
- [x] More tests for a new functionality(existing cover lots of cases, but e.g. we can do maps of structs now)
- [x] Check for cycles
- [x] Populate only exported fields

### DIFF
--- a/config/README.md
+++ b/config/README.md
@@ -181,13 +181,13 @@ Note that any fields you wish to deserialize into must be exported, just like
 Current performance benchmark data:
 
 ```
-BenchmarkYAMLCreateSingleFile-8                       50000     31269 ns/op   11136 B/op     121 allocs/op
-BenchmarkYAMLCreateMultiFile-8                        30000     52378 ns/op   20064 B/op     205 allocs/op
-BenchmarkYAMLSimpleGetLevel1-8                     50000000      27.1 ns/op       0 B/op       0 allocs/op
-BenchmarkYAMLSimpleGetLevel3-8                     50000000      26.8 ns/op       0 B/op       0 allocs/op
-BenchmarkYAMLSimpleGetLevel7-8                     50000000      26.3 ns/op       0 B/op       0 allocs/op
-BenchmarkYAMLPopulateStruct-8                       2000000       861 ns/op     192 B/op      10 allocs/op
-BenchmarkYAMLPopulateStructNested-8                  500000      2616 ns/op     616 B/op      34 allocs/op
-BenchmarkYAMLPopulateStructNestedMultipleFiles-8     500000      3330 ns/op     744 B/op      42 allocs/op
-BenchmarkYAMLPopulateNestedTextUnmarshaler-8         100000     16775 ns/op    3201 B/op     209 allocs/op
+BenchmarkYAMLCreateSingleFile-8                    	   50000	     31345 ns/op	   11144 B/op	     122 allocs/op
+BenchmarkYAMLCreateMultiFile-8                     	   30000	     51927 ns/op	   20080 B/op	     207 allocs/op
+BenchmarkYAMLSimpleGetLevel1-8                     	50000000	        26.8 ns/op	       0 B/op	       0 allocs/op
+BenchmarkYAMLSimpleGetLevel3-8                     	50000000	        27.3 ns/op	       0 B/op	       0 allocs/op
+BenchmarkYAMLSimpleGetLevel7-8                     	50000000	        26.2 ns/op	       0 B/op	       0 allocs/op
+BenchmarkYAMLPopulateStruct-8                      	 1000000	      1598 ns/op	     576 B/op	      16 allocs/op
+BenchmarkYAMLPopulateStructNested-8                	  300000	      3972 ns/op	    1056 B/op	      42 allocs/op
+BenchmarkYAMLPopulateStructNestedMultipleFiles-8   	  300000	      5013 ns/op	    1248 B/op	      52 allocs/op
+BenchmarkYAMLPopulateNestedTextUnmarshaler-8       	  100000	     18209 ns/op	    3368 B/op	     211 allocs/op
 ```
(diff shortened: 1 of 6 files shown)


config/value.go
```diff
@@ -382,216 +394,263 @@ func convertValue(value interface{}, targetType reflect.Type) (interface{}, erro
...
+					value.Set(reflect.New(valueType.Elem()))
+				}
+
+				// We cannot assign reflect.ValueOf(val) to value as is, when field is a user defined type
+				// We need to find the Kind of the custom type and set the value to the specific type
+				// that user defined type is defined with.
+				kind := valueType.Elem().Kind()
+				switch kind {
+				case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+					value.Elem().SetInt(int64(val.(int)))
```
alsamylkin: Elem() is everywhere!

config/value.go
```diff
@@ -382,216 +394,263 @@ func convertValue(value interface{}, targetType reflect.Type) (interface{}, erro
...
+// - for object values, we recurse.
+func (cv Value) valueStruct(key string, target interface{}) error {
+	tarGet := reflect.Indirect(reflect.ValueOf(target))
+	targetType := tarGet.Type()
 
 	for i := 0; i < targetType.NumField(); i++ {
 		field := targetType.Field(i)
-		fieldType := field.Type
 		fieldName := field.Name
+		if fieldName[0] != strings.ToUpper(fieldName)[0] {
```
alsamylkin: Is there anything better to check for private fields?

sectioneight:
General question before I do a real review: this feels like it could be generalized to the point where YAML/Config parsing (separate from UberFx) could be open-sourced as a standalone package. I realize that would increase the maintenance burden on our project, but it strikes me as a bit strange every time I see a bug/PR related to configuration parsing on a generic service framework repo.

config/value.go
```diff
@@ -229,41 +237,41 @@ func (cv Value) TryAsFloat() (float64, bool) {
...
 func (cv Value) AsInt() int {
-	s, ok := cv.TryAsInt()
-	if !ok {
-		panic(fmt.Sprintf("Can't convert to int: %T %v", cv.Value(), cv.Value()))
+	if s, ok := cv.TryAsInt(); ok {
+		return s
 	}
-	return s
+
+	panic(fmt.Sprintf("Can't convert to int: %T %v", cv.Value(), cv.Value()))
```
sectioneight: This stuff all just feels like cleanup (up to this part anyway, I haven't read the rest of the PR). If we find any code towards the bottom that is contentious, I'd vote for extracting these simple changes into a separate PR and narrowing the scope of what we're reviewing to substantive changes only.

Personally, when I open up a 300+/- PR I have a subconscious feeling I'm going to have a bad time, and that affects the quality of my review (this is the same guidance we teach internally at Uber)

alsamylkin: Yeah, I was just fixing everything that was off in this file, and will move it to a second PR.

Unfortunately the headache will remain, ~250 lines need to be added and removed, because we didn't make it right in the first time and kept adding bandaids over it.

config/value.go
```diff
@@ -287,19 +290,19 @@ func (cv Value) Value() interface{} {
 }
 
 const (
-	bucketInvalid   = -1
-	bucketPrimitive = 0
-	bucketArray     = 1
-	bucketObject    = 2
-	bucketMap       = 3
-	bucketSlice     = 4
+	bucketPrimitive = iota
```
anuptalwalkar: why is `bucketInvalid` removed? Do we not need it?

config/value.go
anuptalwalkar: Why do we need to skip un-exported fields? 
Should this be an error to the user to export the fields, since config should always expect exported fields or shouldn't care at all.

config/value.go
alsamylkin: Yes, We are going to just ignore it(because this is what we were doing before)

config/README.md
```diff
@@ -181,13 +181,13 @@ Note that any fields you wish to deserialize into must be exported, just like
...
-BenchmarkYAMLPopulateNestedTextUnmarshaler-8         100000     16775 ns/op    3201 B/op     209 allocs/op
+BenchmarkYAMLCreateSingleFile-8                    	   50000	     31345 ns/op	   11144 B/op	     122 allocs/op
+BenchmarkYAMLCreateMultiFile-8                     	   30000	     51927 ns/op	   20080 B/op	     207 allocs/op
+BenchmarkYAMLSimpleGetLevel1-8                     	50000000	        26.8 ns/op	       0 B/op	       0 allocs/op
+BenchmarkYAMLSimpleGetLevel3-8                     	50000000	        27.3 ns/op	       0 B/op	       0 allocs/op
+BenchmarkYAMLSimpleGetLevel7-8                     	50000000	        26.2 ns/op	       0 B/op	       0 allocs/op
+BenchmarkYAMLPopulateStruct-8                      	 1000000	      1598 ns/op	     576 B/op	      16 allocs/op
+BenchmarkYAMLPopulateStructNested-8                	  300000	      3972 ns/op	    1056 B/op	      42 allocs/op
+BenchmarkYAMLPopulateStructNestedMultipleFiles-8   	  300000	      5013 ns/op	    1248 B/op	      52 allocs/op
+BenchmarkYAMLPopulateNestedTextUnmarshaler-8       	  100000	     18209 ns/op	    3368 B/op	     211 allocs/op
```
alsamylkin: Allocations increased, because we check for cycles now.

PR #341, opened 2017-03-01
Restore logging functionality

glibsm:
There was a flurry of different issues, which meant that as things stood
all logging output was discarded.

The most important one is the populateStruct error was not the corrent
condition to check for use of the default logger.

Secondary, embedding `zap.Config` does not work properly with `PopulateStruct`, 
or indeed yaml serialization at all, I believe (need to test). Essentially we weren't
properly creating log configs out of yaml.

Logging setup function now returns an error if it's not able to properly parse
the config.

Default encoding is now also "console", since not providing one is a zap
error.

before
```
$ ./simple
```

after
```
$ ./simple
1.4883359655306897e+09  info    /Users/glib/gocode/src/go.uber.org/fx/modules/uhttp/http.go:173 Server listening on port     {"trace": {}, "moduleName": "http", "port": 8080}
1.4883359655307891e+09  info    /Users/glib/gocode/src/go.uber.org/fx/service/host.go:350       Module started up cleanly    {"module": "http"}
```

### DIFF
--- a/service/service_setup.go
+++ b/service/service_setup.go
@@ -37,13 +37,28 @@ import (
 func (svc *serviceCore) setupLogging() error {
 	cfg := svc.configProvider.Get("logging")
 	if cfg.HasValue() {
-		// populate struct if config was provided
-		zapCfg := zap.Config{Level: zap.NewAtomicLevel()}
+		// Uhhh... this process is not the most elegant.
+		// Becuase log.Configuration embeds zap, the PopulateStruct
+		// does not work properly as it's unable to serialize fields directly
+		// into the embedded struct, so inner struct has to be treated as a
+		// separate object
+		//
+		// first, use the default zap configuration
+		zapCfg := ulog.DefaultConfiguration().Config
+
+		// override the embedded zap.Config stuct from config
 		if err := cfg.PopulateStruct(&zapCfg); err != nil {
 			return errors.Wrap(err, "unable to parse logging config")
 		}
+
+		// use the overriden zap config
 		svc.logConfig = ulog.Configuration{
 			Config: zapCfg,
+		}
+
+		// override any remaining things fom config, i.e. Sentry
+		if err := cfg.PopulateStruct(&svc.logConfig); err != nil {
+			return errors.Wrap(err, "unable to parse logging config")
 		}
 	} else {
 		// if no config - default to the regular one
--- a/service/service_setup.go
+++ b/service/service_setup.go
@@ -67,7 +67,7 @@ func (svc *serviceCore) setupLogging() error {
 
 	logger, err := svc.logConfig.Build(zap.Hooks(ulog.Metrics(svc.metrics)))
 	if err != nil {
-		return errors.Wrap(err, "Failed to configure logging")
+		return errors.Wrap(err, "failed to configure logging")
 	}
 
 	// TODO(glib): SetLogger returns a deferral to clean up global log which is not used
--- a/service/service_setup.go
+++ b/service/service_setup.go
@@ -34,21 +34,27 @@ import (
 	"go.uber.org/zap"
 )
 
-func (svc *serviceCore) setupLogging() {
-	err := svc.configProvider.Get("logging").PopulateStruct(&svc.logConfig)
-	if err != nil {
-		zap.L().Info("Logging configuration is not provided, using UberFX defaults", zap.Error(err))
+func (svc *serviceCore) setupLogging() error {
+	cfg := svc.configProvider.Get("logging")
+	if cfg.HasValue() {
+		// populate struct if config was provided
+		if err := cfg.PopulateStruct(&svc.logConfig); err != nil {
+			return errors.Wrap(err, "unable to parse logging config")
+		}
+	} else {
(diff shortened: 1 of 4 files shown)


service/service_setup.go
```diff
@@ -34,21 +34,31 @@ import (
...
 		svc.logConfig = ulog.DefaultConfiguration()
 	}
 
 	logger, err := svc.logConfig.Build(zap.Hooks(ulog.Metrics(svc.metrics)))
 	if err != nil {
-		// Typically, this means that we're trying to log to files that don't
-		// exist or don't have appropriate permissions set.
-		zap.L().Info("Failed to configure logger", zap.Error(err))
-		return
+		return errors.Wrap(err, "Failed to configure logging")
```
sectioneight: nit: lowercase Failed -> failed

glibsm:
re: regression

Fixing one of the tests is the regression test :) 
I could write a stand-alone way to validate other parts of the config are picked up properly I suppose.

service/service_setup.go
```diff
@@ -34,21 +34,31 @@ import (
...
+	} else {
+		// if no config - default to the regular one
 		svc.logConfig = ulog.DefaultConfiguration()
 	}
 
 	logger, err := svc.logConfig.Build(zap.Hooks(ulog.Metrics(svc.metrics)))
 	if err != nil {
-		// Typically, this means that we're trying to log to files that don't
-		// exist or don't have appropriate permissions set.
-		zap.L().Info("Failed to configure logger", zap.Error(err))
```
glibsm: note: `zap.L()` *can not* be used until after `ulog.SetLogger`, else it's just a no-op.

alsamylkin:
I've tried to run kv example and it failed:
2017/02/28 20:11:37 Unable to initialize serviceerrorservice instantiation failed: Failed to configure logging: unknown encoding ""

alsamylkin:
In restore-dev-logging branch..
Which seems to be a mandatory field, so I think, we should just add a default value for a yaml tags in ZAP

service/service_setup.go
```diff
@@ -34,21 +34,31 @@ import (
...
 
-func (svc *serviceCore) setupLogging() {
-	err := svc.configProvider.Get("logging").PopulateStruct(&svc.logConfig)
-	if err != nil {
-		zap.L().Info("Logging configuration is not provided, using UberFX defaults", zap.Error(err))
+func (svc *serviceCore) setupLogging() error {
+	cfg := svc.configProvider.Get("logging")
+	if cfg.HasValue() {
+		// populate struct if config was provided
+		zapCfg := zap.Config{Level: zap.NewAtomicLevel()}
```
alsamylkin: This probably need a encoding level, other default values needed to initialize log.

glibsm: I think it might be better to start with `zap.NewDevelopmentLogger()` and then `PopulateStruct`, so we get all the zap defaults from the library directly rather than copying them here.

alsamylkin: Ok, so we'll keep examples not working for now, and pin zap's minimal version later?

glibsm: No, that's not what I meant. Pushing the code that fixes examples

service/service_setup.go
```diff
@@ -34,21 +34,46 @@ import (
...
-	err := svc.configProvider.Get("logging").PopulateStruct(&svc.logConfig)
-	if err != nil {
-		zap.L().Info("Logging configuration is not provided, using UberFX defaults", zap.Error(err))
+func (svc *serviceCore) setupLogging() error {
+	cfg := svc.configProvider.Get("logging")
+	if cfg.HasValue() {
+		// Uhhh... this process is not the most elegant.
+		// Becuase log.Configuration embeds zap, the PopulateStruct
+		// does not work properly as it's unable to serialize fields directly
+		// into the embedded struct, so inner struct has to be treated as a
```
alsamylkin: I think we do have a test for nested structs. Added a task to check with zap config, GFM-415

glibsm: I think those tests probably cover naming the structs explicitly, ie

given

```
type struct logConfig{
   zap.Config
   Sentry
}
```

The following would work:

```
logging:
   config:
      foo: bar
   sentry:
      dsn: asdf
```

But not this (flat):

```
logging:
   foo: bar
   sentry:
      dsn: asdf
```

service/service_setup.go
```diff
@@ -34,21 +34,46 @@ import (
 	"go.uber.org/zap"
 )
 
-func (svc *serviceCore) setupLogging() {
-	err := svc.configProvider.Get("logging").PopulateStruct(&svc.logConfig)
-	if err != nil {
-		zap.L().Info("Logging configuration is not provided, using UberFX defaults", zap.Error(err))
+func (svc *serviceCore) setupLogging() error {
+	cfg := svc.configProvider.Get("logging")
+	if cfg.HasValue() {
```
anuptalwalkar: This block can be moved into ulog as `ulog.Configure()` I don't think service setup needs to understand zapCfg as something separate than ulog config.

glibsm: Done.

PR #342, opened 2017-03-01
gRPC example

peter-edge:
This is a quick example of a module implementation for gRPC based on https://github.com/uber-go/fx/pull/275. It could be extended more, the first thing that comes to mind is adding options for interceptors, but this does the basics. Note that this will fail on travis for now due to not properly setting up protoc and some generated file issues.


breerly:
Wait... what?

Why are we putting gRPC directly into UberFX when YARPC already has a gRPC in progress?

PR #344, opened 2017-03-01
Check return values from provider for maps

alsamylkin:
We panic if a config provider returns an empty value to fill the map type.


alsamylkin:
Skip is to add a feature, that @glibsm mentioned in https://github.com/uber-go/fx/pull/341 :
For embedded structs we use a full path to config provider to resolve the field name:
if type A embeds type B and type B has a field F, then config can look like:
```yaml
A:
   F: something
```

But we require it to be 
```yaml
A:
  B:
     F: something
```

PR #347, opened 2017-03-02
Change new module function naming to be consistent across modules

madhuravi:


service/doc.go
```diff
@@ -79,7 +79,7 @@
 //       kafka.Module("kakfa_topic1", []string{"worker"}),
 //     ).WithModule(
 //       "keyvalue",
-//       rpc.ThriftModule(rpc.CreateThriftServiceFunc(NewYarpcThriftHandler)),
```
anuptalwalkar: Didnt catch it before, but name should be "yarpc" not "keyvalue"

service/README.md
```diff
@@ -57,7 +57,7 @@ func main() {
     kafka.Module("kakfa_topic1", []string{"worker"}),
   ).WithModule(
     "keyvalue",
```
anuptalwalkar: name: yarpc

modules/task/README.md
```diff
@@ -29,7 +29,7 @@ import (
 func main() {
   svc, err := service.WithModule(
     "example",
```
anuptalwalkar: name: task?

madhuravi:
Wow, all green looks strange and new ;)

PR #350, opened 2017-03-02
Only run make coveralls in travis on go 1.8

peter-edge:
Coveralls is making two comments on every PR about code coverage, this will make it just one comment I think.


glibsm:
lets see

PR #352, opened 2017-03-02
Make yarpc module not specific to thrift

peter-edge:
The actual implemention of `yarpc.New` is not thrift-specific, this effectively just renames the associated code to remove the thrift reference.


breerly:
Thank you @peter-edge - this change is a huge step in the right direction.

As usual, please get stamps from the UberFX team :)

breerly:
@peter-edge what do you think about a test that uses a different encoding, like json?

If we had that, we would be confident that proto3 will drop in once we have support in YARPC.

sectioneight:
Going to merge this as-is for now, if Peter or somebody on YARPC wants to add more tests for different encodings that'd be great, but trying to minimize conflicts and code churn here.

peter-edge:
Ya sorry I didn't get to this, we'll add some more as we go.

PR #353, opened 2017-03-02
Add ModuleProvider to help with default Module names

peter-edge:


peter-edge:
Merge this before https://github.com/uber-go/fx/pull/349, https://github.com/uber-go/fx/pull/351, and https://github.com/uber-go/fx/pull/352, I will deal with those PRs later.

peter-edge:
Is this code coverage decrease ok? It's just around the additional lines for the module providers I think.

service/module.go
```diff
@@ -23,10 +23,19 @@ package service
...
 	"sync"
-
-	"github.com/uber-go/tally"
 )
 
+// ModuleProvider provides Modules.
+type ModuleProvider interface {
+	// Name is the name that will be used for a new Module
+	// if no name is given as a ModuleOption.
+	Name() string
```
glibsm: I still think `DefaultName` describes it better and needs less comments to explain what it does.

service/module.go
```diff
@@ -23,10 +23,19 @@ package service
...
+	// Create a new Module. The name of the Host and the scoping
+	// of associated functions on the Host will be done using a name
+	// provided by a ModuleOption, or by the Name on this ModuleProvider.
+	Create(Host) (Module, error)
```
glibsm: This additional method is not actually necessary. You can have require that ModuleCreatFunc confirms to the `Namer` interface. Then all modules provides will have to implement one method on their creation function. 

What are your thoughts on that?

https://golang.org/doc/codewalk/functions/

service/module.go
```diff
@@ -23,10 +23,19 @@ package service
 import (
 	"fmt"
 	"sync"
-
-	"github.com/uber-go/tally"
 )
 
+// ModuleProvider provides Modules.
+type ModuleProvider interface {
```
madhuravi: Can we not add this to the Module interface itself? Do we need ModuleProvider?

service/builder.go
```diff
@@ -22,27 +22,26 @@ package service
...
 type Builder struct {
-	options []Option
-	modules []*moduleCreateInfo
+	options     []Option
+	moduleInfos []*moduleInfo
 }
 
 // WithModule is a helper to create a service without any options
-func WithModule(name string, module ModuleCreateFunc, options ...ModuleOption) *Builder {
+func WithModule(provider ModuleProvider, options ...ModuleOption) *Builder {
```
anuptalwalkar: You have `[]ModuleOption` on ModuleProvider, why do you need separate ModuleOption here?

anuptalwalkar: Since it will be one param, we can have `WithModules(provider ...ModuleProvider)`

service/builder.go
```diff
@@ -22,27 +22,26 @@ package service
 
 import "github.com/pkg/errors"
 
-type moduleCreateInfo struct {
-	name             string
-	moduleCreateFunc ModuleCreateFunc
-	options          []ModuleOption
+type moduleInfo struct {
+	provider ModuleProvider
+	options  []ModuleOption
```
anuptalwalkar: Do we need `ModuleOption` here?

modules/uhttp/http.go
```diff
@@ -83,15 +83,26 @@ type Config struct {
...
-		return newModule(mi, hookup, options...)
-	}
+// New returns a new HTTP ModuleProvider.
+func New(hookup GetHandlersFunc, options ...ModuleOption) service.ModuleProvider {
+	return &moduleProvider{hookup, options}
+}
+
+type moduleProvider struct {
+	hookup  GetHandlersFunc
+	options []ModuleOption
```
anuptalwalkar: I like this approach of keeping ModuleOptions with ModuleProvider, may be we can keep this experience consistent across?

service/module.go
sectioneight: It can be part of the interface but still made simple to implement:

```go
type moduleProviderFunc struct {
  name string
  fn moduleCreateFunc
}

func (m moduleProviderFunc) Name() string { return m.name }
func (m moduleProviderFunc) Create(Host) (Module, error) {
  return m.fn(host)
}

func ModuleProviderFromFunc(f moduleCreateFunc, n name) ModuleProvider {
  return moduleProviderFunc{name: n, fn: f}
}
```

service/module.go
sectioneight: Yes, we need a separate entity, otherwise the chicken-and-egg problem that lead to modules not knowing their name

sectioneight:
sorry peter i missed your comment and put you in merge hell :(

service/module.go
peter-edge: I'd rather just add this as a function to what is already there for easier implementation for others.

service/builder.go
peter-edge: Ya, for names and roles.

service/builder.go
peter-edge: It's not on ModuleProvider :)

modules/uhttp/http.go
peter-edge: I don't see where it was, let's sync up.

service/builder.go
anuptalwalkar: I was referring to [https://github.com/uber-go/fx/pull/353#discussion_r104001122](comment)

peter-edge: link broken

anuptalwalkar: oops/ fixed. Lets sync offline

modules/uhttp/http.go
peter-edge: Oh I see what you mean. These are `yarpc.ModuleOptions`, not `service.ModuleOptions`, that's why they are here.

service/module.go
peter-edge: Added

PR #356, opened 2017-03-02
Recover from DIG resolve panics

glibsm:
TODO--


dig/dig.go
```diff
@@ -85,12 +85,16 @@ func (g *Graph) Register(c interface{}) error {
...
-//
-// TODO(glib): catch any and all panics from this method, as there is a lot of reflect going on
-func (g *Graph) Resolve(obj interface{}) error {
+func (g *Graph) Resolve(obj interface{}) (err error) {
 	g.Lock()
 	defer g.Unlock()
 
+	defer func() {
+		if r := recover(); r != nil {
+			err = fmt.Errorf("panic during Resolve %v", r)
```
glibsm: note that I had to change to using the named return, since that is the only way to return an error after panic recovery

dig/dig_test.go
```diff
@@ -237,3 +237,19 @@ func TestEmptyAfterReset(t *testing.T) {
...
+	type Type1 struct{}
+	c := func() *Type1 {
+		panic("RUH ROH")
+	}
+
+	require.NoError(t, g.Register(c))
+
+	var v *Type1
+	err := g.Resolve(&v)
+	require.Contains(t, err.Error(), "panic during Resolve")
```
alsamylkin: Nit: check error contains RUH ROH?

dig/dig_test.go
```diff
@@ -237,3 +237,19 @@ func TestEmptyAfterReset(t *testing.T) {
...
+
+	type Type1 struct{}
+	c := func() *Type1 {
+		panic("RUH ROH")
+	}
+
+	require.NoError(t, g.Register(c))
+
+	var v *Type1
+	err := g.Resolve(&v)
```
alsamylkin: ResolveAll?

PR #357, opened 2017-03-02
Make static provider recursive

alsamylkin:
Right now static provider is not recursive, which make is hard to use.
We can improve it by simply wrapping the YAML config provider.


glibsm:
neat!

PR #360, opened 2017-03-04
Add panic variations of all DIG functions

glibsm:
Provides a little more flexibility for the developer, without having to add a small wrapper in each service.

I've been using `Must*` in my recent prototypes, so I feel good about upstreaming these.

### DIFF
--- a/dig/dig.go
+++ b/dig/dig.go
@@ -80,6 +80,13 @@ func (g *Graph) Register(c interface{}) error {
 	}
 }
 
+// MustRegister will attempt to register the object and panic if error is encountered
+func (g *Graph) MustRegister(c interface{}) {
+	if err := g.Register(c); err != nil {
+		panic(err)
+	}
+}
+
 // Resolve all of the dependencies of the provided class
 //
 // Provided object must be a pointer
@@ -120,6 +127,13 @@ func (g *Graph) Resolve(obj interface{}) (err error) {
 	return nil
 }
 
+// MustResolve calls Resolve and panics if an error is encountered
+func (g *Graph) MustResolve(obj interface{}) {
+	if err := g.Resolve(obj); err != nil {
+		panic(err)
+	}
+}
+
 // ResolveAll the dependencies of each provided object
 // Returns the first error encountered
 func (g *Graph) ResolveAll(objs ...interface{}) error {
@@ -131,6 +145,13 @@ func (g *Graph) ResolveAll(objs ...interface{}) error {
 	return nil
 }
 
+// MustResolveAll calls ResolveAll and panics if an error is encountered
+func (g *Graph) MustResolveAll(objs ...interface{}) {
+	if err := g.ResolveAll(objs...); err != nil {
+		panic(err)
+	}
+}
+
 // RegisterAll registers all the provided args in the dependency graph
 func (g *Graph) RegisterAll(cs ...interface{}) error {
 	for _, c := range cs {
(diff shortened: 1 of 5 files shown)


dig/dig.go
```diff
@@ -120,6 +127,13 @@ func (g *Graph) Resolve(obj interface{}) (err error) {
 	return nil
 }
 
+// MustResolve calls Resolve and panics if an error is encountered
+func (g *Graph) MustResolve(obj interface{}) {
```
alsamylkin: This is a bit confusing: our current pattern(at least in config) is to have Try methods for no panic behavior and regular verb for methods that can panic.

alsamylkin: Another part is we panic on the code that catches panics.. 
Making a sandwich :)

alsamylkin: Also:
https://github.com/golang/go/wiki/CodeReviewComments#dont-panic

glibsm: Well, we want to catch the panic to better decorate it, right? I'd much rather prefer `failed to register *Type1: something something reflect` message rather than a message `something something reflect went boom`.

So I think we won't be able to get away from deferring a recovery anyways, even if we slightly refactor the catch.

To your earlier point, the Try API is going to read weird and be a breaking change to the existing usage.

We probably could go `TryToRegister` and `TryToResolve`, since they read more naturally. I don't have a strong preference, I just want the option to explicit panic if as I developer I know that I can't proceed without a given Register succeeding

glibsm: The "don't" panic should read more like "don't only panic".

We clearly have an API if users want to validate each error and work around it. Which is also a great argument to keep the API as is to illustrate the "default" behaviour

alsamylkin: I would prefer consistency, even if it requires to update current users(only yarpc module, right?)

glibsm: YARPC and all the other code written for dig in test or real services :)

alsamylkin: We are lucky, there is a tool just for that:
https://godoc.org/golang.org/x/tools/cmd/gorename

CHANGELOG.md
```diff
@@ -18,6 +18,12 @@
 * [Breaking] Rename `yarpc.CreateThriftServiceFunc` to `yarpc.ServiceCreateFunc`
   as it is not thrift-specific.
 * Report version metrics for company-wide version usage information.
+* [Breaking] Rename functions inside `package dig` to match other APIs within UberFx:
+        * `Register` to `TryRegister`
```
breerly: I haven't seen this `Try*` convention before, always thought it was implied by the fact that an error is returned.

glibsm:
*shrug*

```
$ pwd
/usr/local/Cellar/go/1.8/libexec/src
$ ag "func.* try.* {" | wc -l
      23
$ ag "func.* must.* {" | wc -l
      54
```

CHANGELOG.md
madhuravi: Also have a preference for dropping the Try*.
Methods that return an error -> Register, Resolve
Methods than panic -> MustRegister, MustResolve

Just sounds more natural. Maybe I've seen more usages of Must than Try.

akshayjshah: Agreed with @madhuravi. See the regexp and template packages for examples in the standard library.

glibsm: An easy fix, I can just force push head~3 on this branch and we'll be back to `Must`.

`Try*` came from the parity with the Config module

alsamylkin:
Hmm, it looks like I need to rename config for consistency?:)

PR #361, opened 2017-03-04
Emit running version periodically

sectioneight:
So we can collect running versions.

GFM-418

### DIFF
--- a/service/service_core.go
+++ b/service/service_core.go
@@ -48,6 +48,7 @@ type metricsCore struct {
 	statsReporter    tally.CachedStatsReporter
 	metricsCloser    io.Closer
 	runtimeCollector *metrics.RuntimeCollector
+	versionEmitter   versionMetricsEmitter
 }
 
 func (mc *metricsCore) Metrics() tally.Scope {
@@ -184,6 +185,11 @@ func (s *serviceCore) setupRuntimeMetricsCollector() error {
 		s.metrics.SubScope("runtime"), time.Second, runtimeMetricsConfig,
 	)
 	return nil
+}
+
+func (s *serviceCore) setupVersionMetricsEmitter() {
+	s.versionEmitter = newVersionMetricsEmitter(s.metrics)
+	s.versionEmitter.start()
 }
 
 func (s *serviceCore) setupTracer() error {
--- a/service/service_core.go
+++ b/service/service_core.go
@@ -48,7 +48,7 @@ type metricsCore struct {
 	statsReporter    tally.CachedStatsReporter
 	metricsCloser    io.Closer
 	runtimeCollector *metrics.RuntimeCollector
-	versionEmitter   versionMetricsEmitter
+	versionEmitter   *versionMetricsEmitter
 }
 
 func (mc *metricsCore) Metrics() tally.Scope {
--- a/service/version_metrics.go
+++ b/service/version_metrics.go
@@ -35,9 +35,9 @@ type versionMetricsEmitter struct {
 	ticker  *time.Ticker
 }
 
-func newVersionMetricsEmitter(scope tally.Scope) versionMetricsEmitter {
+func newVersionMetricsEmitter(scope tally.Scope) *versionMetricsEmitter {
 	t := time.NewTicker(reportingTime)
-	return versionMetricsEmitter{
+	return &versionMetricsEmitter{
 		counter: scope.Tagged(map[string]string{
 			"uberfx-v": fx.Version,
 			"go-v":     runtime.Version(),
@@ -55,5 +55,7 @@ func (v versionMetricsEmitter) start() {
 }
 
 func (v versionMetricsEmitter) close() {
-	v.ticker.Stop()
+	if v.ticker != nil {
+		v.ticker.Stop()
+	}
 }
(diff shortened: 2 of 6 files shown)


service/service_core.go
```diff
@@ -48,6 +48,7 @@ type metricsCore struct {
 	statsReporter    tally.CachedStatsReporter
 	metricsCloser    io.Closer
 	runtimeCollector *metrics.RuntimeCollector
+	versionEmitter   versionMetricsEmitter
```
glibsm: make this a pointer?

version.go
```diff
@@ -0,0 +1,24 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package fx
+
+// Version is exported for runtime compatibility checks
+const Version = "1.0.0-beta2"
```
glibsm: still `beta1` atm

sectioneight: By the time this code is released, it will be beta2, so I could put it as beta1 but it will never be used as such, right? Was just trying to save the releaser of beta2 a few extra steps

service/service_core.go
sectioneight: why?

version.go
glibsm: derp

service/service_core.go
glibsm: consistency with runtime collector

service/version_metrics.go
```diff
@@ -0,0 +1,59 @@
...
+	counter tally.Counter
+	ticker  *time.Ticker
+}
+
+func newVersionMetricsEmitter(scope tally.Scope) versionMetricsEmitter {
+	t := time.NewTicker(reportingTime)
+	return versionMetricsEmitter{
+		counter: scope.Tagged(map[string]string{
+			"uberfx-v": fx.Version,
+			"go-v":     runtime.Version(),
```
glibsm: will this allow us the most flexibility for aggregation, or should we consider making two separate counters?

sectioneight: Yes, this is what the YARPC team does, see the linked code in the jira ticket.

service/service_core.go
sectioneight: fair enough. seems unnecessary given that there's basically zero data in it, and now the thing can be nil

glibsm: well, maybe we can thange the other one to be an object?:)

sectioneight: nah, already pushing up a change to make this a pointer. nbd.

service/version_metrics.go
dmcaulay: is there anyway we can get the current git hash of the service in this? is stats the best way to do this? how will it be consumed?

sectioneight: See the linked (internal) details in the JIRA ticket in the subject. This is for library versions, not service versions. This is for my team to know who's-running-what-in-prod.

dmcaulay: 👍

service/version_metrics.go
```diff
@@ -0,0 +1,61 @@
...
+type versionMetricsEmitter struct {
+	counter tally.Counter
+	ticker  *time.Ticker
+}
+
+func newVersionMetricsEmitter(scope tally.Scope) *versionMetricsEmitter {
+	t := time.NewTicker(reportingTime)
+	return &versionMetricsEmitter{
+		counter: scope.Tagged(map[string]string{
+			"uberfx-v": fx.Version,
```
alsamylkin: tag with the service name?

sectioneight:
Flaky HTTP test :( re-starting travis a few times

service/version_metrics.go
sectioneight: It's automatically tagged with the service name, since we set up the scope in the manager with the service name. Please see details in the jira ticket :)

service/version_metrics.go
sectioneight: FWIW though yes, we could have a separate thing that emits the current service SHA, but I'm not entirely sure who'd use that or why. Our internal build tool adds the current SHA as a variable (via ldflags) that you can access if that's something you're interested in.

service/version_metrics.go
alsamylkin: NVM, it will be inherited form service metrics.

service/version_metrics.go
```diff
@@ -0,0 +1,61 @@
...
+	"go.uber.org/fx"
+
+	"github.com/uber-go/tally"
+)
+
+const reportingTime = 10 * time.Second
+
+type versionMetricsEmitter struct {
+	counter tally.Counter
+	ticker  *time.Ticker
```
alsamylkin: why is it a pointer?

sectioneight: It's defensive programming -- if somebody calls stop on a vME that wasn't constructed through the ctor, we will know. Probably overkill, but an old habit.

PR #362, opened 2017-03-04
Show the whole cycle path in the DIG cycle error

glibsm:
Another TODO--

Sample error from the test:
`unable to register dig.Type3: detected cycle dig.Type3 -> dig.Type1 -> dig.Type2 -> dig.Type3`

### DIFF
--- a/dig/dig.go
+++ b/dig/dig.go
@@ -22,7 +22,6 @@ package dig
 
 import (
 	"bytes"
-	"container/list"
 	"fmt"
 	"reflect"
 	"sync"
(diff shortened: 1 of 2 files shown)


dig/dig.go
```diff
@@ -211,25 +211,27 @@ func (g *Graph) registerConstructor(c interface{}) error {
...
 
 // DFS and tracking if same node is visited twice
-func (g *Graph) recursiveDetectCycles(n graphNode, visited map[string]bool) error {
-	if visited[n.id()] {
-		return errCycle
-	}
-
+func (g *Graph) recursiveDetectCycles(n graphNode, visited map[string]bool, l *list.List) error {
 	visited[n.id()] = true
+	l.PushBack(n.id())
```
alsamylkin: You need to remove a Type from the list(with defer?), otherwise, you'll include a branch info, that is not a cycle, e.g.:
```
T1
|  \
T2  T3
      \
       T1
```
Will print T1 -> T2 -> T3 -> T1, but I'd expect only relevant info: T1 -> T3 -> T1.

Easier approach would be to return (wrap) type in the error and combine them with the current type it if error is not nil.

alsamylkin: There is a similar issue in config.

dig/dig.go
```diff
@@ -211,25 +211,27 @@ func (g *Graph) registerConstructor(c interface{}) error {
 // When a new constructor is being inserted, detect any present cycles
 func (g *Graph) detectCycles(n *funcNode) error {
 	visited := make(map[string]bool)
-	return g.recursiveDetectCycles(n, visited)
+	l := list.New()
```
alsamylkin: list seems to be redundant here, a regular slice will suffice.

dig/dig.go
glibsm: I was pretty sure i needed to pop, but couldn't write a proper case for it on friday evening heh. I will change the test to break (your example) and then fix the array handling

PR #363, opened 2017-03-04
Fix version metrics licence

alsamylkin:
Somehow we missed an extra line.


sectioneight:
When i ran the license script it gave me extra newlines at the top of the
file. something fishy is going on. so i hand massaged these.

alsamylkin:
I've had the same issue with the recent uber-licence 3.1, don't know what's changed there yet, but 2.* works fine for me.

PR #368, opened 2017-03-06
Clean up service package

peter-edge:
This is some more cleaning up of the service package. The primary point of this PR is to have public types before private types. This also removes some newlines and comments that seem unnecessary due to the code being self-documenting.

### DIFF
--- a/service/manager.go
+++ b/service/manager.go
@@ -115,23 +115,23 @@ func newManager(builder *Builder) (Manager, error) {
 	return m, nil
 }
 
-func (s *manager) IsRunning() bool {
-	return s.closeChan != nil
+func (m *manager) IsRunning() bool {
+	return m.closeChan != nil
 }
 
-func (s *manager) OnCriticalError(err error) {
+func (m *manager) OnCriticalError(err error) {
 	shutdown := true
-	if s.observer == nil {
+	if m.observer == nil {
 		zap.L().Warn(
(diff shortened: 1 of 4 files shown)


service/manager.go
```diff
@@ -267,24 +283,8 @@ func (s *manager) addModule(provider ModuleProvider, options ...ModuleOption) er
...
-func (s *manager) Start() {
-	s.start()
-
-	// block until forced exit
-	s.WaitForShutdown(nil)
+	s.moduleWrappers = append(s.moduleWrappers, moduleWrapper)
+	return nil
 }
 
 func (s *manager) start() Control {
```
alsamylkin: nit: could you remove the error on the following line?
I ran [grinder](https://github.com/rsc/grind) on our code and it was the only issue.

service/manager.go
```diff
@@ -174,6 +141,55 @@ func (s *manager) OnCriticalError(err error) {
 	}
 }
 
+// StartAsync service is used as a non-blocking the call on service start. StartAsync will
+// return the call to the caller with a Control to listen on channels
+// and trigger manual shutdowns.
+func (s *manager) StartAsync() Control {
```
alsamylkin: nit: should it be `m *manager`?:)

peter-edge: Lol this was already there, but ya I will change it.

service/manager.go
peter-edge: Done

PR #369, opened 2017-03-06
[WIP] Consume yarpc config v2

alsamylkin:
YARPC team is creating a new way to configure dispatcher, which would change the config structure and improves the way it is loaded: 
[Config](https://github.com/yarpc/yarpc-go/pull/747/files)
[Examples](https://github.com/yarpc/yarpc-go/pull/751/files)

This change will show, how we can consume it. For FX the pros of a new config is that parsing is done entirely by YARPC, so we don't have to keep types that mimic yarpc.Config.
Also FX doesn't handle outbounds yet, so eliminated code could be even bigger.

### DIFF
--- a/modules/yarpc/yarpc.go
+++ b/modules/yarpc/yarpc.go
@@ -33,7 +33,7 @@ import (
 	"go.uber.org/yarpc"
 	"go.uber.org/yarpc/api/middleware"
 	"go.uber.org/yarpc/api/transport"
-	cfg "go.uber.org/yarpc/x/config"
+	yconfig "go.uber.org/yarpc/x/config"
 	"go.uber.org/zap"
 )
 
@@ -104,25 +104,25 @@ type dispatcherController struct {
 }
 
 // Adds the config to the controller.
-func (controller *dispatcherController) addConfig(config configWrapper) {
-	controller.Lock()
-	defer controller.Unlock()
+func (c *dispatcherController) addConfig(config configWrapper) {
+	c.Lock()
+	defer c.Unlock()
 
-	controller.configs = append(controller.configs, &config)
+	c.configs = append(c.configs, &config)
 }
 
 // Adds the config to the controller.
-func (controller *dispatcherController) appendHandler(handler handlerWithDispatcher) {
-	controller.Lock()
-	defer controller.Unlock()
+func (c *dispatcherController) appendHandler(handler handlerWithDispatcher) {
+	c.Lock()
+	defer c.Unlock()
 
-	controller.handlers = append(controller.handlers, handler)
+	c.handlers = append(c.handlers, handler)
 }
 
 // Apply handlers to the dispatcher.
-func (controller *dispatcherController) applyHandlers() error {
-	for _, h := range controller.handlers {
-		if err := h(&controller.dispatcher); err != nil {
+func (c *dispatcherController) applyHandlers() error {
+	for _, h := range c.handlers {
+		if err := h(&c.dispatcher); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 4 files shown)


modules/yarpc/yarpc.go
```diff
@@ -281,16 +247,17 @@ func newYARPCModule(
...
 	}
-	if err := host.Config().Scope("modules").Get(host.Name()).PopulateStruct(&module.config); err != nil {
-		return nil, errs.Wrap(err, "can't read inbounds")
-	}
+
+	val := host.Config().Scope("modules").Get(host.Name()).Value()
 
 	// iterate over inbounds
-	transportsIn, err := prepareInbounds(module.config.Inbounds, host.Name())
+	c, err := cfg.New().LoadConfig(val)
```
alsamylkin: @abhinav I wasn't able to make yarpc config to load our configs(it is a map[interface{}]interface{}) yet. Need to sync about yarpc config decoder offline.

breerly: You'll want to pass in the `map[string]interface{}` for *just* the `yarpc:` key - is that what you're trying to do?

modules/yarpc/yarpc.go
```diff
@@ -132,31 +98,31 @@ type dispatcherController struct {
...
+	configs    []*configWrapper
 	handlers   []handlerWithDispatcher
 	dispatcher yarpc.Dispatcher
 }
 
 // Adds the config to the controller.
-func (c *dispatcherController) addConfig(config yarpcConfig) {
-	c.Lock()
-	defer c.Unlock()
+func (controller *dispatcherController) addConfig(config configWrapper) {
```
abhinav: This change is unrelated to this PR. Please make this change in a different PR, otherwise it adds too much noise to the PR.

Also, FWIW `c` or `ctl` would be considered more idiomatic to Go: https://github.com/golang/go/wiki/CodeReviewComments#receiver-names

modules/yarpc/yarpc.go
```diff
@@ -35,8 +33,7 @@ import (
 	"go.uber.org/yarpc"
 	"go.uber.org/yarpc/api/middleware"
 	"go.uber.org/yarpc/api/transport"
-	"go.uber.org/yarpc/transport/http"
-	tch "go.uber.org/yarpc/transport/tchannel"
+	cfg "go.uber.org/yarpc/x/config"
```
abhinav: Optional: Using `yconfig` might be more obvious

modules/yarpc/yarpc.go
alsamylkin: Yes, in case of :
```yaml
modules:
  hello:
   inbounds:
     http:
       address: ":8080"
     redis:
       queueKey: requests
       processingKey: processing
   transports:
     redis:
       address: localhost:6379
```
We'll pass this map(which is map[string]map[string]map[string]string :) ):
```yaml
   inbounds:
     http:
       address: ":8080"
     redis:
       queueKey: requests
       processingKey: processing
   transports:
     redis:
       address: localhost:6379
```

modules/yarpc/yarpc.go
alsamylkin: Thanks! I was out of c's, cfg's and configs in dispatcherController.Start() :)

modules/yarpc/yarpc.go
```diff
@@ -322,36 +289,10 @@ func newYARPCModule(
 		return nil
 	})
 
-	module.log.Info("Module successfuly created", zap.Any("inbounds", module.config.Inbounds))
+	module.log.Info("Module successfuly created", zap.Reflect("config", c))
```
anuptalwalkar: Reflect?

modules/yarpc/yarpc.go
```diff
@@ -281,16 +247,17 @@ func newYARPCModule(
 		statsClient: newStatsClient(host.Metrics()),
 		log:         ulog.Logger(context.Background()).With(zap.String("module", host.Name())),
 	}
-	if err := host.Config().Scope("modules").Get(host.Name()).PopulateStruct(&module.config); err != nil {
-		return nil, errs.Wrap(err, "can't read inbounds")
-	}
+
+	val := host.Config().Scope("modules").Get(host.Name()).Value()
```
anuptalwalkar: Should we get service name rather than using `host.Name()` here? Also, scoped config doesn't add any value here.

breerly:
Direction is looking good.

@abhinav is still working on getting all the transports ported to this new paradigm.

We'll check back in when this PR needs to be updated.

alsamylkin:
Closing this in favor of https://github.com/uber-go/fx/pull/482, because we are doing some DI refactor, which makes this PR obsolete.

PR #370, opened 2017-03-06
Increase module startup timeout to 10 seconds

madhuravi:
Test service is failing on startup because it;s unable to startup modules in 1 second.
"msg":"Error starting the module","error":"module didn't start after 1s"

### DIFF
--- a/service/manager.go
+++ b/service/manager.go
@@ -410,7 +410,10 @@ func (s *manager) startModules() []error {
 					}
 				case <-time.After(defaultStartupWait):
 					lock.Lock()
-					results = append(results, fmt.Errorf("module didn't start after %v", defaultStartupWait))
+					results = append(
+						results,
+						fmt.Errorf("module: %s didn't start after %v", m.Name(), defaultStartupWait),
+					)
 					lock.Unlock()
 				}
 			}
--- a/service/manager.go
+++ b/service/manager.go
@@ -35,7 +35,7 @@ import (
 )
 
 const (
-	defaultStartupWait = 20 * time.Second
+	defaultStartupWait = 10 * time.Second
 )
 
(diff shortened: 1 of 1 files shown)


peter-edge:
Maybe 10? 20 seems like a lot

madhuravi:
@peter-edge I thought about 10 originally but decided to be more liberal on the timeout since different modules behave differently. We could have users create custom modules for instance that might take longer to start and we don't have a way to override it right now. Having a larger timeout than required would just mean waiting longer in a deployment for a timeout and that seems like an ok tradeoff instead of have the deployment fail when it takes longer to start up.

glibsm:
I'm ok with this as stopgap, though 10 seems better.

Ideally each module should provide its own SLAs on timeouts. Wel never be
able to get one number that suits all use cases

At the very least we should allow service owners to override it, it's not
something the framework should dictate

alsamylkin:
Something is wrong if we need > 1 second to start a test service..
@madhuravi what module is taking so long?

madhuravi:
@alsamylkin The module failing locally was YARPC I think, we don't have clarity in the error message. Added it here.
Also this could be a laptop being slow and might actually be ok on the prod host but the same thing is likely to affect our users as well.

dmcaulay:
@madhuravi should we create an issue so we can track down the why part of this pr? it might be something that our users see, but we should know what's causing this. we should at least point to the culprit when we see a timeout.

madhuravi:
Good point Dan. I personally didn't see it as a big issue. Service and in turn module startup can take in the order of seconds. That seems like a fair SLA to me based on expectation from deployments/startup latencies. What do you think the startup SLA should be? Feel free to cut a task if you think this needs follow up.

PR #371, opened 2017-03-07
Clean up yarpc package

peter-edge:
Continuing cleanup work/procrastinating on writing docs. This has **no change in functionality**, it just moves stuff around and one or two private renames.


modules/yarpc/yarpc_options.go
```diff
@@ -1,47 +0,0 @@
-// Copyright (c) 2017 Uber Technologies, Inc.
```
alsamylkin: Why do you want to combine yarpc and yarpc_options.go? I prefer smaller files..

peter-edge: yarpc_options.go is so small (and logically connected to `Module`) that in this case it makes sense to have the code right next to `Module` and newModule. It makes more sense for a developer in my opinion to immediately see these 20ish lines of code next to where `ModuleOptions` are used.

PR #372, opened 2017-03-07
Stop running make examples explicitly on travis

peter-edge:
make lint calls make examples, so there is no need to spend more time on travis builds for this

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,6 @@ install:
 script:
   - make lint
   - make test
-  - make examples
 
 after_success:
    - '[ "${TRAVIS_GO_VERSION}" == "1.8" ] && make coveralls'


peter-edge:
That's fine, but note this is one of those changes I don't want us to hold
up in paperwork too much - it's clear from the code that this changes
nothing, it's the type of change I would argue doesn't even need review in
general

glibsm:
For me the question is, why does "make lint" run examples?

peter-edge:
Ya I don't know either, I think it might be because it needs to build
examples for linting. I'd prefer if it didnt, and we had a make ci command:

.PHONY: ci
ci: test lint examples

peter-edge:
Yes, this is the problem - what needs to happen to not do this is either 1. the `PKGS` and `LIST_PKGS` variables need to be filtered for lint or 2. the examples generated files need to be checked in. The first option will mean, however, that things like checking for license headers will not happen for example files. I am in favor of 2.

PR #375, opened 2017-03-08
dig: Constructors can now return errors

### LINKED ISSUE #283: dig: constructor returns
Right now it's required that constructors return one and only one value: a pointer.

@dmcaulay pointed out it might be useful to return `*ptr, error`, in case constructors fail to initialize for one reason or another.

abhinav:
This allows constructors to fail with an error.

    func NewMyResult(...) (*MyResult, error) {
        // ..
    }

The error is wrapped for every layer so that you get,

    unable to resolve Foo:
      unable to resolve *MyResult:
        root cause

This works by changing `funcNode`s to always be functions in the form,

    func(...) ($ResultType, error)

When constructors are initially registered, we wrap them to always return a
`nil` error if they are in the no-error form.

---

CC @glibsm @breerly

I would also like feedback on whether we want error wrapping or not. I think
that if you have a missing dependency, there isn't much you can do to recover
from it so getting the original error back doesn't serve much value. Getting an
error message wrapped in 5 layers of context tells you exactly what failed, why
it failed, and why it was even needed in the first place.

Resolves #283

### DIFF
--- a/dig/dig.go
+++ b/dig/dig.go
@@ -31,37 +31,13 @@ import (
 
 var (
 	errParamType     = errors.New("graph injection must be done through a pointer or a function")
-	errReturnCount   = errors.New("constructor function must return up to two values")
+	errReturnCount   = errors.New("constructor function must one or two values")
 	errReturnKind    = errors.New("constructor return type must be a pointer")
 	errReturnErrKind = errors.New("second return value of constructor must be error")
 	errArgKind       = errors.New("constructor arguments must be pointers")
 
 	_typeOfError = reflect.TypeOf((*error)(nil)).Elem()
 )
-
-// Wraps a function that returns a single value to return that value and a nil
-// error.
-func addErrorReturn(c interface{}) interface{} {
-	v := reflect.ValueOf(c)
-	t := v.Type()
-
-	if t.NumOut() != 1 {
-		panic(fmt.Sprintf("addErrorReturn called with invalid function: %v", t))
-	}
-
-	returnTypes := []reflect.Type{t.Out(0), _typeOfError}
-	argTypes := make([]reflect.Type, t.NumIn())
-	for i := 0; i < t.NumIn(); i++ {
-		argTypes[i] = t.In(i)
-	}
-
-	return reflect.MakeFunc(
-		reflect.FuncOf(argTypes, returnTypes, t.IsVariadic()),
-		func(args []reflect.Value) []reflect.Value {
-			return append(v.Call(args), reflect.Zero(_typeOfError))
-		},
-	).Interface()
-}
 
 // New returns a new Dependency Injection Graph
 func New() *Graph {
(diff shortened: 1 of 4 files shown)


glibsm:
Thanks for jumping in on this!

Thoughts on returning `$type` OR `($type, error)`?

glibsm:
I also think it might be worthwhile to consider wrapping `$type, error` under `dig.Object` struct or something like that, so we have some room to maneuver in the future.

Then the rule becomes `dig constructors must return dig.Object` (or whatever we call it). That will allow us in the future to extend the API to add a named type, or scope, for instance.

dig/dig.go
```diff
@@ -30,12 +30,39 @@ import (
...
+	returnTypes := []reflect.Type{t.Out(0), _typeOfError}
+	argTypes := make([]reflect.Type, t.NumIn())
+	for i := 0; i < t.NumIn(); i++ {
+		argTypes[i] = t.In(i)
+	}
+
+	return reflect.MakeFunc(
+		reflect.FuncOf(argTypes, returnTypes, t.IsVariadic()),
+		func(args []reflect.Value) []reflect.Value {
+			return append(v.Call(args), reflect.Zero(_typeOfError))
```
alsamylkin: Should we actually check, that returned pointer is not nil?

dig/dig.go
```diff
@@ -62,15 +89,23 @@ func (g *Graph) Register(c interface{}) error {
 
 	switch ctype.Kind() {
 	case reflect.Func:
-		if ctype.NumOut() != 1 {
+		if ctype.NumOut() == 0 || ctype.NumOut() > 2 {
```
anuptalwalkar: > Provided argument must be a function that returns exactly one pointer argument

update the documentation on line 82?

anuptalwalkar: I meant to add something similar to "second return value of constructor must be error"

abhinav:
@glibsm This supports returning `$type` or `($type, error)`. Constructors can be in either form but they're internally adapted to the latter form so that the rest of the internal code stays simple. Note that none of the existing tests had to be changed.

dig/dig.go
```diff
@@ -30,12 +30,39 @@ import (
...
+	errReturnKind    = errors.New("constructor return type must be a pointer")
+	errReturnErrKind = errors.New("second return value of constructor must be error")
+	errArgKind       = errors.New("constructor arguments must be pointers")
+
+	_typeOfError = reflect.TypeOf((*error)(nil)).Elem()
 )
 
+// Wraps a function that returns a single value to return that value and a nil
+// error.
+func addErrorReturn(c interface{}) interface{} {
```
glibsm: This is pretty neat :)

I don't actually think we need to wrap every non-error function though. We can move that logic all the way down to `value()` method on the `funcNode`

At the moment it does `err, _ := vs[1].Interface().(error)`. Instead of doing that we can check if there is a second value and if there is, then check it. Otherwise just proceed as normal with one value

abhinav: Ah, that's a good point

glibsm:
re: wrapping, I'm all for wrapping @abhinav. 

I think the more information we can provide to the user as far as why or when something failed the better, since a lot of this is in the "reflect magic" category and provides no feedback until runtime.

dig/dig.go
abhinav: Validation should probably happen in funcNode instead. This function was deleted based on @glibsm's comment so this isn't necessary anyway.

breerly:
🎉

dig/dig.go
abhinav: Fixed

dig/dig.go
abhinav: Yeah, I updated this. It is now,

> The provided argument must be a function that accepts its dependencies as
arguments and returns a single result, which must be a pointer type.
The function may optionally return an error as a second result.

glibsm:
that last commit is 💯

PR #380, opened 2017-03-08
Use go.uber.org/tools/update-license

peter-edge:


alsamylkin:
Will it affect our current license? I thought the new package adds an extra line on top..

peter-edge:
It doesn't, and this PR is just a change to use the new tools package.

PR #382, opened 2017-03-08
Prototype: interpolation of env vars in YAML

glibsm:
Here is what I'm proposing to completely remove the `envProvider`. I've hard some concerns about going with Regexp and I'm not set in stone on this. We can write a more sophisticated parser, but I'd prefer to keep the functionality to just this use-case. I don't want to add an open-ended template support.

Seems to have worked out pretty well, there is one TODO to think through the syntax of what it looks like if people want to use `${FOO}` as-is (super rare).

/cc @breerly @abhinav for comments

### DIFF
--- a/config/yaml.go
+++ b/config/yaml.go
@@ -41,8 +41,10 @@ type yamlConfigProvider struct {
 }
 
 var (
-	_envSeparator          = []byte(":")
-	_             Provider = &yamlConfigProvider{}
+	_envSeparator = []byte(":")
+	_emptyString  = []byte(`""`)
+
+	_ Provider = &yamlConfigProvider{}
 )
 
 func newYAMLProviderCore(files ...io.ReadCloser) Provider {
@@ -357,7 +359,7 @@ func interpolateEnvVars(data []byte) ([]byte, error) {
 		}
 
 		// return empty byte slices for "" as the default
-		if bytes.Compare(def, []byte(`""`)) == 0 {
+		if bytes.Compare(def, _emptyString) == 0 {
 			return nil
 		}
 
--- a/config/yaml.go
+++ b/config/yaml.go
@@ -327,7 +327,7 @@ func unmarshalYAMLValue(reader io.ReadCloser, value interface{}) error {
 // TODO: what if someone wanted a literal ${FOO} in config? need a small escape hatch
 func interpolateEnvVars(data []byte) ([]byte, error) {
 	// ${VAR_NAME:default} or ${VAR_NAME}
-	reg := regexp.MustCompile(`\$\{\w+:?[\w"]*}`)
+	reg := regexp.MustCompile(`\$\{\w+:?[^}]*}`)
 
 	var regErr error
 	data = reg.ReplaceAllFunc(data, func(in []byte) []byte {
--- a/config/yaml.go
+++ b/config/yaml.go
@@ -327,7 +327,7 @@ func unmarshalYAMLValue(reader io.ReadCloser, value interface{}) error {
 // TODO: what if someone wanted a literal ${FOO} in config? need a small escape hatch
 func interpolateEnvVars(data []byte) ([]byte, error) {
 	// ${VAR_NAME:default} or ${VAR_NAME}
-	reg := regexp.MustCompile(`\$\{\w+(:\w+)?}`)
+	reg := regexp.MustCompile(`\$\{\w+:?[\w"]*}`)
 
 	var regErr error
 	data = reg.ReplaceAllFunc(data, func(in []byte) []byte {
@@ -350,7 +350,18 @@ func interpolateEnvVars(data []byte) ([]byte, error) {
 		}
 
 		// return the default
-		return split[1]
+		def := split[1]
+		if len(def) < 1 {
+			regErr = fmt.Errorf(`default is empty for %s (use "" for empty string)`, key)
+			return in
+		}
+
+		// return empty byte slices for "" as the default
+		if bytes.Compare(def, []byte(`""`)) == 0 {
+			return nil
+		}
+
+		return def
 	})
 
 	if regErr != nil {
--- a/config/yaml.go
+++ b/config/yaml.go
@@ -350,7 +350,7 @@ func interpolateEnvVars(data []byte) ([]byte, error) {
 		}
 
 		// return the default
-		def := split[1]
+		def := bytes.Join(split[1:], _envSeparator)
 		if len(def) == 0 {
 			errs = append(errs, fmt.Sprintf(`default is empty for %s (use "" for empty string)`, key))
 			return in
--- a/config/yaml.go
+++ b/config/yaml.go
@@ -25,11 +25,14 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"os"
 	"reflect"
+	"regexp"
+	"strconv"
 	"strings"
 
+	"github.com/pkg/errors"
 	"gopkg.in/yaml.v2"
-	"strconv"
 )
 
 type yamlConfigProvider struct {
@@ -37,7 +40,10 @@ type yamlConfigProvider struct {
 	vCache map[string]Value
 }
 
-var _ Provider = &yamlConfigProvider{}
+var (
+	_envSeparator          = []byte(":")
+	_             Provider = &yamlConfigProvider{}
+)
 
 func newYAMLProviderCore(files ...io.ReadCloser) Provider {
 	var root interface{} = make(map[interface{}]interface{})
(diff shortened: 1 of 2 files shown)


config/yaml.go
```diff
@@ -290,15 +296,70 @@ func (n *yamlNode) Children() []*yamlNode {
...
+		key := string(split[0])
+
+		if envVal, ok := os.LookupEnv(key); ok {
+			return []byte(envVal)
+		}
+
+		// value not found in the env, lets check if there is a default
+		if len(split) < 2 {
+			regErr = fmt.Errorf("environment variable %s not found and default is not provided", key)
+			return in
```
anuptalwalkar: should we panic?

config/yaml.go
```diff
@@ -290,15 +296,70 @@ func (n *yamlNode) Children() []*yamlNode {
...
+	data = reg.ReplaceAllFunc(data, func(in []byte) []byte {
+		// remove ${ and }, just look inside the braces
+		valAndMaybeDefault := in[2 : len(in)-1]
+
+		// ${VALUE:DEFAULT} -> VALUE, DEFAULT
+		// ${VALUE} -> VALUE, empty default
+		split := bytes.Split(valAndMaybeDefault, _envSeparator)
+		key := string(split[0])
+
+		if envVal, ok := os.LookupEnv(key); ok {
```
anuptalwalkar: My only fear here is that we have a caching layer on top of it. (or below..Im not sure now). This will be called only on startup time, and its not really updatable using environment variable.

config/yaml_test.go
```diff
@@ -56,6 +56,39 @@ func TestYamlSimple(t *testing.T) {
...
+func TestYAMLEnvInterpolation(t *testing.T) {
+	t.Parallel()
+	defer env.Override(t, "OWNER_EMAIL", "hello@there.yasss")()
+
+	cfg := []byte(`
+name: some name here
+owner: ${OWNER_EMAIL}
+module:
+  fake:
+    number: ${FAKE_NUMBER:321}`)
```
anuptalwalkar: test for incomplete? `${FAKE_NUMBER:}`

config/yaml.go
```diff
@@ -25,19 +25,25 @@ import (
...
 
 type yamlConfigProvider struct {
 	root   *yamlNode
 	vCache map[string]Value
 }
 
-var _ Provider = &yamlConfigProvider{}
+var (
+	_envSeparator          = []byte(":")
+	_             Provider = &yamlConfigProvider{}
```
anuptalwalkar: nit: new line here will make formatting go away.

config/yaml.go
glibsm: it's part of a chain of functions that return an error (the outermost function panics), and it think it's ok to error wrap all the way up and panic there

config/yaml_test.go
glibsm: good idea

config/yaml.go
glibsm: That is the intended behavior. Env variables need to be available to the binary on start, preferable through prepending them to the command line on binary launch (which is how Mesos works, for instance).

config/yaml_test.go
glibsm: Actually, I'm not sure if it's a good idea anymore :) What if empty string is the default? 🤔

What then, force `${FOO:""}`?

anuptalwalkar: Test for not failing on default being empty?

anuptalwalkar: I am not sure. I don't think it should fail anyway. I wanted some validation for not panicing.

glibsm: I think empty default is legit, i.e. `fullTelephone: 1-800-LOLZ${OPTIONAL_EXTENSION:}`

Should result in `1-800-LOLZ` if `OPTIONAL_EXTENSION` is not present in env. I am happy to discuss if we want a special syntax for that, but right now doing  `fullTelephone: 1-800-LOLZ${OPTIONAL_EXTENSION}` will panic, since there is no fallback at all

glibsm: Figured it out. Pushing a commit with this

config/yaml.go
```diff
@@ -25,19 +25,27 @@ import (
...
 	root   *yamlNode
 	vCache map[string]Value
 }
 
-var _ Provider = &yamlConfigProvider{}
+var (
+	_envSeparator = []byte(":")
+	_emptyString  = []byte(`""`)
+
+	_ Provider = &yamlConfigProvider{}
```
alsamylkin: No need to do this, we already check it indirectly with constructor.

config/yaml_test.go
```diff
@@ -56,6 +56,63 @@ func TestYamlSimple(t *testing.T) {
 	assert.Equal(t, ":28941", c.AsString())
 }
 
+func TestYAMLEnvInterpolation(t *testing.T) {
+	t.Parallel()
+	defer env.Override(t, "OWNER_EMAIL", "hello@there.yasss")()
+
+	cfg := []byte(`
+name: some name here
+owner: ${OWNER_EMAIL}
```
alsamylkin: test case when env variable OWNER_EMAIL is ${OWNER_EMAIL}? :)

config/yaml.go
```diff
@@ -290,15 +298,81 @@ func (n *yamlNode) Children() []*yamlNode {
...
+// port, the config would look like this:
+//
+//   modules:
+//     http:
+//       port: ${HTTP_PORT:8080}
+//
+// In the case that HTTP_PORT is not provided, default value (in this case 8080)
+// will be used
+//
+// TODO: what if someone wanted a literal ${FOO} in config? need a small escape hatch
```
alsamylkin: in this PR?

config/yaml.go
```diff
@@ -290,15 +298,81 @@ func (n *yamlNode) Children() []*yamlNode {
...
+func interpolateEnvVars(data []byte) ([]byte, error) {
+	// ${VAR_NAME:default} or ${VAR_NAME}
+	reg := regexp.MustCompile(`\$\{\w+:?[^}]*}`)
```
alsamylkin: Can be moved to globals?

config/yaml.go
```diff
@@ -290,15 +298,81 @@ func (n *yamlNode) Children() []*yamlNode {
...
+
+		// value not found in the env, lets check if there is a default
+		if len(split) < 2 {
+			regErr = fmt.Errorf("environment variable %s not found and default is not provided", key)
+			return in
+		}
+
+		// return the default
+		def := split[1]
+		if len(def) < 1 {
```
alsamylkin: == 0?:)

config/yaml.go
```diff
@@ -290,15 +298,81 @@ func (n *yamlNode) Children() []*yamlNode {
...
+
+		// return empty byte slices for "" as the default
+		if bytes.Compare(def, _emptyString) == 0 {
+			return nil
+		}
+
+		return def
+	})
+
+	if regErr != nil {
```
alsamylkin: Simply `return data, regErr`?

config/yaml.go
```diff
@@ -290,15 +298,81 @@ func (n *yamlNode) Children() []*yamlNode {
...
+		// value not found in the env, lets check if there is a default
+		if len(split) < 2 {
+			regErr = fmt.Errorf("environment variable %s not found and default is not provided", key)
+			return in
+		}
+
+		// return the default
+		def := split[1]
+		if len(def) < 1 {
+			regErr = fmt.Errorf(`default is empty for %s (use "" for empty string)`, key)
```
alsamylkin: This gives you only the last error, not all errors in the config..

config/yaml.go
```diff
@@ -290,15 +298,81 @@ func (n *yamlNode) Children() []*yamlNode {
...
+//
+// TODO: what if someone wanted a literal ${FOO} in config? need a small escape hatch
+func interpolateEnvVars(data []byte) ([]byte, error) {
+	// ${VAR_NAME:default} or ${VAR_NAME}
+	reg := regexp.MustCompile(`\$\{\w+:?[^}]*}`)
+
+	var regErr error
+	data = reg.ReplaceAllFunc(data, func(in []byte) []byte {
+		// remove ${ and }, just look inside the braces
+		valAndMaybeDefault := in[2 : len(in)-1]
```
alsamylkin: Could you use grouping in the regex? All these magic numbers look too hacky.

glibsm:
There is no groups support for replacing data within a byte slice. Doing groups will require a lot more indexing work, which you will like even less..

config/yaml.go
glibsm: there are no groups in this usage pattern

config/yaml.go
```diff
@@ -290,15 +298,81 @@ func (n *yamlNode) Children() []*yamlNode {
...
+		}
+
+		// value not found in the env, lets check if there is a default
+		if len(split) < 2 {
+			regErr = fmt.Errorf("environment variable %s not found and default is not provided", key)
+			return in
+		}
+
+		// return the default
+		def := split[1]
```
alsamylkin: if a value contains `:` are we going to return only the part between first 2 colons?

glibsm: now you're talking

config/yaml.go
glibsm: maybe, maybe not

glibsm:
going to open another PR using `os.Expand` for comparison purposes of the two approaches

config/yaml.go
```diff
@@ -290,15 +299,77 @@ func (n *yamlNode) Children() []*yamlNode {
...
+// TODO: what if someone wanted a literal ${FOO} in config? need a small escape hatch
+func interpolateEnvVars(data []byte) ([]byte, error) {
+	var errs []string
+	data = _interpolationRegex.ReplaceAllFunc(data, func(in []byte) []byte {
+		// remove ${ and }, just look inside the braces
+		valAndMaybeDefault := in[2 : len(in)-1]
+
+		// ${VALUE:DEFAULT} -> VALUE, DEFAULT
+		// ${VALUE} -> VALUE, empty default
+		split := bytes.Split(valAndMaybeDefault, _envSeparator)
```
alsamylkin: You don't need a split, just a first occurrence of _envSeparator..

config/yaml_test.go
alsamylkin: Actually it is not needed, because replace returns a copy of the input slice.

glibsm:
Closing in favor of #384 , as it's using a system library. 
Can iterate from there as necessary

PR #383, opened 2017-03-09
Allow modifiable struct handler names in thriftsync

anuptalwalkar:
Providing some additional functionality with naming. Currently the handler structs are named as "YARPCHandler". Now this can be configured in thriftsync args.


glibsm:
just for my own curiosity, is there always one struct per yarpc module, or can there be multiple handler structs?

anuptalwalkar:
It's one per service. and as where it stands, thriftsync can only since one service at a time.

PR #384, opened 2017-03-09
Interpolate environment variables into YAML config

glibsm:
Different way to write #382 

This does two extra conversions of `[]byte` to `string`. Since it only happens once at startup I'm willing to overlook that.

@alsamylkin


config/yaml.go
```diff
@@ -329,36 +325,34 @@ func unmarshalYAMLValue(reader io.ReadCloser, value interface{}) error {
...
+		var key string
+		var def string
+
+		if sep == -1 {
+			// separator missing - everything is the key ${KEY}
+			key = in
+		} else {
+			// ${KEY:DEFAULT}
+			key = in[:sep]
+			def = in[sep+1:]
```
glibsm: this index logic can easily replace the `split/join` on the other diff, and seems much cleaner.

alsamylkin:
I prefer this version, because there is no need for a regexp, just standard functions.

config/yaml.go
```diff
@@ -41,11 +40,8 @@ type yamlConfigProvider struct {
 }
 
 var (
-	_envSeparator = []byte(":")
-	_emptyString  = []byte(`""`)
-
-	// ${VAR_NAME:default} or ${VAR_NAME}
-	_interpolationRegex = regexp.MustCompile(`\$\{\w+:?[^}]*}`)
+	_envSeparator = ":"
+	_emptyDefault = `""`
```
alsamylkin: Do we need to have it explicitly? We can simply use ${PATH}, instead of ${PATH:""}

glibsm: `${PATH}` is currently defined as invalid if the envvar is missing. We shouldn't assume that it's fine that's it's not there and force the service owner to tell us "it's ok if the variable is missing"

sectioneight: Huge +1, let's not emulate bash here, let's be as explicit as possible

config/yaml.go
```diff
@@ -329,36 +325,34 @@ func unmarshalYAMLValue(reader io.ReadCloser, value interface{}) error {
...
-		valAndMaybeDefault := in[2 : len(in)-1]
-
-		// ${VALUE:DEFAULT} -> VALUE, DEFAULT
-		// ${VALUE} -> VALUE, empty default
-		split := bytes.Split(valAndMaybeDefault, _envSeparator)
-		key := string(split[0])
-
-		if envVal, ok := os.LookupEnv(key); ok {
-			return []byte(envVal)
+	// Is this conversion ok?
```
sectioneight: Is this a TODO? A performance concern? We could consider using a bytes.Buffer if performance is the issue

config/yaml.go
```diff
@@ -329,36 +325,34 @@ func unmarshalYAMLValue(reader io.ReadCloser, value interface{}) error {
...
+	str := string(data)
+	errs := []string{}
```
sectioneight: I like this syntax better but I see people like @prashantv doing:

```go
var errs []string
```

and I assume he has a good reason

config/yaml.go
```diff
@@ -329,36 +325,34 @@ func unmarshalYAMLValue(reader io.ReadCloser, value interface{}) error {
...
+
+	str = os.Expand(str, func(in string) string {
```
sectioneight: If we broke this anonymous function out into a named func it may be easier to test

config/yaml.go
alsamylkin: I think it is to avoid unnecessary allocation:
https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

config/yaml.go
glibsm: Was just a note from myself to myself to make sure it's not too wasteful. There are a lot of similar APIs offered between `[]byte` and `string`, so there may be a way to do it without this conversion at all, but the again, that might not be all that necessary from the get go, especially since this only happens once

config/yaml.go
alsamylkin: I would prefer to use regular errors and recursively wrap them.

glibsm: I haven't seen that pattern before, do you have any examples?
I think array of errors is pretty standard when doing something like this, but I may be wrong.

alsamylkin: I meant to use errors.Wrap(), that is going to have a message for values that don't have valid defaults. It would require an additional if though, so it is up to you.

PR #385, opened 2017-03-09
Set root name as default host.Name and separate the module name

madhuravi:
Post the module re-write PR https://github.com/uber-go/fx/pull/275, the default behavior for yarpc module has changed to use the module name "yarpc" as the service name. 
We want the root level "name:" on config to be used as the default service. This is the majority use case.
Separated the concepts of module name and service name, cleaned up some of the module structs based on this.

### DIFF
--- a/modules/uhttp/http.go
+++ b/modules/uhttp/http.go
@@ -55,8 +55,6 @@ const (
 
 	// default healthcheck endpoint
 	healthPath = "/health"
-
-	_modName = "uhttp"
 )
 
 var _ service.Module = &Module{}
@@ -87,12 +85,14 @@ type GetHandlersFunc func(service service.Host) []RouteHandler
 
 // New returns a new HTTP ModuleProvider.
 func New(hookup GetHandlersFunc, options ...ModuleOptionFn) service.ModuleProvider {
-	return service.ModuleProviderFromFunc(func(host service.Host) (service.Module, error) {
-		return newModule(host, hookup, options...)
+	modName := "uhttp"
+	return service.ModuleProviderFromFunc(modName, func(host service.Host) (service.Module, error) {
+		return newModule(modName, host, hookup, options...)
 	})
 }
 
 func newModule(
+	modName string,
 	host service.Host,
 	getHandlers GetHandlersFunc,
 	options ...ModuleOptionFn,
@@ -108,8 +108,8 @@ func newModule(
 		Port:    defaultPort,
 		Timeout: defaultTimeout,
 	}
-	log := ulog.Logger(context.Background()).With(zap.String("module", _modName))
-	if err := host.Config().Scope("modules").Get(_modName).PopulateStruct(&cfg); err != nil {
+	log := ulog.Logger(context.Background()).With(zap.String("module", modName))
+	if err := host.Config().Scope("modules").Get(modName).PopulateStruct(&cfg); err != nil {
 		log.Error("Error loading http module configuration", zap.Error(err))
 	}
 	module := &Module{
--- a/modules/uhttp/http.go
+++ b/modules/uhttp/http.go
@@ -84,7 +84,7 @@ type Config struct {
 type GetHandlersFunc func(service service.Host) []RouteHandler
 
 // New returns a new HTTP ModuleProvider.
-func New(hookup GetHandlersFunc, options ...ModuleOptionFn) service.ModuleProvider {
+func New(hookup GetHandlersFunc, options ...ModuleOption) service.ModuleProvider {
 	modName := "uhttp"
 	return service.ModuleProviderFromFunc(modName, func(host service.Host) (service.Module, error) {
 		return newModule(modName, host, hookup, options...)
@@ -95,11 +95,11 @@ func newModule(
 	modName string,
 	host service.Host,
 	getHandlers GetHandlersFunc,
-	options ...ModuleOptionFn,
+	options ...ModuleOption,
 ) (*Module, error) {
-	moduleOption := &moduleOption{}
+	moduleOptions := &moduleOptions{}
 	for _, option := range options {
-		if err := option(moduleOption); err != nil {
+		if err := option(moduleOptions); err != nil {
 			return nil, err
 		}
 	}
@@ -119,7 +119,7 @@ func newModule(
 		config:   cfg,
 		log:      log,
 	}
-	module.mcb = module.mcb.AddMiddleware(moduleOption.inboundMiddleware...)
+	module.mcb = module.mcb.AddMiddleware(moduleOptions.inboundMiddleware...)
 	return module, nil
 }
 
--- a/modules/uhttp/http.go
+++ b/modules/uhttp/http.go
@@ -55,6 +55,8 @@ const (
 
 	// default healthcheck endpoint
 	healthPath = "/health"
+
+	_modName = "uhttp"
 )
 
 var _ service.Module = &Module{}
@@ -84,8 +86,8 @@ type Config struct {
 type GetHandlersFunc func(service service.Host) []RouteHandler
 
 // New returns a new HTTP ModuleProvider.
-func New(hookup GetHandlersFunc, options ...ModuleOption) service.ModuleProvider {
-	return service.ModuleProviderFromFunc("http", func(host service.Host) (service.Module, error) {
+func New(hookup GetHandlersFunc, options ...ModuleOptionFn) service.ModuleProvider {
+	return service.ModuleProviderFromFunc(func(host service.Host) (service.Module, error) {
 		return newModule(host, hookup, options...)
 	})
 }
@@ -93,11 +95,11 @@ func New(hookup GetHandlersFunc, options ...ModuleOption) service.ModuleProvider
 func newModule(
 	host service.Host,
 	getHandlers GetHandlersFunc,
-	options ...ModuleOption,
+	options ...ModuleOptionFn,
 ) (*Module, error) {
-	moduleOptions := &moduleOptions{}
+	moduleOption := &moduleOption{}
 	for _, option := range options {
-		if err := option(moduleOptions); err != nil {
+		if err := option(moduleOption); err != nil {
 			return nil, err
 		}
 	}
@@ -106,8 +108,8 @@ func newModule(
 		Port:    defaultPort,
 		Timeout: defaultTimeout,
 	}
-	log := ulog.Logger(context.Background()).With(zap.String("module", host.Name()))
-	if err := host.Config().Scope("modules").Get(host.Name()).PopulateStruct(&cfg); err != nil {
+	log := ulog.Logger(context.Background()).With(zap.String("module", _modName))
+	if err := host.Config().Scope("modules").Get(_modName).PopulateStruct(&cfg); err != nil {
 		log.Error("Error loading http module configuration", zap.Error(err))
 	}
 	module := &Module{
@@ -117,7 +119,7 @@ func newModule(
 		config:   cfg,
 		log:      log,
 	}
-	module.mcb = module.mcb.AddMiddleware(moduleOptions.inboundMiddleware...)
+	module.mcb = module.mcb.AddMiddleware(moduleOption.inboundMiddleware...)
 	return module, nil
 }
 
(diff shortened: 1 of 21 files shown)


madhuravi:
Some weird glue "cannot detect VCS" errors are causing the build to fail, will dig into that tomorrow.

service/module.go
```diff
@@ -108,47 +106,53 @@ type moduleWrapper struct {
...
 	module, err := moduleProvider.Create(scopedHost)
 	if err != nil {
 		return nil, err
 	}
 	if module == nil {
 		return nil, nil
 	}
-	return &moduleWrapper{module: module, scopedHost: scopedHost}, nil
+	return &moduleWrapper{
+		name:       reflect.TypeOf(module).PkgPath(),
```
madhuravi: Another way to solve this is bring back the Name() function on the Module interface. That's more explicit and clear rather than picking up the package name via reflect. Thoughts?

modules/uhttp/http_options.go
```diff
@@ -20,17 +20,17 @@
 
 package uhttp
 
-// ModuleOption is a function that configures module creation.
-type ModuleOption func(*moduleOptions) error
+// ModuleOptionFn is a function that configures module creation.
+type ModuleOptionFn func(*moduleOption) error
```
peter-edge: This should still be named ModuleOption - this is consistent with most golang libraries. That this happens to be a function is of no concern to the user.

modules/yarpc/yarpc.go
```diff
@@ -93,21 +94,26 @@ type Module struct {
 	controller  *dispatcherController
 }
 
-// ModuleOption is a function that configures module creation.
-type ModuleOption func(*moduleOptions) error
+// ModuleOptionFn is a function that configures module creation.
+type ModuleOptionFn func(*moduleOption) error
```
peter-edge: Same comment

service/module.go
```diff
@@ -47,58 +45,58 @@ type Module interface {
...
-func ModuleProviderFromFunc(name string, createFunc func(Host) (Module, error)) ModuleProvider {
-	return &moduleProviderFunc{name, createFunc}
+func ModuleProviderFromFunc(createFunc func(Host) (Module, error)) ModuleProvider {
+	return &moduleProvider{createFunc: createFunc}
 }
 
-// ModuleOption is a function that configures module creation.
-type ModuleOption func(*moduleOptions) error
+// ModuleOptionFn is a function that configures module creation.
+type ModuleOptionFn func(*moduleOption) error
```
peter-edge: Same comment

service/module.go
```diff
@@ -47,58 +45,58 @@ type Module interface {
...
 
-// WithModuleName will override the name given by the ModuleProvider.
-func WithModuleName(name string) ModuleOption {
-	return func(o *moduleOptions) error {
-		o.name = name
+// WithName will override the name given by the ModuleProvider.
+func WithName(name string) ModuleOptionFn {
```
peter-edge: I think this should still be WithModuleName - you have WithModule (not With) on service, and there is more than one construct in the service package.

service/module.go
```diff
@@ -47,58 +45,58 @@ type Module interface {
...
 	}
 }
 
-// WithModuleRole will add a role to the Module.
+// WithRole will add a role to the Module.
 //
 // If the role was already added, this will be a no-op.
-func WithModuleRole(role string) ModuleOption {
-	return func(o *moduleOptions) error {
+func WithRole(role string) ModuleOptionFn {
```
peter-edge: Same comment

service/module.go
peter-edge: We already discussed using reflect and ruled this out.

madhuravi: Cool, I will add Name() function back on the Module interface instead.

alsamylkin: Yeah, that's dangerous, e.g. if module is an unnamed type, it will be an empty string..

service/module.go
madhuravi: It's not the module name though, it is the service name and service.WithServiceName stutters.

service/module.go
madhuravi: same reply ;)

service/module.go
```diff
@@ -47,58 +45,58 @@ type Module interface {
...
 				return nil
 			}
 		}
-		o.roles = append(o.roles, role)
+		o.Roles = append(o.Roles, role)
 		return nil
 	}
 }
 
+// moduleOption specifies options for module name and role
```
glibsm: this comment is invalid

service/module.go
peter-edge: We discussed name on module extensively I believe, this was ruled out and this is why we had ModuleProvider.

modules/uhttp/http_options.go
madhuravi: The user does not use this directly. I prefer Fn for aliases that represent the function. We can add an interface and name it ModuleOption and expose that if you prefer that.

alsamylkin: Lets separate this change from the intended changes in the PR title?

There are some other options we can use and implement. I do like Zap interface with private functions.

madhuravi: I don't think every rename should be a new PR :( Kills productivity.

modules/uhttp/http.go
```diff
@@ -106,8 +106,8 @@ func newModule(
 		Port:    defaultPort,
 		Timeout: defaultTimeout,
 	}
-	log := ulog.Logger(context.Background()).With(zap.String("module", host.Name()))
-	if err := host.Config().Scope("modules").Get(host.Name()).PopulateStruct(&cfg); err != nil {
+	log := ulog.Logger(context.Background()).With(zap.String("module", host.ModuleName()))
```
anuptalwalkar: Suggestion: since there is no context, you can use `zap.L().With(...` instead of `ulog.Logger(ctx)`

service/manager.go
```diff
@@ -385,25 +385,25 @@ func (m *manager) startModules() []error {
 	// calls to return
 	wg.Add(len(m.moduleWrappers))
 	for _, mod := range m.moduleWrappers {
-		go func(m *moduleWrapper) {
-			if !m.IsRunning() {
+		go func(modWrapper *moduleWrapper) {
```
alsamylkin: why do you need this change?
https://github.com/golang/go/wiki/CodeReviewComments#variable-names

service/module.go
```diff
@@ -119,16 +129,26 @@ func newModuleWrapper(
...
+		moduleOptions.ModuleName = moduleProvider.DefaultName()
 	}
-	scopedHost, err := newScopedHost(
-		host,
-		moduleOptions.name,
-		moduleOptions.roles...,
-	)
-	if err != nil {
-		return nil, err
+	if moduleOptions.ServiceName != "" {
```
anuptalwalkar: nit: In similar pattern with moduleOptions.ModuleName, you can do - 
```
if moduleOptions. ServiceName == "" {
  moduleOptions.ServiceName = host.Name()
}

if len(moduleOptions.Roles) == 0 {
  moduleOptions.Roles = host.Roles()
}

scopedHost := &scopedHost{
  Host:  host,
  serviceName:  moduleOptions.ServiceName,
  moduleName:  moduleOptions.ModuleName,
  roles:  moduleOptions.Roles,
}
```

service/module.go
```diff
@@ -27,8 +27,7 @@ import (
 
 // ModuleProvider provides Modules.
 type ModuleProvider interface {
-	// DefaultName is the name that will be used for a new Module
-	// if no name is given as a ModuleOption.
+	// DefaultName returns the module name
```
alsamylkin: Either comment is incorrect, or method name should change.
DefaultName that returns module name sounds weird..

service/module.go
```diff
@@ -48,57 +47,68 @@ type Module interface {
...
+
 // NewScopedHost returns a new Host scoped to a module. This should generally be used for testing.
-func NewScopedHost(host Host, name string, roles ...string) (Host, error) {
-	return newScopedHost(host, name, roles...)
+func NewScopedHost(host Host, moduleName, serviceName string, roles ...string) (Host, error) {
+	return newScopedHost(host, moduleName, serviceName, roles...), nil
 }
 
-type moduleProviderFunc struct {
+type moduleProvider struct {
```
alsamylkin: Thank you!

service/module.go
```diff
@@ -48,57 +47,68 @@ type Module interface {
...
-
-type moduleOptions struct {
-	name  string
-	roles []string
-}
+func (m *moduleProvider) DefaultName() string              { return m.name }
+func (m *moduleProvider) Create(host Host) (Module, error) { return m.createFunc(host) }
 
 type moduleWrapper struct {
+	name       string
```
alsamylkin: shouldn't it be serviceName?

service/stub_module.go
```diff
@@ -29,17 +29,17 @@ type StubModuleProvider struct {
...
 
 // NewStubModuleProvider returns a new StubModuleProvider with the given name
 // and create function from NewStubModuleCreateFunc.
 func NewStubModuleProvider(name string, stubModule *StubModule) *StubModuleProvider {
-	return &StubModuleProvider{name, NewStubModuleCreateFunc(stubModule)}
+	return &StubModuleProvider{NameVal: name, CreateVal: NewStubModuleCreateFunc(stubModule)}
 }
 
-// DefaultName returns the default name.
+// DefaultName returns the module name
```
alsamylkin: Again DefaultName() that returns module name is very confusing.

modules/yarpc/yarpc.go
```diff
@@ -331,9 +336,9 @@ func (c *dispatcherController) addDefaultMiddleware(host service.Host, statsClie
 	c.addConfig(cfg)
 }
 
-// Merge all the YARPC configs in the collection: transports and middleware are going to be shared.
+// Create the YARPC config : transports and middleware are going to be shared.
 // The name comes from the first config in the collection and is the same among all configs.
```
alsamylkin: I think this comment. line can be deleted

service/module.go
```diff
@@ -48,57 +47,68 @@ type Module interface {
 
 // ModuleProviderFromFunc creates a new ModuleProvider from a name and create function.
 func ModuleProviderFromFunc(name string, createFunc func(Host) (Module, error)) ModuleProvider {
-	return &moduleProviderFunc{name, createFunc}
+	return &moduleProvider{name: name, createFunc: createFunc}
```
peter-edge: nit: i didn't put keys on purpose, we should talk about this

service/module.go
```diff
@@ -48,57 +47,68 @@ type Module interface {
...
 }
 
 // ModuleOption is a function that configures module creation.
 type ModuleOption func(*moduleOptions) error
 
-// WithModuleName will override the name given by the ModuleProvider.
-func WithModuleName(name string) ModuleOption {
+// WithName will override the root service name specified in config.
+func WithName(name string) ModuleOption {
```
anuptalwalkar: names are confusing since we have WithModuleName. Can we make this "WithServiceName"?

service/stub_module.go
glibsm: The comment is incorrect, i've explained the function naming already. It returns the default, fallback, whatever you want to call it, if a name option is not explicitly provided.

I can't stress it enough how much i'm tired of talking about this, but can we please stop renaming things backwards and forwards and just move on? If proven confusing, easy to fix in the future. Lets stop guessing

service/manager.go
madhuravi: manager itself is called m, I was hoping to avoid confusion here.

service/module.go
madhuravi: this is moduleName, I did not explicitly call it so because the struct itself says moduleWrapper.

service/service.go
```diff
@@ -59,6 +59,8 @@ type Host interface {
 	Observer() Observer
 	Config() config.Provider
 	Tracer() opentracing.Tracer
+	// TODO: Will be removed once log/metrics scoping is moved to the module provider
```
glibsm: yasssss

service/module.go
madhuravi: service.WithServiceName stutters. service.WithName should imply service name.

service/module.go
madhuravi: It's not the same though. There is an else here too to set to host's values. Optionally, I can move this to the function implementation in ScopedHost but couldn't decide which was better and stuck with this :)

PR #386, opened 2017-03-09
Remove /dev/null from glide version output

glibsm:
@breerly

### DIFF
--- a/.build/deps.mk
+++ b/.build/deps.mk
@@ -1,7 +1,7 @@
 .PHONY: libdeps
 libdeps:
 	@$(call label,Installing Glide and locked dependencies...)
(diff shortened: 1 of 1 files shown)


.build/deps.mk
```diff
@@ -1,7 +1,7 @@
 .PHONY: libdeps
 libdeps:
 	@$(call label,Installing Glide and locked dependencies...)
-	$(ECHO_V)glide --version 2>/dev/null ||	go get -v github.com/Masterminds/glide
+	$(ECHO_V)glide --version ||	go get -v github.com/Masterminds/glide
```
sectioneight: you added some weird spaces after the or-or

glibsm:
This is not terribly important to know, but might come in handy should we be debugging more dep issues

PR #387, opened 2017-03-09
Use docker on travis

peter-edge:
This makes it so we no longer are using travis' env and now completely control the environment that fx is tested on. This also means that if it passes locally in docker, it passes remotely.


.travis.yml
```diff
@@ -1,28 +1,14 @@
...
-go_import_path: go.uber.org/fx
-
+sudo: required
+services:
+  - docker
 env:
-  - V=1 COVERMODE=atomic
-
+  - DOCKER_GO_VERSION=1.7 V=1 COVERMODE=atomic
+  - DOCKER_GO_VERSION=1.8 V=1 COVERMODE=atomic
```
anuptalwalkar: Does this mean we can't run travis against `tip`?

Makefile
```diff
@@ -14,7 +14,10 @@ all: lint test
 
 COV_REPORT := overalls.coverprofile
 
-DOCKER_IMAGE := uber/fx
+DOCKER_MAKE_CMD ?= test
```
anuptalwalkar: Should this be part of `.build/docker.mk` ?

.travis.yml
peter-edge: Correct, there's no golang tip image

Makefile
peter-edge: I actually dislike having separate build files like this, it makes things less explicit, I think it's fine here, not much logic and it directly links to logic in the Makefile

.travis.yml
peter-edge: Personally this not only doesn't bother me, but I rather not be needing to test against tip - it's not like the one time every 6 months that golang is released we can't re-run `make dockerrun` :)

Makefile
glibsm: `make ci` here, right?

peter-edge: Really I would not want any default tbh...actually I'm going to push something

Makefile
```diff
@@ -128,15 +131,11 @@ benchreset:
 
 .PHONY: dockerbuild
 dockerbuild:
-	docker build -t $(DOCKER_IMAGE) .
+	docker build -t $(DOCKER_IMAGE) -f $(DOCKERFILE) .
```
breerly: Why not build *inside* of the Docker container?

Makefile
```diff
@@ -14,7 +14,10 @@ all: lint test
 
 COV_REPORT := overalls.coverprofile
 
-DOCKER_IMAGE := uber/fx
+DOCKER_MAKE_CMD ?= test
+DOCKER_GO_VERSION ?= 1.8
+DOCKER_IMAGE := uber/fx-$(DOCKER_GO_VERSION)
```
breerly: Why not name the image the same thing each build?

breerly:
Totally in favor of this - reproducible builds is something we should lean on Mr. Docker for.

Makefile
peter-edge: Cuz once I get to it, I will add caching http://rundef.com/fast-travis-ci-docker-build

Makefile
peter-edge: Wait what docker container haha.

peter-edge:
@breerly I can set this up for zap, sally, etc too...this is what I generally like to do sometimes anyways, especially for non-trivial stuff https://github.com/peter-edge/protoeasy-go/blob/master/circle.yml

glibsm:
Im ok to give up tip and see if that significantly improves our build times. Always easy to put it back.

peter-edge:
This PR is going to slow them down temporarily. Setting up caching is not fun. But it still is an improvement IMO.

peter-edge:
Slow down = from 2 mins to 3 mins

PR #388, opened 2017-03-09
Add configurable host and service names to changelog

madhuravi:


CHANGELOG.md
```diff
@@ -19,6 +19,8 @@
 * [Breaking] Rename `yarpc.CreateThriftServiceFunc` to `yarpc.ServiceCreateFunc`
   as it is not thrift-specific.
 * Report version metrics for company-wide version usage information.
+* [Breaking] Use default config name as service name and allow both service name and module name
```
glibsm: this doesn't belong in the changelog, and is also not breaking - it restores previous functionality

madhuravi: Sure, it's breaking from previous change but not from beta1, I can remove. Aiden has been getting me in the habit of doing this. Why does it not belong in the changelog?
There is difference in functionality, users can specify custom module and service name in options.

glibsm: If you want to highlight that there are some new options for module configuration, sure. Lets focus on that.

The changelog here is beta1 -> beta2, so there is definitely nothing breaking.

CHANGELOG.md
```diff
@@ -19,6 +19,9 @@
 * [Breaking] Rename `yarpc.CreateThriftServiceFunc` to `yarpc.ServiceCreateFunc`
   as it is not thrift-specific.
 * Report version metrics for company-wide version usage information.
+* Allow configurable service name and module name via service options.
+* [Breaking] Use "modules.uhttp" config block rather than "modules.http" to match
```
glibsm: Default configuration key for `modules.http` was renamed to `modules.uhttp`

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 ## v1.0.0-beta2 (unreleased)
 
 * [Breaking] Remove `ulog.Logger` interface and expose `*zap.Logger` directly.
+* [Breaking] Config for `modules/rpc` was renamed to `modules/yarpc`
+* [Breaking] Config for `modules.http` was renamed to `modules.uhttp`
```
glibsm: lets use either `/` or `.` in both

PR #390, opened 2017-03-10
Remove code duplication in scalar function

alsamylkin:
Removed lines are already covered by [convertValueFromStruct](https://github.com/uber-go/fx/blob/master/config/decoder.go#L61).

### DIFF
--- a/config/decoder.go
+++ b/config/decoder.go
@@ -136,24 +136,7 @@ func (d *decoder) scalar(childKey string, value reflect.Value, def string) error
 					value.Set(reflect.New(valueType.Elem()))
 				}
 
-				// We cannot assign reflect.ValueOf(val) to value as is, when field is a user defined type
-				// We need to find the Kind of the custom type and set the value to the specific type
-				// that user defined type is defined with.
-				kind := valueType.Elem().Kind()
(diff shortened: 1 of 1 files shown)


config/decoder.go
```diff
@@ -136,24 +136,7 @@ func (d *decoder) scalar(childKey string, value reflect.Value, def string) error
...
-				case reflect.Float32, reflect.Float64:
-					value.Elem().SetFloat(val.(float64))
-				case reflect.Bool:
-					value.Elem().SetBool(val.(bool))
-				case reflect.String:
-					value.Elem().SetString(val.(string))
-				default:
-					value.Elem().Set(reflect.ValueOf(val))
-				}
+				return d.scalar(childKey, value.Elem(), def)
```
madhuravi: This is a recursive call but the removed code doesn't seem to be doing the same thing?

alsamylkin: Well, it does more :) What we are doing is checking if a value is a pointer and then repeating steps from convertValueFromStruct, but there are actually more cases to cover than just numbers(convertValue). What we need to do is to allocate memory and do all necessary conversions with scalar values in one place, instead of two.

PR #391, opened 2017-03-10
Fix travis

peter-edge:
Travis was broken with how it is set up with Docker. This fixes some environment variables, and gets coveralls working again, along with simplifying the Makefile so that `make dockerci` now does almost the same thing locally as it does remotely.


glibsm:
`make ci` and `make travis` seem like the same thing now

glibsm:
seems like bash syntax is still borked

Makefile
```diff
@@ -137,6 +137,14 @@ dockerbuild:
 dockerrun: dockerbuild
 	docker run -e V -e COVERMODE $(DOCKER_IMAGE) make $(DOCKER_MAKE_CMD)
 
+.PHONY: travis
+travis: dockerbuild
+	if [ "$(DOCKER_GO_VERSION)" == "1.8" ]; then \
+		docker run -e V -e COVERMODE $(DOCKER_IMAGE) make coveralls; \
+	else; \
```
glibsm: no need for this semi

peter-edge:
Trying caching for now, it may slow it down, if so I'll remove it.

alsamylkin:
Coveralls is working!:)

.dockerignore
```diff
@@ -1 +1,5 @@
+.cache
+examples/dig/dig
```
madhuravi: Why is this necessary? We want to make sure that examples still compile.

peter-edge: This ignores them when building the docker container, ie it doesn't pass (usually darwin) binaries to the (linux) container. They are still built by `make ci` within the docker container.

Short: this has no effect on the build other than making the tarball sent to the docker daemon smaller. https://docs.docker.com/engine/reference/builder/#dockerignore-file

PR #392, opened 2017-03-10
Fix race condition that caused flaky task test failure

madhuravi:
TestConsumeWithoutRegister was failing intermittently depending on whether the enqueued message was consumed before or after clearing the register cache. Fixed this by starting consume only after the register cache is cleared.

### DIFF
--- a/modules/task/backend_test.go
+++ b/modules/task/backend_test.go
@@ -31,6 +31,15 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func withBackendSetup(t *testing.T, backend Backend) func() {
+	_globalBackend = backend
+	_ = _globalBackend.Start()
+	return func() {
+		assert.NoError(t, _globalBackend.Stop())
+		_globalBackend = NopBackend{}
+	}
+}
+
 func TestNopBackend(t *testing.T) {
 	b := &NopBackend{}
 	assert.NotNil(t, b.Encoder())
@@ -41,11 +50,11 @@ func TestNopBackend(t *testing.T) {
 
 func TestInMemBackend(t *testing.T) {
 	b := NewInMemBackend(newTestHost(t)).(*inMemBackend)
+	defer withBackendSetup(t, b)()
 	assert.NotNil(t, b.Encoder())
 	assert.NoError(t, b.Start())
 	publishEncodedVal(t, b)
 	publishEncodedVal(t, b)
-	assert.NoError(t, b.Stop())
 }
 
 func TestInMemBackendStartTimeout(t *testing.T) {
--- a/modules/task/backend_test.go
+++ b/modules/task/backend_test.go
@@ -33,7 +33,7 @@ import (
 
 func withBackendSetup(t *testing.T, backend Backend) func() {
 	_globalBackend = backend
-	_ = _globalBackend.Start()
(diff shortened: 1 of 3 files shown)


madhuravi:
@alsamylkin

modules/task/execution_test.go
```diff
@@ -114,21 +117,33 @@ func TestEnqueueWithoutRegister(t *testing.T) {
 }
 
 func TestConsumeWithoutRegister(t *testing.T) {
+	host, _ := service.NewScopedHost(service.NopHost(), "task", "hello")
+	_testScope = host.Metrics()
+	_globalBackend = NewInMemBackend(host)
+	_errorCh = _globalBackend.(*inMemBackend).ErrorCh()
```
alsamylkin: I thought tasks are not doing error channels anymore and track function executions through logging?

madhuravi: the way Peter changed it is he kept channels around for in memory backend and added a method specifically for this implementation to expose it. Don't see a reason to change it.

alsamylkin: The reason would be to simplify testing for not in-memory backend, boilerplate that other backends could follow.

madhuravi: hmm? reason for what?

alsamylkin: To test logged messages?

madhuravi: not sure what you're talking about at all. explain in atleast a few sentences :) This is testing execution, not the cherami backend.

alsamylkin: Cherami backend is doing testing through log messages and in-memory backend is doing testing through error channels. Should we unify it? In the end all you want is to know if a function was executed and what was the result, but methods are very different and we have an extra functionality(the error channels) just for testing.

madhuravi: It does not really matter how different backends implement propagating the error. In both cases we are checking the error manifests as expected irrespective of whether is comes through a channel or a log message. In memory backend is primarily used for testing and channels make it easy to wait on and test so I prefer keeping this around.
For cherami backend itself, I want to iterate on logging vs error channels based on user feedback.

modules/task/backend_test.go
```diff
@@ -31,6 +31,15 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func withBackendSetup(t *testing.T, backend Backend) func() {
+	_globalBackend = backend
+	_ = _globalBackend.Start()
```
alsamylkin: should we have `require.NoError(t, _globalBackend.Start())`?

modules/task/execution_test.go
```diff
@@ -114,21 +117,33 @@ func TestEnqueueWithoutRegister(t *testing.T) {
...
+	}()
 	fn := func(ctx context.Context, num float64) error { return nil }
 	err := Register(fn)
 	require.NoError(t, err)
 	err = Enqueue(fn, _ctx, float64(1.0))
 	require.NoError(t, err)
+	// This clears the previous register
 	fnLookup.setFnNameMap(make(map[string]interface{}))
+	// Start after clearing the register to avoid race condition where consume happens before clearing
+	_ = _globalBackend.Start()
```
alsamylkin: Again, require.NoError?

PR #394, opened 2017-03-10
Remove getBucket function to clean up unmarshal method

alsamylkin:
We have a custom bucketing function, that is not adding any value, but introduces an extra layer of logic along with extra constants. We use standard reflect.'Kind' everywhere now and keep dispatch logic clean for all possible value types.

To keep unmarshal clean, we introduce a new method to handle reflect.Ptr recursively: if there is a value in config, allocate memory for the value and use it, so all other types don't have to worry about it.

### DIFF
--- a/config/decoder.go
+++ b/config/decoder.go
@@ -29,14 +29,6 @@ import (
 	"github.com/go-validator/validator"
 )
 
-const (
-	bucketPrimitive = iota
-	bucketArray
-	bucketObject
-	bucketMap
-	bucketSlice
-)
-
 type fieldInfo struct {
 	FieldName    string
 	DefaultValue string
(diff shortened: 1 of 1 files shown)


config/decoder.go
```diff
@@ -369,18 +315,35 @@ func (d *decoder) unmarshal(name string, value reflect.Value, def string) error
...
-	case bucketSlice:
+	case reflect.Slice:
 		return d.sequence(name, value)
-	case bucketMap:
+	case reflect.Map:
+		return d.mapping(name, value, def)
+	case reflect.Interface:
 		return d.mapping(name, value, def)
+	case reflect.Chan:
+		return nil
```
glibsm: return error, or am i missing something.

config/decoder.go
```diff
@@ -284,23 +232,7 @@ func (d *decoder) mapping(childKey string, value reflect.Value, def string) erro
...
-
-	v2 := global.Get(childKey)
-
-	if !v2.HasValue() && valueType.Kind() == reflect.Ptr {
-		// in this case we will keep the pointer value as not defined.
-		return nil
-	}
-
-	if !value.CanSet() {
-		return nil
```
glibsm: i think `CanSet` should stay and return error on false

glibsm:
overall looks like a very welcome change, especially since there is no impact on the tests

config/decoder.go
alsamylkin: We iterate over fields, so if a struct has a channel, we want to simply skip it.

Actually we should check that this channel also implements TextUnmarshaller to possibly construct it, but this is a subject of other change. Example is zap.Config, some of it's fields are actually functions that implement unmarshaller to read configs.

config/decoder.go
alsamylkin: We always can set value now, if it is not the case the unmarshall method did something really wrong.

PR #395, opened 2017-03-10
Do not run make lint on pushes with overcommit

peter-edge:
While I am working on speeding up the dev process here, I want to remove `make lint` from being run on pushes. @glibsm has argued that it prevents bad builds from going into CI, but I argue it just adds 40 seconds to the push process - this is what CI is for. If you really want to do this and make sure there are not bad builds, you really should be running `make dockerci` - `make lint` only catches some stuff. So right now, the push hook checks some stuff but not all, so there still can be bad builds, and if you want the push hook to check for bad builds, it should just run `make dockerci`, but this is exactly what travis does anyways, so why make the time to check a push twice as long? Just let travis do what it does.

### DIFF
--- a/.build/check_lint.sh
+++ b/.build/check_lint.sh
@@ -1,4 +1,0 @@
-#!/bin/bash
-
-# Overcommit requires custom commands to be checked in
-make lint
--- a/.overcommit.yml
+++ b/.overcommit.yml
@@ -48,11 +48,6 @@ PreCommit:
     exclude:
       - '.build/*.sh'
 
-PrePush:
-  UberfxGoLint:
-    enabled: true
-    required_executable: './.build/check_lint.sh'
-
 #PostCheckout:
 #  ALL: # Special hook name that customizes all hooks of this type
 #    quiet: true # Change all post-checkout hooks to only display output on failure


alsamylkin:
I would side with @glibsm because I don't want to see my build failed later because of some formatting issue, so I need to run an extra command to check that.
I think we should make linting faster instead.

peter-edge:
Note that lint just does part of it, if you argue for lint, you should argue for dockerci completely. That's kind of the thing. For me, CI is the public verification that you can merge. When I do things locally, I want control over it. Generally, I run `make ci` before pushing anyways, and if you want to know beforehand, just get into that habit. Don't enforce it on pushes.

madhuravi:
We might miss running it locally Peter. Especially with additive changes. 

>> "it should just run make dockerci, but this is exactly what travis does anyways, so why make the time to check a push twice as long? Just let travis do what it does."

are you saying travis already runs dockerci and lint as part of it?

peter-edge:
`dockerci` runs `lint examples test`, so it's already covered. I'm saying remove this being automatic on pushes, it makes no difference in terms of what ends up getting merged in the end. I have `alias gpfx="git push --no-verify"` already

glibsm:
This mimics the behavior of `arc diff` - it runs lint before you open a diff.

It looks like the slow part is running the errcheck.

With: `make lint  50.61s user 30.90s system 391% cpu 20.815 total`
Without: `make lint  9.67s user 9.18s system 225% cpu 8.365 total`

I would be ok to defer running errcheck to CI, but I want the other lint to run before PRs. I dont' actually think it slows people down, quite the opposite. You don't need to wait for 5m for CI to tell you a function is missing a public comment.

peter-edge:
You're right, you don't but a lot of PRs I do I know that lint doesn't do anything and/or lint won't fail. If I think lint will fail, I run `make lint`, lol. I'm saying I want opt-in instead of opt-out - there's nothing preventing you from saving yourself the time of a failed build by running `make lint` if you are concerned.

madhuravi:
Got it, lint still runs as part of the build but just not when you push a commit to the branch to be reflected here. That sounds fine to me. 

I've also never had git push actually detect any of these lint failures. I always end up seeing it in travis so this is probably not that useful anyway.

glibsm:
I don't see what the issue is if we speed up lint and punt error checking to CI. You already have the alias, everyone is happy. We don't necessarily need to push this philosophy onto the rest of the team, or even external contributors rigth?

peter-edge:
Ya it's up to you really, I would argue you'll end up liking it even though you think you don't now :) ie I would ask you to try it for two days and see if you end up liking it more or hating it more. BUT, I can just close this too, its fine.

On the linter, I really need to write a version of https://github.com/alecthomas/gometalinter for uber-go. Basically have consistent linters across all uber-go repositories, and do things as parallel as possible (not just parallel by binary, but by package).

glibsm:
lets give it a shot, see how it feels to be in the free world

sectioneight:
For the record I fully support this. I've been running `git push origin HEAD --no-verify` because of the extra time -- I'd rather let CI do its job or run `make lint` explicitly before I push a branch.

PR #398, opened 2017-03-10
Speed up make lint by about 2x

peter-edge:
On my local machine, this reduces the time for make lint from about 19.7 seconds to about 9.9 seconds.

### DIFF
--- a/.build/flags.mk
+++ b/.build/flags.mk
@@ -3,7 +3,9 @@ PKGS ?= $(shell glide novendor)
 LIST_PKGS ?= $(shell go list ./... | grep -v /vendor/)
 
 # Many Go tools take file globs or directories as arguments instead of packages.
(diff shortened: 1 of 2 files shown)


.build/flags.mk
```diff
@@ -3,7 +3,9 @@ PKGS ?= $(shell glide novendor)
 LIST_PKGS ?= $(shell go list ./... | grep -v /vendor/)
 
 # Many Go tools take file globs or directories as arguments instead of packages.
-PKG_FILES ?= *.go auth config dig internal metrics modules service testutils tracing ulog
+ROOT_PKG_FILES := $(wildcard *.go)
+PKG_NAMES := auth config dig internal metrics modules service testutils tracing ulog
```
glibsm: this should be something like `go list | grep -v vendor`, rather than enumerating them all by hand

peter-edge: This is already in the existing code, this diff is not trying to fix all the Makefiles

PR #399, opened 2017-03-10
Dont enumerate packages in Makefile, get list from disk

glibsm:
Follow up to my comment from #398

### DIFF
--- a/.build/flags.mk
+++ b/.build/flags.mk
@@ -4,7 +4,7 @@ LIST_PKGS ?= $(shell go list ./... | grep -v /vendor/)
 
 # Many Go tools take file globs or directories as arguments instead of packages.
 ROOT_PKG_FILES := $(wildcard *.go)
-PKG_NAMES := auth config dig internal metrics modules service testutils tracing ulog
+PKG_NAMES := $(shell find . -d 1 -type d | cut -f 2 -d"/" | grep -ve "vendor\|\.\|examples" | paste -sd " " -)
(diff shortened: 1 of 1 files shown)


glibsm:
Ugh looks like docker has a bunch of other crap laying around, let me rewrite to use `go list` rather than `find`

.build/flags.mk
```diff
@@ -4,7 +4,8 @@ LIST_PKGS ?= $(shell go list ./... | grep -v /vendor/)
 
 # Many Go tools take file globs or directories as arguments instead of packages.
 ROOT_PKG_FILES := $(wildcard *.go)
-PKG_NAMES := auth config dig internal metrics modules service testutils tracing ulog
+# Get all the go packages, ignore vendor and take fx/* directories
+PKG_NAMES := $(shell go list ./... | grep -v "vendor\|examples" |  cut -d"/" -f 3 | uniq | paste -sd " " -)
```
sectioneight: I guess we're well past any of this working on Windows, but I wonder if there's a more natural Makefile syntax to accomplish this. I assume this is at least portable across GNU/Darwin, BSD/Darwin, and GNU/Linux?

sectioneight: The grep -v seems a bit fragile. Can we do something like `$(shell go list $(glide nv) | grep -v /examples` in case we have some other examples package that we do want to test?

glibsm: hmmmm

PR #400, opened 2017-03-11
Add ability to use unmarshaller on a channel or a function

alsamylkin:
Struct fields can be functions or channels and we skip them right now, but some libraries e.g. Zap uses fields with these types and implement TextUnmarshaller to read configs: [LevelEncoder](https://github.com/uber-go/zap/blob/14d784558fd634fc33059b52572035ed2a1a99f3/zapcore/encoder.go#L68)

With this PR we can read these types.

### DIFF
--- a/config/decoder.go
+++ b/config/decoder.go
@@ -318,6 +318,10 @@ func (d *decoder) textUnmarshaller(key string, value reflect.Value, str string) 
 	}
 
 	// Value has to have a pointer receiver to be able to modify itself with TextUnmarshaller
+	if !value.CanAddr() {
+		return fmt.Errorf("can't use TextUnmarshaller because %s is not addressable", key)
+	}
+
 	switch t := value.Addr().Interface().(type) {
 	case encoding.TextUnmarshaler:
 		return t.UnmarshalText([]byte(str))
--- a/config/decoder.go
+++ b/config/decoder.go
@@ -22,8 +22,8 @@ package config
 
 import (
 	"bytes"
-	"fmt"
 	"encoding"
+	"fmt"
 	"reflect"
 	"strconv"
 
@@ -309,7 +309,7 @@ func (d *decoder) textUnmarshaller(key string, value reflect.Value, str string) 
 	v := d.getGlobalProvider().Get(key)
 	if v.HasValue() {
 		str = v.String()
-	} else  if str == "" {
+	} else if str == "" {
 		return nil
 	}
 
@@ -318,7 +318,7 @@ func (d *decoder) textUnmarshaller(key string, value reflect.Value, str string) 
 	}
 
 	// Value has to have a pointer receiver to be able to modify itself with TextUnmarshaller
-	switch t := value.Addr().Interface().(type){
+	switch t := value.Addr().Interface().(type) {
 	case encoding.TextUnmarshaler:
 		return t.UnmarshalText([]byte(str))
 	}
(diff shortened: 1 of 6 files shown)


config/decoder.go
```diff
@@ -294,6 +295,37 @@ func (d *decoder) pointer(name string, value reflect.Value, def string) error {
...
+func (d *decoder) textUnmarshaller(key string, value reflect.Value, str string) error {
+	v := d.getGlobalProvider().Get(key)
+	if v.HasValue() {
+		str = v.String()
+	} else if str == "" {
+		return nil
+	}
+
+	if value.IsNil() {
+		value.Set(reflect.New(value.Type()).Elem())
```
peter-edge: you might want to do a bit more checking here, reflect can panic like crazy

config/decoder.go
```diff
@@ -294,6 +295,37 @@ func (d *decoder) pointer(name string, value reflect.Value, def string) error {
...
+	}
+
+	// Value has to have a pointer receiver to be able to modify itself with TextUnmarshaller
+	switch t := value.Addr().Interface().(type) {
```
peter-edge: same here

config/decoder.go
alsamylkin: We simply allocate memory for value with its own type :)

config/decoder.go
alsamylkin: Make sense.

config/decoder.go
```diff
@@ -294,6 +295,41 @@ func (d *decoder) pointer(name string, value reflect.Value, def string) error {
...
+		return nil
+	}
+
+	if value.IsNil() {
+		value.Set(reflect.New(value.Type()).Elem())
+	}
+
+	// Value has to have a pointer receiver to be able to modify itself with TextUnmarshaller
+	if !value.CanAddr() {
+		return fmt.Errorf("can't use TextUnmarshaller because %s is not addressable", key)
```
sectioneight: Prefer %q to %s to make it obvious which part is interpolated (it adds double quotes around the value)

config/decoder.go
```diff
@@ -294,6 +295,41 @@ func (d *decoder) pointer(name string, value reflect.Value, def string) error {
...
+	}
+
+	switch t := value.Addr().Interface().(type) {
+	case encoding.TextUnmarshaler:
+		return t.UnmarshalText([]byte(str))
```
sectioneight: Naive question: is any of this code going to be used in the "hot path"? Allocating a new byte slice may be small but non-negligable, same question about all the reflect stuff. I'm far enough away from the code right now that it's unclear to me how deep we're digging here.

IOW, do we need to add this to our benchmarks?

alsamylkin: Our config is being loaded with a service startup, so there is no hot path(at least for static configs). 
There is some room for improvement, e.g use a pointer receiver for a value, maybe cache types we populate with configs(we already try to cache nodes in config providers) Unfortunately, we use reflection a lot and it is always done on heap: https://golang.org/src/reflect/value.go?s=62046:62079#L2085, so performance will be always slow..

alsamylkin: I'll add benchmarks though.

sectioneight: Doesn't sound necessary :)

sectioneight:
I'd still like to see the key names quotes in the error messages but I can make a follow up PR if you don't want to

PR #401, opened 2017-03-11
Move interface parsing from map to its own method

alsamylkin:
Parsing of interface fields is done through fields right now. This PR moves it to its own method and add a check, that a value from config provider actually implements the interface.

### DIFF
--- a/config/decoder.go
+++ b/config/decoder.go
@@ -22,9 +22,13 @@ package config
 
 import (
 	"bytes"
+<<<<<<< HEAD
 	"encoding"
 	"fmt"
+=======
+>>>>>>> Fix fmt
 	"encoding"
+	"fmt"
 	"reflect"
 	"strconv"
 
@@ -311,10 +315,14 @@ func (d *decoder) textUnmarshaller(key string, value reflect.Value, str string) 
 	if v.HasValue() {
 		str = v.String()
 <<<<<<< HEAD
+<<<<<<< HEAD
 	} else if str == "" {
 =======
 	} else  if str == "" {
 >>>>>>> Add ability to unmarshal channel and functional fields
+=======
+	} else if str == "" {
+>>>>>>> Fix fmt
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


config/decoder.go
```diff
@@ -231,6 +224,23 @@ func (d *decoder) mapping(childKey string, value reflect.Value, def string) erro
...
+	if !v.HasValue() || v.Value() == nil {
+		return nil
+	}
+
+	src := reflect.ValueOf(v.Value())
+	if src.Type().Implements(value.Type()) {
+		value.Set(src)
+		return nil
+	}
+
```
peter-edge: nit: i don't like having so many newlines haha

config/decoder.go
```diff
@@ -231,6 +224,23 @@ func (d *decoder) mapping(childKey string, value reflect.Value, def string) erro
...
+	return fmt.Errorf("%v doesn't implement to %v", src.Type(), value.Type())
```
peter-edge: nit: remove "to "

config/decoder.go
alsamylkin: It is the msft style and I think google java style: line breaks after closing non terminal brackets :)

sectioneight: @peter-edge: what's the argument against the newlines? most of the idiomatic code I've seen here, in other uber teams, and in the stdlib uses them to make the code more parsable at a glance. is there really a downside to them? just that the files get long?

config/decoder.go
```diff
@@ -231,6 +224,22 @@ func (d *decoder) mapping(childKey string, value reflect.Value, def string) erro
...
+		return nil
+	}
+
+	src := reflect.ValueOf(v.Value())
+	if src.Type().Implements(value.Type()) {
+		value.Set(src)
+		return nil
+	}
+
+	return fmt.Errorf("%v doesn't implement %v", src.Type(), value.Type())
```
sectioneight: consider adding quotes around the %vs, either with %q or doing:

```go
return fmt.Errorf(`"%v" doesn't implement "%v"`, ...)
```

to make it clear which parts are constant and which parts are variant on the problematic type

alsamylkin:
No changes for tests is needed, because interface was resolved with maps before, all I did here is extracted interface part to a specialized method.

Well, I've also added error check for interface assignment, we panic in these scenarios now ;)

PR #402, opened 2017-03-11
Remove scope method from the provider interface

alsamylkin:
@peter-edge mentioned, that using Scope in Provider is very clumsy and we actually don't really need every Provider to implement Scope method, Get should be enough. Users can either use ScopedProvider to create a new scope for a provider or we can add this functionality to the Value. Users can simply call: 
`config.Get("yarpc").Get("http").Get("port")`
instead of 
`config.Scope("yarpc").Scope("http").Get("port")`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## v1.0.0-beta3 (unreleased)
+* Remove `Scope` method from the `config.Provider` interface.
 
 ## v1.0.0-beta2 (09 Mar 2017)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 # Changelog
 
 ## v1.0.0-beta3 (unreleased)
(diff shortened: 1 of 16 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## v1.0.0-beta3 (unreleased)
+* Remove `Scope` method from the `config.Provider` interface.
```
sectioneight: Maybe mention why briefly? "because it was unused" or whatever

sectioneight: Also this is potentially `[BREAKING]` right, since people could be calling Scope() in their code?

alsamylkin:
Scope supposed to return a provider that appends a prefix to the current provider. We implemented it for every provider differently, but I've changed to use the scoped provider underneath everywhere a couple month ago, and with addition of Get method to Value, we can simply remove the repeated code and method from the interface.

PR #403, opened 2017-03-11
Add task.MustRegister which panics on error

sectioneight:
Although panics are generally frowned upon, there are certain cases
where they can make developers' lives easier. Particularly for the task
module, the general use case is that in the `main` func (or some other
func called by `main` during initialization), there is no way to recover
from the error and we want to crash the process.

In abscence of MustRegister, a developer who wishes to register many
tasks must check the error of each, and upon encountering an error will
almost certainly want to exit the program with a non-zero exit code
since the service should not start if the task functions are not
properly defined.

In other words, it turns this:

```go
  if err := task.Register(FirstTaskFunc); err != nil {
    log.Fatal("Could not register task", err.Error())
  }
  if err := task.Register(SecondTaskFunc); err != nil {
    log.Fatal("Could not register task", err.Error())
  }
```

Into this:

```go
  task.MustRegister(FirstTaskFunc)
  task.MustRegister(SecondTaskFunc)
```

Both of which behave semantically similar (although the output may be
slightly different).

This gives developers the flexibility to choose which behavior they
want, similar to `regexp.MustCompile` from the standard library, where a
program should not start if the developer incorrectly constructed a
regular expression.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## v1.0.0-beta3 (unreleased)
 
-* Add `task.MustRegister` convenience function which fails fast by panic-ing
+* Add `task.MustRegister` convenience function which fails fast by panicking
   Note that this should only be used during app initialization, and is provided
   to avoid repetetive error checking for services which register many tasks.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,16 @@
 
 ## v1.0.0-beta3 (unreleased)
 
+* Add `task.MustRegister` convenience function which fails fast by panic-ing
+  Note that this should only be used during app initialization, and is provided
+  to avoid repetetive error checking for services which register many tasks.
+
 ## v1.0.0-beta2 (09 Mar 2017)
 
 * [Breaking] Remove `ulog.Logger` interface and expose `*zap.Logger` directly.
 * [Breaking] Rename config and module from `modules.rpc` to `modules.yarpc`
-* [Breaking] Rename config key from `modules.http` to `modules.uhttp` to match the module name
+* [Breaking] Rename config key from `modules.http` to `modules.uhttp` to match
+  the module name
 * [Breaking] Upgrade `zap` to `v1.0.0-rc.3` (now go.uber.org/zap, was
     github.com/uber-go/zap)
 * Remove now-unused `config.IsDevelopmentEnv()` helper to encourage better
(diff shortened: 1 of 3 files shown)


sectioneight:
Note: I'm open to calling `log.Fatal` instead, but this mimics the behavior of the stdlib Must* functions (or at least regexp.MustCompile)

CHANGELOG.md
```diff
@@ -2,11 +2,16 @@
 
 ## v1.0.0-beta3 (unreleased)
 
+* Add `task.MustRegister` convenience function which fails fast by panic-ing
```
sectioneight: not sure if "panic-ing" is right here. panicing? spell checker tells me that's wrong. panicking?

CHANGELOG.md
```diff
@@ -2,11 +2,16 @@
...
+  Note that this should only be used during app initialization, and is provided
+  to avoid repetetive error checking for services which register many tasks.
+
 ## v1.0.0-beta2 (09 Mar 2017)
 
 * [Breaking] Remove `ulog.Logger` interface and expose `*zap.Logger` directly.
 * [Breaking] Rename config and module from `modules.rpc` to `modules.yarpc`
-* [Breaking] Rename config key from `modules.http` to `modules.uhttp` to match the module name
+* [Breaking] Rename config key from `modules.http` to `modules.uhttp` to match
```
sectioneight: Somebody isn't running overcommit, causing commit failures on the changelog (max line length). I would like to move this to a separate PR, but seems overkill in this case.

modules/task/execution.go
```diff
@@ -133,6 +133,15 @@ func Register(fn interface{}) error {
 	return nil
 }
 
+// MustRegister registers a function for an async task or panics
+// Since Task registration is performed in main() we want to fail-fast and
+// reduce the amount of error-checking
+func MustRegister(fn interface{}) {
```
alsamylkin: This assumes that Register never panics, which I think it is(e.g. when GlobalBackend() is nil?)
Could you add a todo, to make Register panic free then?

CHANGELOG.md
alsamylkin: Panicking is what urban dictionary says :)

sectioneight: Ah yes, just checked stdlib and they say "panicking" as well. Updating.

modules/task/execution.go
sectioneight: Interesting, I'll add one. Yet another reason why I think this is valuable -- fail fast in main if stuff isn't set up right rather than logging and doing... what exactly?

sectioneight: Actually, just read the code. It defaults to a noop backend. If the programmer registers a nil backend somehow, that means they created a backend provider that returns a nil interface **and** a nil error, which should not be the case (and is not the case in our internal implementation). So I'm going to skip the TODO for now.

sectioneight:
Going to wait for @madhuravi to take a look since it's the weekend and this is her module, want to make sure she is OK with the functionality.

CHANGELOG.md
madhuravi: Does travis not run this and report errors? Is changelog max line length different from the rest of the code?

PR #405, opened 2017-03-13
Load additional config files for service usage

anuptalwalkar:
GFM-438

Presently, config package is not supporting all the files needed for a service. This change should accommodate configurations such as deployment name based configs,  mesos and monopod configurations.


config/config.go
```diff
@@ -39,21 +40,39 @@ const (
...
+	_deploymentPrefix = "deployment"
+	_monopod          = "monopod"
+	_secretsFile      = "secrets"
 )
 
 var (
 	_setupMux sync.Mutex
 
 	_envPrefix            = "APP"
+	_deploymentEnv        = "APP_DEPLOYMENT_NAME"
```
alsamylkin: Should it be `_envPrefix + "_DEPLOYMENT_NAME"`?

config/config.go
```diff
@@ -88,23 +107,57 @@ func ResolvePath(relative string) (string, error) {
...
 		baseFiles = append(baseFiles, fmt.Sprintf("%s-%s", env, dc))
 	}
 
-	var files []string
-	dirs := []string{".", _configRoot}
-	for _, dir := range dirs {
-		for _, baseFile := range baseFiles {
-			files = append(files, fmt.Sprintf("%s/%s.yaml", dir, baseFile))
-		}
+	if deployment() != "" {
```
alsamylkin: cache deployment() on stack? No need to read env variable every time..

config/config.go
```diff
@@ -88,23 +107,57 @@ func ResolvePath(relative string) (string, error) {
...
+	if strings.Contains(deployment(), _monopod) {
+		baseFiles = append(
+			baseFiles, _monopod, fmt.Sprintf("%s-%s", _monopod, dc),
+		)
+	}
+
+	if os.Getenv(_mesosContainerName) != "" {
+		baseFiles = append(
+			baseFiles,
+			"mesos",
```
alsamylkin: "mesos" should be a const..

config/config.go
```diff
@@ -39,21 +40,39 @@ const (
...
+	_appRoot            = "APP_ROOT"
+	_environment        = "_ENVIRONMENT"
+	_datacenter         = "_DATACENTER"
+	_mesosContainerName = "MESOS_CONTAINER_NAME"
+	_configDir          = "_CONFIG_DIR"
+
+	_configRoot       = "./config"
+	_baseFile         = "base"
+	_deploymentPrefix = "deployment"
+	_monopod          = "monopod"
```
glibsm: left comment on the task, this should not be in the public repo

config/config.go
anuptalwalkar: Its not _env_prefix in the backend, but completely different environment variable. I'll update.

config/config.go
anuptalwalkar: Sure. Its only called on `Load()` :)

anuptalwalkar:
abandoning for greater good.

PR #406, opened 2017-03-13
Added test provider

jborlum:
Added a simple test provider that fits our needs. This is untested and should be matured.

### DIFF
--- a/config/test_provider.go
+++ b/config/test_provider.go
@@ -1,0 +1,1 @@
+package config


config/test_provider.go
```diff
@@ -0,0 +1 @@
+package config
```
anuptalwalkar: Does this mean to have source code?

btw, license is missing. You need to run `make lint`.

alsamylkin:
This is not needed anymore, because https://github.com/uber-go/fx/pull/410 is merged.

PR #407, opened 2017-03-13
Allow overriding list of config files

anuptalwalkar:
* removed dc specific details from base config files so that fx users don't have to follow env-dc type config naming. 
* `SetConfigFiles` introduced to allow downstream dependencies to inject set of config files for use

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -41,7 +41,6 @@ const (
 const (
 	_appRoot     = "APP_ROOT"
 	_environment = "_ENVIRONMENT"
-	_datacenter  = "_DATACENTER"
 	_configDir   = "_CONFIG_DIR"
 	_configRoot  = "./config"
 	_baseFile    = "base"
@@ -54,6 +53,7 @@ var (
 	_envPrefix            = "APP"
 	_staticProviderFuncs  = []ProviderFunc{YamlProvider(), EnvProvider()}
 	_dynamicProviderFuncs []DynamicProviderFunc
+	_configFiles          []string
 )
 
 var (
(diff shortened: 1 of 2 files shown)


config/config_test.go
```diff
@@ -363,15 +363,22 @@ func TestGetConfigFiles(t *testing.T) {
...
-	files := getConfigFiles()
+	files := getConfigFiles(baseFiles()...)
 	assert.Contains(t, files, "./base.yaml")
 	assert.Contains(t, files, "./development.yaml")
 	assert.Contains(t, files, "./secrets.yaml")
-	assert.Contains(t, files, "./development-dc.yaml")
 	assert.Contains(t, files, "./config/base.yaml")
 	assert.Contains(t, files, "./config/development.yaml")
 	assert.Contains(t, files, "./config/secrets.yaml")
-	assert.Contains(t, files, "./config/development-dc.yaml")
```
glibsm: i think we can ditch the "dc" files as well from default, right?

config/config.go
```diff
@@ -201,6 +204,10 @@ func UnregisterProviders() {
 // Load creates a Provider for use in a service
 func Load() Provider {
 	var static []Provider
+
+	if len(_configFiles) == 0 {
+		_configFiles = getConfigFiles(baseFiles()...)
```
alsamylkin: use SetConfigFiles instead?

config/config_test.go
anuptalwalkar: no `dc` here.

config/config.go
```diff
@@ -201,6 +204,10 @@ func UnregisterProviders() {
 // Load creates a Provider for use in a service
 func Load() Provider {
 	var static []Provider
+
+	if len(_configFiles) == 0 {
```
alsamylkin: So you can't enforce to have empty list of config files?:)

config/config.go
anuptalwalkar: oh you are right.

config/config.go
anuptalwalkar: Do we need to? You can simply not have config files in the /config dir. The default is here just to tell the yaml provider where to look for configurations.

alsamylkin: Can we set baseFiles() in init() and then skip this if all together?

anuptalwalkar: I don't like in init, but I can set it as a default values in variable and skip this step.

PR #408, opened 2017-03-13
Override provider in value returned by Get for group provider

alsamylkin:
Value returned by Get has the provider that returned it, which limit following lookups to only that provider.


alsamylkin:
It is a bug, I've missed when was removing Scope

PR #409, opened 2017-03-14
Add nop config provider

alsamylkin:
We have Nop implementations for most interfaces, but not yet for a config provider.

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -407,3 +407,16 @@ func TestEnvProviderWithEmptyPrefix(t *testing.T) {
 	scope := emptyScope.Get("key")
 	require.Equal(t, "value", scope.Get("").AsString())
 }
+
+func TestNopProvider_Get(t *testing.T) {
+	t.Parallel()
+	p := NopProvider{}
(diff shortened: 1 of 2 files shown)


alsamylkin:
If users want to test a specific functionality, e.g. RegisterCallback always returns an error, the easiest way to do that will be embedding a NopConfigProvier, and they can't do this with a staticProvider, because it is private.

Another part is type preservation: we serialize objects to yaml and loose type information, uints, int8s will be converted to ints, float32s to float64.

anuptalwalkar:
I see. You can simply define a variable to provider or provider group. something like this -

`var NopProvider = NewYAMLProviderFromBytes([]byte(""))`

then access it like -
`	NopProvider.RegisterChangeCallback("", nil)`

alsamylkin:
How would it help in the above cases? This register change callback always returns nil, types still are going to be simplified..

alsamylkin:
Yes, it is supposed to help writing tests by embedding in structs, and e.g. I am going to use this provider to test numerical conversions.

PR #412, opened 2017-03-15
Update to zap 1.0

akshayjshah:
I just cut the 1.0 release of zap; this diff updates UberFX to work with the handful of changes introduced and locks to `^1`.

This should reduce the number of dependency conflicts that FX's users have to resolve.

### DIFF
--- a/ulog/sentry/sentry.go
+++ b/ulog/sentry/sentry.go
@@ -136,6 +136,11 @@ func (c *core) Write(ent zapcore.Entry, fs []zapcore.Field) error {
 	return nil
 }
 
+func (c *core) Sync() error {
+	c.client.Wait()
+	return nil
+}
+
 func (c *core) with(fs []zapcore.Field) *core {
 	// Copy our map.
 	m := make(map[string]interface{}, len(c.fields))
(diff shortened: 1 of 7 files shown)


glibsm:
I saw "changes" and I was like uh-oh. Saw the diff and was very pleased.

Congrats on the 1.0!!

akshayjshah:
Sweet. At some point, FX may want to rename `testutils` to `fxtest` for the same reason.

alsamylkin:
We want to move `testutils` out of the global let each package to have their own testutils.

madhuravi:
Nice, that makes sense but will bother the anti-stuttering folks. Objectively this does look better for users IMO.

akshayjshah:
The stuttering advice isn't typically applied to package names; see `net/http` and `net/http/httptest`, for example.

ulog/sentry/sentry.go
```diff
@@ -136,6 +136,11 @@ func (c *core) Write(ent zapcore.Entry, fs []zapcore.Field) error {
 	return nil
 }
 
+func (c *core) Sync() error {
```
alsamylkin: Why do we need this?

glibsm: zap interface was expanded and this implements the new method

sectioneight:
@akshayjshah thanks for the PR! We really appreciate the collaboration.

PR #414, opened 2017-03-15
DIG: Add the ability to register and resolve interfaces

dmcaulay:
We ran into some issues when we tried to register and resolve `service.Host`. We were able to get it working if we did something like the following:

```
var h service.Host = // some valid host
g.Register(&h)

var registered *Host
g.Resolve(&registered)
```

Unfortunately, this approach won't work with constructors unless your constructors ask for a `*service.Host`. This is obviously awkward. This PR fixes the problem. Here's the new usage:

```
var h service.Host = // some valid host
g.Register(&h)

var registered Host
g.Resolve(&registered)
```

### DIFF
--- a/dig/node.go
+++ b/dig/node.go
@@ -60,17 +60,11 @@ func (n node) id() string {
 
 type objNode struct {
 	node
-
-	obj interface{}
 }
 
 // Return the earlier provided instance
 func (n *objNode) value(g *Graph) (reflect.Value, error) {
-	v := reflect.ValueOf(n.obj)
-	if n.objType.Kind() == reflect.Interface {
-		v = v.Elem()
-	}
-	return v, nil
+	return n.cachedValue, nil
 }
 
 func (n objNode) dependencies() []interface{} {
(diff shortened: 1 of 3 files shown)


glibsm:
Change looks good to me, there are some docs that need to be modified, but I could do that in a later PR. I distinctly remember writing something like `// notice the pointer to an interface`

glibsm:
Was able to repro the problem locally. Looks like there were some missing test cases around direct interface registration.

Thanks for the fix :)

```go
package main

import (
	"fmt"

	"go.uber.org/fx/dig"
)

type s struct {
	bar string
}

func (s s) Foo() string {
	return s.bar
}

type i interface {
	Foo() string
}

func newFooI() i {
	return s{bar: "bar"}
}

func main() {
	g := dig.New()

	// constructor: working fine
	g.MustRegister(newFooI)

	var fi i
	g.MustResolve(&fi)
	fmt.Println(fi)

	// directly: broken
	var ri i = s{bar: "baz"}
	g.MustRegister(ri)

	var fi2 i
	g.MustResolve(&fi2)
	fmt.Println(fi2)
}
```

dig/node.go
```diff
@@ -60,13 +60,11 @@ func (n node) id() string {
 
 type objNode struct {
```
glibsm: shouldn't refactor now, because it will probably change very soon, but having a struct that only embeds one other struct and nothing else looks odd :)

PR #415, opened 2017-03-15
Use type switch to convert numeric values

alsamylkin:
We do type assertions when try to convert numeric values, which leads to panics, when certain values were parsed, e.g. by yaml marshaler as ints and PopulateStruct is expecting float64.

In this change we iterate over possible numerical types and assign values if there are no overflow errors.


config/decoder.go
```diff
@@ -51,24 +52,273 @@ func derefType(t reflect.Type) reflect.Type {
...
-func convertValueFromStruct(value interface{}, targetType reflect.Type, fieldType reflect.Type, fieldValue reflect.Value) error {
+func convertSignedInts(src interface{}, dst *reflect.Value) error {
+	switch t := src.(type) {
+	case int:
+		v := int64(t)
+		if !dst.OverflowInt(v) {
+			dst.SetInt(v)
+			return nil
+		}
+	case uint:
```
sectioneight: naive question: does it not compile to collapse these cases?
```go
case int, uint, ...:
```

alsamylkin: It does, but then 't' becomes an 'interface{}' and I can't convert to 'int64' easily.

config/decoder_test.go
```diff
@@ -279,3 +279,22 @@ func TestIdenticalFuzzing(t *testing.T) {
...
+	require.NoError(t, p.Get(Root).PopulateStruct(&f32))
+	require.Equal(t, f32, float32(i32))
+	require.Equal(t, f32, float32(i32+1))
+
+	var i64 int64 = 1 << 53
+	p = newValueProvider(i64)
+	var f64 float64
+	require.NoError(t, p.Get(Root).PopulateStruct(&f64))
+	require.Equal(t, f64, float64(i64))
+	require.Equal(t, f64, float64(i64+1))
```
glibsm: i don't get this line. is this an overflow of 64 float?

alsamylkin: It is an accuracy issue, floating points have 23/52 bits for the fraction part, we are just documenting via test, that if a provider returns such a big integer, converting it to float has some limitations.

config/decoder_test.go
```diff
@@ -0,0 +1,300 @@
...
+			return p.Get(Root).Populate(&x)
+		},
+		"uintptr": func(p Provider) error {
+			var x uintptr
+			return p.Get(Root).Populate(&x)
+		},
+	}
+
+	for to, f := range conversions {
+		assert.True(t, t.Run(fmt.Sprintf("%q overflow", to), func(t *testing.T) {
```
sectioneight: maybe make these sub-tests?

alsamylkin: They are already subtests :) I was a bit confused about testing.T.Run returning a bool, reporting whether a subtest failed or not, but it looks like it already marks parent:
https://golang.org/src/testing/testing.go?s=15898:15921#L427

PR #413, opened 2017-03-15
Add options to task module to disable execution

madhuravi:

PR #417, opened 2017-03-16
Add task docs for disable execution and some cleanup

madhuravi:
Followup from https://github.com/uber-go/fx/pull/413

### DIFF
--- a/modules/task/README.md
+++ b/modules/task/README.md
@@ -28,9 +28,9 @@ import (
 
 func main() {
   svc, err := service.WithModule(task.New(newBackend)).Build()
-	if err != nil {
-		log.Fatal("Failed to initialize module", err)
-	}
+  if err != nil {
+    log.Fatal("Failed to initialize module", err)
+  }
   if err := task.Register(updateCache); err != nil {
     ulog.Logger().Fatal("could not register task", "error", err)
   }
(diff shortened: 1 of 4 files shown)


modules/task/README.md
```diff
@@ -55,6 +58,20 @@ The async task module is a singleton and a service can initialize
...
+It is possible to enqueue and execute tasks on different clusters if you want
+to separate workloads. The default configuration spins up workers that will
+execute the task. In order to spin up just task enqueueing without execution,
+initialize the module as follows:
+
+```go
+  svc, err := service.WithModule(
+    task.New(newBackend, task.DisableExecution()),
+  ).Build()
+	if err != nil {
```
sectioneight: indentation is off here

modules/task/README.md
```diff
@@ -28,6 +28,9 @@ import (
 
 func main() {
   svc, err := service.WithModule(task.New(newBackend)).Build()
+	if err != nil {
```
sectioneight: indentation is off here

modules/task/README.md
```diff
@@ -55,6 +58,20 @@ The async task module is a singleton and a service can initialize
...
 and encodings for message passing.
 
+It is possible to enqueue and execute tasks on different clusters if you want
+to separate workloads. The default configuration spins up workers that will
+execute the task. In order to spin up just task enqueueing without execution,
+initialize the module as follows:
+
+```go
+  svc, err := service.WithModule(
+    task.New(newBackend, task.DisableExecution()),
```
sectioneight: Hmm, should this be taken care of by module roles instead of a one-off for the task module? E.g. shouldn't they just start the service with roles="yarpc" and not "task" if they don't want to start the task module?

madhuravi: In the case you describe yes. This is where they want to run the task module on both clusters, but one they only enqueue and the other they want to consume on.

PR #418, opened 2017-03-16
Rename PopulateStruct to Populate

alsamylkin:
Rename `PopulateStruct` method in Value to `Populate`. The method can now populate not only structs, but anything: slices, maps, builtin types and maps.

### DIFF
--- a/config/decoder.go
+++ b/config/decoder.go
@@ -215,18 +215,19 @@ func (d *decoder) mapping(childKey string, value reflect.Value, def string) erro
 		childKey = addSeparator(childKey)
 
 		for key := range v {
-
-			// Skip empty keys, because lookups in provider can toggle infinite loops/spill internals.
-			if mapKey := childKey + fmt.Sprintf("%v", key); mapKey != "" {
-				itemValue := reflect.New(valueType.Elem()).Elem()
-
-				// Try to unmarshal value and save it in the map.
-				if err := d.unmarshal(mapKey, itemValue, def); err != nil {
-					return err
-				}
-
-				destMap.SetMapIndex(reflect.ValueOf(key), itemValue)
+			subKey := fmt.Sprintf("%v", key)
+			if subKey == "" {
+				return fmt.Errorf("empty key leads to ambiguity for path: %q", childKey)
 			}
+
+			itemValue := reflect.New(valueType.Elem()).Elem()
+
+			// Try to unmarshal value and save it in the map.
+			if err := d.unmarshal(childKey+subKey, itemValue, def); err != nil {
+				return err
+			}
+
+			destMap.SetMapIndex(reflect.ValueOf(key), itemValue)
 		}
 
 		value.Set(destMap)
--- a/config/decoder.go
+++ b/config/decoder.go
@@ -73,6 +73,13 @@ func convertValueFromStruct(value interface{}, targetType reflect.Type, fieldTyp
 	return nil
 }
 
+func addSeparator(key string) string {
+	if key != "" {
+		key += _separator
+	}
+	return key
+}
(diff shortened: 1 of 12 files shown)


anuptalwalkar:
Can you please open the pr against `dev`?

config/decoder.go
```diff
@@ -125,7 +125,9 @@ func (d *decoder) sequence(childKey string, value reflect.Value) error {
 
 	// start looking for child values.
 	elementType := derefType(valueType).Elem()
-	childKey += _separator
+	if childKey != "" {
```
madhuravi: Move this out to a function, it's in quite a few places.

config/decoder.go
```diff
@@ -205,9 +209,18 @@ func (d *decoder) mapping(childKey string, value reflect.Value, def string) erro
...
-		childKey += _separator
+		if childKey != "" {
+			childKey += _separator
+		}
+
 		for key := range v {
 			mapKey := childKey + fmt.Sprintf("%v", key)
+
+			// Skip empty keys, because lookups in provider can toggle infinite loops/spill internals.
+			if mapKey == "" {
```
madhuravi: Should this not be an error?

alsamylkin: Why? Users can have an empty key in a map, we shouldn't error on that.

alsamylkin:
Well, we load from something, so I am afraid, this can lead to a confusion. PopulateStruct on the other hand was already used, we are simply expanding it's scope. Another possibility is Fill, but I don't want to completely rename, simply adjust.

config/decoder.go
```diff
@@ -205,9 +212,16 @@ func (d *decoder) mapping(childKey string, value reflect.Value, def string) erro
...
 	if v, ok := val.(map[interface{}]interface{}); ok {
-		childKey += _separator
+		childKey = addSeparator(childKey)
+
 		for key := range v {
 			mapKey := childKey + fmt.Sprintf("%v", key)
+
+			// Skip empty keys, because lookups in provider can toggle infinite loops/spill internals.
+			if mapKey == "" {
+				continue
```
anuptalwalkar: minor: Instead of skip/continue checking for not empty is more clear. we can combine with line 218.

config/yaml.go
```diff
@@ -74,11 +74,11 @@ func newYAMLProviderCore(files ...io.ReadCloser) Provider {
 // * in all the remaining cases B will overwrite A.
 func mergeMaps(dst interface{}, src interface{}) interface{} {
 	if dst == nil {
-		panic("Destination node is nil")
+		return src
```
anuptalwalkar: This looks like change of behavior. Can we have tests? What if both are null?

config/decoder.go
```diff
@@ -205,17 +212,21 @@ func (d *decoder) mapping(childKey string, value reflect.Value, def string) erro
...
 		for key := range v {
-			mapKey := childKey + fmt.Sprintf("%v", key)
-			itemValue := reflect.New(valueType.Elem()).Elem()
 
-			// Try to unmarshal value and save it in the map.
-			if err := d.unmarshal(mapKey, itemValue, def); err != nil {
-				return err
-			}
+			// Skip empty keys, because lookups in provider can toggle infinite loops/spill internals.
+			if mapKey := childKey + fmt.Sprintf("%v", key); mapKey != "" {
```
madhuravi: suggest adding the empty check here and in addSeparator into a new PR so we can test it out.

config/decoder.go
alsamylkin: Actually you right, it is an error, because it leads to ambiguity between root element and element with an empty key.

PR #424, opened 2017-03-18
Handle user-defined types for populate struct

shawnburke:
I discovered that PopulateStruct fails in cases where there is:

1) A secondary provider that's serving values as strings (e.g. the env provider)
2) A user-defined type in a struct.

This handles this by falling back to comparing reflect.Kind rather than the type of the value to be set for string cases.

Added the YARPC test case that triggered this.

PR #426, opened 2017-03-20
Parse config type aliases

alsamylkin:
In convertValue we have a type switch that is using output direct type, thus failing to set value if a target type is e.g. an alias. Shawn proposed a fix https://github.com/uber-go/fx/pull/424, but we should keep numerical parsing in one place and remove it from generic convertValue function.


config/config_test.go
```diff
@@ -426,3 +426,128 @@ func TestNopProvider_Get(t *testing.T) {
...
+
+	require.NoError(t, v.Populate(cfg))
+}
+
+func TestPointerChildTypedField(t *testing.T) {
+	t.Parallel()
+
+	type Port int
+	type childPort struct {
+		Port *Port
```
anuptalwalkar: test with non pointer type?

config/value.go
```diff
@@ -282,10 +281,6 @@ func convertValue(value interface{}, targetType reflect.Type) (interface{}, erro
 	case string:
 		target := reflect.New(targetType).Interface()
 		switch t := target.(type) {
-		case *int:
-			return strconv.Atoi(v)
-		case *bool:
-			return strconv.ParseBool(v)
 		case *time.Duration:
```
anuptalwalkar: what if time.Duration has an alias? Should we add test for it?

config/config_test.go
alsamylkin: We already have plenty of them, e.g.: https://github.com/uber-go/fx/blob/dev/config/decoder_test.go#L233

This is a regression test from Shawn's PR.

config/value.go
alsamylkin: Well, time.Duration is an alias https://golang.org/pkg/time/#Duration :)

PR #427, opened 2017-03-20
Return a key with the error on Value.Populate failures

alsamylkin:
We provide only information about types, omitting the offending key, which can be confusing for users. The easiest way would be to return error from `decoder.unmarshal`, but because it is recursive, we need to track leafs and attach the offending key on leafs.


config/decoder.go
```diff
@@ -206,6 +207,10 @@ func addSeparator(key string) string {
 	return key
 }
 
+func addKeyToError(key string, err error) error {
```
madhuravi: nit: errorWithKey(err error, key string) is more meaningful.

config/decoder.go
```diff
@@ -342,7 +347,8 @@ func (d *decoder) mapping(childKey string, value reflect.Value, def string) erro
 		for key := range v {
 			subKey := fmt.Sprintf("%v", key)
 			if subKey == "" {
-				return fmt.Errorf("empty key leads to ambiguity for path: %q", childKey)
+				// We can confuse an empty map key with a root element.
+				return addKeyToError(childKey, errors.New("empty map key is ambigious"))
```
madhuravi: nit: is it worth adding another function, newErrorWithKey(msg string, key string), so we don't have to create error twice? A small performance win but not sure how much of a difference it makes. Will leave this to you.

config/decoder.go
alsamylkin: Yeah, make sense

config/decoder.go
alsamylkin: I would rather to keep an error here, so stacktrace is not going to get too long.

PR #428, opened 2017-03-21
Introduce a cached provider

alsamylkin:
We cache values only for yaml config provider(not concurrently safe). We can introduce a simple wrapper, that can be used by any other provider to cache values, even DynamicProviders, that have callbacks by registering a callback for the underlying provider for all values that were ever retrieved.

### DIFF
--- a/config/cache_provider.go
+++ b/config/cache_provider.go
@@ -1,0 +1,88 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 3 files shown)


config/cache_provider.go
```diff
@@ -0,0 +1,88 @@
...
+func (p *cachedProvider) Get(key string) Value {
+	p.RLock()
+	if v, ok := p.cache[key]; ok {
+		p.RUnlock()
+		return v
+	}
+
+	p.RUnlock()
+	err := p.Provider.RegisterChangeCallback(key, func(key string, provider string, data interface{}) {
+		p.Lock()
```
anuptalwalkar: Not sure if locking/unlocking saving us anything here. May be just have a `p.RLock()` with `defer p.RUnlock()`?

config/cache_provider.go
```diff
@@ -0,0 +1,88 @@
...
+// The value is cached only if it is found.
+func (p *cachedProvider) Get(key string) Value {
+	p.RLock()
+	if v, ok := p.cache[key]; ok {
+		p.RUnlock()
+		return v
+	}
+
+	p.RUnlock()
+	err := p.Provider.RegisterChangeCallback(key, func(key string, provider string, data interface{}) {
```
anuptalwalkar: We provide RegisterChangeCallback  for users. Will this method override the way callbacks are setup by users?

alsamylkin: It depends on the underlying provider implementation. Usually there is only one callback allowed to store, but I can create another provider wrapper, that can multiplex them.

config/cache_provider.go
alsamylkin: We write to a map, effectively changing it, thats why we use write lock, instead of a simple read, when we retrieve a value from the cache.

config/cache_provider.go
```diff
@@ -0,0 +1,88 @@
...
+		p.cache[key] = NewValue(p, key, data, true, GetType(data), nil)
+		p.Unlock()
+	})
+
+	if err != nil {
+		return NewValue(p, key, err, true, GetType(err), nil)
+	}
+
+	v := p.Provider.Get(key)
+	if v.found {
```
alsamylkin: It may be the case for a static provider, that never changes, which we can add, but in general case the value in dynamic provider can change on the following requests.

PR #430, opened 2017-03-21
Add more async task tests to cover interfaces

madhuravi:
Added tests to cover:
1. An async func defined on a struct and uses member variables from the struct that is not a direct argument to the function itself.
2. An async func that takes an interface as argument, only works when the implementation is registered for encoding.

### DIFF
--- a/modules/task/execution_test.go
+++ b/modules/task/execution_test.go
@@ -242,7 +242,7 @@ func TestEnqueueFnClosure(t *testing.T) {
 	require.NoError(t, err)
 }
 
-func TestEnqueueFnWithStructArgReturnsError(t *testing.T) {
+func TestEnqueueFnWithStructArg(t *testing.T) {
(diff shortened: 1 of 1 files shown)


madhuravi:
@glibsm This covers registering the implementation when an async task takes an interface as input.

modules/task/execution_test.go
```diff
@@ -242,7 +242,7 @@ func TestEnqueueFnClosure(t *testing.T) {
 	require.NoError(t, err)
 }
 
-func TestEnqueueWithStructFnWithError(t *testing.T) {
+func TestEnqueueFnWithStructArgReturnsError(t *testing.T) {
```
alsamylkin: Returns no error?:)

madhuravi: Ah that was about the execution returning error, will change wording.

PR #431, opened 2017-03-21
Nuke the env provider

glibsm:
The other side of #384

Env provider is being split up into two things:
1. Ability to interpolate env variables directly into YAML 
2. A small wrapper function around `os.Getenv` which will return `config.Value` to leverage the existing config infrastructure


glibsm:
Side note: this will have been a lot easier to rip out if env specific tests were in `env_test.go` as opposed to all lumped into `config.go` :(

config/config.go
```diff
@@ -51,7 +51,7 @@ var (
 	_setupMux sync.Mutex
```
alsamylkin: Could you remove all this appRoot, environment vars above?

config/config.go
```diff
@@ -120,13 +120,6 @@ func YamlProvider() ProviderFunc {
...
 
-// EnvProvider returns function to create environment based config provider
-func EnvProvider() ProviderFunc {
-	return func() (Provider, error) {
-		return NewEnvProvider(defaultEnvPrefix, nil), nil
-	}
-}
-
 // Environment returns current environment setup for the service
 func Environment() string {
```
alsamylkin: This should be removed as well? Also Path(), EnvironmentPrefix, Key etc..

glibsm: I thought unused functions were a compiler warning. Let me take a sweep and see what's left lingering

glibsm: - `Environment()` is used for something else
- `Path()` is used for YAML file loading (not affected by this)
- `EnvironmentPrefix` is also used as part of YAML loading

Can't remove environment stuff as part of the YAML loading :) 
Once we create the env package, I expect some of that might change.

alsamylkin:
Changelog missing?

PR #433, opened 2017-03-22
Pass a tracer to the uhttp client constructor

alsamylkin:
We use a global tracer in the uhttp client, which makes impossible
to write parallel tests for users(also make our tests flaky). In this change 
we pass a tracer explicitly to the client constructor, so e.g. its dependencies 
can be resolved by DIG, if users need it.


modules/uhttp/uhttpclient/client.go
```diff
@@ -24,15 +24,16 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/opentracing/opentracing-go"
```
glibsm: wrong imports section

modules/uhttp/uhttpclient/client.go
```diff
@@ -24,15 +24,16 @@ import (
...
 	"go.uber.org/fx/auth"
 )
 
 // New creates an http.Client that includes 2 extra outbound middleware: tracing and auth
 // they are going to be applied in following order: tracing, auth, remaining outbound middleware
 // and only if all of them passed the request is going to be send.
 // Client is safe to use by multiple go routines, if global tracer is not changed.
-func New(info auth.CreateAuthInfo, middleware ...OutboundMiddleware) *http.Client {
-	defaultMiddleware := []OutboundMiddleware{tracingOutbound(), authenticationOutbound(info)}
+func New(info auth.CreateAuthInfo, tracer opentracing.Tracer, middleware ...OutboundMiddleware) *http.Client {
```
glibsm: i think we should check for nils for info and tracer and set to nops, otherwise there may be crashes if called with `(nil, nil)` :(

alsamylkin: Nice catch!

modules/uhttp/uhttpclient/client.go
```diff
@@ -25,14 +25,23 @@ import (
...
 // and only if all of them passed the request is going to be send.
 // Client is safe to use by multiple go routines, if global tracer is not changed.
-func New(info auth.CreateAuthInfo, middleware ...OutboundMiddleware) *http.Client {
-	defaultMiddleware := []OutboundMiddleware{tracingOutbound(), authenticationOutbound(info)}
+func New(tracer opentracing.Tracer, info auth.CreateAuthInfo, middleware ...OutboundMiddleware) *http.Client {
+	defaultMiddleware := make([]OutboundMiddleware, 0, 2+len(middleware))
+	if tracer != nil {
+		defaultMiddleware = append(defaultMiddleware, tracingOutbound(tracer))
+		if info != nil {
+			defaultMiddleware = append(defaultMiddleware, authenticationOutbound(info))
```
glibsm: what if tracer is nil, but info isn't? :P

modules/uhttp/uhttpclient/client.go
```diff
@@ -25,14 +25,23 @@ import (
...
+	"github.com/opentracing/opentracing-go"
 )
 
 // New creates an http.Client that includes 2 extra outbound middleware: tracing and auth
 // they are going to be applied in following order: tracing, auth, remaining outbound middleware
 // and only if all of them passed the request is going to be send.
 // Client is safe to use by multiple go routines, if global tracer is not changed.
-func New(info auth.CreateAuthInfo, middleware ...OutboundMiddleware) *http.Client {
-	defaultMiddleware := []OutboundMiddleware{tracingOutbound(), authenticationOutbound(info)}
+func New(tracer opentracing.Tracer, info auth.CreateAuthInfo, middleware ...OutboundMiddleware) *http.Client {
```
madhuravi: I think CreateAuthInfo was meant to be a collection of the base objects. It has auth and metrics right now. Either add tracer to it (and rename interface) or make all of them separate arguments.

madhuravi: @alsamylkin wanted to do it in a later PR, approving.

modules/uhttp/uhttpclient/client.go
alsamylkin: Well, then we can't use the auth. DIG will always have both, right?;)

PR #434, opened 2017-03-22
Add a provider wrapper to multiplex callbacks

alsamylkin:
Some providers(like MockDynamicProvider) allow to register only one callback. We can easily extend them by wrapping in another provider, that maintains a slice of callbacks that would to be executed on a change.

### DIFF
--- a/config/stack_callback_provider.go
+++ b/config/stack_callback_provider.go
@@ -1,0 +1,91 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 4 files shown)


config/stack_callback_provider.go
```diff
@@ -0,0 +1,91 @@
...
+// THE SOFTWARE.
+
+package config
+
+import (
+	"fmt"
+	"sync"
+)
+
+type stackCallbackProvider struct {
```
alsamylkin: I am open to a better name :)

dmcaulay: `mutliCallbackProvider`?

dmcaulay:
this looks good to me!

glibsm:
I think it may just be the late hour, but the reasoning for the particular approach is eluding me

I think it would be better to provide a way for dynamic providers to embed a struct that manages these callbacks. 

Wrapping it into a private struct is not ideal, as at that point control over the provider is lost somewhat - we just returned a private instance.

alsamylkin:
@glibsm I don't see, how do we loose a control over a provider, we still have the instance that we are wrapping, so another provider can embed this one and use it together with a provider that doesn't allow multiple callbacks per key.

I do think though, that we shouldn't make a user to register callbacks manually, Value can do it inside, but there are lots of things to consider: we copy Values a lot, so registering callbacks there can potentially explode our memory usage, concurrent safety is not here yet, etc.

config/stack_callback_provider.go
glibsm: +1 for multi

glibsm:
@alsamylkin my suggestion was to consider using composition instead. I should be easy for a dynamic provider to allow this out of the box, without the user having to wrap their providers.

I don't think it's an easy change, and certainly not API compatible.

PR #436, opened 2017-03-23
Move DIG code into external repo

glibsm:
I will follow up with moving all the tasks to the new repo as well. 

For our customers, a simple fmt to rewrite the import path should be sufficient.


glibsm:
code coverage drop is expected, as I'm removing a 93% covered package.

PR #440, opened 2017-03-23
Propagate tracing context through cherami backend

madhuravi:
When a task is enqueued, the context is dropped on the floor. This change is to extract values (SpanContext) from the context and pass this on to the cherami. When the cherami message is consumed, the context values is retrieved, decoded and passed on to the function execution,

### DIFF
--- a/modules/task/cherami/cherami.go
+++ b/modules/task/cherami/cherami.go
@@ -267,6 +267,7 @@ func (b *Backend) consumeAndExecute() {
 			b.consumeAndExecute()
 		}
 	}()
+
 	for delivery := range b.deliveryCh {
 		messageData := delivery.GetMessage().GetPayload().GetData()
 		b.withContext(delivery, func(ctx context.Context) {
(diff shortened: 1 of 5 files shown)


madhuravi:
Writing more tests, will update soon.

modules/task/cherami/cherami_test.go
```diff
@@ -49,74 +50,80 @@ var (
...
+		t,
+		lines,
+		map[string]int{"forget to register": _defaultClientConfig.ConsumeWorkerCount + 1},
+	)
+}
+
+func TestBackendWorkflowWithTask(t *testing.T) {
+	m := newMock()
+	defer m.AssertExpectations(t)
+	//bknd := createNewBackend(t, m, _host)
```
glibsm: unfinished test

modules/task/cherami/cherami_test.go
```diff
@@ -261,13 +268,70 @@ func TestStartBackendOpenPublisherError(t *testing.T) {
...
+		cBknd := bknd.(*Backend)
+		m.Delivery.On("GetMessage").Return(
+			&cherami_gen.ConsumerMessage{
+				Payload: &cherami_gen.PutMessage{
+					Data:        _publishMsg,
+					UserContext: map[string]string{_ctxKey: ""},
+				},
+			},
+		)
+		m.Delivery.On("Nack").Return(testArg.nackError).Once()
```
glibsm: ack/nack into consts?
we should also check if cherami team exports those

alsamylkin: This is a method name, it comes from mock library to specify what to return on the "Nack" method call.

modules/task/cherami/cherami.go
```diff
@@ -261,20 +269,43 @@ func (b *Backend) consumeAndExecute() {
...
+		if err != nil {
+			ulog.Logger(ctx).Error("Unable to decode context", zap.Error(err))
 			b.taskFailure.Inc(1)
-			ulog.Logger(ctx).Error("Task run failed", zap.Error(err))
-			_ = delivery.Nack()
-		} else {
-			b.taskSuccess.Inc(1)
-			if err = delivery.Ack(); err != nil {
-				ulog.Logger(ctx).Error("Task ack to cherami failed", zap.Error(err))
+			if err := delivery.Nack(); err != nil {
```
alsamylkin: If we fail to acknowledge delivery/decode context, we still continue to execute task?

alsamylkin: This is actually part of a global run..

PR #442, opened 2017-03-24
Refactor http module

anuptalwalkar:
Middlewares are defined on router than the handler. Now they can be defined and build separately from handlers. Following to this change, we want to separate and build middlewares specific to the defined routes.

### DIFF
--- a/modules/uhttp/router.go
+++ b/modules/uhttp/router.go
@@ -22,17 +22,42 @@ package uhttp
 
 import (
 	"net/http"
-
-	"github.com/gorilla/mux"
+	"regexp"
+	"sync"
 
 	"go.uber.org/fx/service"
 )
 
+// The HandlerFunc type is an adapter to allow the use of
+// functions as http handlers.
+type HandlerFunc func(w http.ResponseWriter, r *http.Request)
+
+// ServeHTTP calls f(w, r).
+func (f HandlerFunc) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	f(w, r)
+}
+
+// A RequestMatcher is a matching predicate for an HTTP Request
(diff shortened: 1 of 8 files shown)


blampe:
Let's chat about uhttp before landing this.

cc @breerly @akshayjshah @glibsm @madhuravi

modules/uhttp/router.go
```diff
@@ -22,17 +22,42 @@ package uhttp
...
+	}
+}
+
+type route struct {
+	matches RequestMatcher
+	handler http.Handler
+}
+
 // Router is wrapper around gorila mux
 type Router struct {
```
alsamylkin: I don't think we should have a router/custom interfaces in the http module. All we need to do is accept http.Handler interface and attach some metrics to it, then users can attach any routing frameworks they want.

anuptalwalkar: the whole point of having the wrapper is to inject middlewares that we have, not necessarily some metrics, but additional libraries such as rate limiting, etc. current mux.Router doesn't have the middleware chaining functionality.

alsamylkin: Rate limiter can also implement http.Handler, so we can chain middleware.

anuptalwalkar: I am confused. we are chaining using http.Handler only, except http.Handler doesn't have chaining. Thats why we have middleware to provide the chaining. you can't chain handlers with implementing only http.Handler as is.

alsamylkin: Will this work?
```golang
type AlexMiddleware struct {
	next http.Handler
}

func (a *AlexMiddleware) ServeHTTP(w ResponseWriter, rq *Request) {
	fmt.Println("Alex was here")
	a.next.ServeHTTP(w, rq)
}
```

PR #444, opened 2017-03-24
Add more defaults in cherami for owner, timeout, retention

madhuravi:
Cherami defaults to 0 if retention is not set. Setting defaults on retention period, consumer group timeout and owner email.

@datoug

### DIFF
--- a/modules/task/cherami/cherami.go
+++ b/modules/task/cherami/cherami.go
@@ -46,12 +46,12 @@ const (
 )
 
 var (
-	_hyperbahnMu                 sync.RWMutex
-	_hyperbahnHostsFile          string
-	_cheramiClientFunc                 = cherami.NewHyperbahnClient
-	_consumedMessagesRetention   int32 = 1
-	_unconsumedMessagesRetention int32 = 7
-	_defaultClientConfig               = clientConfig{
+	_hyperbahnMu                          sync.RWMutex
+	_hyperbahnHostsFile                   string
+	_cheramiClientFunc                          = cherami.NewHyperbahnClient
+	_consumedMessagesRetentionInSeconds   int32 = 1 * 24 * 60 * 60 // 1 day
+	_unconsumedMessagesRetentionInSeconds int32 = 7 * 24 * 60 * 60 // 7 days
+	_defaultClientConfig                        = clientConfig{
 		ConsumerName:       "uberfx-async",
 		PrefetchCount:      10,
 		ConsumeWorkerCount: 10,
@@ -179,8 +179,8 @@ func createDestination(client cherami.Client, config clientConfig, ownerEmail st
 	if _, err := client.CreateDestination(
 		&cherami_gen.CreateDestinationRequest{
 			Path: &config.Destination,
-			ConsumedMessagesRetention:   &_consumedMessagesRetention,
-			UnconsumedMessagesRetention: &_unconsumedMessagesRetention,
+			ConsumedMessagesRetention:   &_consumedMessagesRetentionInSeconds,
+			UnconsumedMessagesRetention: &_unconsumedMessagesRetentionInSeconds,
 			OwnerEmail:                  &ownerEmail,
 		},
 	); err != nil && !alreadyExistsError(err) {
--- a/modules/task/cherami/cherami.go
+++ b/modules/task/cherami/cherami.go
@@ -25,6 +25,7 @@ import (
 	"sync"
 	"time"
 
+	"go.uber.org/fx/config"
 	"go.uber.org/fx/modules/task"
 	"go.uber.org/fx/service"
 	"go.uber.org/fx/ulog"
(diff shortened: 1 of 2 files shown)


modules/task/cherami/cherami.go
```diff
@@ -46,15 +46,18 @@ const (
...
 
 var (
-	_hyperbahnMu         sync.RWMutex
-	_hyperbahnHostsFile  string
-	_cheramiClientFunc   = cherami.NewHyperbahnClient
-	_defaultClientConfig = clientConfig{
+	_hyperbahnMu                 sync.RWMutex
+	_hyperbahnHostsFile          string
+	_cheramiClientFunc                 = cherami.NewHyperbahnClient
+	_consumedMessagesRetention   int32 = 1
```
datoug: the unit of this is seconds

modules/task/cherami/cherami.go
```diff
@@ -46,15 +46,18 @@ const (
 )
 
 var (
-	_hyperbahnMu         sync.RWMutex
```
glibsm: this section looks like a mix of `const`s and `var`s

modules/task/cherami/cherami.go
```diff
@@ -46,15 +46,18 @@ const (
...
-	_hyperbahnHostsFile  string
-	_cheramiClientFunc   = cherami.NewHyperbahnClient
-	_defaultClientConfig = clientConfig{
+	_hyperbahnMu                 sync.RWMutex
+	_hyperbahnHostsFile          string
+	_cheramiClientFunc                 = cherami.NewHyperbahnClient
+	_consumedMessagesRetention   int32 = 1
+	_unconsumedMessagesRetention int32 = 7
```
glibsm: can people override this?

modules/task/cherami/cherami.go
alsamylkin: can we use time.Duration instead of ints?

modules/task/cherami/cherami.go
```diff
@@ -96,7 +100,11 @@ func NewBackend(host service.Host) (task.Backend, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to parse config for cherami")
 	}
-	return newBackendWithConfig(host, config)
+	var ownerEmail string
+	if err = host.Config().Get("owner").Populate(&ownerEmail); err != nil {
```
alsamylkin: You can simply use `host.Config().Get(config.ServiceOwnerKey).AsString()`

modules/task/cherami/cherami.go
```diff
@@ -96,7 +100,11 @@ func NewBackend(host service.Host) (task.Backend, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to parse config for cherami")
 	}
-	return newBackendWithConfig(host, config)
+	var ownerEmail string
```
alsamylkin: why ownerEmail is not part of clientConfig?

madhuravi: ownerEmail is at the root always, since services need it configured too.

modules/task/cherami/cherami.go
madhuravi: yes! I was looking for this.

modules/task/cherami/cherami.go
madhuravi: unfortunately, the retention needs to be a var so I can pass it in as a reference to cherami. Cherami takes int pointers.

modules/task/cherami/cherami.go
madhuravi: cherami takes input as int32, keeping it consistent with that.

alsamylkin: We can convert them to int32 when we pass it to cherami, right now it doesn't even have a measurement unit in the name..

madhuravi: latest update has it in the name, refresh.

modules/task/cherami/cherami.go
```diff
@@ -125,60 +138,68 @@ func newBackendWithConfig(host service.Host, config clientConfig) (task.Backend,
...
-		&cherami_gen.CreateDestinationRequest{Path: &destination},
+		&cherami_gen.CreateDestinationRequest{
+			Path: &cc.Destination,
+			ConsumedMessagesRetention:   &_consumedMessagesRetentionInSeconds,
+			UnconsumedMessagesRetention: &_unconsumedMessagesRetentionInSeconds,
+			OwnerEmail:                  &ownerEmail,
+		},
 	); err != nil && !alreadyExistsError(err) {
-		return errors.Wrapf(err, "unable to create destination: %s", destination)
+		return errors.Wrapf(err, "unable to create destination: %s", cc.Destination)
```
alsamylkin: nit: use %q for interpolation

PR #447, opened 2017-03-27
Remove "\n" from README text

glibsm:
This is technicality about how we pull in README into docs.
Needs to be fixed downstream, but for the time being it's ok.

### DIFF
--- a/config/README.md
+++ b/config/README.md
@@ -61,10 +61,8 @@ if err := cfg.Get("stuff.server").Populate(target); err != nil {
   // fail, we didn't find it.
 }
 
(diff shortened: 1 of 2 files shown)


config/README.md
```diff
@@ -61,11 +61,9 @@ if err := cfg.Get("stuff.server").Populate(target); err != nil {
   // fail, we didn't find it.
 }
 
-fmt.Printf("Port is: %v\n", target.Port)
+fmt.Printf("Port is: %v", target.Port) // "Port is 8081"
```
alsamylkin: You can also do fmt.Println("Port is:", target.Port) or if you want to keep the format, please use %q.

PR #450, opened 2017-03-28
Introduce config loader

alsamylkin:
We load configs in old fashioned way through a mix of functions and global variables. Extracting this functionality to a config loader helps us to manage loading and testing. Changes:
* Wrap global functions to methods in Loader
* Override dirs to load files from with SetDirs
* Make loader concurrent safe
* All tests in config package are parallel and can rerun multiple times

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@
   of using a global tracer. This will allow to use http client in parallel tests.
 - **[Breaking]** Introduce a config loader, this will allow to override config loading
   and use custom dirs to load from. In order to load configs calls to `config.Load()`
-  should be replaced with `config.DefaultLoader.Load()`.
+  should be replaced with `config.NewLoader().Load()`.
 
 ## v1.0.0-beta2 (09 Mar 2017)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## v1.0.0-beta4 (unreleased)
+
+- **[Breaking]** Introduce a config loader, this will allow to override config loading
+  and use custom dirs to load from. In order to load configs calls to `config.Load()`
+  should be replaced with `config.NewLoader().Load()`.
+
 ## v1.0.0-beta3 (unreleased)
 
 - **[Breaking]** Environment config provider was removed. If you were using
@@ -24,9 +30,6 @@
   at `go.uber.org/dig`.
 - **[Breaking]** Pass a tracer the `uhttp/uhttpclient` constructor explicitly, instead
   of using a global tracer. This will allow to use http client in parallel tests.
-- **[Breaking]** Introduce a config loader, this will allow to override config loading
-  and use custom dirs to load from. In order to load configs calls to `config.Load()`
-  should be replaced with `config.NewLoader().Load()`.
 
 ## v1.0.0-beta2 (09 Mar 2017)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@
   at `go.uber.org/dig`.
 - **[Breaking]** Pass a tracer the `uhttp/uhttpclient` constructor explicitly, instead
   of using a global tracer. This will allow to use http client in parallel tests.
+- **[Breaking]** Introduce a config loader, this will allow to override config loading
+  and use custom dirs to load from.
 
 ## v1.0.0-beta2 (09 Mar 2017)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,8 @@
 - **[Breaking]** Pass a tracer the `uhttp/uhttpclient` constructor explicitly, instead
   of using a global tracer. This will allow to use http client in parallel tests.
 - **[Breaking]** Introduce a config loader, this will allow to override config loading
-  and use custom dirs to load from.
+  and use custom dirs to load from. In order to load configs calls to `config.Load()`
+  should be replaced with `config.DefaultLoader.Load()`.
 
 ## v1.0.0-beta2 (09 Mar 2017)
 
--- a/config/config.go
+++ b/config/config.go
@@ -77,7 +77,7 @@ func newDefaultLoader() *Loader {
 	}
 
 	l.configFiles = l.baseFiles()
-	l.RegisterProviders(l.YamlProvider(l.configFiles...))
+	l.RegisterProviders(l.YamlProvider())
 
 	return l
 }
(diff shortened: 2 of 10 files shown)


config/decoder.go
```diff
@@ -386,14 +386,7 @@ func (d *decoder) iface(key string, value reflect.Value, def string) error {
 
 // Sets value to an object type.
 func (d *decoder) object(childKey string, value reflect.Value) error {
-	value = value.Addr()
-
-	if value.IsNil() {
```
alsamylkin: This code is unreachable.

CHANGELOG.md
```diff
@@ -24,6 +24,8 @@
   at `go.uber.org/dig`.
 - **[Breaking]** Pass a tracer the `uhttp/uhttpclient` constructor explicitly, instead
   of using a global tracer. This will allow to use http client in parallel tests.
+- **[Breaking]** Introduce a config loader, this will allow to override config loading
```
glibsm: is it actually breaking? if so, we should add some notes as to what the user should expect to do to migrate

config/config.go
```diff
@@ -29,186 +29,248 @@ import (
...
+	l.lock.Lock()
+
+	l.configFiles = files
+
+	l.lock.Unlock()
+}
+
+func (l *Loader) getFiles() []string {
+	l.lock.RLock()
+	l.lock.RUnlock()
```
glibsm: this one should be a defer unlock

config/config.go
```diff
@@ -29,186 +29,248 @@ import (
...
-// SetEnvironmentPrefix sets environment prefix for the application
-func SetEnvironmentPrefix(envPrefix string) {
-	_envPrefix = envPrefix
+// SetEnvironmentPrefix sets environment prefix for the application.
+func (l *Loader) SetEnvironmentPrefix(envPrefix string) {
+	l.lock.Lock()
+
+	l.envPrefix = envPrefix
+
+	l.lock.Unlock()
```
glibsm: lets be consistent: either use defer unlock, or this one

config/config_test.go
```diff
@@ -266,64 +282,76 @@ boolean:
...
+	l.SetEnvironmentPrefix("TEST")
+
+	files := l.joinFilepaths(l.baseFiles()...)
+	expected := []string{
+		"base.yaml",
+		"development.yaml",
+		"secrets.yaml",
+		"config/base.yaml",
+		"config/development.yaml",
+		"config/secrets.yaml"}
```
glibsm: `}` on new line

config/config_test.go
```diff
@@ -473,14 +512,105 @@ rpc:
...
+	f := func(dir string) {
+		l.SetDirs(dir)
+		p := l.Load()
+		assert.Equal(t, "bane", p.Get("vilain").String())
+	}
+
+	wg := sync.WaitGroup{}
+	wg.Add(2)
+	op := func() {
+		withBase(t, f, "vilain: bane")
```
glibsm: are you sure bane is a villain?

service/manager.go
```diff
@@ -83,7 +83,7 @@ func newManager(builder *Builder) (Manager, error) {
 		// If the user didn't pass in a configuration provider, load the standard.
 		// Bypassing standard config load is pretty much only used for tests, although it could be
 		// useful in certain circumstances.
-		m.configProvider = config.Load()
+		m.configProvider = config.DefaultLoader.Load()
```
glibsm: we should preserve the `Load()` package level function which just calls DefaultLoader, if this is one of the "breaking" changes to the API

CHANGELOG.md
alsamylkin: It breaks our API, because we change config methods signatures.

config/config.go
alsamylkin: Nice catch!

config/config.go
alsamylkin: Well, I was doing for performance, but consistency is probably going to be safer :)

config/config_test.go
alsamylkin: yes, superheroes don't do any testing.

config/config_test.go
```diff
@@ -473,14 +513,105 @@ rpc:
...
+		wg.Done()
+	}
+
+	go op()
+	go op()
```
anuptalwalkar: why call twice?

config/config.go
```diff
@@ -29,186 +29,243 @@ import (
...
+	dirs []string
+
+	// Where to look for environment variables.
+	lookUp lookUpFunc
+}
+
+// DefaultLoader will be used if config is not defined for a service.
+var DefaultLoader = newDefaultLoader()
+
+func newDefaultLoader() *Loader {
```
anuptalwalkar: Feels like this should be exported. Users shouldn't have to go through DefaultLoader to load config.

service/manager.go
alsamylkin: While it is tempting to keep it consistent, I'd prefer to not to do that:
1. There are other methods moved that have a receiver, so it will not fully help us.
2. It will be harder to maintain in the future: config can be loaded from multiple places, instead of one, so I'd prefer to bite a bullet now.

config/config.go
```diff
@@ -29,186 +29,243 @@ import (
...
+
+	// Dirs to load from.
+	dirs []string
+
+	// Where to look for environment variables.
+	lookUp lookUpFunc
+}
+
+// DefaultLoader will be used if config is not defined for a service.
+var DefaultLoader = newDefaultLoader()
```
anuptalwalkar: If Im reading this correctly, we are anyway having this is a global object. Can we keep Load() method on package level? (May be just `Load()` that calls `DefaultLoader.Load()`) Then we don't have to change much of the usage.

config/config.go
alsamylkin: Hmm, yeah, make sense, let me update remove the global object.

config/config_test.go
alsamylkin: To test for race conditions.

config/config.go
anuptalwalkar: I realized that my both comments are conflicting for/against global objects. I prefer having global variable and exported `New` simply for accessibility, but either way is fine.

glibsm:
With the removal of the global loader, how do people override the set of files they want to use?

alsamylkin:
Yeah, rolling it back and keeping the default loader to keep setup in private repos.

config/config.go
alsamylkin: Glib mentioned the same issue above, lets keep discussions there in one place.

PR #453, opened 2017-03-28
v1.0.0-beta3 ready code into master

glibsm:


glibsm:
this was not correctly merge, should be a merge commit, rather than a rebase merge (which rewrote all the hashes)

PR #454, opened 2017-03-28
Update version to beta4

alsamylkin:


alsamylkin:
Already done manually.

PR #457, opened 2017-04-05
Do not interpolate secrets.yaml

glibsm:
Secrets often contain auto-generated passwords and things of that nature, which can contain combinations that look very much like an environment interpolation syntax.

This still doesn't resolve the need to give user control should they actually want literal `${}` in their config.

### DIFF
--- a/config/yaml.go
+++ b/config/yaml.go
@@ -288,9 +288,20 @@ func unmarshalYAMLValue(reader io.ReadCloser, value interface{}, lookUp lookUpFu
 		return errors.Wrap(err, "failed to read the yaml config")
 	}
 
-	data, err := interpolateEnvVars(raw, lookUp)
-	if err != nil {
-		return errors.Wrap(err, "failed to interpolate environment variables")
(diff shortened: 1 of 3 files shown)


config/yaml.go
```diff
@@ -288,9 +288,20 @@ func unmarshalYAMLValue(reader io.ReadCloser, value interface{}, lookUp lookUpFu
 		return errors.Wrap(err, "failed to read the yaml config")
 	}
 
-	data, err := interpolateEnvVars(raw, lookUp)
-	if err != nil {
-		return errors.Wrap(err, "failed to interpolate environment variables")
+	var data []byte
+	skipInterpolate := false
+	if f, ok := reader.(*os.File); ok {
+		if strings.Contains(f.Name(), "secrets.yaml") {
```
anuptalwalkar: may be just check for secrets? This doesn't cover secrets-dc.yaml, etc.

glibsm: yeah good point

Issue #425, opened 2017-03-19
Support for HTTP module with Auth and non Auth clients?


dbainbridge:
My service exposes both authenticated and non-authenticated HTTP endpoints.  The `defaultInboundMiddlewareChainBuilder` is automatically adding auth client from the `service` level and `uauth` appears to only allow a single auth client per service. 

Is there a way for me to provide both authenticated and non-authenticated endpoints without checking for specific paths in a custom auth client?

sectioneight:
Yes, we're reviewing our middleware strategy both for HTTP and YARPC to figure out the best way to express endpoint-specific middlewares/decorators/interceptors/whatever-you-call-them. Will have a better story at the end of this week, and have something before the 1.0 release that allows endpoint-specific auth (and other middleware) configuration for both uhttp and yarpc modules.

breerly:
Closing - auth will be implemented as a module in a separate repo.

PR #458, opened 2017-04-05
[prototype] Add granularity to middleware on handlers

anuptalwalkar:
Introducing a way to configure middleware based on handler definition, and gathering thoughts on this proposal. This is inline with issue #425 

As of today, yarpc (or even uhttp) middlewares are defined as static defaults in UberFx. We want to expand on how middleware are defined for a service, making them configurable and more granular to the users. Middleware per procedure are useful for number of cases - 

1. Services that need to configure middleware (statically or dynamically) per handler.
2. Services want specific middleware implemented for certain handlers. (rate limit, error rate detection based on handlers)
3. Easy to configure middleware at runtime rather than having to update the code.

For each request, middleware will be executed in the following order -
1. default middleware- Default middleware is a middleware that should be integrated with the transport layer regardless of the handler definition.
Example default middleware - logging, tracing middleware, panic handler

2. handler middleware - Handler middleware is not generic to all handlers in the transport layer. This middleware can be configured for any number of procedures. 
Configuration setup - 
```
------------------------------------
modules:
  yarpc:
    procedures:
      handler1:
        middleware:
          - recovery
          - ratelimit
      handler2:
        middleware:
------------------------------------
```

### DIFF
--- a/modules/yarpc/middleware.go
+++ b/modules/yarpc/middleware.go
@@ -25,7 +25,7 @@ import (
 	"sync"
 
 	"go.uber.org/fx/auth"
-	"go.uber.org/fx/modules/decorator"
+	"go.uber.org/fx/modules/yarpc/middleware"
 	"go.uber.org/fx/service"
 	"go.uber.org/fx/ulog"
 	"go.uber.org/zap"
@@ -40,13 +40,13 @@ var (
 	mu sync.Mutex
 )
 
-// TransportUnaryMiddleware keeps all the decorator layers defined in the configuration
+// TransportUnaryMiddleware keeps all the middleware layers defined in the configuration
 type TransportUnaryMiddleware struct {
-	procedures map[string][]decorator.Decorator
-	decorators map[string]transport.UnaryHandler
+	procedures  map[string][]middleware.Middleware
+	middlewares map[string]transport.UnaryHandler
 }
 
-// Handle all decorators for Unary transport
+// Handle all middlewares for Unary transport
 func (l TransportUnaryMiddleware) Handle(
 	ctx context.Context,
 	req *transport.Request,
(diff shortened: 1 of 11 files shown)


anuptalwalkar:
@glibsm @breerly @peter-edge

modules/yarpc/middleware.go
```diff
@@ -34,33 +36,57 @@ import (
...
-		Tagged(map[string]string{_tagProcedure: req.Procedure}).
-		Timer(req.Procedure).
-		Start()
-	defer stopwatch.Stop()
+	if _, ok := l.procedures[req.Procedure]; !ok {
+		return handler.Handle(ctx, req, resw)
+	}
 
-	return handler.Handle(ctx, req, resw)
+	mu.Lock()
```
dmcaulay: I'd rather see us load at init time and remove the lock. Locking on every call seems wasteful.

modules/yarpc/yarpc_test.go
```diff
@@ -48,13 +48,19 @@ modules:
          port: 0
      - http:
          port: 0
+    procedures:
```
dmcaulay: This config looks different than the key value example config.

dmcaulay: Maybe the example config needs to be updated?

modules/yarpc/middleware.go
amckinney: With the current implementation, we will only lock upon the first call to `Handle`. When Anup and I explored this earlier, it seemed that the initialization needed to happen here (given the current state of FX+YARPC). There may be a way to refactor this in the future, but it will potentially require changes in both repositories.

dmcaulay:
Also @akshayjshah, since he was asking about yarpc middleware usage.

modules/yarpc/middleware.go
anuptalwalkar: Yes, the reason is that we simply don't know what handlers are being initialized (unless we parse the thrift file)

modules/yarpc/yarpc_test.go
anuptalwalkar: Fixing the keyvalue example.

anuptalwalkar:
@breerly I updated the PR for now, but I don't intent to merge/close it. It makes complete sense to move this under yarpc and only keep configurable middleware to fx. This is mostly a prototype to demonstrate how configuration would look like, based on middleware per procedure and default middleware. 
What do I need to make this happen in yarpc-go?

anuptalwalkar:
closing since the changes will be needed in yarpc.

PR #459, opened 2017-04-10
Remove stats from yarpc module

anuptalwalkar:
Cleaning up any repeated counters/timers added to the yarpc module. counters and timer stats will be reported from yarpc library itself.


modules/yarpc/middleware.go
```diff
@@ -34,22 +34,14 @@ import (
 
 const _panicResponse = "Server Error"
 
-type contextInboundMiddleware struct {
-	statsClient *statsClient
-}
+type contextInboundMiddleware struct{}
```
alsamylkin: Do we need this Middleware anymore?

modules/yarpc/middleware.go
```diff
@@ -109,36 +98,31 @@ func authorize(ctx context.Context, host service.Host, statsClient *statsClient)
...
 ) error {
-	defer panicRecovery(ctx, p.statsClient)
+	defer panicRecovery(ctx)
 	return handler.HandleOneway(ctx, req)
 }
 
-func panicRecovery(ctx context.Context, statsClient *statsClient) {
+func panicRecovery(ctx context.Context) {
 	if err := recover(); err != nil {
-		statsClient.RPCPanicCounter().Inc(1)
```
alsamylkin: Does YARPC already have it? In order to check panics users will have to set alerts for log messages, instead of metrics, not sure if it is always possible.

alsamylkin: Hmm, looks like there is no need for this middleware either, YARPC will handle it.

anuptalwalkar: you are right. Will remove both.

glibsm:
does this need an update in the changelog?

alsamylkin:
I don't think any API changed, only some counters are removed..

glibsm:
Changelogs are not only for the API changes, or only *breaking* api changes. It's to keep track of what the differences between the releases are.

Removal of metrics fits that bill for me

PR #461, opened 2017-04-11
Introduce a command line provider

alsamylkin:
Our documentation states that service roles can be overridden from command line,
but we never actually had a config provider that takes data from arguments
passed to a program.

We use "github.com/ogier/pflag" package to read command line flags, because
the standard one isn't POSIX compatible(it can be easily replaced with the standard
though by replacing import to "flag", function signatures are exactly the same).

By default flags don't have a support for list values, so we assume list elements
are comma separated. If users want to have different separators, the can create
their own type and set flags via [Var](https://godoc.org/github.com/ogier/pflag#FlagSet.Var)

If users want to introduce their own command line parameters, they can use the default pflag
[CommandLine](https://godoc.org/github.com/ogier/pflag#pkg-variables) variable and
extend the provider.

Documentation update is coming in the following PRs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,10 @@
 - **[Breaking]** Introduce a config loader, this will allow to override config loading
   and use custom dirs to load from. In order to load configs calls to `config.Load()`
   should be replaced with `config.NewLoader().Load()`.
-- Added `metrics.NopScope` for tests on service.NopHost with tagging capabilities turned
-  on by default
+- Added `metrics.NopScope` for tests on service.NopHost with tagging capabilities
+  turned on by default
+- Added a command line provider `config.NewCommandLineProvider()`, which can be used
+  to pass configuration parameters through command line.
 
 ## v1.0.0-beta3 (28 Mar 2017)
 
--- a/config/command_line_provider.go
+++ b/config/command_line_provider.go
@@ -70,7 +70,7 @@ func NewCommandLineProvider(flags *flag.FlagSet, args []string) Provider {
 func assignValues(m map[string]interface{}, key string, value flag.Value) {
 	if ss, ok := value.(*StringSlice); ok {
 		slice := []string(*ss)
-		tmp := map[string]interface{}{}
+		tmp := make(map[string]interface{}, len(slice))
 		m[key] = tmp
 		for i, str := range slice {
 			tmp[fmt.Sprint(i)] = str
@@ -95,6 +95,8 @@ func traversePath(m map[string]interface{}, f *flag.Flag) (prev map[string]inter
 		if tmp, ok := curr[item].(map[string]interface{}); ok {
 			curr = tmp
 		} else {
+			// This should never happen, because pflag/flag sort flags before calling a visitor,
+			// but it is better to be safe then sorry.
 			curr = map[string]interface{}{}
 		}
 	}
--- a/config/command_line_provider.go
+++ b/config/command_line_provider.go
@@ -83,7 +83,7 @@ func assignValues(m map[string]interface{}, key string, value flag.Value) {
 }
 
 // Traverse map with the flag name used as path.
-func traversePath(m map[string]interface{}, f *flag.Flag) (prev map[string]interface{}, last string) {
+func traversePath(m map[string]interface{}, f *flag.Flag) (map[string]interface{}, string) {
 	curr, prev := m, m
 	path := strings.Split(f.Name, _separator)
 	for _, item := range path {
--- a/config/command_line_provider.go
+++ b/config/command_line_provider.go
@@ -32,18 +32,19 @@ type StringSlice []string
 
 var _ flag.Value = (*StringSlice)(nil)
 
+// String returns slice elements separated by comma.
 func (s *StringSlice) String() string {
 	return strings.Join(*s, ",")
 }
 
+// Set splits val using comma as separators.
 func (s *StringSlice) Set(val string) error {
 	*s = StringSlice(strings.Split(val, ","))
 	return nil
 }
 
 type commandLineProvider struct {
-	p Provider
-	NopProvider
+	Provider
 }
 
 // NewCommandLineProvider returns a Provider that is using command line parameters as config values.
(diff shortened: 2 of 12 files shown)


config/command_line_provider.go
```diff
@@ -0,0 +1,107 @@
...
+		panic(err)
+	}
+
+	m := make(map[string]interface{})
+	flags.VisitAll(func(f *flag.Flag) {
+		prev, last := traversePath(m, f)
+		assignValues(prev, last, f.Value)
+	})
+
+	return commandLineProvider{Provider: NewStaticProvider(m)}
```
glibsm: well this is pretty nifty

CHANGELOG.md
```diff
@@ -5,8 +5,10 @@
 - **[Breaking]** Introduce a config loader, this will allow to override config loading
   and use custom dirs to load from. In order to load configs calls to `config.Load()`
   should be replaced with `config.NewLoader().Load()`.
-- Added `metrics.NopScope` for tests on service.NopHost with tagging capabilities turned
-  on by default
+- Added `metrics.NopScope` for tests on service.NopHost with tagging capabilities
+  turned on by default
```
glibsm: thanks for tidying this up

config/command_line_provider.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// Assign values to a map element based on value type.
+// If value is a StringSlice - create a new map and with keys - indices and values - StringSlice elements.
+// Otherwise just assign it's string value.
+func assignValues(m map[string]interface{}, key string, value flag.Value) {
+	if ss, ok := value.(*StringSlice); ok {
+		slice := []string(*ss)
+		tmp := map[string]interface{}{}
```
glibsm: nit: you can pre-make it with the size

config/command_line_provider_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+	c := NewCommandLineProvider(f, []string{"--Sushi.Tools.1=Fork", "--Sushi.Tools=Fukin, Hashi"})
+	type Sushi struct {
+		Tools []string
+	}
+
+	var v Sushi
+	require.NoError(t, c.Get("Sushi").Populate(&v))
+	assert.Equal(t, Sushi{
+		Tools: []string{"Fukin", "Fork"},
```
glibsm: ಠ_ಠ

config/command_line_provider.go
alsamylkin: Yeah, need to write more docs to spread the word about staticProvider.

config/command_line_provider_test.go
alsamylkin: https://en.wikipedia.org/wiki/Sushi#Etiquette

glibsm: I know, but the implication is clearly there. Lets keep it clean of innuendo :)

config/config.go
```diff
@@ -71,18 +73,22 @@ type Loader struct {
 }
 
 // DefaultLoader is going to be used by a service if config is not specified.
-var DefaultLoader = NewLoader()
+// First values are going to be looked in dynamic providers, then in command line provider
+// and YAML provider is going to be the last.
+var DefaultLoader = NewLoader(commandLineProviderFunc)
```
anuptalwalkar: Should we wrap `commandLineProviderFunc` inside NewLoader as default rather than passing it in?

config/command_line_provider.go
```diff
@@ -0,0 +1,109 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package config
+
+import (
+	"fmt"
+	"strings"
+
+	flag "github.com/ogier/pflag"
```
alsamylkin: It is in description, but in general stdlib flags are don't support shorthands and use one dash for everything, so we'll have to use `./service -roles=a,b` instead of `./service --roles=a,b`

config/command_line_provider.go
```diff
@@ -0,0 +1,109 @@
...
+		}
+
+		return
+	}
+
+	m[key] = value.String()
+}
+
+// Traverse map with the flag name used as path.
+func traversePath(m map[string]interface{}, f *flag.Flag) (prev map[string]interface{}, last string) {
```
alsamylkin: https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

config/command_line_provider.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+type commandLineProvider struct {
+	Provider
+}
+
+// NewCommandLineProvider returns a Provider that is using command line parameters as config values.
+// In order to address nested elements one can use dots in flag names which are considered separators.
+// One can use StringSlice type to work with a list of comma separated strings.
+func NewCommandLineProvider(flags *flag.FlagSet, args []string) Provider {
```
alsamylkin: It is being used in `commandLineProviderFunc()` for DefaultLoader.
Docs are in a separate [PR](https://github.com/uber-go/fx/pull/462)

config/command_line_provider.go
glibsm: This is dangerous, as a missing return would still get implied return through named parameters. The only time named parameters should be used is if you want to return an error on panic defer - that's literally it.

In all other cases we should avoid using named returns

alsamylkin: Peer pressured..

glibsm: Yeah but this is the *good* kind of peer pressure :)

config/command_line_provider.go
```diff
@@ -0,0 +1,109 @@
...
+// Otherwise just assign it's string value.
+func assignValues(m map[string]interface{}, key string, value flag.Value) {
+	if ss, ok := value.(*StringSlice); ok {
+		slice := []string(*ss)
+		tmp := make(map[string]interface{}, len(slice))
+		m[key] = tmp
+		for i, str := range slice {
+			tmp[fmt.Sprint(i)] = str
+		}
+
```
anuptalwalkar: nit: new line

alsamylkin: This is an msft style habit: extra line after closing multiline scopes :)

PR #463, opened 2017-04-12
Add handler struct to file refresh

anuptalwalkar:
This change is made for refresh to create valid go files. Code generation started to fail after thrift 1.1.0 due to stricter validation.

### DIFF
--- a/modules/yarpc/thriftrw-plugin-thriftsync/update.go
+++ b/modules/yarpc/thriftrw-plugin-thriftsync/update.go
@@ -121,7 +121,7 @@ func (u *Updater) compare(service *api.Service, filepath string, handlerDir stri
 
 // RefreshAll creates new funcs if they are missing from *.go file, and updates
 // all the existing functions from the idl.
-func (u *Updater) RefreshAll(service *api.Service, filepath string) error {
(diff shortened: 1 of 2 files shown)


modules/yarpc/thriftrw-plugin-thriftsync/update.go
```diff
@@ -176,12 +176,12 @@ func (u *Updater) createExpr(filepath string, service *api.Service, f *api.Funct
...
 
-func (u *Updater) generateSingleFunction(goFilePath string, service *api.Service, f *api.Function) ([]byte, error) {
-	var funcs []*api.Function
+func (u *Updater) generateSingleFunction(goFilePath string, service *api.Service, f *api.Function, handlerStructName string) ([]byte, error) {
 	newData := &updatedData{
-		Service:   service,
-		Functions: append(funcs, f),
+		Service: service,
 	}
+	newData.Functions = append(newData.Functions, f)
```
alsamylkin: Why do you want to move it out of initializer?

anuptalwalkar: Its more streamlined than having to create funcs var and append.

PR #466, opened 2017-04-14
Remove authClient from host and cleanup CreateAuthInfo

anuptalwalkar:
There is no need of Host to utilize or initialize authClient, so now modules can load auth client by calling `auth.Load()`. Also removed CreateAuthInfo to simplify initialization.  

In the future, the auth client will be provided to modules via dig as part of module construction.

### DIFF
--- a/modules/uhttp/http.go
+++ b/modules/uhttp/http.go
@@ -101,6 +101,7 @@ func newModule(host service.Host, handler http.Handler) (*Module, error) {
 	serveMux := http.NewServeMux()
 	serveMux.Handle(healthPath, healthHandler{})
 
+	// TODO: pass in the auth client as part of module construction
 	authClient := auth.Load(host.Config(), host.Metrics())
 	stats := newStatsClient(host.Metrics())
 
--- a/modules/uhttp/http.go
+++ b/modules/uhttp/http.go
@@ -29,6 +29,7 @@ import (
 	"sync"
 	"time"
 
+	"go.uber.org/fx/auth"
 	"go.uber.org/fx/service"
 	"go.uber.org/fx/ulog"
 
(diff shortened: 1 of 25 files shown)


modules/uhttp/http.go
```diff
@@ -110,10 +111,11 @@ func newModule(
 	if err := host.Config().Get("modules").Get(host.ModuleName()).Populate(&cfg); err != nil {
 		log.Error("Error loading http module configuration", zap.Error(err))
 	}
+
 	module := &Module{
 		Host:     host,
 		handlers: addHealth(getHandlers(host)),
-		mcb:      defaultInboundMiddlewareChainBuilder(log, host.AuthClient(), newStatsClient(host.Metrics())),
+		mcb:      defaultInboundMiddlewareChainBuilder(log, auth.Load(host.Config(), host.Metrics()), newStatsClient(host.Metrics())),
```
alsamylkin: Why are we creating a new client instead of reusing the existing one?

modules/uhttp/uhttpclient/client_middleware.go
```diff
@@ -85,9 +86,9 @@ func tracingOutbound(tracer opentracing.Tracer) OutboundMiddlewareFunc {
 
 // authenticationOutbound on client side calls authenticate, and gets a claim that client is who they say they are
 // We only authorize with the claim on server side
-func authenticationOutbound(info auth.CreateAuthInfo) OutboundMiddlewareFunc {
-	authClient := auth.Load(info)
-	serviceName := info.Config().Get(config.ServiceNameKey).AsString()
+func authenticationOutbound(cfg config.Provider, scope tally.Scope) OutboundMiddlewareFunc {
```
alsamylkin: Can we accept an auth.Client, instead of loading a new client for each middleware?

anuptalwalkar:
> I don't understand, why do we need to load client separately in each middleware..

Sorry, I should have been more clear in description. The loading will go away once we make the module constructor accept more objects instead of host. This is a stop gap for auth initialization. I will add a TODO here to pass the client in.

modules/uhttp/uhttpclient/client_middleware.go
anuptalwalkar: fixed.

anuptalwalkar:
ping

PR #467, opened 2017-04-15
Error on Populate a non pointer type

alsamylkin:
Right now we panic if user passes a non pointer type, error is more helpful.

### DIFF
--- a/config/static_provider_test.go
+++ b/config/static_provider_test.go
@@ -208,3 +208,13 @@ func TestPopulateForNestedMaps(t *testing.T) {
 	assert.Contains(t, err.Error(), `empty map key is ambigious`)
 	assert.Contains(t, err.Error(), `a.`)
 }
+
+func TestPopulateNonPointerType(t *testing.T) {
+	t.Parallel()
+
+	p := NewStaticProvider(42)
+	x := 13
+	err := p.Get(Root).Populate(x)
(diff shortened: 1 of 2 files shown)


glibsm:
good.... goooood.... 

```
$ ag reflect | wc -l
     107
```

config/static_provider_test.go
```diff
@@ -208,3 +208,13 @@ func TestPopulateForNestedMaps(t *testing.T) {
 	assert.Contains(t, err.Error(), `empty map key is ambigious`)
 	assert.Contains(t, err.Error(), `a.`)
 }
+
+func TestPopulateNonPointerType(t *testing.T) {
+	t.Parallel()
+
+	p := NewStaticProvider(42)
+	x := 13
+	err := p.Get(Root).Populate(x)
```
glibsm:
```
var foo Interface
foo = structImplementingInterface{}
Populate(&foo)
```

does that work?

alsamylkin: Only if a value returned by a provider implements the Interface:
https://github.com/uber-go/fx/blob/dev/config/decoder.go#L371

alsamylkin: https://github.com/uber-go/fx/blob/dev/config/yaml_test.go#L691

PR #468, opened 2017-04-17
	Add metrics, logging, tracing to di-first

anuptalwalkar:


service2/service2.go
```diff
@@ -52,6 +52,9 @@ func (s *Service) Start() {
 func (s *Service) Stop() {
 	// close all dig stuff
 	log.Println("Stopping...")
+	for _, closer := range _closers {
```
alsamylkin: Nit: this probably should be done in reverse order.

service2/service_core.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"github.com/opentracing/opentracing-go"
+	"github.com/pkg/errors"
+	"github.com/uber-go/tally"
+	jconfig "github.com/uber/jaeger-client-go/config"
+	"go.uber.org/zap"
+)
+
+var (
+	_metricsStatsReporter tally.CachedStatsReporter
```
alsamylkin: Lets try to not use global vars, it should be part of the design.

anuptalwalkar: I agree. We currently don't support registering multiple object returns.

PR #471, opened 2017-04-18
More efficient group provider initialization

alsamylkin:
1. We are using a really inefficient way to reverse a slice.
2. WithProvider can't be used outside of the `config` package, because `providerGroup` is private.

### DIFF
--- a/config/provider_group.go
+++ b/config/provider_group.go
@@ -25,22 +25,18 @@ type providerGroup struct {
 	providers []Provider
 }
 
-// NewProviderGroup creates a configuration provider from a group of backends
+// NewProviderGroup creates a configuration provider from a group of backends.
+// The highest priority provider is the last.
 func NewProviderGroup(name string, providers ...Provider) Provider {
-	group := providerGroup{
-		name: name,
+	l := len(providers)
+	for i := 0; i < l/2; i++ {
+		last := l - i - 1
+		providers[i], providers[last] = providers[last], providers[i]
 	}
-	for _, provider := range providers {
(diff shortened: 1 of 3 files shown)


modules/task/cherami/cherami_test.go
```diff
@@ -101,7 +101,7 @@ func TestBackendWorkflowWorkerPanic(t *testing.T) {
 	}
 	// Publish valid message
 	publish(t, m, bknd, deliveryCh, nil, nil)
-	time.Sleep(30 * time.Millisecond)
+	time.Sleep(100 * time.Millisecond)
```
alsamylkin: https://travis-ci.org/uber-go/fx/jobs/223243446

glibsm: what's this got to do with group init of providers, though? :(

config/provider_group.go
```diff
@@ -25,22 +25,18 @@ type providerGroup struct {
...
 }
 
-// NewProviderGroup creates a configuration provider from a group of backends
+// NewProviderGroup creates a configuration provider from a group of backends.
+// The highest priority provider is the last.
 func NewProviderGroup(name string, providers ...Provider) Provider {
-	group := providerGroup{
-		name: name,
+	l := len(providers)
+	for i := 0; i < l/2; i++ {
```
glibsm: stared at this for way to long to understand what's going on

modules/task/cherami/cherami_test.go
alsamylkin: Nothing, sleep is flaky as predicted.

PR #472, opened 2017-04-19
Make loader to look for config files in the current folder

alsamylkin:
Loader checks only config folder when it is not set by environment variable:
somehow it was still controlled by a global variable, not the `Loader.dirs` variable.

We remove `joinFilepaths` because this part is redundant and should be done by
`FileResolver`.


config/config.go
```diff
@@ -166,12 +147,12 @@ func (l *Loader) Environment() string {
 }
 
 // Path returns path to the yaml configurations
-func (l *Loader) Path() string {
+func (l *Loader) Path() []string {
```
glibsm: Path -> Paths?

config/config_test.go
```diff
@@ -636,3 +607,15 @@ func TestLoader_StaticProviderOrder(t *testing.T) {
...
+	t.Parallel()
+	f := func(dir string) {
+		l := NewLoader()
+		l.SetConfigFiles(dir+"/base.yaml")
+		p := l.Load()
+		assert.Equal(t, "base", p.Get("value").AsString())
+	}
+
+	withBase(t, f, "value: base")
+}
```
glibsm: newline

config/config.go
```diff
@@ -166,12 +147,12 @@ func (l *Loader) Environment() string {
...
 	if path, ok := l.lookUp(l.EnvironmentPrefix() + _configDir); ok {
-		return path
+		return []string{path}
 	}
 
-	return _configRoot
```
alsamylkin: CHANGELOGging..

glibsm:
lint fails

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
   turned on by default
 - Added a command line provider `config.NewCommandLineProvider()`, which can be used
   to pass configuration parameters through command line.
+- Refactored `Loader.Path()` to `Loader.Paths()`, because we actually look for config
```
glibsm: wording here is iffy.

@Lupie, suggestions?

alsamylkin: can make it `it`

config/config.go
```diff
@@ -123,36 +122,18 @@ func (l *Loader) ResolvePath(relative string) (string, error) {
...
-			files = append(files, filepath.Join(dir, fmt.Sprintf("%s.yaml", baseFile)))
-		}
-	}
-
-	return files
-}
-
 func (l *Loader) baseFiles() []string {
-	return []string{_baseFile, l.Environment(), _secretsFile}
+	return []string{_baseFile + ".yaml", l.Environment() + ".yaml", _secretsFile + ".yaml"}
```
anuptalwalkar: You can also update the consts as base.yaml, secrets.yaml, etc.

CHANGELOG.md
glibsm: - `Loader.Path() string` is now `Loader.Paths() []string`, to better reflect that configuration is loaded from multiple directories.

config/config.go
```diff
@@ -165,13 +146,13 @@ func (l *Loader) Environment() string {
 	return _devEnv
 }
 
-// Path returns path to the yaml configurations
-func (l *Loader) Path() string {
+// Paths returns paths to the yaml configurations
+func (l *Loader) Paths() []string {
 	if path, ok := l.lookUp(l.EnvironmentPrefix() + _configDir); ok {
-		return path
+		return []string{path}
```
anuptalwalkar: When will this be multiple strings?

alsamylkin: If ennvironment variable is set, then never, but by default it is multiple strings: we look in `.` and `./config`

PR #474, opened 2017-04-25
Add development logging config to sample apps

sectioneight:
This uses a much friendlier output during development


sectioneight:
image

sectioneight:
Well, we don't run the sample apps during CI, so I think it actually would show up in travis we just never exercise these

PR #476, opened 2017-04-26
Add start and stop timeout options

alsamylkin:
We have a fixed timeout for starting modules and nothing for stopping, that can be a hindrance, for e.g. a developer machine.

Moreover there are couple bugs:
* We are trying to stop modules, that are not running
* Ignore errors from stoping modules


service/manager.go
```diff
@@ -426,24 +430,28 @@ func (m *manager) startModules() []error {
...
+				return
 			}
-			wg.Done()
+			errC <- nil
 		}(mod)
+
+		select {
+		case err := <-errC:
+			if err != nil {
+				results = append(results,
```
madhuravi: Looks like we still need lock synchronization here.

service/manager.go
```diff
@@ -426,24 +430,28 @@ func (m *manager) startModules() []error {
...
+					fmt.Errorf("module %q stopped with error %q", mod.Name(), err))
+			}
+		case <-time.After(m.stopTimeout):
+			results = append(results,
```
madhuravi: and here

service/manager.go
```diff
@@ -265,8 +270,7 @@ func (m *manager) shutdown(err error, reason string, exitCode *int) (bool, error
 	}
 
 	m.transitionState(Stopped)
-
-	return true, err
+	return true, multierr.Combine(err, multierr.Combine(errs...))
```
madhuravi: Is it worth taking this dependency? Looks like multierr is in "beta".

service/options.go
```diff
@@ -63,3 +65,19 @@ func WithObserver(observer Observer) Option {
 		return nil
 	}
 }
+
+// WithStartTimeout configures a manager with a timeout to start modules
```
madhuravi: Config seems a better fit for specifying timeouts.

service/manager.go
alsamylkin: Why? Select chooses one channel to receive and we append results slice only once per iteration.

service/manager.go
alsamylkin: yarpc is using it: https://github.com/yarpc/yarpc-go/blob/dev/glide.yaml#L66
and they are on 1.0

service/manager.go
madhuravi: You're right, I thought this was inside the goroutine.

config/decoder.go
```diff
@@ -401,7 +401,7 @@ func (d *decoder) valueStruct(key string, target interface{}) error {
 		field := targetType.Field(i)
 
 		// Check for the private field
-		if field.PkgPath != "" && !field.Anonymous {
+		if field.PkgPath != "" || field.Anonymous {
```
madhuravi: related?

service/options.go
alsamylkin: Make sense, moved to a config and discovered a nice bug :)

service/manager.go
```diff
@@ -45,6 +42,10 @@ type ExitCallback func(shutdown Exit) int
 // Implements Manager interface
 type manager struct {
 	serviceCore
+
+	StartTimeout time.Duration `yaml:"start_timeout" default:"10s"`
```
madhuravi: we seem to be using a mix of underscores and camel case in yaml files. we should settle on one to be consistent.

alsamylkin: We never use camel case in yaml, where did you see it?

service/manager.go
```diff
@@ -337,19 +341,24 @@ func (m *manager) start() Control {
...
+			}
+
+			zap.L().Error("Error starting the module", zap.Error(e))
+			// return first service error
+			if serviceErr == nil {
+				serviceErr = e
+			}
+
 			return Control{
 				ExitChan:     errChan,
```
madhuravi: are these channels closed somewhere?

config/decoder.go
alsamylkin: yes, it is a bug in config, that was uncovered by populating manager struct that has a private embedded serviceCore..
It is really hard to test it outside the config package :)

service/manager.go
alsamylkin: This object is returned to a user to get errors and close channels.

service/manager.go
```diff
@@ -85,6 +86,10 @@ func newManager(builder *Builder) (Manager, error) {
 		return nil, err
 	}
 
+	if err := m.configProvider.Get(config.Root).Populate(m); err != nil {
```
madhuravi: is root the right place? can there be two fx services in one codebase?

service/manager.go
madhuravi: Check highgarden config, I see a few. Also with the cherami module, there is deploymentCluster (I originally had a yaml mapping for it but removed based on your code review comment)

service/manager.go
alsamylkin: builder produces only one service, so there could be multiple services, but they need to be created differently.

service/manager.go
alsamylkin: Thank you! Lets be consistent then, I'll keep then camel case then.

service/manager.go
madhuravi: they would all have to share the timeout then? that's ok for the time being if that's not a popular use case.

alsamylkin: No, if a user wants to have multiple services, they have to have a custom config and provide separate configs with an option: we can't have multiple services returned by a builder, because we have `name`, `owner`, `modules` etc. under `config.Root` already.

PR #477, opened 2017-04-28
Add dotted keys support to yaml provider

alsamylkin:
We split the path to a node based on dots, which makes impossible to work with e.g. maps with keys containing dots.

In this change we recursively try to find the longest key that matches dotted path to a node. For example, if we want to find a key `a.b` in a `map["a":map["b":"c"], "a.b":"d"]` we want to return `d` instead of `c`.

### DIFF
--- a/config/yaml.go
+++ b/config/yaml.go
@@ -214,33 +214,24 @@ func (n yamlNode) Type() reflect.Type {
 	return reflect.TypeOf(n.value)
 }
 
+// Find the first longest match in child nodes for the dottedPath.
 func (n *yamlNode) Find(dottedPath string) *yamlNode {
-	node := n
-	parts := strings.Split(dottedPath, ".")
+	for curr := dottedPath; len(curr) != 0; {
+		for _, v := range n.Children() {
+			if strings.EqualFold(v.key, curr) {
+				if curr == dottedPath {
+					return v
+				}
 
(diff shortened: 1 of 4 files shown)


config/yaml.go
```diff
@@ -214,33 +214,24 @@ func (n yamlNode) Type() reflect.Type {
...
-				break
+				if node := v.Find(dottedPath[len(curr)+1:]); node != nil {
+					return node
+				}
 			}
 		}
 
-		if !found {
+		if last := strings.LastIndex(curr, _separator); last > 0 {
+			curr = curr[0:last]
```
anuptalwalkar: you can just do curr[:last]

config/yaml_test.go
```diff
@@ -889,3 +889,129 @@ x:
...
+	type a struct {
+		B b
+	}
+
+	bytes := []byte(`
+a:
+  b:
+    s: Mozart
+    i: 1756
+a.b.i: 1791
```
anuptalwalkar: is this recursive? What about 
```
a:
  b.i: 1756
a.b.i: 1234
a.b:
  i: 1234```

alsamylkin: Yes, it is recursive

alsamylkin:
Ping..

PR #478, opened 2017-04-29
Create fx service and module using dig to share components

madhuravi:
A new service entrypoint (parallel to service.Manager) and new Module interface in package fx, that refactors module logic using dig to share components.


madhuravi:
@uber-go/service-framework

fx.go
```diff
@@ -18,4 +18,150 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-package fx // import "go.uber.org/fx"
```
sectioneight: You lost the ``// import "go.uber.org/fx"` comment which is important to keep for canonical import paths

fx.go
```diff
@@ -18,4 +18,150 @@
...
+func New(modules ...Module) *Service {
+	s := &Service{
+		g:       dig.New(),
+		modules: modules,
+	}
+
+	// load config for module creation and include it in the graph
+	cfg := config.DefaultLoader.Load()
+	// TODO: need to pull latest dig for direct Interface injection fix
+	s.g.MustRegister(func() config.Provider { return cfg })
```
alsamylkin: we can simply register the `config.DefaultLoader.Load()`

fx.go
```diff
@@ -18,4 +18,150 @@
...
+	Constructor() []Component
+	Stop()
+}
+
+// Service controls the lifecycle of an fx service
+type Service struct {
+	g              *dig.Graph
+	modules        []Module
+	components     []Component
+	scopeCloser    io.Closer
```
alsamylkin: Combine closers? And close them on stop in reverse order..

fx.go
```diff
@@ -18,4 +18,150 @@
...
+			}
+		}
+		s.logger.Info("Module started", zap.String("module", m.Name()))
+	}
+	s.logger.Info("Service has started")
+}
+
+// Stop stops the service
+func (s *Service) Stop() {
+	for _, m := range s.modules {
```
alsamylkin: I'd think we need to Stop them in a reversed order, otherwise it can lead to surprises on stop/shutdowns if there are interdependencies between modules.

fx.go
```diff
@@ -19,3 +19,149 @@
...
+			s.g.MustRegister(ctor)
+		}
+	}
+
+	return s
+}
+
+func createLogger(cp config.Provider, scope tally.Scope) (*zap.Logger, func(), error) {
+	cfg := cp.Get("logging")
+	logConfig := ulog.Configuration{}
```
alsamylkin:
```golang
cfg := ulog.DefaultConfiguration()
if err := cp.Get("logging").Populate(&cfg); err != nil {
   return nil, nil, errors.Wrap(err, "failed to initialize logging from config")
}
```
?

fx.go
```diff
@@ -19,3 +19,149 @@
...
+		s.g.MustRegister(c)
+	}
+	return s
+}
+
+// Start starts the service
+func (s *Service) Start() {
+	// TODO: move to dig, perhaps #Call(constructor) function
+	for _, m := range s.modules {
+		for _, ctor := range m.Constructor() {
```
alsamylkin: add
```golang
s.logger.Info("Starting module", zap.String("module", m.Name()))
```
?

fx.go
```diff
@@ -19,3 +19,149 @@
...
+	"go.uber.org/zap"
+)
+
+// Component is a function that initializes and returns objects to be shared across the code
+type Component interface{}
+
+// Module is a building block or primary function performed by a service built on fx
+type Module interface {
+	Name() string
+	Constructor() []Component
```
glibsm: Just remember a piece of feedback I forgot to include earlier.

I think it's critical for `Constructor()` to take in `config.Provider` because it will allow modules to consider config before returning components. For example, maybe module can always return components A and B, but an optional C needs to be enabled through config.

alsamylkin: The object implementing Module can take it in it's own constructor, otherwise it can lead to a confusion with a config that is passed to Components.

fx.go
madhuravi: right now, we only have one closer of type io.Closer. I intentionally punted this to when we have more than one so we can make a decision based on that.

fx.go
madhuravi: fx is using an older version of dig that doesnt work with interfaces very well. See https://github.com/uber-go/dig/issues/27. There is a TODO already to change this once latest version of dig is pulled in.

fx.go
madhuravi: Good point, makes sense.

fx.go
madhuravi: What Alex said. Also this contradicts with what you said earlier Glib about Constructor taking no args primarily so modules can use their own config/logging/metrics implementation if they choose to.

fx.go
```diff
@@ -19,3 +19,149 @@
...
+}
+
+// Service controls the lifecycle of an fx service
+type Service struct {
+	g              *dig.Graph
+	modules        []Module
+	components     []Component
+	scopeCloser    io.Closer
+	logger         *zap.Logger
+	loggerCloserFn func()
```
anuptalwalkar: Why do we need this?

fx.go
```diff
@@ -19,3 +19,149 @@
...
+		g:       dig.New(),
+		modules: modules,
+	}
+
+	// load config for module creation and include it in the graph
+	cfg := config.DefaultLoader.Load()
+	// TODO: need to pull latest dig for direct Interface injection fix
+	s.g.MustRegister(func() config.Provider { return cfg })
+
+	scope, _, scopeCloser := metrics.RootScope(&scopeInit{config: cfg})
```
anuptalwalkar: you should register statsReporter as well.

fx.go
```diff
@@ -19,3 +19,149 @@
...
 package fx // import "go.uber.org/fx"
+
+import (
+	"io"
+	"reflect"
+
+	"github.com/pkg/errors"
+	"github.com/uber-go/tally"
+
+	"go.uber.org/dig"
```
anuptalwalkar: imports ordering

fx.go
anuptalwalkar: I agree that constructor should take no arguments. This can free up users to write modules freely without requiring to import config, if they don't want to use it.

fx.go
```diff
@@ -19,3 +19,149 @@
...
+	Name() string
+	Constructor() []Component
+	Stop()
+}
+
+// Service controls the lifecycle of an fx service
+type Service struct {
+	g              *dig.Graph
+	modules        []Module
+	components     []Component
```
anuptalwalkar: We don't need components. We are registering them into dig within `WithComponents` call.

fx.go
glibsm: That's not true. Taking configuration as part of the component means that a module can not control the *number* of modules provided based on the configuration.

For example, consider that yarpc module would automatically expose all the clients in the graph:

```
modules:
  yarpc:
    outbounds:
    - foo
    - bar
```

We'd want to expose `foo` and `bar` in the graph based on the configuration provided to the initial module and dynamically expose two components.

Otherwise, how would a module react to the configuration and dynamically expose various numbers of components?

glibsm: If someone is able to use the above example of a configuration and show how modules can expose a various number of strongly typed components, then it's ok that `Constructor` takes no arguments.

> Also this contradicts with what you said earlier Glib about Constructor taking no args

Yes, I know that I've previously argued for it taking no arguments, but the configuration is very special, even more so than the tracing, logging and metrics.

madhuravi: Can you write some pseudocode here for the case you're describing?

fx.go
madhuravi: zap logger returns this as a closer. Right now it unsets the globals we set. This is not strictly required since at service stop, we don't care about this anymore. But it's just good practice to call the closer fn. It can be extended in future for instance and we will be covering our bases this way.

fx.go
madhuravi: It's not being used right now. I'd prefer to add it when we are using it.

fx.go
madhuravi: Damn I never get imports right 😃

fx.go
madhuravi: We do a HasValue also. Updated with a version that incorporates that.

fx.go
glibsm:
```
func (yarpc) Constructors(cfg config.Provider) []fx.Component {
    components := []fx.Component{}
    if cfg.outbounds {
        for outbound in outbounds {
            append(components, make(outbound))
        }
    }
    if cfg.inbounds {
        append(components, dispatcher)
    }
    return components
}
```

How else can this be achieved?

anuptalwalkar: Wait. We don't support having multiple `make(outbound)` right now without dispatcher being available, but we can totally do this -  

```go
func (yarpc) Constructors() []fx.Component {
    return []fx.Component {
    func(cfg config.Provider) *outbound1, *outbound2 {
        if cfg.outbounds {
                // make outbounds
                return outbound1, outbound2
            }
        }  
    }, 
    func(cfg config.Provider) *yarpc.Dispatcher {
        if cfg.inbounds {
            return dispatcher;
        }
    }
    }
}
```

madhuravi: The drawback with what you're describing Anup is we don't know at compile time how many outbound objects should be returned.

anuptalwalkar: I wrote two, but it can be `n` outbound objects :)
soon/later once dig will start supporting slices, we can make the func return `[]outbounds`

madhuravi: To Glib's question:

Right now, outbounds are very specific to the client i.e. you need the outbound specific import and the type is also specific to the outbound client so the use case doesn't exist yet.
It is possible that in future there is a generic way to create outbounds.

If so, there is nothing that limits this to config. It could be any other dependency as well or the module could have their own config library based on which they want to do this. If we really want to support this, we should talk about sending the dig graph in to the constructor or to the module construction for more flexibility.

glibsm: I'm less talking about the specifics of YARPC or outbounds in general, but rather illustrating how the configuration can impact the number of components that the module has to return. I'm ok to YAGNI and punt on this for now, but I feel pretty strongly that configuration is a special thing that should be used as part of the module creation process, rather than a dependency of a component.

alsamylkin: This can be easily achieved with an option: `WithConfigurableModule` that takes a module ctor with a config. It is true, that our control flow is a bit confusing right now: we make user to create modules for our service, instead of e.g. giving us a module ctor.

madhuravi:
ping @glibsm

PR #479, opened 2017-05-01
Merge dev to master

alsamylkin:
Master is protected and I can't push the merge without review.


anuptalwalkar:
lgtm!

glibsm:
Are we doing a release or something? Tip of master should contain the
latest release.

alsamylkin:
Master "is 21 commits behind dev". Not a release, just a long time no changes.

glibsm:
This defeats part of the reason to have master. It means the tip no longer
points to a stable release, but rather an arbitrary point in the
development history.

alsamylkin:
It is stable right now, but with the new module interface changes it will become unstable, so I simply want to fix that point first.

alsamylkin:
No urge to do this now.

PR #480, opened 2017-05-02
Update dig to version 0.3

anuptalwalkar:
1. Deleted YARPC module which is using dig 0.1 on package level APIs
2. Added separate fx_components.go with necessary components registered in the service. (runtime metrics collector, version metrics, and tracing)
3. Updated fx.go with `dig.Invoke` for simplified initialization


fx.go
```diff
@@ -46,104 +41,86 @@ type Module interface {
 
 // Service controls the lifecycle of an fx service
 type Service struct {
-	g              *dig.Graph
+	c              *dig.Container
 	modules        []Module
-	scopeCloser    io.Closer
-	logger         *zap.Logger
+	closers        []io.Closer
```
alsamylkin: Thank you!

examples/keyvalue/server/main.go
```diff
@@ -23,15 +23,11 @@ package main
...
-	"go.uber.org/fx/modules/yarpc"
 	"go.uber.org/fx/service"
 )
 
 func main() {
-	svc, err := service.WithModule(
-		yarpc.New(yarpc.ServiceCreateFunc(NewYARPCThriftHandler)),
-		service.WithRole("service"),
-	).WithOptions(
+	svc, err := service.WithModule(nil).WithOptions(
```
glibsm: lets just delete this example entirely, since it no longer makes sense

fx.go
```diff
@@ -46,104 +41,86 @@ type Module interface {
...
-	// TODO: move to dig, perhaps #Call(constructor) function
-	for _, m := range s.modules {
-		s.logger.Info("Starting module", zap.String("module", m.Name()))
-		for _, ctor := range m.Constructor() {
-			ctype := reflect.TypeOf(ctor)
-			switch ctype.Kind() {
-			case reflect.Func:
-				objType := ctype.Out(0)
-				s.g.MustResolve(reflect.New(objType).Interface())
+	if err := s.c.Invoke(func(l *zap.Logger) {
```
glibsm: woah!!!!!!!!!

metrics/metrics_test.go
```diff
@@ -93,8 +93,12 @@ func getScope() (tally.Scope, tally.CachedStatsReporter, io.Closer) {
 	return RootScope(scopeInit())
 }
 
-func scopeInit() (string, config.Provider) {
-	return "SomeName", config.NewStaticProvider(map[string]interface{}{"foo": "bar"})
+func scopeInit() config.Provider {
+	return config.NewStaticProvider(
+		map[string]interface{}{
+			config.ServiceNameKey: "somename",
+			"foo": "bar"},
```
glibsm: nit: `}` on new line

version_metrics.go
```diff
@@ -0,0 +1,60 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package fx // import "go.uber.org/fx"
+
+import (
+	"runtime"
+	"time"
+
+	"github.com/uber-go/tally"
```
glibsm: not sure we want to marry fx package to tally, right?

fx_components.go
```diff
@@ -0,0 +1,94 @@
...
+	if err := cfg.Get("metrics.runtime").Populate(&runtimeMetricsConfig); err != nil {
+		return nil, errors.Wrap(err, "unable to load runtime metrics configuration")
+	}
+	runtimeCollector := metrics.StartCollectingRuntimeMetrics(
+		scope.SubScope("runtime"), time.Second, runtimeMetricsConfig,
+	)
+	return runtimeCollector, nil
+}
+
+func setupVersionMetricsEmitter(scope tally.Scope) *versionMetricsEmitter {
```
glibsm: imo should be a separate comonent that we pull in by default for the service scaffold, but this doesn't belong directly in fx

examples/keyvalue/server/main.go
anuptalwalkar: I wanted to keep it just in case we add a new yarpc module, which Alex is planning, we don't have to rewrite bunch examples again.

glibsm: Yeah but we have version contol. We can easily bring it back, lets not leave the stuff around in case we might want to use it sometime soon.

It's the same reason we don't comment out a whole file but leave it checked in, right?

version_metrics.go
anuptalwalkar: thought I will introduce version metrics here so service has it by default. I'll remove now.

alsamylkin: Lets call it fally!:)

glibsm: I'm probably conflating a lot of the things together in my feedback right now, so it's probably ok to merge as is, especially as this PR focuses on updating dig.

Everything should become a lot clearer once we've been able to sync.

fx_components.go
glibsm: Same as above comment. If this hasn't changed, just moved to a new file lets leave it and merge the dig change. We can interate from there

version_metrics.go
anuptalwalkar: oh why not? This is a good way to introduce runtime metrics collector?

anuptalwalkar: tally is alredy a dependency in fx.go this simply moves it to fx_components.

fx_components.go
anuptalwalkar: I thought we want it enabled by default regardless of where its being added. Its good for tracking fx versions?

anuptalwalkar:
Merging for further iterations. I suppose we will follow up on deleting lot of code from here:)

PR #481, opened 2017-05-04
Move uhttp module to dig

alsamylkin:
Refactor uhttp module to implement to fx.Module interface:
* Module constructor accepts a component that should return an http handler.
* removed service.Host from everywhere in modules/uhttp.
* updated docs to show how new handler constructor can accept a config.
* added more tests.

### DIFF
--- a/modules/uhttp/http.go
+++ b/modules/uhttp/http.go
@@ -71,13 +71,15 @@ func New(handlerCtor fx.Component) *Module {
 	return &Module{handlerCtor: handlerCtor}
 }
 
+// Name returns uhttp.
 func (m *Module) Name() string {
 	return "uhttp"
 }
 
 type starter struct{}
 
-func (m *Module) Constructors() []fx.Component {
+// Constructor returns module components: handler constructor and harness for it.
+func (m *Module) Constructor() []fx.Component {
 	return []fx.Component{
 		m.handlerCtor,
 		func(provider config.Provider, l *zap.Logger, scope tally.Scope, handler http.Handler) (*starter, error) {
--- a/modules/uhttp/http.go
+++ b/modules/uhttp/http.go
@@ -29,31 +29,17 @@ import (
 	"sync"
 	"time"
 
+	"go.uber.org/fx"
 	"go.uber.org/fx/auth"
-	"go.uber.org/fx/service"
+	"go.uber.org/fx/config"
 	"go.uber.org/fx/ulog"
 
 	"github.com/pkg/errors"
+	"github.com/uber-go/tally"
 	"go.uber.org/zap"
 )
 
 const (
-	// ContentType is the header key that contains the body type
-	ContentType = "Content-Type"
-	// ContentLength is the length of the HTTP body
-	ContentLength = "Content-Length"
-	// ContentTypeText is the plain content type
-	ContentTypeText = "text/plain"
-	// ContentTypeJSON is the JSON content type
-	ContentTypeJSON = "application/json"
-
-	// HTTP defaults
-	defaultTimeout = 60 * time.Second
-	defaultPort    = 3001
-
-	// Reporter timeout for tracking HTTP requests
-	defaultReportTimeout = 90 * time.Second
-
 	// default healthcheck endpoint
 	healthPath = "/health"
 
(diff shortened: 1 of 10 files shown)


alsamylkin:
Do I need to update the changelog?:)

modules/uhttp/http.go
```diff
@@ -29,122 +29,111 @@ import (
...
-func newModule(host service.Host, handler http.Handler) (*Module, error) {
-	// setup config defaults
-	cfg := Config{
-		Port:    defaultPort,
-		Timeout: defaultTimeout,
-	}
-	log := ulog.Logger(context.Background()).With(zap.String("module", host.ModuleName()))
-	if err := host.Config().Get("modules").Get(host.ModuleName()).Populate(&cfg); err != nil {
-		log.Error("Error loading http module configuration", zap.Error(err))
+func New(handlerCtor fx.Component) *Module {
```
alsamylkin: I would really prefer to have it as an abstract function, that returns modules, we can pass configs/loggers/etc directly to modules then.

modules/uhttp/http.go
```diff
@@ -29,122 +29,111 @@ import (
...
-	}, nil
+// Name returns uhttp.
+func (m *Module) Name() string {
+	return "uhttp"
 }
 
-// Start begins serving requests over HTTP
-func (m *Module) Start() error {
-	return nil
+type starter struct{}
```
anuptalwalkar: Why do you need empty starter struct? This is going to be a module construct, you can just return error.

modules/uhttp/http.go
```diff
@@ -29,122 +29,111 @@ import (
...
+
+// Constructor returns module components: handler constructor and harness for it.
+func (m *Module) Constructor() []fx.Component {
+	return []fx.Component{
+		m.handlerCtor,
+		func(provider config.Provider, l *zap.Logger, scope tally.Scope, handler http.Handler) (*starter, error) {
+			// setup config defaults
+			cfg := Config{}
+
+			m.l = ulog.Logger(context.Background()).With(zap.String("module", m.Name()))
```
anuptalwalkar: `m.l = l.With(zap.String("module", m.Name())` ?

modules/uhttp/http_test.go
```diff
@@ -79,48 +91,81 @@ func TestHTTPModule_Tracer(t *testing.T) {
...
-	)
-	require.NoError(t, err)
-	mod, err := newModule(host, handler)
-	if expectError {
-		require.Error(t, err, "Expected error instantiating module")
-		fn(nil)
-		return
+	di := dig.New()
+	p := config.NewYAMLProviderFromBytes(_httpConfig)
+	di.MustRegister(&p)
```
anuptalwalkar: you can use `di.ProvideAll`.. which returns an err

modules/uhttp/http_test.go
```diff
@@ -79,48 +91,81 @@ func TestHTTPModule_Tracer(t *testing.T) {
...
+	mod := New(&handler)
+	ctors := mod.Constructor()
+	for i := range ctors {
+		di.MustRegister(ctors[i])
 	}
-	require.NoError(t, err, "Unable to instantiate module")
-	assert.NoError(t, mod.Start(), "Got error from starting")
+
+	var s *starter
+	di.MustResolve(&s)
```
anuptalwalkar: `Resolve` and check for error

modules/uhttp/http_test.go
```diff
@@ -79,48 +91,81 @@ func TestHTTPModule_Tracer(t *testing.T) {
...
+	di := dig.New()
+	p := config.NewYAMLProviderFromBytes(_httpConfig)
+	di.MustRegister(&p)
+	di.MustRegister(&tally.NoopScope)
+	di.MustRegister(zap.NewNop())
+
+	mod := New(&handler)
+	ctors := mod.Constructor()
+	for i := range ctors {
+		di.MustRegister(ctors[i])
```
anuptalwalkar: `err := di.Invoke(ctors[i])` and you don't need to resolve starter below. :)

modules/uhttp/http.go
alsamylkin: Oh yes, Service, just calls Invoke()

PR #482, opened 2017-05-05
[WIP] Bring YARPC back

alsamylkin:
This PR is how YARPC module could look like with the DI refactor.
Caveat: we do need the new YARPC config, otherwise we'll need to implement converters
from map[string]interface{} to YARPC types ourselves.

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -7,7 +7,7 @@ import:
 - package: github.com/gorilla/context
   version: ^1.1.0
 - package: go.uber.org/yarpc
-  version: ^v1.4.0
+  version: config
 - package: go.uber.org/dig
   version: ~0.3
 - package: go.uber.org/thriftrw
--- a/modules/yarpc/yarpc.go
+++ b/modules/yarpc/yarpc.go
@@ -2,17 +2,22 @@ package yarpc
 
 import (
 	"go.uber.org/fx"
-	"go.uber.org/yarpc"
-	"go.uber.org/zap"
-	"go.uber.org/yarpc/api/transport"
 	"go.uber.org/fx/config"
+
+	"go.uber.org/yarpc"
+	"go.uber.org/yarpc/api/transport"
+	"go.uber.org/yarpc/transport/http"
+	"go.uber.org/yarpc/transport/tchannel"
+	yconfig "go.uber.org/yarpc/x/config"
+
 	"github.com/uber-go/tally"
+	"go.uber.org/zap"
 )
 
 type Module struct {
-	l *zap.Logger
+	l           *zap.Logger
 	handlerCtor fx.Component
-	d *yarpc.Dispatcher
+	d           *yarpc.Dispatcher
 }
 
 type Transports struct {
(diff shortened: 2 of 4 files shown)


alsamylkin:
I know, that it is maybe a throwaway PR, but it does look like a very nice and simple module.

glide.yaml
```diff
@@ -7,7 +7,7 @@ import:
 - package: github.com/gorilla/context
   version: ^1.1.0
 - package: go.uber.org/yarpc
-  version: ^v1.4.0
+  version: config
```
glibsm: is this PR considered blocked until that branch merges?

modules/yarpc/yarpc.go
```diff
@@ -0,0 +1,111 @@
...
+
+// Module represents a collection of YARPC components.
+type Module struct {
+	l           *zap.Logger
+	handlerCtor fx.Component
+	d           *yarpc.Dispatcher
+}
+
+// Transports represent a collection Procedures that will be registered with a dispatcher.
+type Transports struct {
```
glibsm: what about middleware?

modules/yarpc/yarpc_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+		return &Transports{Ts: proc}, nil
+	}
+
+	module := New(fn)
+	for _, component := range module.Constructor() {
+		di.MustRegister(component)
+	}
+
+	cfg := map[string]interface{}{
```
glibsm: i find your preference for this static syntax vs yaml baffling

modules/yarpc/yarpc_test.go
```diff
@@ -0,0 +1,124 @@
...
+		},
+	}
+
+	provider := config.NewStaticProvider(cfg)
+	di.MustRegister(&provider)
+	tracer := opentracing.Tracer(opentracing.NoopTracer{})
+	di.MustRegister(&tracer)
+	di.MustRegister(&metrics.NopScope)
+	logger := zap.NewNop()
+	di.MustRegister(logger)
```
glibsm: I thought our general guidance was "dig does not appear in tests".
You're still able to test the functions if you memorized them on a module, then you can call them directly after obtaining test/mock objects of tracer, logger, etc

glide.yaml
alsamylkin: yes, unless we want out own config parsing, which we should avoid.

modules/yarpc/yarpc.go
alsamylkin: We can have a constructor for it, not in this PR, but I don't see any problem adding it.
Middleware could also be populated from the yarpc's `Configurator.LoadConfig`.

modules/yarpc/yarpc_test.go
alsamylkin: I do like types and it is hard to see them in yaml. It also breaks my indentation if I am going to use ``.

modules/yarpc/yarpc_test.go
alsamylkin: Well, it depends what tests are we talking. I think I can replace dig here with a service, but it seems for module tests(not a user level functionality) we'll need dig sometimes.

glibsm: Right, so the service tests definitely need to use dig. However, the modules might not necessarily need to.

Imagine that all the constructor functions were either named or memorized. Then, you can test them individually just by calling them direclty, without the need to involve dig, or a service.

glibsm:
The yarpc module will make it's way to a new repo

PR #484, opened 2017-05-11
Remove service.Host

madhuravi:
Removed Host interface. Pass in explicit config/tracer/metrics etc. where required.
Certain constructors such as task backend implementations have a TODO to be fixed after the module refactor.


modules/task/execution_test.go
```diff
@@ -197,13 +193,16 @@ func TestEnqueueSimpleFn(t *testing.T) {
 	require.Error(t, err)
 	assert.Contains(t, err.Error(), "Simple error")
 
-	snapshot := _testScope.(tally.TestScope).Snapshot()
-	timers := snapshot.Timers()
-	counters := snapshot.Counters()
+	// TODO(madhu): Add back after module refactor
```
alsamylkin: why it doesn't work anymore?

madhuravi: Will address this in a separate PR, coming soon.

PR #485, opened 2017-05-12
Bring back stats to the task module

madhuravi:
This was deleted with the earlier commit to remove service.Host. Adding back now with passing in a metrics scope to enable task module stats.

### DIFF
--- a/service/module.go
+++ b/service/module.go
@@ -23,6 +23,8 @@ package service
 import (
 	"fmt"
 	"sync"
+
+	"github.com/uber-go/tally"
 )
 
 // ModuleProvider provides Modules.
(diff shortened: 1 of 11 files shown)


service/module.go
```diff
@@ -32,7 +34,7 @@ type ModuleProvider interface {
 	// Create a new Module. The name of the Host and the scoping
 	// of associated functions on the Host will be done using a name
 	// provided by a ModuleOption, or by the DefaultName on this ModuleProvider.
-	Create() (Module, error)
+	Create(tally.Scope) (Module, error)
```
alsamylkin: This requires a changelog as well..

PR #486, opened 2017-05-12
Expand environment variables only in static yaml files

alsamylkin:
YAML provider replaces all the strings in form ${var} $var everywhere, even in comments and there is no way to opt out of it.

Instead of a blank substitution we introduce separate constructors for interpolation. `WithExpand` constructors will iterate over all the yaml nodes and replace only values, that need to be interpolated.

Loader is updated to reflect this changes, it has a staticFiles field with a corresponding `Set` method to set files that will not be interpolated, remaining base file will be expanded with the `os.Lookup` function.


config/config.go
```diff
@@ -130,7 +133,7 @@ func (l *Loader) ResolvePath(relative string) (string, error) {
 }
 
 func (l *Loader) baseFiles() []string {
-	return []string{_baseFile, l.Environment() + ".yaml", _secretsFile}
+	return []string{_baseFile, l.Environment() + ".yaml"}
```
madhuravi: Add comment that the order matters.

config/config.go
```diff
@@ -140,7 +143,9 @@ func (l *Loader) getResolver() FileResolver {
 // YamlProvider returns function to create Yaml based configuration provider
 func (l *Loader) YamlProvider() ProviderFunc {
 	return func() (Provider, error) {
-		return NewYAMLProviderFromFiles(false, l.getResolver(), l.getFiles()...), nil
+		static := NewYAMLProviderFromFiles(false, l.getResolver(), l.getStaticFiles()...)
+		interpolated := NewYAMLProviderWithExpand(false, l.getResolver(), os.LookupEnv, l.getFiles()...)
+		return NewProviderGroup("yaml", interpolated, static), nil
```
madhuravi: Add comment that order matters.

config/config.go
```diff
@@ -163,19 +168,28 @@ func (l *Loader) Paths() []string {
...
 	l.lock.Lock()
 	defer l.lock.Unlock()
 
 	l.configFiles = files
 }
 
+// SetStaticConfigFiles overrides the set of available config files for the service.
+// Config values will not be interpolated.
+func (l *Loader) SetStaticConfigFiles(files ...string) {
+	l.lock.Lock()
```
madhuravi: if you embed sync.RWMutex, you can just call Lock on loader directly. So l.Lock() instead of l.lock.Lock()

config/yaml.go
```diff
@@ -273,36 +306,34 @@ func (n *yamlNode) Children() []*yamlNode {
 	return nodes
 }
 
-func unmarshalYAMLValue(reader io.ReadCloser, value interface{}, lookUp lookUpFunc) error {
-	raw, err := ioutil.ReadAll(reader)
-	if err != nil {
-		return errors.Wrap(err, "failed to read the yaml config")
+func (n *yamlNode) dfs(expand func(string) string) {
```
madhuravi: better to name it something like applyOnAllNodes or expandNodeValues etc. its cool that it uses dfs but does not add much to the readability.

config/config.go
alsamylkin: Then Lock() will become a public method for Loader ;)

madhuravi: good point.

config/config.go
```diff
@@ -140,7 +144,11 @@ func (l *Loader) getResolver() FileResolver {
 // YamlProvider returns function to create Yaml based configuration provider
 func (l *Loader) YamlProvider() ProviderFunc {
 	return func() (Provider, error) {
-		return NewYAMLProviderFromFiles(false, l.getResolver(), l.getFiles()...), nil
+		static := NewYAMLProviderFromFiles(false, l.getResolver(), l.getStaticFiles()...)
+		interpolated := NewYAMLProviderWithExpand(false, l.getResolver(), os.LookupEnv, l.getFiles()...)
```
madhuravi: nit: prefer consistent words i.e. either interpolate or expand. Rename to NewYAMLProviderWithInterpolate.

config/static_provider.go
```diff
@@ -29,12 +34,17 @@ type staticProvider struct {
...
-	if err != nil {
-		panic(err)
+	return staticProvider{
+		Provider: NewYAMLProviderFromReader(toReadCloser(data)),
 	}
+}
 
-	return staticProvider{Provider: NewYAMLProviderFromBytes(b)}
+// NewStaticProviderWithExpand returns a static provider with values replaced by a mapping function.
+func NewStaticProviderWithExpand(data interface{}, mapping func(string) (string, bool)) Provider {
```
madhuravi: Same, use one - interpolate or expand.

config/config.go
```diff
@@ -140,7 +144,11 @@ func (l *Loader) getResolver() FileResolver {
...
+
+		// Interpolated files will have higher priority than static.
+		return NewProviderGroup("yaml", static, interpolated), nil
```
glibsm: doesn't this mean that secrets will no longer override base/prod?

config/yaml.go
```diff
@@ -314,14 +334,9 @@ func unmarshalYAMLValue(reader io.ReadCloser, value interface{}, lookUp lookUpFu
 // will be used
 //
 // TODO: what if someone wanted a literal ${FOO} in config? need a small escape hatch
```
glibsm: thoughts on this?

config/config.go
alsamylkin: os.Expand is a a function that we use to do replacements..

config/config.go
alsamylkin: Yes and the reason is that if you want something to be dynamic, it should have a higher priority(same as dynamic providers have an upper hand over static ones)

config/config.go
alsamylkin: I am trying to understand, what is the confusion? we don't have any interpolate provider anywhere..

config/yaml.go
alsamylkin: Move it to a different file :) It is hard to achieve this with expand, but sometimes you don't have an access to edit a file(e.g. if a config file is not controlled by you), so using a loader for this task seems preferable.

config/config.go
```diff
@@ -140,7 +144,11 @@ func (l *Loader) getResolver() FileResolver {
 // YamlProvider returns function to create Yaml based configuration provider
 func (l *Loader) YamlProvider() ProviderFunc {
 	return func() (Provider, error) {
-		return NewYAMLProviderFromFiles(false, l.getResolver(), l.getFiles()...), nil
+		static := NewYAMLProviderFromFiles(false, l.getResolver(), l.getStaticFiles()...)
+		expanded := NewYAMLProviderWithExpand(false, l.getResolver(), os.LookupEnv, l.getFiles()...)
+
+		// Expanded files will have higher priority than static.
+		return NewProviderGroup("yaml", static, expanded), nil
```
glibsm: Continuing the conversation - this will not work.

Secrets must be able to override keys from other files, that's what they are there for...

alsamylkin: Ok, lets break the contract later.

config/yaml.go
glibsm: At least after this PR, that syntax is only applicable in the values, right? I think most of the time "unwanted" expands of it are going to occur in comments for people explaining exactly what the `${FOO:BAR}` syntax means

alsamylkin: Thanks, I've forgot to update the changelog.
Yes, syntax is applicable only in values, in order to escape these values, we'll need to use something other than os.Expand.

PR #487, opened 2017-05-12
Introduce an expand config provider

alsamylkin:
Sometimes it is hard to expand variables in a provider constructor. 
ExpandConfig provider can be used as a wrapper to lazy interpolate
values from any configuration provider.

### DIFF
--- a/config/expand_provider.go
+++ b/config/expand_provider.go
@@ -1,0 +1,80 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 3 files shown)


config/expand_provider.go
```diff
@@ -0,0 +1,80 @@
...
+	p       Provider
+	mapping func(string) string
+}
+
+// NewExpandProvider returns a config provider that uses
+// a mapping function to expand ${var} or $var values in
+// the values returned by the provider p.
+// Provider uses a lazy substitution i.e. a value will be
+// interpolated iff it was requested.
+func NewExpandProvider(p Provider, mapping func(string) string) Provider {
```
madhuravi: how is this different from YamlProviderInterpolation and StaticProviderInterpolation?
Is lazy fetch the only difference? Worth calling it that then?

config/expand_provider.go
```diff
@@ -0,0 +1,80 @@
...
+// THE SOFTWARE.
+
+package config
+
+import (
+	"fmt"
+	"os"
+)
+
+type expandProvider struct {
```
madhuravi: use name consistency from other PR. eg. interpolationProvider.

glibsm:
Do we want to replace the service init with YAML wrapped in this provider, or is that for next step?

config/expand_provider.go
alsamylkin: Well, yaml/static provider works only for yaml and static provider, this wrapper works for any provider, but with a lazy get.

alsamylkin:
We don't want to replace the yaml provider, because provider with expand can tell us a that there are wrong keys on load not when we'll fetch a value. I plan to have both, because some our teams want to use templates and this wrapper allows them to templatize providers.

config/expand_provider.go
alsamylkin: I use WithExpand everywhere for constructors, sometimes comments/private functions have interpolate in them, it is hard to use expand all the time :)

glibsm:
I know it's a bit of extra work, but lets move over these two new files into the new repo and open a PR.

alsamylkin:
Sure

PR #490, opened 2017-05-23
Extract config package out

glibsm:
Config has moves to it's new self-contained home at uber-go/config.


glide.yaml
```diff
@@ -1,25 +1,18 @@
...
   # TODO: cut v0.4 of dig, and get ready to solidify v1.0
   version: master
 testImport:
-- package: github.com/google/gofuzz
 - package: github.com/stretchr/testify
   version: ~1.1.4
   subpackages:
   - assert
   - require
-- package: go.uber.org/zap
```
alsamylkin: Why do we throw zap away with the config?

glibsm: fx does not rely or has anything to do with zap.
But you bring a good point, why does config need zap? Lets open a PR in the config repo to remove zap usage

glibsm: https://github.com/uber-go/config/pull/2

PR #491, opened 2017-05-24
Rename app Start to Run

madhuravi:
Keep it consistent with RunForever

### DIFF
--- a/app.go
+++ b/app.go
@@ -83,11 +83,12 @@ func (s *App) Provide(constructors ...interface{}) {
 	}
 }
 
-// Start the app by explicitly invoking all the user-provided constructors.
+// Run the app by explicitly invoking all the user-provided constructors.
 //
 // See dig.Invoke for moreinformation.
-func (s *App) Start(ctx context.Context, funcs ...interface{}) error {
-	return withTimeout(ctx, func() error { return s.start(funcs...) })
+func (s *App) Run(ctx context.Context, funcs ...interface{}) error {
+	log.Print("Starting the app")
+	return withTimeout(ctx, func() error { return s.run(funcs...) })
 }
 
 func withTimeout(ctx context.Context, f func() error) error {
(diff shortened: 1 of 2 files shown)


madhuravi:
@uber-go/service-framework

breerly:
As you probably already considered, it's not really an option to have `app.StartForever` to match `app.Start`. I think that it's OK that these don't match in naming, since they actually do different things:

* `App.Start` - Process startup
* `App.Stop` - Process shutdown
* `App.RunForever` - Process startup, block until SIGTERM/SIGITN, process shutdown

Totally open to naming changes, but lets make sure that start/stop map have naming symmetry and map conceptually to process startup and shutdown.

madhuravi:
I like Start and Stop having symmetry as well. It's Start and RunForever that bothers me. They do the same thing, RunForever just calls Start and then waits for shutdown :)

If you have ideas on how to name them consistently, let me know.

breerly:
Hmm... let me think about that today! I'll come back w/ a few ideas - maybe one of those will strike our interest :)

anuptalwalkar:
I liked our previous naming `Start` `StartAsync` and `Stop` where `Start` is blocking start, and `StartAsync` gives back control to main.

madhuravi:
I like Start and StartAsync as well :)

breerly:
The `*Async` naming conventions strikes me as odd in Golang... let me dig around and see if there are any prevailing conventions - we aren't the first one to write libs w/ this type of functionality.

breerly:
@madhuravi @anuptalwalkar lets jam in the meetin' tomorrow - I'll bring some notes/ideas

madhuravi:
ping @breerly Can you share your ideas doc?

breerly:
Hey @madhuravi, was OOO! Will follow up today!

madhuravi:
sounds good, thanks!

madhuravi:
reminder @breerly , want to get this closed out soon since it's one of the entry functions that is used everywhere.  Changes will be on multiple repos now :)

PR #492, opened 2017-05-25
Allow to provide objects for the App

alsamylkin:
Right now we can register only constructors and the only reason we panic is because of trying to read the function information from a pointer.

Container.Provide lets us to register any object.

### DIFF
--- a/app.go
+++ b/app.go
@@ -181,7 +181,7 @@ func funcLocation(c interface{}) (string, int) {
 }
 
 func logCaller(name string) {
-	_, file, line, ok := runtime.Caller(2)
+	_, file, line, ok := runtime.Caller(3)
 	if !ok {
 		panic("can't determine caller")
 	}
--- a/app.go
+++ b/app.go
@@ -181,10 +181,7 @@ func funcLocation(c interface{}) (string, int) {
 }
 
 func logCaller(name string) {
-	_, file, line, ok := runtime.Caller(3)
-	if !ok {
-		panic("can't determine caller")
+	if _, file, line, ok := runtime.Caller(3); ok {
+		log.Printf("%s was called %s %d\n", name, file, line)
 	}
-
-	log.Printf("%s was called %s %d\n", name, file, line)
 }
--- a/app.go
+++ b/app.go
@@ -77,9 +77,13 @@ var (
 // to all other constructors, and called lazily at startup
 func (s *App) Provide(constructors ...interface{}) {
 	for _, c := range constructors {
-		// Print the constructor signature, file and line number from where it has come
-		file, line := funcLocation(c)
-		log.Printf("Providing constructor %q from %v:%d", reflect.TypeOf(c).String(), file, line)
+		if reflect.TypeOf(c).Kind() == reflect.Func {
+			// Print the constructor signature, file and line number from where it has come
+			file, line := funcLocation(c)
+			log.Printf("Providing constructor %q from %v:%d", reflect.TypeOf(c).String(), file, line)
+		} else {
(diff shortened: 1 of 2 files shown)


app.go
```diff
@@ -77,9 +77,13 @@ var (
...
 	for _, c := range constructors {
-		// Print the constructor signature, file and line number from where it has come
-		file, line := funcLocation(c)
-		log.Printf("Providing constructor %q from %v:%d", reflect.TypeOf(c).String(), file, line)
+		if reflect.TypeOf(c).Kind() == reflect.Func {
+			// Print the constructor signature, file and line number from where it has come
+			file, line := funcLocation(c)
+			log.Printf("Providing constructor %q from %v:%d", reflect.TypeOf(c).String(), file, line)
+		} else {
+			log.Printf("Providing object %q", c)
```
glibsm: any reflect/runtime witchcraft can be done to see which file/package it came from? I think objects support `PkgPath()` on their types

alsamylkin: Well, yeah, we can do better and add something like a caller of a provide.

glibsm: Looking forward to it ;)

breerly: Working on a PR now that will make the line-level logging of this method and invoke a bit more obvious.

alsamylkin:
Test logs:
```
2017/05/25 11:54:56 Provide was called /src/go.uber.org/fx/app.go 63
2017/05/25 11:54:56 Provide was called /src/go.uber.org/fx/app_test.go 144
2017/05/25 11:54:56 Providing object "*fx.empty"
```

app.go
```diff
@@ -172,3 +179,12 @@ func funcLocation(c interface{}) (string, int) {
 	mfunc := runtime.FuncForPC(reflect.ValueOf(c).Pointer())
 	return mfunc.FileLine(mfunc.Entry())
 }
+
+func logCaller(name string) {
+	_, file, line, ok := runtime.Caller(2)
```
glibsm: If we want to go down this direction, need to remove the top-most call as it will always be app.go

alsamylkin: Sure, new sample log:
```
2017/05/25 12:52:44 Provide was called /src/go.uber.org/fx/app_test.go 144
2017/05/25 12:52:44 Provide was called /src/go.uber.org/fx/vendor/github.com/stretchr/testify/assert/assertions.go 678
2017/05/25 12:52:44 Providing object "*fx.empty"
```

app.go
```diff
@@ -172,3 +179,12 @@ func funcLocation(c interface{}) (string, int) {
 	mfunc := runtime.FuncForPC(reflect.ValueOf(c).Pointer())
 	return mfunc.FileLine(mfunc.Entry())
 }
+
+func logCaller(name string) {
+	_, file, line, ok := runtime.Caller(3)
+	if !ok {
+		panic("can't determine caller")
```
glibsm: instead of panicing, lets just not log :)

app_test.go
```diff
@@ -139,4 +139,8 @@ func TestApp(t *testing.T) {
 		err = s.Stop(stopCtx)
 		assert.Contains(t, err.Error(), "context deadline exceeded")
 	})
+	t.Run("ProvideDoesn'tPanicForObjectInstances", func(t *testing.T) {
```
breerly: suggest `Doesnt` - the ' will make the console output a bit harder to read

breerly:
@alsamylkin dig is supposed to be a low level building block - are you OK w/ moving the logging statements to Fx instead? This will allow us to control the logging output to be something an Fx user would expect, but not necessarily someone building something w/ Dig.

breerly:
oh

PR #498, opened 2017-05-30
NewTestLifecycle should return *TestLifecycle

abhinav:
TestLifecycle provides a couple extra methods. These are inaccessible if
the function returns a Lifecycle instead of a TestLifecycle.

### DIFF
--- a/lifecycle.go
+++ b/lifecycle.go
@@ -98,7 +98,7 @@ func (l *lifecycle) stop() error {
 }
 
 // NewTestLifecycle creates a new test lifecycle
-func NewTestLifecycle() Lifecycle {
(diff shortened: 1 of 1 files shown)


abhinav:
Merging to unblock. Please leave comments if there are any concerns.

lifecycle.go
```diff
@@ -98,7 +98,7 @@ func (l *lifecycle) stop() error {
 }
 
 // NewTestLifecycle creates a new test lifecycle
-func NewTestLifecycle() Lifecycle {
+func NewTestLifecycle() *TestLifecycle {
```
alsamylkin: Looks ok, could you add a compile time check, that TestLifecycle implements Lifecycle?

PR #499, opened 2017-06-06
modular -> dev

glibsm:
I think we've been sitting on the modular branch long enough, it's time to upstream all that into dev. 

Since this is large enough I didn't feel right pushing directly into dev.
Will temporarily enable merge commit landing so we don't squash the commit history.

README and CHANGELOG need a lot of work, naturally, which can be done subsequently, before the next release.


breerly:
Also, please don't squash this; I want to preserve the commits.

PR #500, opened 2017-06-08
App: Add Invoke method

abhinav:
This adds an Invoke method to `fx.App` which is equivalent to dig's
`Invoke` method.

The reasoning for adding this method is that you expect to be able to do
things with your fully initialized running service after it has been
started, in a serial fashion.

That is,

```go
// Arguments to Start are invoked *before* Lifecycle.Start is
// called. So these are dependencies for Starting the service.
app.Start(...)
defer app.Stop()

// Over here, we expect to be able to do things with our fully
// initialized and running service.
err := app.Invoke(func(d *yarpc.Dispatcher) error {
  // ..
})
```

An alternative to this is a `Resolve` method,

```go
var d *yarpc.Dispatcher
app.Resolve(&d)
```

But in my opinion, that's easier to misuse. People will pass `*App` around and
possibly resolve things on the request path. With Invoke, you have to
explicitly choose to do the "wrong" thing:

```go
var d *yarpc.Dispatcher
app.Invoke(func(dispatcher *yarpc.Dispatcher) {
  d = dispatcher
})
```

### DIFF
--- a/app.go
+++ b/app.go
@@ -130,6 +130,23 @@ func (s *App) start(funcs ...interface{}) error {
 	return nil
 }
 
+// Invoke calls the given functions in-order with objects from the D.I.
+// container. The function blocks until all functions have executed.
+//
+// Functions may optionally return errors. The error returned by the first
+// failing function will be returned to the caller as-is.
+//
+// This function MUST NOT be called before Start or after Stop.
+func (s *App) Invoke(funcs ...interface{}) error {
+	for _, fn := range funcs {
+		if err := s.container.Invoke(fn); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
 // Stop the app
 func (s *App) Stop(ctx context.Context) error {
 	return withTimeout(ctx, s.lifecycle.stop)
--- a/app_test.go
+++ b/app_test.go
@@ -22,6 +22,7 @@ package fx
 
 import (
 	"context"
+	"errors"
 	"testing"
 	"time"
 
@@ -102,7 +103,7 @@ func TestApp(t *testing.T) {
 		s.Start(context.Background(), biz) // this is invoked explicitly
 		assert.Equal(t, 3, count)
 	})
-	t.Run("InvokeRequiresConstructors", func(t *testing.T) {
+	t.Run("StartRequiresConstructors", func(t *testing.T) {
 		s := New()
 		err := s.Start(context.Background(), &type1{})
 		require.Error(t, err)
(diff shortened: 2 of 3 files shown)


app.go
```diff
@@ -35,8 +35,19 @@ import (
...
+const (
+	stopped appState = iota // initial state
+	starting
+	started
+	stopping
+)
+
 // App models a modular application
 type App struct {
+	state appState
```
glibsm: since it's not an embed, why is it in a separate section?

app.go
```diff
@@ -35,8 +35,19 @@ import (
 	"go.uber.org/multierr"
 )
 
+type appState int
+
+const (
+	stopped appState = iota // initial state
```
glibsm: typically iota is `uninitialized`, right?

app.go
```diff
@@ -35,8 +35,19 @@ import (
 	"go.uber.org/multierr"
 )
 
+type appState int
```
glibsm: i wonder if adding state is extra friction for this PR. Do we want to get just the Invoke part in and then talk about semantics for when other API functions are OK to be called?

app.go
abhinav: No, it's reset to 0 at the start of each const block.

app.go
breerly: Let's jam in person. 

I think that `fx.App` might actually want a state machine repr like this so that we can help people interact with it correctly. If we can come to the same thinking here, then I'd probably want to see this commit  broken up into a stack:

1. Add `fx.App.state` with validations
2. Add `fx.App.Invoke` to invoke a func while running

app.go
```diff
@@ -130,9 +153,39 @@ func (s *App) start(funcs ...interface{}) error {
...
 }
 
+// Invoke calls the given functions in-order with objects from the D.I.
+// container. The function blocks until all functions have executed.
+//
+// Functions may optionally return errors. The error returned by the first
+// failing function will be returned to the caller as-is.
+//
+// This function MUST NOT be called before Start or after Stop.
+func (s *App) Invoke(funcs ...interface{}) error {
```
breerly: Feels a bit like this should be a single func, also, I'm not 100% sure if it should return an `err`.

app.go
abhinav: Splitting. This PR is just Invoke now.

app.go
abhinav: Fixed in the app state PR.

app.go
abhinav: We definitely need to be able to return an error. The idea is that the invoked function can return an error:

```
app.Invoke(func(d *yarpc.Dispatcher) error {
  if err := doSomethingWith(d); err != nil { return err }
  doSomethingElse()
})
```

abhinav:
Now with tests.

app.go
```diff
@@ -130,6 +130,23 @@ func (s *App) start(funcs ...interface{}) error {
...
+	for _, fn := range funcs {
+		if err := s.container.Invoke(fn); err != nil {
```
alsamylkin: We do check that `funcs` are actually functions in `Start`, but not in `Invoke`. Could you add some consistency between them?

abhinav:
@alsamylkin The reason we can't rely on `Start` and Lifecycle for this is:

- We want ordering guarantees. Appending to Lifecycle doesn't come with
  ordering guarantees if different things are instantiated in a different
  order.
- That is an overly complicated of doing something simple that people will
  expect to be able to do: Start the application and then do things with the
  resulting application *serially* -- not out of band in a different goroutine.

That said, @akshayjshah has talked me out of this approach because it
introduces some complications with regards to lifecycles (#501 plus what
happens if you Invoke on something that wasn't requested in Start and it ends
up appending a hook to Lifecycle that never gets called). The approach
suggested by @akshayjshah, which I'll be implementing in a different PR is,

```
var deps struct {
  Dispatcher *yarpc.Dispatcher
  Buffer *bytes.Buffer
}

app := fx.New()
app.Provide(...)

app.Start(
  ..,
  fx.Populate(&deps),
)
```

`fx.Populate` will essentially generate a function equivalent to,

```
func(d *yarpc.Dispatcher, buff *bytes.Buffer) {
  // set attributes on `deps`
}
```

This approach has a couple advantages:

- No lifecycle complications because you're requesting everything in `Start`.
  If it was requested, it will be started.
- No new API surface to `fx.App`. This function can theoretically be
  implemented outside the `fx` library. Although given its utility, it should
  be implemented in `fx`.
- `fx.App` doesn't have to become stateful (we can drop #501).

PR #501, opened 2017-06-08
App: Track state

abhinav:
This changes App to start tracking whether it has been started or not.
The state machine we enforce is:

     New()
       |
       v
    Stopped <------- Stopping
       | Start()         ^
       v                 | Stop()
    Starting -------> Started

Invoke from #500 is available only after `Start()` and before `Stop()`.

### DIFF
--- a/app.go
+++ b/app.go
@@ -35,8 +35,18 @@ import (
 	"go.uber.org/multierr"
 )
 
+type appState int
+
+const (
+	stopped appState = iota // initial state
+	starting
+	started
+	stopping
+)
+
 // App models a modular application
 type App struct {
+	state     appState
 	container *dig.Container
 	lifecycle *lifecycle
 	logger    fxlog.Logger
@@ -74,6 +84,10 @@ var (
 // Provide constructors into the D.I. Container, their types will be available
 // to all other constructors, and called lazily at startup
 func (s *App) Provide(constructors ...interface{}) {
+	if s.state != stopped {
+		s.logger.Panic(errors.New("Provide can only be called before an App is started"))
+	}
+
 	for _, c := range constructors {
 		fxlog.PrintProvide(s.logger, c)
 		err := s.container.Provide(c)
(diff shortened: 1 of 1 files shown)


akshayjshah:
I'm on board with introducing a simple notion of lifecycle here, at least to explore the idea.

I'm somewhat concerned that we're developing a very similar lifecycle state machine for the health coordinator. The two don't need to have all the same behaviors, but it does look like they're converging on a very similar state machine. Perhaps there's an opportunity to unify them before 1.0.

breerly:
@akshayjshah my hope is that other libs can *lean* on the state machine here in some way... I'm not exactly sure what that entails, but that's what I've been generically thinking.

app.go
```diff
@@ -74,6 +84,10 @@ var (
 // Provide constructors into the D.I. Container, their types will be available
 // to all other constructors, and called lazily at startup
 func (s *App) Provide(constructors ...interface{}) {
+	if s.state != stopped {
+		s.logger.Panic(errors.New("Provide can only be called before an App is started"))
```
alsamylkin: I don't quite understand panic over errors, you simply don't want to change the function signature?

app.go
```diff
@@ -87,7 +101,15 @@ func (s *App) Provide(constructors ...interface{}) {
 //
 // See dig.Invoke for moreinformation.
 func (s *App) Start(ctx context.Context, funcs ...interface{}) error {
-	return withTimeout(ctx, func() error { return s.start(funcs...) })
+	if s.state != stopped {
+		s.logger.Panic(errors.New("Start can not be called if the App is already running"))
+	}
+	s.state = starting
+	if err := withTimeout(ctx, func() error { return s.start(funcs...) }); err != nil {
```
alsamylkin: hmm, so if there is an error during start(e.g. timeout), next time we'll retry the start will panic, because the state is not stopped?

abhinav:
Closing in favor of a different approach. See https://github.com/uber-go/fx/pull/500#issuecomment-307459407

PR #503, opened 2017-06-08
Make stacks composable

akshayjshah:
This PR sketches out one approach to making stacks composable. It introduces a `Group` type (bikeshed away), that let modules transparently provide multiple types or offer an all-in-one starter.

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -98,7 +98,7 @@ func TestApp(t *testing.T) {
 			return nil
 		}
 		s := New()
-		s.Provide(Constructors{new1, new2, new3})
+		s.Provide(Group{new1, new2, new3})
 		require.NoError(t, s.Start(context.Background(), biz))
 		assert.Equal(t, 3, count)
 	})
@@ -131,7 +131,7 @@ func TestApp(t *testing.T) {
 		require.Error(t, err)
 		assert.Contains(t, err.Error(), "*fx.type1 &{} is not a function")
 	})
-	t.Run("StartStack", func(t *testing.T) {
+	t.Run("StartGroup", func(t *testing.T) {
 		count := 0
 		new1 := func() *type1 {
 			return &type1{}
@@ -152,7 +152,7 @@ func TestApp(t *testing.T) {
 		}
 		s := New()
 		s.Provide(new1, new2, new3)
-		require.NoError(t, s.Start(context.Background(), Starters{biz1, biz3}))
+		require.NoError(t, s.Start(context.Background(), Group{biz1, biz3}))
 		assert.Equal(t, 2, count)
 	})
 	t.Run("StartTimeout", func(t *testing.T) {
--- a/app_test.go
+++ b/app_test.go
@@ -98,7 +98,7 @@ func TestApp(t *testing.T) {
 			return nil
 		}
 		s := New()
-		s.Provide(Group{new1, new2, new3})
+		s.Provide(Group{new1, Group{new2, new3}})
 		require.NoError(t, s.Start(context.Background(), biz))
 		assert.Equal(t, 3, count)
 	})
--- a/app_test.go
+++ b/app_test.go
@@ -79,7 +79,7 @@ func TestApp(t *testing.T) {
 		s.Start(context.Background(), biz)
 		assert.Equal(t, 4, initOrder)
 	})
-	t.Run("ProvideStack", func(t *testing.T) {
+	t.Run("ProvideGroup", func(t *testing.T) {
 		count := 0
 		new1 := func() *type1 {
 			t.Error("this module should not init: no provided type relies on it")
--- a/app_test.go
+++ b/app_test.go
@@ -29,11 +29,11 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-type type1 struct{}
-type type2 struct{}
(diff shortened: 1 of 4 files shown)


stack.go
```diff
@@ -0,0 +1,52 @@
...
+		} else {
+			ret = append(ret, f)
+		}
+	}
+	return ret
+}
+
+// Constructors composes multiple constructors into a single Provide-able
+// object.
+type Constructors []interface{}
```
breerly: Instead of calling this `Constructors`, I'd rather have an explicit name like `ConstructorGroup`.

stack.go
```diff
@@ -0,0 +1,52 @@
...
+
+func (c Constructors) stack() []interface{} {
+	return []interface{}(c)
+}
+
+// Starters composes multiple functions into a single Start-able object.
+type Starters []interface{}
```
breerly: Hmmm... this name is rough.... maybe... `InvokeGroup`?

akshayjshah: Sure, sounds good.

stack.go
```diff
@@ -0,0 +1,45 @@
...
+		} else {
+			ret = append(ret, f)
+		}
+	}
+	return ret
+}
+
+// A Group composes multiple constructors or invokable functions into a single
+// object.
+type Group []interface{}
```
alsamylkin: Out of curiosity, why do we need this type?

abhinav: I think the idea is to make it opt-in rather than treating all slices as lists of constructors.

akshayjshah: Yep, that's it. We can go a little further and do 

```
type Group interface {
   stack() []interface{}
}

func NewGroup(...interface{}) Group
```

which gives us a little wiggle room. We could, for example, add `NewNamedGroup(string, ...interface{})` later on.

glibsm:
Shall we hold off on this until after beta4? We're pretty close to cutting it (even though we know there are some of these things we want to get done for future-proofing).

That would also remove some of the pressure to merge merge merge, as the train is about to leave.

Thoughts?

breerly:
That makes sense to me G

akshayjshah:
Sure, works for me.

app_test.go
```diff
@@ -79,6 +79,29 @@ func TestApp(t *testing.T) {
...
+		}
+		biz := func(s2 *type2, s3 *type3) error {
+			count++
+			return nil
+		}
+		s := New()
+		s.Provide(Group{new1, new2, new3})
+		require.NoError(t, s.Start(context.Background(), biz))
+		assert.Equal(t, 3, count)
+	})
```
abhinav: Should we test for nested stacks?

akshayjshah: Sure, will add.

glibsm:
Let me make sure I understand this correctly. 

Lets say I have `m1.Module` and `m2.Module` in packages `m1`, and `m2` respectively and I want to add a convenience thing that will combine both of them into a stack together.

Can I wrote an `m3.Modules` helper that will look like this?

```
package m3
func Module() fx.Group {
    return fx.Group{m1.Module, m2.Module}
}
```

akshayjshah:
```
package m3

var Module = fx.Group{
  m1.Module,
  m2.Module,
}
```

glibsm:
:mind_blown:

breerly:
The more I look at it, the more I like `fx.Group`.

This API is growing on me quickly.

breerly:
Although I wonder if this should go directly in dig, along with `dig.In` and `dig.Out`...

abhinav:
I think we can get away with starting with it in fx. If/when we port the feature to dig natively, fx can simply delegate to it.

app_test.go
akshayjshah: done.

stack.go
```diff
@@ -0,0 +1,45 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package fx
+
+type stack interface {
+	stack() []interface{}
+}
+
+func unstack(funcs []interface{}) []interface{} {
```
glibsm: nit: ungroup? To keep the terminology consistent

akshayjshah: sure, can do.

breerly:
Instead of `fx.Group`, how about `fx.Stack` - I want to keep the stack terminology around, and it feels smoother that fxstack *is* an fx.Stack.

Also, I think we've already started to ingrain "stack" in our terminology - I worry introducing "group" will actually move us a step away from clear and concise understandability.

alsamylkin:
Stack has a very different meaning when we are talking about collections :)

glibsm:
We're mixing and matching concepts here.

Stack is something that composes multiple modules together, and gets a change to affect startup, etc.

Group is something that ties multiple modules together *to be provided to the stack* and make the "break the glass" scenarios a lot easier, as well as tying together multiple constructors logically.

breerly:
that feels wrong - we should shoot for a single concept

breerly:
Let's discuss these concepts on a ca

breerly:
Call*

akshayjshah:
Sure, sounds good.

PR #504, opened 2017-06-10
Add fx.Inject to pull values from the container

abhinav:
This adds an `fx.Inject(target)` function, the output of which may be
passed to `App.Start()` to fill the target struct with values from the
DI container.

    var x struct {
        Dispatcher *yarpc.Dispatcher
    }

    app.Start(ctx, Inject(&x))

This enables something very simple that users will expect to be able to
do: start the application, do things with the objects inside the started
application, and then stop it--all in a serial manner rather than
out-of-band in a separate goroutine. Need for this has already become
apparent while writing integration tests that interact with Fx.
See #500 for more discussion.

This works by generating a function that accepts a single argument for
each exported field in the struct. The generated function copies values
from the received arguments back to the user-provided struct, leaving
unexported fields alone.


glibsm:
@abhinav Question on the approach: why are we adding this to Fx as opposed to Dig directly? Are we trying to keep the surface level of the dig api in check, or this is something that's super specific to Fx. Looking at the implementation, it will be much easier to execute in a dig function that has the access to the internal graph, and just overall seems like a useful feature.

Another thing to consider: I get that  `Populate` is a mirror of the config API, but this type of thing exists in pretty much all DI frameworks and is called `Inject`

populate.go
```diff
@@ -0,0 +1,133 @@
...
+//
+// The target MUST be a pointer to a struct. Only exported fields will be
+// filled. All field tags supported by dig.In are supported on Populate
+// targets.
+func Populate(target interface{}) interface{} {
+	v := reflect.ValueOf(target)
+
+	if t := v.Type(); t.Kind() != reflect.Ptr || t.Elem().Kind() != reflect.Struct {
+		panic(fmt.Sprintf("expected a pointer to a struct, got a %v", t))
+	}
```
akshayjshah: Rather than panic, what do you think of returning a function like this:

```
func() error {
   return fmt.Sprintf("Populate expected a pointer to a struct, got a %v", t)
}
```

abhinav:
Bad news: With Go 1.9, the behavior is fixed regarding wrapper methods for
embedded fields. The following from [StructOf](https://golang.org/pkg/reflect/#StructOf)
docs is now definitely accurate for private methods.

> StructOf currently does not generate wrapper methods for embedded fields. 

Given that, we will have to generate a function which accepts one argument per
field rather than a single struct with `dig.In`. This also means that we can't
automatically get support for `dig.In` tags unless we alter how `dig`
recognizes `In` structs.

abhinav:
@glibsm Yeah, I like the idea of keeping dig API surface small. Plus this change is almost "external": you could totally implement this outside of dig and fx and it would work just as fine. 

No strong feelings about the naming. I can rename it to `Inject`.

populate.go
abhinav: Sounds reasonable. Applications will explode at `app.Start` instead.

abhinav:
Renamed.

inject_test.go
```diff
@@ -0,0 +1,185 @@
...
+			return gave1
+		}
+
+		new2 := func() *type2 { panic("new2 must not be called") }
+
+		app := New()
+		app.Provide(new1, new2)
+
+		var out struct {
+			T1 *type1
```
willhug: what if T1 was already set?

inject_test.go
```diff
@@ -0,0 +1,185 @@
...
+		new1 := func() *type1 {
+			gave1 = &type1{}
+			return gave1
+		}
+
+		app := New()
+		app.Provide(new1)
+
+		var out struct {
+			Result struct {
```
willhug: this would work even if it wasn't a dig.In right?

willhug: oh wait, nvm it wouldn't work because dig.In tells fx that it can be decoded.  Should we try a non-dig.In embedded struct?

inject_test.go
abhinav: It would be stomped upon. There's no opt-out mechanism for exported fields. I'll add a test for this.

glibsm:
great, now all i can do for the rest of the way is look at the turtle gif

akshayjshah:
Will's gif game is strong.

glibsm:
Q on dropping the dig.In support: this means we're losing the ability to use tags, which shortly will mean we lose the ability for named types.

Is this purely out of the 1.9 frustrations, or this is literally going to be impossible in 1.9? Seems like a pretty big drawback of this approach

inject_test.go
abhinav: Added tests

abhinav:
Yeah, the behavior will just break in Go 1.9; no workaround. The
drawback applies only to the struct passed to `Inject`. All fields of
that struct are like any other `dig.Invoke` function parameters. Any
field could be a struct which embeds `dig.In` and get all the features
that come with it.

If we really want to support `dig.In` on Inject structs directly, we can
specially handle the case where the provided struct embeds `dig.In`,
generating a function that accepts just that struct.

glibsm:
@abhinav that makes sense, I didn't realize that embedding the `dig.In` and the using `Inject` on it works.

Can we please add a test for that? Maybe try out an optional tag to make sure we don't accidentally regress there.

inject.go
```diff
@@ -0,0 +1,110 @@
...
+// to App.Start.
+//
+// 	var target struct {
+// 		Dispatcher *yarpc.Dispatcher
+// 	}
+// 	err := app.Start(ctx, Inject(&target))
+//
+// The target MUST be a pointer to a struct. Only exported fields will be
+// filled.
+func Inject(target interface{}) interface{} {
```
glibsm: why is return type not an error?

abhinav:
@glibsm There is indeed a test for `dig.In` fields. There is no test for
top-level `dig.In` objects because that is not yet supported. The second
part of my comment was suggesting a way to implement this functionality
if we find need for it.

inject.go
glibsm: Nvm. I'z dumb.
I see how you tie this into the API

glibsm:
@abhinav I should proofread my comments better. I saw the test, but I don't see tags used as part of it :P 

I was suggesting to add that too, but maybe I'm just too skeptical

PR #506, opened 2017-06-11
Rename RunForever to Run

breerly:
As discussed offline.


glibsm:
Two more usages in comments: line 67 and 70.
I'll push a commit then merge

PR #507, opened 2017-06-11
Add fx.App.Done()

breerly:
As discussed offline, this makes the expanded use-case of calling Start and Stop manually a bit easier, since the signal blocking code doesn't need to be duplicated.

### DIFF
--- a/app.go
+++ b/app.go
@@ -135,6 +135,13 @@ func (s *App) Stop(ctx context.Context) error {
 	return withTimeout(ctx, s.lifecycle.stop)
 }
 
+// Done allows blocking on SIGINT or SIGTERM
+func (App) Done() <-chan os.Signal {
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, syscall.SIGINT, syscall.SIGTERM)
+	return c
+}
+
 // RunForever starts the app, blocks for SIGINT or SIGTERM, then gracefully stops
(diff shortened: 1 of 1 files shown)


app.go
```diff
@@ -135,6 +135,13 @@ func (s *App) Stop(ctx context.Context) error {
 	return withTimeout(ctx, s.lifecycle.stop)
 }
 
+// Done allows blocking on SIGINT or SIGTERM
+func (App) Done() <-chan os.Signal {
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, syscall.SIGINT, syscall.SIGTERM)
+	return c
+}
```
akshayjshah: Merge train seems to have sailed, I think we should consider a different signature in the final release. Rather than exposing `<-chan os.Signal`, we may want to follow the lead of `context.Context` and expose `<-chan struct{}`. This also gives us more flexibility down the road.

glibsm: Train is shoveling coal right now, ready to depart. Haven't left yet though!

Could you elaborate? We'd still listen on signals notify via empty struct?
How does that allow us to future-proof?

akshayjshah: Part of the goal here is, I think, to hide away from application developers what the source of the stop signal is. If we expose a channel of `os.Signal`, we lose some of that benefit. We also lose the possibility of making the app satisfy the `context.Context` interface.

PR #508, opened 2017-06-12
travis: Test against Go tip

abhinav:
This adds Go tip to the list of Go versions being tested against. This
will ensure we have forwards compatibility with upcoming Go versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ go:
   - 1.7
   - 1.8
   - 1.8.x
+  - tip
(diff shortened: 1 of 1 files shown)


glibsm:
The main reason we took out tip is that the build sometimes takes a very long time.
I have no other objects of keeping it in and if it starts acting up, we can drop it

abhinav:
Reasonable. I added it back because I found out about a change in `reflect`
that would break #504.

PR #510, opened 2017-06-12
Pin to ~0.4.0 dig

glibsm:


glibsm:
Looks like a test needs to be updated, since we now only allow constructors

PR #512, opened 2017-06-13
Provide current version into container

akshayjshah:
Give `Version` a meaningful type and provide it into the container. This allows other packages (e.g., our internal version-heartbeating module) to do useful things.

### DIFF
--- a/app.go
+++ b/app.go
@@ -57,9 +57,8 @@ func New(opts ...Option) *App {
 		lifecycle: lifecycle,
 		logger:    logger,
 	}
-	app.Provide(func() Lifecycle {
-		return lifecycle
-	})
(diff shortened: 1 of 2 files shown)


app.go
```diff
@@ -57,9 +57,8 @@ func New(opts ...Option) *App {
 		lifecycle: lifecycle,
 		logger:    logger,
 	}
-	app.Provide(func() Lifecycle {
-		return lifecycle
-	})
+	app.Provide(func() Lifecycle { return lifecycle })
+	app.Provide(func() SemanticVersion { return Version })
```
breerly: Don't provide in container, it's a constant. Other things should just use it directly.

akshayjshah: Derp, fair point. :/

PR #514, opened 2017-06-16
Proposal: top-level API and stacks

akshayjshah:
This PR proposes a tweak of the top-level application API that makes
stacks much nicer.

The meat of the change is simple: it makes constructors and invokes
options passed to `New`. This has a few benefits:

- Conceptually, it fully defines the application during construction.
This reduces the temptation to pass the application object out of main.
- It lets us mirror dig's `Provide` and `Invoke` terminology, which
makes it easier to navigate docs.
- It naturally leads to the `Options` constructor, which lets a stack
define both provides and invokes in a single object.

You can see an example application in the included example. A stack
might look like this:

```
package fxstack

var Stack = fx.Options(
  fx.Provide(
    foo.Module,
    bar.Module,
    baz.Module,
  ),
  fx.Invoke(
    foo.Run,
    baz.Run,
  ),
)
```

In a user's main function, we'd see this:

```
app := fx.New(fxstack.Stack)
app.Run()
```

To reduce the wordiness of users' code, this PR also provides a small
context construction helper and makes `Inject` a standalone `Option`.

### DIFF
--- a/app.go
+++ b/app.go
@@ -24,142 +24,238 @@ import (
 	"context"
 	"os"
 	"os/signal"
-	"reflect"
 	"syscall"
 	"time"
 
-	"github.com/pkg/errors"
 	"go.uber.org/dig"
 	"go.uber.org/fx/internal/fxlog"
 	"go.uber.org/fx/internal/fxreflect"
 	"go.uber.org/multierr"
 )
 
-// App models a modular application
+// DefaultTimeout is the start and stop timeout used by Run.
+const DefaultTimeout = 15 * time.Second
+
+// Timeout is a convenience function to construct a context with a timeout.
+// It's only intended to reduce noise in the main function; since it doesn't
+// expose context.CancelFunc, it may leak resources.
+func Timeout(d time.Duration) context.Context {
+	ctx, cancel := context.WithTimeout(context.Background(), d)
+	// Assign to the blank identifier on a separate line to convince the linter
+	// that we really don't want the cancel function.
+	_ = cancel
+	return ctx
+}
+
+// An Option configures an App.
+type Option interface {
+	apply(*App)
+}
+
+type optionFunc func(*App)
+
+func (f optionFunc) apply(app *App) { f(app) }
+
+// Provide registers constructors with the application's dependency injection
+// container. Constructors provide one or more types, can depend on other
+// types available in the container, and may optionally return an error. For
+// example:
+//
+//  // Provides type *C, depends on *A and *B.
+//  func(*A, *B) *C
+//
+//  // Provides type *C, depends on *A and *B, and indicates failure by
+//  // returning an error.
+//  func(*A, *B) (*C, error)
+//
+//  // Provides type *B and *C, depends on *A, and can fail.
+//  func(*A) (*B, *C, error)
+//
+// The order in which constructors are provided doesn't matter. Constructors
+// are called lazily and their results are cached for reuse.
+//
+// Taken together, these properties make it perfectly reasonable to Provide a
+// large number of standard constructors even if only a fraction of them are
+// used.
+//
+// See the documentation for go.uber.org/dig for further details.
+func Provide(constructors ...interface{}) Option {
+	return optionFunc(func(app *App) {
+		for _, c := range constructors {
+			app.provide(c)
+		}
+	})
+}
+
+// Invoke registers functions that are executed eagerly on application start.
+// Arguments for these functions are provided from the application's
+// dependency injection container, so the arguments form the leaves of the
+// dependency resolution graph. That is, only constructors whose returned
+// types are required to invoke these functions are executed.
+//
+// Unlike constructors, invoked functions are always executed, and they're
+// always run in order. Invoked functions may have any number of returned
+// values. If the final returned object is an error, it's assumed to be a
+// success indicator. All other returned values are discarded.
+//
+// See the documentation for go.uber.org/dig for further details.
+func Invoke(funcs ...interface{}) Option {
+	return optionFunc(func(app *App) {
+		app.invokes = append(app.invokes, funcs...)
+	})
+}
+
+// Options composes a collection of Options into a single Option.
+func Options(opts ...Option) Option {
+	return optionFunc(func(app *App) {
+		for _, opt := range opts {
+			opt.apply(app)
+		}
+	})
+}
+
+// An App is a modular application built around dependency injection.
 type App struct {
+	optionErr error
 	container *dig.Container
 	lifecycle *lifecycle
+	invokes   []interface{}
 	logger    fxlog.Logger
 }
 
-// Option allows App to be customized
-type Option func(*App)
-
-// New creates a new modular application
+// New creates and initializes an App. All applications begin with the
+// Lifecycle type available in their dependency injection container.
 func New(opts ...Option) *App {
 	logger := fxlog.New()
-
-	container := dig.New()
 	lifecycle := newLifecycle(logger)
 
 	app := &App{
-		container: container,
+		container: dig.New(),
 		lifecycle: lifecycle,
 		logger:    logger,
 	}
-	app.Provide(func() Lifecycle {
-		return lifecycle
-	})
+
+	app.provide(func() Lifecycle { return lifecycle })
+
+	for _, opt := range opts {
+		opt.apply(app)
+	}
 
 	return app
 }
 
-var (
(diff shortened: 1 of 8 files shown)


inject_test.go
```diff
@@ -269,11 +272,32 @@ func TestInject(t *testing.T) {
...
+		)
+
+		require.NoError(t, app.Start(context.Background()), "failed to start")
 		assert.NotNil(t, out.Result.T1, "T1 must not be nil")
 		assert.Nil(t, out.Result.T2, "T2 must be nil")
 		assert.True(t, gave1 == out.Result.T1, "T1 must match")
 	})
 }
+
+func ExampleInject() {
```
glibsm: what's going on here, is this part of how documentation is done through tests?

.travis.yml
```diff
@@ -3,7 +3,6 @@ language: go
 go_import_path: go.uber.org/fx
 
 go:
-  - 1.7
```
glibsm: Completely unrelated to the proposal. I'm ok with the change, but this can be insta-merged in another PR with reasoning as to why we drop 1.7

example_test.go
```diff
@@ -0,0 +1,102 @@
...
+		fx.Invoke(Register),
+	)
+
+	// In a real application, we could just use app.Run() here. Since we don't
+	// want this example to run forever, we'll use Start and Stop.
+	if err := app.Start(fx.Timeout(time.Second)); err != nil {
+		log.Fatal(err)
+	}
+
+	// Normally, we'd block here with <-app.Done().
```
glibsm: add a get here to see the handler firing?

glibsm:
Are we still planning to support groups #503, or would this completely shift the model ?

inject_test.go
akshayjshah: Yep! 
screen shot 2017-06-16 at 11 00 55 am

See https://blog.golang.org/examples for details.

glibsm: Well that sure is good to know...

.travis.yml
akshayjshah: Sure, no problem. The other option is to not close the server on app close and leave a TODO to add that once all our supported Go versions have a `Close` or `Shutdown`.

I'd prefer to make the example realistic, since we don't have any users on Go 1.7.

example_test.go
akshayjshah: Can do, but it'll require either (1) starting the application on a fixed port, or (2) using `httptest` and propagating the listening port in some gross way. I'll push a commit with (1), see what you think.

glibsm: Well if the cost of doing so outweighs the benefits, then there is no reason to bother. Was just an idea, I don't even know if it's going to show up well in godoc

akshayjshah:
Thanks for the prompt review! I think #503 is still important, but it's sort of orthogonal - it makes it much nicer for modules to provide multiple constructors. We don't need it for composable stacks anymore, though.

We _could_ drop that PR and shift the way we build modules a bit. Instead of `zapfx.Module`, which needs to supply both `func(...stuff...) *zap.Logger` and `func(*zap.Logger) *zap.SugaredLogger`, we could explicitly have `zapfx.Logger` and `zapfx.SugaredLogger`. Stacks and users would need to provide both constructors explicitly.

Let's get through this, and then see how we can pivot that PR to make modules nicer but not introduce overlapping concepts.

.travis.yml
akshayjshah: Open as #516.

example_test.go
akshayjshah: I think you're right - it'd be nicer if we could show off a request.

akshayjshah: Added in https://github.com/uber-go/fx/pull/514/commits/6e1611a50065dffafdbc362f04c09b44fbc2db1b - let me know what you think.

akshayjshah:
Tip is broken; otherwise tests are passing.

app.go
```diff
@@ -24,142 +24,238 @@ import (
...
+		for _, c := range constructors {
+			app.provide(c)
+		}
+	})
+}
+
+// Invoke registers functions that are executed eagerly on application start.
+// Arguments for these functions are provided from the application's
+// dependency injection container, so the arguments form the leaves of the
+// dependency resolution graph. That is, only constructors whose returned
```
abhinav: nit: Complexity consistency. The docs for Provide are very approachable but the docs for Invoke are throwing graph resolution at the user.

app.go
```diff
@@ -24,142 +24,238 @@ import (
...
 )
 
-// App models a modular application
+// DefaultTimeout is the start and stop timeout used by Run.
+const DefaultTimeout = 15 * time.Second
+
+// Timeout is a convenience function to construct a context with a timeout.
+// It's only intended to reduce noise in the main function; since it doesn't
+// expose context.CancelFunc, it may leak resources.
+func Timeout(d time.Duration) context.Context {
```
abhinav: Opportunity for misuse is high here if someone decides that they can just use this instead of `context.WithTimeout`.

This will be used only if people use `Start` and `Stop` directly at which point they can build their own contexts.

app.go
```diff
@@ -24,142 +24,238 @@ import (
...
-		s.logger.Printf("ERROR\t\tStart failed, rolling back: %v", err)
-		if stopErr := s.lifecycle.stop(); stopErr != nil {
-			s.logger.Printf("ERROR\t\tCouldn't rollback cleanly: %v", stopErr)
+	// Attempt to start cleanly.
+	if err := app.lifecycle.start(); err != nil {
+		// Start failed, roll back.
+		app.logger.Printf("ERROR\t\tStart failed, rolling back: %v", err)
+		if stopErr := app.lifecycle.stop(); stopErr != nil {
+			app.logger.Printf("ERROR\t\tCouldn't rollback cleanly: %v", stopErr)
 			return multierr.Combine(err, stopErr)
```
abhinav: nit: use multierr.Append if there are only two arguments

inject_test.go
```diff
@@ -269,11 +272,32 @@ func TestInject(t *testing.T) {
...
+	var target struct {
+		Logger *log.Logger
+	}
+
+	app := New(
+		Provide(func() *log.Logger { return log.New(os.Stdout, "", 0) }),
+		Inject(&target),
+	)
+
+	app.Start(context.Background())
```
abhinav: Handle Start error?

lifecycle.go
```diff
@@ -97,16 +97,15 @@ func (l *lifecycle) stop() error {
...
 	return &TestLifecycle{
 		newLifecycle(nil),
 	}
 }
 
-// TestLifecycle makes testing funcs that rely on Lifecycle
-// possible be exposing a Start and Stop func which can be
-// called manually in the context of a unit test.
+// TestLifecycle makes testing funcs that rely on Lifecycle possible by
+// exposing a Start and Stop func which can be called manually in unit tests.
```
abhinav: Nit:

> [..] makes testing **functions** that [..] exposing Start and Stop **methods**
> which [..]

app.go
```diff
@@ -24,142 +24,238 @@ import (
...
+		}
+	})
+}
+
+// An App is a modular application built around dependency injection.
 type App struct {
+	optionErr error
 	container *dig.Container
 	lifecycle *lifecycle
+	invokes   []interface{}
```
glibsm: At one point we had `dig.Fn` as an alias for `interface{}`, to make room for better documentation and make it clearer for usages just like this.

Now looking at this I think `dig.Invocable` or something like that might also be on the cards.

Nothing to do with the PR, just thinking out loud.

example_test.go
glibsm: I like it!

app.go
breerly: I dunno, I kind of like the convenience function. It's low cost on our end, and still flexible for users, e.g. they can just not use it.

app.go
akshayjshah: Will reword, but let's still plan on a final pass through GoDoc - I'm sure Lucie can make this much better.

app.go
akshayjshah: :+1:

app.go
akshayjshah: Good point. The team's general perception seems to be that the native context API is wordy, so the difference between using `Run` and `Start`/`Stop` is large. This was an attempt to reduce that delta, at least in `main`, without introducing functional options for `Start` and `Stop`.

inject_test.go
akshayjshah: :+1:

PR #516, opened 2017-06-16
Stop CI on Go 1.7

akshayjshah:
In order to write good examples of application startup and shutdown,
we'll want to use some server type that's common in open source. The
natural candidate is `http.Server`. Unfortunately, `Server` doesn't have
`Close` or `Shutdown` in Go 1.7.

Since we don't currently have any users on 1.7, we should optimize for
good examples and limit support to Go 1.8 and above.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ language: go
 go_import_path: go.uber.org/fx
 
 go:
-  - 1.7
   - 1.8
   - 1.8.x
   - tip


akshayjshah:
Broken out from #514.

glibsm:
Thanks. I know one line is a lot to ask to be made into it's own PR, but it's much better this way for context, and we won't block the other PR by discussing these minor points.

akshayjshah:
No worries at all. PRs are easy and cheap - makes sense to keep the commit history nice.

Issue #511, opened 2017-06-12
Readme is almost completely bare


glibsm:
It was completely removed as part of the modular refactor, need to bring back the relevant content. 

Perhaps a basic example of what things look like is a good starting point.

akshayjshah:
Do we have to keep the markdown-to-GoDoc converter?

I've generally seen successful projects use README.md and GoDoc for different purposes: the README is basically a sales pitch, while the GoDoc is user documentation.

glibsm:
We can try out to see what the life looks like without it, particularly if we're not willing to invest in it to make it better and take it across the finish line. It doesn't work well for all the md cases.

akshayjshah:
Just to cut down on pre-release blockers, maybe start without it and see whether we want to invest more down the road?

Lupie:
+1 to Akshay

breerly:
README is heading in the right direction as of #555 - closing for now.

PR #518, opened 2017-06-16
Bring back coverage and lint

### LINKED ISSUE #495: Switch to codecov.io
In practice I've found codecov.io to be a much better tool than coveralls:

* Able to run locally, and the coverage *always* matches what codecov.io says
* The coverage comments on PRs gamifies not lowering coverage in a way I've seen lead to better PRs
* The browser plugin shows inline lines that aren't covered, leading to better code reviews

I'd like to switch to this while the source code is still small and relatively simple.

glibsm:
Regardless of coveralls/codecov bikeshed we have no coverage information on PRs at all, and there was no lint.

Bringing all that back.

#511 
Fixes #495


glibsm:
oh dear. lots of work to do on coverage

PR #522, opened 2017-06-20
Move TestLifecycle into an fxtest subpackage

### LINKED ISSUE #520: Move TestLifecycle to an fxtest package
Separating `TestLifecycle` into a separate package will let us add actual testing helpers (think `TestLifecycle.MustStart`) without compiling the testing package into applications.

akshayjshah:
This PR moves the testing spy for `fx.Lifecycle` into a separate
`fxtest` subpackage. This cleans up the documentation for the main
package and lets us (in the future) take dependencies on the `testing`
package in our test helpers.

Fixes #520.

### DIFF
--- a/fxtest/fxtest.go
+++ b/fxtest/fxtest.go
@@ -28,30 +28,33 @@ import (
 // TB is a subset of the standard library's testing.TB interface. It's
 // satisfied by both *testing.T and *testing.B.
 type TB interface {
+	// TODO: Use Logf to implement an fxlog.Logger that doesn't spam the console
+	Logf(string, ...interface{})
 	Errorf(string, ...interface{})
 	FailNow()
 }
+
+var _ fx.Lifecycle = (*Lifecycle)(nil)
 
 // Lifecycle is a testing spy for fx.Lifecycle. It exposes Start and Stop
 // methods (and some test-specific helpers) so that unit tests can exercise
 // hooks.
 type Lifecycle struct {
-	*lifecycle.Lifecycle
-
-	t TB
+	t  TB
+	lc *lifecycle.Lifecycle
 }
 
 // NewLifecycle creates a new test lifecycle.
 func NewLifecycle(t TB) *Lifecycle {
 	return &Lifecycle{
-		Lifecycle: lifecycle.New(nil),
-		t:         t,
+		lc: lifecycle.New(nil),
+		t:  t,
 	}
 }
 
 // Start executes all registered OnStart hooks in order, halting at the first
 // hook that doesn't succeed.
-func (l *Lifecycle) Start() error { return l.Lifecycle.Start() }
+func (l *Lifecycle) Start() error { return l.lc.Start() }
 
 // MustStart calls Start, failing the test if an error is encountered.
 func (l *Lifecycle) MustStart() {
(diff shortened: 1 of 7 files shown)


fxtest/fxtest.go
```diff
@@ -0,0 +1,94 @@
...
+	Start() error
+}
+
+type stopper interface {
+	Stop() error
+}
+
+// Fataler is a subset of the standard library's testing.TB interface. It's
+// satisfied by both *testing.T and *testing.B.
+type Fataler interface {
```
abhinav: Optional: I'm more partial to the `Errorf` plus `FatalNow`
APIs because it makes aborting the test optional.

Plus it gives us testify integration:
https://godoc.org/github.com/stretchr/testify/require#TestingT

fxtest/fxtest.go
```diff
@@ -0,0 +1,94 @@
...
+func NewLifecycle(t Fataler) *Lifecycle {
+	var deps struct {
+		fx.Lifecycle
+	}
+	app := fx.New(fx.Inject(&deps))
+	if err := app.Start(context.Background()); err != nil {
+		t.Fatalf("failed to create new lifecycle: %v", err)
+	}
+	if err := app.Stop(context.Background()); err != nil {
+		t.Fatalf("failed to create new lifecycle: %v", err)
```
abhinav: This seems really jank. We're basically adding a hard requirement that `Lifecycle` can be started and stopped multiple times.

fxtest/fxtest.go
akshayjshah: Sure, no objections to that.

fxtest/fxtest.go
akshayjshah: None of that is part of the public API, so I wouldn't call it a hard requirement. The current implementation is safe to start and stop repeatedly, so this seemed reasonable to me.

I don't mind moving the lifecycle implementation into a separate package, but it's going to move around a bunch more code.

lifecycle.go
```diff
@@ -62,62 +62,35 @@ func (l *lifecycle) Append(hook Hook) {
 	l.hooks = append(l.hooks, hook)
 }
 
-// start calls all OnStarts in order, halting on the first OnStart that fails
-// and marking that position so that stop can rollback.
-func (l *lifecycle) start() error {
+func (l *lifecycle) Start() error {
```
glibsm: I don't think this change is desirable.

The reason @breerly and I made it private to begin with is so module authors don't feel compelled to *accidentally* start/stop a lifecycle, which will obviously have a catastrophic effect on the service.

Apart from trusting module authors to do the right thing, I'd like to *completely* take away the possibility of that happening, if that makes sense.

Is this change required to move the test lifecycle to another package?

akshayjshah: Since the `Lifecycle` interface doesn't expose these methods, it's impossible for module authors to call them.

fxtest/fxtest.go
```diff
@@ -0,0 +1,86 @@
...
+
+// Lifecycle is a testing spy for fx.Lifecycle. It exposes Start and Stop
+// methods (and some test-specific helpers) so that unit tests can exercise
+// hooks.
+type Lifecycle struct {
+	*lifecycle.Lifecycle
+
+	t TB
+}
+
```
abhinav: compile time check to ensure we satisfy the interface?

```go
var _ fx.Lifecycle = (*Lifecycle)(nil)
```

fxtest/fxtest_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+package fxtest
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"go.uber.org/fx"
```
abhinav: reverse ordering

PR #532, opened 2017-06-26
Clean up hanging functions in fxlog

akshayjshah:
The internal fxlog package is structured around an interface that (a)
isn't satisfied by any other logging library, and (b) has only one
implementation. This PR simplifies things a bit by dropping the
interface and making the helper functions into methods. Along the way,
it rationalizes the pointer/value semantics of the logger type - `New`
returns a pointer, and all the methods have pointer receivers.

If we want to make fx's logging pluggable in the future, we'll still
need these changes: we'll want the user to supply an object that
satisfies the same interface as the standard library's `*log.Logger`,
and we'll wrap it in the same concrete type that we're using now.

### DIFF
--- a/internal/fxlog/fxlog.go
+++ b/internal/fxlog/fxlog.go
@@ -30,63 +30,53 @@ import (
 	"go.uber.org/fx/internal/fxreflect"
 )
 
-// Logger logs human-readable output
-type Logger interface {
-	Println(string)
-	Printf(string, ...interface{})
-	Panic(error)
-	Fatalf(string, ...interface{})
+// New returns a new Logger.
+func New() *Logger {
+	return &Logger{log.New(os.Stderr, "", log.LstdFlags)}
 }
 
-// New returns a new StdLogger
-func New() *StdLogger {
-	return &StdLogger{
-		l: log.New(os.Stderr, "", log.LstdFlags),
+// A Logger writes output to standard error.
+type Logger struct {
+	std *log.Logger
+}
+
+// Println logs a single Fx line.
+func (l *Logger) Println(str string) {
+	l.std.Println(prepend(str))
+}
+
+// Printf logs a formatted Fx line.
+func (l *Logger) Printf(format string, v ...interface{}) {
+	l.std.Printf(prepend(format), v...)
+}
+
+// PrintProvide logs a type provided into the dig.Container.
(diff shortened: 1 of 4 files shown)


breerly:
My intention with the interface was to enable testing the log output, something I wanted to do for RC2.

akshayjshah:
This interface still doesn't allow testing log output, since it encapsulates the very details that need testing. Instead, we'll want a private interface that's satisfied by the standard library's `*log.Logger`, which allows us to test the logic in the `fxlog.Logger` type.

breerly:
Let's talk when I'm in - I'm flexy

internal/fxlog/fxlog.go
```diff
@@ -30,63 +30,60 @@ import (
...
 }
 
-// New returns a new StdLogger
-func New() *StdLogger {
-	return &StdLogger{
-		l: log.New(os.Stderr, "", log.LstdFlags),
-	}
+// New returns a new Logger.
+func New() *Logger {
+	return &Logger{log.New(os.Stderr, "", log.LstdFlags)}
```
glibsm: why standard error, is this what the loggers default to?

internal/fxlog/fxlog_test.go
```diff
@@ -0,0 +1,94 @@
+package fxlog
```
glibsm: this file needs a license

internal/fxlog/fxlog.go
akshayjshah: yes

internal/fxlog/fxlog_test.go
akshayjshah: :+1:

internal/fxlog/fxlog_test.go
```diff
@@ -0,0 +1,94 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+type spy struct {
+	*bytes.Buffer
+}
+
+func newSpy() *spy {
```
breerly: yay, ty sir

akshayjshah:
🤦‍♂️  Too many logging tests...

akshayjshah:
I don't know what's happening here - I've pushed a commit with the license change but it's not showing up here. Oddly, it does show up in the commit list for this branch: https://github.com/uber-go/fx/commits/ajs-logs

PR #534, opened 2017-06-26
Improve test coverage

akshayjshah:
Improves test coverage throughout.


breerly:
This feels a bit sloppy - are we just trying to increase code coverage without testing any behavior?

If there is a behavior to verify, then we should write a standalone test that ensures the behavior works.

akshayjshah:
Re: `Options` - sure, I can break out a separate test.

PR #536, opened 2017-06-27
Provide a testing harness for apps

### LINKED ISSUE #535: Don't log to stderr in tests
In tests, we should log to the `testing.T` instead of standard error. Right now, when one test fails, the user is presented with *all* the output from *all* tests.

akshayjshah:
Currently, it's irritating for module owners to integration test their code - they have to use the plain `fx.App` type, which is a bit verbose for common test assertions. More importantly, `fx.App` writes verbose logs to standard error. Since Go's testing library buffers all console output and displays it when tests fail, this makes every test failure print a wall of Fx logs.

This PR exposes a `Logger` option that lets users replace an application's log destination, then adds an `fxtest.App` type that directs all output to the user's `testing.T`. Under this setup, test failures will only show the log output from that one test. Along the way, it adds `MustStart` and `MustStop` helpers.

Fixes #535.

### DIFF
--- a/app.go
+++ b/app.go
@@ -113,6 +113,20 @@ func Options(opts ...Option) Option {
 	})
 }
 
+// Printer is the interface required by fx's logging backend. It's implemented
+// by most loggers, including the standard library's.
+type Printer interface {
+	Printf(string, ...interface{})
+}
+
+// Logger redirects the application's log output to the provided printer.
+func Logger(p Printer) Option {
+	return optionFunc(func(app *App) {
+		app.logger = &fxlog.Logger{Printer: p}
+		app.lifecycle = &lifecycleWrapper{lifecycle.New(app.logger)}
+	})
+}
+
 // An App is a modular application built around dependency injection.
 type App struct {
 	optionErr error
@@ -134,12 +148,11 @@ func New(opts ...Option) *App {
 		logger:    logger,
 	}
 
-	app.provide(func() Lifecycle { return lc })
-
 	for _, opt := range opts {
 		opt.apply(app)
 	}
 
+	app.provide(func() Lifecycle { return app.lifecycle })
 	return app
 }
 
--- a/fxtest/fxtest.go
+++ b/fxtest/fxtest.go
@@ -21,6 +21,8 @@
 package fxtest
 
 import (
+	"context"
+
 	"go.uber.org/fx"
 	"go.uber.org/fx/internal/lifecycle"
 )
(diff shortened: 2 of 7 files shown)


internal/fxlog/fxlog.go
```diff
@@ -32,28 +32,29 @@ import (
...
 // A Logger writes output to standard error.
 type Logger struct {
-	std printer
+	Printer
 }
 
 // Println logs a single Fx line.
 func (l *Logger) Println(str string) {
-	l.std.Printf(prepend(str))
+	l.Printer.Printf(prepend(str))
```
glibsm: maybe just `l.Printf` here now that we embed?

fxtest/fxtest.go
```diff
@@ -34,6 +36,50 @@ type TB interface {
...
+func (p *testPrinter) Printf(format string, args ...interface{}) {
+	p.Logf(format, args...)
+}
+
+// App is a wrapper around fx.App that provides some testing helpers. By
+// default, it uses the provided TB as the application's logging backend.
+type App struct {
+	*fx.App
+
+	tb TB
```
glibsm: is this meant to signal `testing.TB`?

app.go
```diff
@@ -113,6 +113,20 @@ func Options(opts ...Option) Option {
 	})
 }
 
+// Printer is the interface required by fx's logging backend. It's implemented
+// by most loggers, including the standard library's.
```
glibsm: `(type *zap.Logger has no field or method Printf)`

If we wanted to use zap as an actual internal logger for the app as well, would we just build a bridge?

fxtest/fxtest.go
```diff
@@ -34,6 +36,50 @@ type TB interface {
 	FailNow()
 }
 
+type testPrinter struct {
+	TB
```
abhinav: There's a TODO on `TB interface` that is resolved by this issue.

fxtest/fxtest.go
```diff
@@ -34,6 +36,50 @@ type TB interface {
...
+// default, it uses the provided TB as the application's logging backend.
+type App struct {
+	*fx.App
+
+	tb TB
+}
+
+// New creates a new test application.
+func New(tb TB, opts ...fx.Option) *App {
+	allOpts := make([]fx.Option, 0, len(opts))
```
abhinav: nit: `len(opts) + 1`?

internal/fxlog/fxlog.go
akshayjshah: :+1: good point.

fxtest/fxtest.go
akshayjshah: Yep, it's a subset of that interface. Depending on just a subset makes it easier to write a small testing spy, and it also avoids any chance of mistakenly compiling the testing package into production binaries.

app.go
akshayjshah: `*zap.SugaredLogger` implements this interface.

internal/fxlog/fxlog.go
akshayjshah: Actually, it turns out that we just don't need this method at all.

fxtest/fxtest.go
akshayjshah: :+1: good catch - fixed.

PR #542, opened 2017-06-30
First batch for beta3

glibsm:
The main goal here is to try to fix the logrus imports. 

Metrics had to be changed due to tally changing the keys for the snapshots to include the tags.

Cherami mocks were also updated to reflect the API changes.

Once the v1-beta-patches branch is in a good state, a `v1.0.0-beta3.1` will be cut to ease the dep management for existing users, until they can take the time to migrate to RC1.


glibsm:
Updating a few repos using this branch. Will merge and cut the release is things look good

glibsm:
ignore the revert commit - accidental click

PR #543, opened 2017-06-30
Simplify build scripts

akshayjshah:
While revising `CONTRIBUTING.md`, I realized that Fx's build and dev setup is much more elaborate than the current project needs. This is stack of 5 PRs to simplify things so that we're friendlier to new contributors. (Note that this means that PRs must be merged in reverse order: 5, then 4, etc.)

This PR removes unused Makefiles and Docker config.


.build/colors.mk
```diff
@@ -1,18 +0,0 @@
-# Colorization rules
```
glibsm: RIP my proudest Makefile moment :(. 

I know it's not used post di refactor, but this file makes the make output kind of awesome. You're welcome to drop it but it will make a return 🤘

akshayjshah: I'm not saying I wasn't impressed :)

I'm really hoping we get to a place where we mostly just `go test ./...` and the Makefile's for new contributors and Travis.

PR #546, opened 2017-06-30
[4/5] Make Travis builds faster

akshayjshah:
We don't need to run in VMs, so we'll get some speed improvements by
setting `sudo:false` and using their container-based infra instead. We
should also cache the vendor directory and check licensing in CI.

Context: While revising `CONTRIBUTING.md`, I realized that Fx's build and dev setup is much more elaborate than the current project needs. This is stack of 5 PRs to simplify things so that we're friendlier to new contributors. (Note that this means that PRs must be merged in reverse order: 5, then 4, etc.)


.travis.yml
```diff
@@ -1,14 +1,23 @@
...
+cache:
+  directories:
+    - vendor
+
 install:
   - make dependencies
 
 script:
   - make lint
+  - make license
```
glibsm: make lint already runs license in dry mode:

```
@echo "Checking for license headers..."
@DRY_RUN=1 .build/check_license.sh | tee -a lint.log
```

akshayjshah: :facepalm: okay

akshayjshah:
Good point, I'll amend it to do that.

.travis.yml
```diff
@@ -1,11 +1,15 @@
 language: go
+sudo: false
 
 go_import_path: go.uber.org/fx
-
 go:
   - 1.8.x
   - tip
```
peter-edge: I wouldn't do this, or at least allow it to fail - can you do 1.9?

akshayjshah: I can definitely add `1.9.x`. I'd prefer not to allow tip to fail unless it becomes flaky.

akshayjshah: Actually, gimme doesn't know what to do with 1.9.x, and I don't want to use the beta explicitly. We can come back to this after 1.9 is released.

Separately, I *really* wish gimme supported `stable`, `previous-stable`, and `next-stable`, since so many projects have the same support policies.

PR #547, opened 2017-06-30
[5/5] Remove Overcommit

akshayjshah:
To be friendly to open-source contributors, let's drop Overcommit. We
lose a little linting power, but we run all the important checks in CI.
Personally, I found it odd that I needed a system-global Ruby gem
installed to work on a Go project.

Context: While revising `CONTRIBUTING.md`, I realized that Fx's build and dev setup is much more elaborate than the current project needs. This is stack of 5 PRs to simplify things so that we're friendlier to new contributors. (Note that this means that PRs must be merged in reverse order: 5, then 4, etc.)


glibsm:
Probably because I installed it years ago that it never bothered me, but I really like it's functionality, even though it is a global gem :(

I find that the most useful part for me is the markdown lint. Next after that is the commit message validation, but we squash everything anyways.

akshayjshah:
Yeah - I'm not saying that it's useless, just that it's extremely odd in a Go project.

Since we're squashing everything, the person clicking the merge button is on the hook to write a good commit message (Github's default are aaaaawwwwwful).

Agreed that the markdown linting is nice, but since we're not syncing the readme and the GoDoc it doesn't seem like a big win.

PR #548, opened 2017-06-30
Make contributing guide more friendly

akshayjshah:
Update `CONTRIBUTING.md` to (1) reflect changes in Fx's build tooling,
and (2) hand-hold a bit more.

The current guide is pretty focused on nuts and bolts that most people
won't run into, and it glosses over some important details: how to clone
the repo, how to create a feature branch, etc. Basically, experienced
Gophers and git users don't need any of this, and newbies need more
hand-holding.

Along the way, I've tried to strike a more welcoming tone. Rather than a
laundry list of the things contributors must do to avoid having a PR
rejected, I've tried to focus on how they can be successful.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,84 +1,80 @@
-# Developing Fx
+# Contributing
 
-This doc is intended for contributors to Fx (hopefully that's you!).
+Thanks for helping to make Fx better for everyone!
 
-## Development environment
+If you'd like to add new exported APIs, please [open an issue][open-issue]
+describing your proposal &mdash; discussing API changes ahead of time makes
+pull request review much smoother.
 
-* **Go**. Install on OS X with `brew install go`. Make sure `go version` returns at
-  least `1.8` since we're going to be using 1.8+ features like subtests.
+Note that you'll need to sign [Uber's Contributor License Agreement][cla]
+before we can accept any of your contributions. If necessary, a bot will remind
+you to accept the CLA when you open your pull request.
 
-* **[overcommit](https://github.com/brigade/overcommit)**, a git hook manager.
-  Install `overcommit` into your path with `sudo gem install overcommit`.
-  Enable it on the Fx repo with `overcommit --install`.
-  We use overcommit to enforce a variety of style, semantic, and legal things
-  (for example, license headers on all source files).
+## Setup
 
-## Checking out the code
+[Fork][fork], then clone the repository:
 
-Make sure the repository is cloned to the correct location:
-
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1,84 +1,80 @@
...
-// FIXME(ai) make this actually work
+```
+make test
+make lint
 ```
 
-Your initials in parens are optional but good practice. This is better
-than a TODO because our CI checks for unresolved FIXMEs. If you forget to fix
-a FIXME, your code won't get merged.
+If you're not using the minor version of Go specified in the Makefile's
```
Lupie: I'd phrase this in the positive, so like:
For `make lint` to work, you must be using the minor version of Go specified in the Makefile's `LINTABLE_MINOR_VERSIONS`. If you're using another version, that's fine, but you'll only discover style issues after opening your pull request.

CONTRIBUTING.md
```diff
@@ -1,84 +1,80 @@
...
+`LINTABLE_MINOR_VERSIONS` variable, `make lint` doesn't do anything. This is
+fine, but it means that you'll only discover style violations after you open
+your pull request.
 
-## Testing
+## Making Changes
```
Lupie: Let me double-check, but I think we've agreed on sentence case for H2s and smaller. Title case only for H1s.

CONTRIBUTING.md
```diff
@@ -1,84 +1,80 @@
...
+git checkout master
+git fetch upstream
+git rebase upstream/master
+git checkout -b cool_new_feature
 ```
 
-### Disabling the race detector
-
-The race detector makes the tests run way slower. To disable it:
+Make your changes, then ensure that `make lint` and `make test` still pass. If
```
Lupie: Make your changes, and then check that

CONTRIBUTING.md
Lupie: Ya, GitHub lacks a real set standard on style like this, but it definitely leans toward H1s = title, H2s and below = sentence. I'm probably the only one who cares, but I'm for embracing this rule consistently.

CONTRIBUTING.md
akshayjshah: Sounds good.

CONTRIBUTING.md
akshayjshah: Fair enough, happy to comply.

CONTRIBUTING.md
akshayjshah: :+1:

PR #551, opened 2017-07-05
Run Invokes during New, not during Start

### LINKED ISSUE #550: Invokes are tied to lifecycle
## Problem: 
Fx Invokes don't happen until Start() is called and thus are tied to lifecycle.

## Background
There is not an opportunity for the service author to interact with the fully resolved graph until all the `Start()` hooks have been executed.

##  Possible fix
Run Invokes at fx.New() time rather than at fx.Start().

This seems like the simplest solution. The only downside I see is if users are misusing the system and putting code that should be in a lifecycle event into their invokes than that code would be run at New() time which doesn't have timeouts.

glibsm:
This builds on top of a lot of previous work, most notably the refactor
of the top level API to no longer expose `Provide` and `Invoke` on the
app object. Now that all options have to atomically be provided on
`New`, it no longer makes sense to wait until the application start to
resolve.

All the heavy-weight work, like listening for the network, will still be
done as part of the `OnStart` hooks.

Fixes #550.

Essentially this introduces a new touch point: when the graph is fully
resolved, but the start Hooks have not yet run. I see some potential
downsides of this, such as metrics collectors not being started and
someone trying to interact with them...


app.go
```diff
@@ -138,7 +138,13 @@ type App struct {
 
 // New creates and initializes an App. All applications begin with the
 // Lifecycle type available in their dependency injection container.
+//
+// It then executes all functions supplied via the Invoke option. Supplying
+// arguments to these functions requires calling some of the constructors
+// supplied by the Provide option. If any invoked function fails, an error is
+// returned immediately.
 func New(opts ...Option) *App {
+	// TODO(glib): New should probably return (*App, error)
```
breerly: `New` should be easy as possible to use - lets just panic instead of returning an error, especially since there is nothing that can be done to recover.

app.go
```diff
@@ -153,6 +159,24 @@ func New(opts ...Option) *App {
...
 
 	app.provide(func() Lifecycle { return app.lifecycle })
+
+	if app.optionErr != nil {
+		// Some provides failed, short-circuit immediately.
+		app.logger.Printf("Error after options were applied: %v", app.optionErr)
+		return nil
+	}
+
+	// Execute invokes.
```
breerly: Lets move this entire block into a private func `App.invoke`.

app.go
```diff
@@ -153,6 +159,24 @@ func New(opts ...Option) *App {
...
+	// TODO(glib): wrap in timeout?
+	for _, fn := range app.invokes {
+		fname := fxreflect.FuncName(fn)
+		app.logger.Printf("INVOKE\t\t%s", fname)
+		if err := app.container.Invoke(fn); err != nil {
+			app.logger.Printf("Error during %q invoke: %v", fname, err)
```
breerly: Use `app.logger.Fatalf()` here.

app.go
```diff
@@ -153,6 +159,24 @@ func New(opts ...Option) *App {
 	}
 
 	app.provide(func() Lifecycle { return app.lifecycle })
+
+	if app.optionErr != nil {
+		// Some provides failed, short-circuit immediately.
+		app.logger.Printf("Error after options were applied: %v", app.optionErr)
```
breerly: use `app.logger.Fatalf()` here

app.go
```diff
@@ -153,6 +159,24 @@ func New(opts ...Option) *App {
...
+		return nil
+	}
+
+	// Execute invokes.
+	// TODO(glib): wrap in timeout?
```
breerly: Shouldn't need to. If we run into this in the future we can add it though.

app.go
alsamylkin: Why? 
1. It is so much easier to follow the same pattern as the constructors.
2. Panic means that something went really wrong, so we should do it when the internal logic is inconsistent, users invoke's can easily return errors.

Another solution could be provide to an option, to skip/panic/exit on errors(like pflag: https://github.com/spf13/pflag/blob/master/flag.go#L119)

breerly: Mainly because there is nothing that can be recovered, so this just makes the API harder to use.

Let's get @akshayjshaw's opinion.

akshayjshah: Think of a panic as a marginally more polite segfault; just because there's nothing that the user can do to fix this app object doesn't mean that we shouldn't shut the process down gracefully. At the very least, any other long-lived goroutines should have the chance to drain, checkpoint their state, and exit. `Run` gets to cheat, but `New`, `Start`, `Done`, and `Stop` should behave themselves and let the service owner decide how to handle errors.

I'm not opposed to returning an error from `New`, but we don't necessarily need to.

We can instead add an `Err` method that returns `app.optionErr` (which we should rename to `app.err`). This is the same approach used to keep [`bufio.Scanner`](https://godoc.org/bufio#Scanner) easy to use without resorting to panics. It should let 

```
fx.New(...some options...).Run()
```
continue to work, since `Start` can continue to short-circuit if `New` failed.

Users who go the fully-manual route have to handle the error somewhere, so they may end up with

```
app := fx.New(...options...)
if err := app.Err(); err != nil {
  log.Fatalf(err)
}
...usual context creation/Start/Done/Stop dance...
```
If they're just going to call `Start` right away, though, they can skip the explicit `Err` check.

app.go
akshayjshah: See above.

app.go
akshayjshah: See above.

app.go
alsamylkin: My last resort :) https://github.com/golang/go/wiki/PanicAndRecover

By convention, no explicit panic() should be allowed to cross a package boundary. Indicating error conditions to callers should be done by returning error value.

app.go
```diff
@@ -173,16 +197,11 @@ func (app *App) Run() {
 	}
 }
 
-// Start starts the application.
+// Start executes all the onStart hooks of the fx.Lifecycle.
```
akshayjshah: Can we explain what this means to a typical service owner, who probably won't ever interact with lifecycle? At the least, we can add a second sentence: "This typically starts long-running goroutines, like network servers or message queue consumers."

Small nit: `onStart` should have initial caps to match the `Hook` field name.

app.go
```diff
@@ -224,19 +243,6 @@ func (app *App) provide(constructor interface{}) {
 }
 
 func (app *App) start() error {
-	if app.optionErr != nil {
-		// Some provides failed, short-circuit immediately.
-		return app.optionErr
-	}
```
akshayjshah: Start should still short-circuit if `New` encountered an error.

app.go
glibsm: Yeah. I was trying to cut some corners for the sake of opening this PR, totally agree on making this godoc better.

glibsm:
Apologies for long-running PR. Cleaning up right now and merging with all the changes that went in

breerly:
@glibsm I've been working under the impression that this is WIP. 

Once it's ready for review, give us a heads up.

app.go
```diff
@@ -193,9 +198,43 @@ func New(opts ...Option) *App {
...
+	}
+
+	if err := app.runInvokes(); err != nil {
+		app.err = err
+	}
+
 	return app
 }
 
+// runInvokes will execute invokes in order.
```
breerly: "in the order provided"

inject_test.go
```diff
@@ -204,33 +203,6 @@ func TestInject(t *testing.T) {
 		assert.True(t, gave3 == out.T3, "T3 must match")
 	})
 
-	t.Run("OverwritesExisting", func(t *testing.T) {
```
glibsm: Note that this test no longer applies. It was trying to overwrite a value of an already-injected struct. 

This no longer applies since Start no longer modifies the graph, it's fully resolved.

app.go
glibsm: That was the original wording, but I didn't want to use the word "provide" to describe part of the "invoke" functionality related to the order.

app.go
```diff
@@ -193,9 +198,43 @@ func New(opts ...Option) *App {
...
+	}
+
+	if err := app.executeInvokes(); err != nil {
+		app.err = err
+	}
+
 	return app
 }
 
+// Execute invokes in order.
```
breerly: Perhaps "in the order supplied" is a little more clear

glibsm: There we go

glibsm:
@breerly It was WIP, but still not cool to leave a looming PR for a week.

It's now ready to review, I've addressed all the comments from the discussions and prototypes. The resulting LOC change is actually quite small.

app.go
```diff
@@ -193,9 +198,43 @@ func New(opts ...Option) *App {
...
+		if _, ok := fn.(Option); ok {
+			return fmt.Errorf("fx.Option should be passed to fx.New directly, not to fx.Invoke: fx.Invoke received %v", fn)
+		}
+
+		if err := app.container.Invoke(fn); err != nil {
+			app.logger.Printf("Error during %q invoke: %v", fname, err)
+			return err
+		}
+	}
+
```
breerly: nit - space likely unneeded here

app.go
```diff
@@ -213,16 +252,15 @@ func (app *App) Run() {
 	}
 }
 
-// Start starts the application.
+// Start executes all the OnStart hooks of the resolved object graph
+// in the intantiation order.
```
breerly: instantiation*

inject_test.go
breerly: This is actually a really positive sign for this PR.

app.go
```diff
@@ -193,9 +198,43 @@ func New(opts ...Option) *App {
...
+// so we can handle the infinite-invokes.
+//
+// Returns the first error encountered. Maybe shoul multiErr them together.
+func (app *App) executeInvokes() error {
+	for _, fn := range app.invokes {
+		fname := fxreflect.FuncName(fn)
+		app.logger.Printf("INVOKE\t\t%s", fname)
+
+		if _, ok := fn.(Option); ok {
+			return fmt.Errorf("fx.Option should be passed to fx.New directly, not to fx.Invoke: fx.Invoke received %v", fn)
```
bayesianmind: nice, appreciate the rough edge detection here!

app.go
```diff
@@ -193,9 +198,43 @@ func New(opts ...Option) *App {
...
 	}
 	app.provide(func() Lifecycle { return app.lifecycle })
+
+	if app.err != nil {
+		app.logger.Printf("Error after options were applied: %v", app.err)
+		return app
+	}
+
+	if err := app.executeInvokes(); err != nil {
+		app.err = err
```
bayesianmind: I'd really like to be able to get at this err without having to call Start()/trigger lifecycle.
Ideas:
- fx.TryNew() method which returns (*App, error)
- fx.Option that is an error callback
- app.Error() method

breerly: I really dont want to return an err from New if possible.

I'm OK with the idea of `fx.TryNew` or `app.Error`.

@glibsm @akshayjshah @abhinav can you weigh in here?

abhinav: I like `app.Err` better than `TryNew`.

```
app := New(..)
app.Run()
```

Is equivalent to,

```
app := New(..)

if err := app.Err(); err != nil {
  panic(err)
}

// Usual Start/Stop logic
```

`Start` would need to account for `Err` not having been checked manually by the user.

(Also, let's call it `Err()` instead. There's precedent: https://golang.org/pkg/bufio/#Scanner.Err)

app.go
```diff
@@ -193,9 +198,43 @@ func New(opts ...Option) *App {
...
+// Execute invokes in order supplied to New.
+//
+// It might be worthwhile to consider adding context.Context to this function
+// so we can handle the infinite-invokes.
+//
+// Returns the first error encountered. Maybe shoul multiErr them together.
+func (app *App) executeInvokes() error {
+	for _, fn := range app.invokes {
+		fname := fxreflect.FuncName(fn)
+		app.logger.Printf("INVOKE\t\t%s", fname)
```
alsamylkin: should we log the invoke after type check for the Option?

app.go
alsamylkin: why this error is not logged, but the next one is?

app.go
alsamylkin: I am a bit confused which errors are logged and which are not.

glibsm: I will clean up so everything is logged.

breerly: Nice, look like we have `Err()` now :)

breerly: Does that mean @abhinav and @bayesianmind are going to stamp? :1st_place_medal:

app.go
```diff
@@ -193,9 +198,55 @@ func New(opts ...Option) *App {
...
+
+		if _, ok := fn.(Option); ok {
+			err = fmt.Errorf("fx.Option should be passed to fx.New directly, not to fx.Invoke: fx.Invoke received %v", fn)
+		} else {
+			err = app.container.Invoke(fn)
+		}
+
+		if err != nil {
+			app.logger.Printf("Error during %q invoke: %v", fname, err)
```
bayesianmind: This is bugged, you need a break here. Otherwise the next invoke will just nuke the err.

app.go
```diff
@@ -193,9 +198,55 @@ func New(opts ...Option) *App {
...
+
 	return app
 }
 
+// Err returns an error that may have been encountered during the
+// graph resolution.
+//
+// This includes things like incomplete graphs, circular dependencies,
+// missing dependencies, invalid constructors, etc.
+func (app *App) Err() error {
```
abhinav: Actually, this also includes errors encountered during invokes.

app.go
```diff
@@ -193,9 +198,55 @@ func New(opts ...Option) *App {
...
+	return app.err
+}
+
+// Execute invokes in order supplied to New.
+//
+// It might be worthwhile to consider adding context.Context to this function
+// so we can handle the infinite-invokes.
+//
+// Returns the first error encountered. Maybe shoul multiErr them together.
```
abhinav: outdated doc?

app.go
abhinav: +1

glibsm: Will do.

glibsm: Fixed. Added a test to cover, too

glibsm:
@bayesianmind I kept the `gsm-new-invokes` branch around in case you used that commit hash. Let me know when you can swap back to master.

PR #554, opened 2017-07-07
Run provides after other options

### LINKED ISSUE #552: fx.Logger() option ordering is significant
In general the order of most fx options in fx.New(...) call is not important (there are edge cases with invokes that matter but generally it should not depend on order). However, if you don't put the fx.Logger option first than the other provides will output using the default logger rather than the logger the user specified.

## Possible Fix
I plan to put a PR that modifies application of fx.Options to be:
1) All other options
2) All Provides
3) All Invokes in order

bayesianmind:
This removes positional dependence of options like fx.Logger()
Fixes #552

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -49,6 +49,23 @@ func TestNewApp(t *testing.T) {
 		require.NoError(t, app.Start(context.Background()))
 		assert.True(t, found)
 	})
+
+	t.Run("OptionsHappensBeforeProvides", func(t *testing.T) {
+		optionRun := false
+
+		p := func() struct{} {
+			assert.True(t, optionRun, "Option must run before provides")
+			return struct{}{}
+		}
+		inv := func(struct{}) {}
+
+		anOption := optionFunc(func(app *App) {
+			optionRun = true
+		})
+
+		app := New(Provide(p), Invoke(inv), anOption)
+		app.Start(Timeout(1 * time.Second))
+	})
 }
 
 func TestOptions(t *testing.T) {
--- a/app.go
+++ b/app.go
@@ -82,9 +82,7 @@ func (f optionFunc) apply(app *App) { f(app) }
 // See the documentation for go.uber.org/dig for further details.
 func Provide(constructors ...interface{}) Option {
 	return optionFunc(func(app *App) {
-		for _, c := range constructors {
-			app.provide(c)
-		}
+		app.provides = append(app.provides, constructors...)
 	})
 }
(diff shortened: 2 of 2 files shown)


glibsm:
As Alex said, a regression test that validates all logging is uniform in the spy wouldn't go amiss

bayesianmind:
I've added a test for this. Logging tests tend to be brittle so I just made so options in general are run before provides.

app_test.go
```diff
@@ -49,6 +49,24 @@ func TestNewApp(t *testing.T) {
 		require.NoError(t, app.Start(context.Background()))
 		assert.True(t, found)
 	})
+
+	t.Run("OptionsHappensBeforeProvides", func(t *testing.T) {
+		optionRun := false
+
+		type type1 struct{}
```
akshayjshah: nit: no need for the alias, we can just use `struct{}` here and below.

app_test.go
```diff
@@ -49,6 +49,24 @@ func TestNewApp(t *testing.T) {
...
+			assert.True(t, optionRun, "Option must run before provides")
+			return type1{}
+		}
+		inv := func(type1) {}
+
+		anOption := optionFunc(func(app *App) {
+			optionRun = true
+		})
+
+		app := New(Provide(p), Invoke(inv), anOption)
```
akshayjshah: Can we start using `fxtest.New` instead of the plain `New`? As you may have noticed, using `New` results in a wall of output when a single test fails. Then we can also use `app.MustStart()` and `app.MustStop()` below.

akshayjshah:
One other thought: since we don't need (and definitely shouldn't be using) the `provides` field outside of `New`, it's a bit out of place - OO would generally advise only keeping state that we'll use in most other methods. An alternative approach would modify the `Options` interface to also take `[]interface{}` and scope the slice of provides to `New`.

app_test.go
bayesianmind: Sure. That's less readable IMO, but if you prefer it I'm fine with it.

app_test.go
breerly: I see. No worries, land away :)

bayesianmind:
Akshay and I jamed on trying to use fxtest.New rather than New directly. The issue is that I can't create my own optionfunc since the apply method is unexported. So in the end the simplest thing was to leave it where it is now.

glibsm:
Don't think Ben has access to land, merging.

PR #556, opened 2017-07-08
Shorten pitch

breerly:
Spent a bit iterating on the pitch in `README.md` in order to make the value proposition more obvious to the average Gopher. Think this is a step in the right direction.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,9 @@
 
 An application framework for Go that:
 
-* Makes it easy to write composable and testable apps using dependency injection.
-* Removes boilerplate in main and the need for global state and package-level init functions.
-* Eliminates the need for service owners to manually install and manage individual libraries.
+* Makes dependency injection easy.
+* Eliminates the need for global state and `func init()`.
+* Consolidates setup in `main()` across your apps.
 
 ## Status
 
--- a/README.md
+++ b/README.md
@@ -4,7 +4,6 @@ An application framework for Go that:
 
 * Makes dependency injection easy.
 * Eliminates the need for global state and `func init()`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,9 +2,9 @@
 
 An application framework for Go that:
 
-* Makes it easy to write composable and testable apps using dependency injection.
-* Removes boilerplate in main and the need for global state and package-level init functions.
-* Eliminates the need for service owners to manually install and manage individual libraries.
+* Makes dependency injection easy.
+* Eliminates the need for global state and `func init()`.
+* Consolidates setup in `main()` across your apps.
```
akshayjshah: Love the first two points. This last bit isn't super-clear to me, though. Perhaps we could say, "Makes it easy to build reusable, consistent platforms" or "Helps infrastructure teams keep applications consistent."

breerly: I think the terms platforms and infrastructure teams don't have a well defined definition across the industry...

I was reaching for something here that was literal... can you think of anything else?

dmcaulay: Brainstorming...

* Accepts bundled depependcies

Or more words:

* Provides composable dependency bundles for consistency across projects

Less feature specific:

* Enables centralized dependency management

breerly: I think I'm going to write a quickstart... it might be that in doing that I can make this point more clear.

breerly:
I switched to "Enables a consistent platform across your apps." for now.

After I land this, I'm going to see if I can get a quickstart going. Hopefully that will give me something to anchor off of.

breerly:
Removing the last bullet point for now - @abhinav doesnt get it, and others wont either.

PR #559, opened 2017-07-10
Verify App.Run

breerly:
This integration test shamelessly verifies that sending `os.Interrupt` or `syscall.SIGTERM` does indeed cause `App.Run` to stop blocking.

I say shamelessly because it introduces a 100ms sleep and cannot be run concurrently.

That being said, it's nice to have some sort of verification that this does what it says, and gets us above 90% CC.

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -233,11 +233,9 @@ func TestAppRun(t *testing.T) {
 		go func() {
 			app.Run()
 		}()
-
 		// 100ms sleep makes sure that app is in RUNNING state
 		time.Sleep(100 * time.Millisecond)
 	}
-
 	t.Run("BlocksOnSignals", func(t *testing.T) {
 		app := New()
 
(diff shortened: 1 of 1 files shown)


app_test.go
```diff
@@ -226,6 +228,28 @@ func TestAppStop(t *testing.T) {
 	})
 }
 
+func TestAppRun(t *testing.T) {
+	run := func(app *App) {
+		go func() {
+			app.Run()
+		}()
+		// 100ms sleep makes sure that app is in RUNNING state
+		time.Sleep(100 * time.Millisecond)
```
alsamylkin: Could you add a simple function, with start and stop hooks on a wait group. Start adds something to a wait group, stop executes `Done()` and go routine with the running app waits for the group.

breerly:
Fair - I'll rework the PR to remove the use of sleep lol 💃

PR #560, opened 2017-07-10
Take a context in lifecycle hooks

akshayjshah:
Lifecycle hooks are often trying to make RPCs, shut down a server
gracefully, or perform other operations that take a timeout.
Idiomatically, timeouts are passed using the `context.Context` type, so we
should pass a context into our hooks.

Note that this *immediately* makes the package-level example more realistic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
   sink.
 - Add `fxtest.App`, which redirects log output to the user's `testing.TB` and
   provides some lifecycle helpers.
+- Alter lifecycle hook signatures to accept a context.
 
 ## v1.0.0-rc1 (20 Jun 2017)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
   sink.
 - Add `fxtest.App`, which redirects log output to the user's `testing.TB` and
   provides some lifecycle helpers.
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
   sink.
 - Add `fxtest.App`, which redirects log output to the user's `testing.TB` and
   provides some lifecycle helpers.
+- Alter lifecycle hook signatures to accept a context.
```
breerly: Should note that this is breaking.

glibsm: +1. This breaks all the hooks, albeit very slightly

example_test.go
```diff
@@ -52,14 +53,14 @@ func NewMux(lc fx.Lifecycle, logger *log.Logger) *http.ServeMux {
...
+		OnStart: func(context.Context) error {
 			logger.Print("Starting HTTP server.")
 			go server.ListenAndServe()
 			return nil
 		},
-		OnStop: func() error {
+		OnStop: func(ctx context.Context) error {
 			logger.Print("Stopping HTTP server.")
-			return server.Close()
+			return server.Shutdown(ctx)
```
glibsm: nice

CHANGELOG.md
akshayjshah: Sounds good.

PR #562, opened 2017-07-11
fx.Invoke: Fix panic if value is not a func

abhinav:
This fixes a panic on app.Start (due to logging) which occurs if the
argument to fx.Invoke was not a function.

Before:

    panic: reflect: call of reflect.Value.Pointer on struct Value

    goroutine 8 [running]:
    reflect.Value.Pointer(0x134c4a0, 0x1584f88, 0x99, 0x1584f88)
            [..]

After:

    err := app.Start()
    fmt.Println(err)
    // Output: can't invoke non-function {} (type struct {})

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -189,6 +189,13 @@ func TestAppStart(t *testing.T) {
 		assert.Contains(t, err.Error(), "OnStart fail")
 		assert.Contains(t, err.Error(), "OnStop fail")
 	})
+
+	t.Run("InvokeNonFunction", func(t *testing.T) {
+		app := fxtest.New(t, Invoke(struct{}{}))
+		err := app.Start(context.Background())
+		require.Error(t, err, "expected start failure")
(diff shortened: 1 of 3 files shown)


alsamylkin:
I don't understand, why there was a panic on invoke, we already check, that a passed value should be a function:
https://github.com/uber-go/fx/blob/master/app.go#L237
via
https://github.com/uber-go/dig/blob/master/dig.go#L121

alsamylkin:
My bad, can we make sure that invoke option arguments are functions instead?
It is weird to construct a valid option out of invalid argument..

abhinav:
The panic happens when we try to figure out the function name before logging it.

We let dig verify that the arguments are valid. Duplicating the logic in fx would be counterproductive, especially if dig widens its definition of valid values in the future. We've already done this in the past (just pointers and interfaces to basically everything) so it's not a completely unlikely change.

PR #563, opened 2017-07-11
Detect Provide/Invoke on fx.Option

### LINKED ISSUE #553: Provide: Detect options
It's easy for someone to do,

```
fx.Provide(
  foo,
  bar,
  anotherthing.Module,
)
```

Where,

```
package anotherthing

var Module = fx.Provide(...)
```

In such a case, the user will get a cryptic error like,

```
Failed to start: can't provide fx.optionFunc: unable to collect return types of a constructor: must provide at least one non-error type
```

We should detect if the value being provided is an `fx.Option` and perhaps treat is as such, or at least give a more obvious error message.

abhinav:
It's easy to accidentally do,

    Module = fx.Provide(..)

    app := fx.New(
        fx.Provide(foo, bar, Module),
    )

Currently this results in a cryptic error message:

    can't provide fx.optionFunc: unable to collect return types of a constructor: must provide at least one non-error type

With this change, the error message will be similar to,

    cannot Provide fx.Option: fx.Provide(go.uber.org/fx_test.TestAppStart.func5.2(), go.uber.org/fx_test.TestAppStart.func5.3())

Resolves #553

### DIFF
--- a/app.go
+++ b/app.go
@@ -260,7 +260,7 @@ func (app *App) provide(constructor interface{}) {
 	app.logger.PrintProvide(constructor)
 
 	if _, ok := constructor.(Option); ok {
-		app.optionErr = fmt.Errorf("cannot Provide fx.Option: %v", constructor)
+		app.optionErr = fmt.Errorf("fx.Option should be passed to fx.New directly, not to fx.Provide: fx.Provide received %v", constructor)
 		return
 	}
 
@@ -280,7 +280,7 @@ func (app *App) start(ctx context.Context) error {
 		app.logger.Printf("INVOKE\t\t%s", fxreflect.FuncName(fn))
 
 		if _, ok := fn.(Option); ok {
-			return fmt.Errorf("cannot Invoke fx.Option: %v", fn)
+			return fmt.Errorf("fx.Option should be passed to fx.New directly, not to fx.Invoke: fx.Invoke received %v", fn)
 		}
 
 		if err := app.container.Invoke(fn); err != nil {
--- a/app.go
+++ b/app.go
@@ -22,8 +22,10 @@ package fx
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"os/signal"
+	"strings"
 	"syscall"
 	"time"
 
@@ -81,9 +83,21 @@ func (f optionFunc) apply(app *App) { f(app) }
 //
 // See the documentation for go.uber.org/dig for further details.
 func Provide(constructors ...interface{}) Option {
-	return optionFunc(func(app *App) {
-		app.provides = append(app.provides, constructors...)
-	})
+	return provideOption(constructors)
+}
+
+type provideOption []interface{}
+
+func (po provideOption) apply(app *App) {
+	app.provides = append(app.provides, po...)
+}
+
+func (po provideOption) String() string {
+	items := make([]string, len(po))
+	for i, c := range po {
+		items[i] = fxreflect.FuncName(c)
+	}
+	return fmt.Sprintf("fx.Provide(%v)", strings.Join(items, ", "))
 }
 
 // Invoke registers functions that are executed eagerly on application start.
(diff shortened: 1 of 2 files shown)


app.go
```diff
@@ -220,8 +258,14 @@ func (app *App) provide(constructor interface{}) {
...
 	app.logger.PrintProvide(constructor)
+
+	if _, ok := constructor.(Option); ok {
+		app.optionErr = fmt.Errorf("cannot Provide fx.Option: %v", constructor)
+		return
+	}
+
 	if err := app.container.Provide(constructor); err != nil {
-		app.optionErr = multierr.Append(app.optionErr, err)
+		app.optionErr = err
```
abhinav: note that the multierr.Append we have here will never actually "append" because this function *always* short circuits if `optionErr` was non-nil on entry.

app.go
```diff
@@ -81,9 +83,21 @@ func (f optionFunc) apply(app *App) { f(app) }
...
-	return optionFunc(func(app *App) {
-		app.provides = append(app.provides, constructors...)
-	})
+	return provideOption(constructors)
+}
+
+type provideOption []interface{}
+
+func (po provideOption) apply(app *App) {
+	app.provides = append(app.provides, po...)
```
alsamylkin: Can we change the Option interface to return errors on apply?
So in that case we'll return an error if `po` is not a collection of functions.
As I've mentioned in https://github.com/uber-go/fx/pull/562,
we should try to detect errors earlier, during construction time.

abhinav: That change will require returning an error from `New`, which I think is
undesirable.

Instead of a minimal app being,

```go
app := fx.New()
app.Run()
```

It will become,

```go
app, err := fx.New()
if err != nil {
  log.Fatal(err)
}

app.Run()
```

alsamylkin: Well, there was a discussion about it: https://github.com/uber-go/fx/pull/551#discussion_r125991994 ,
but I am not sure if we are committed to any particular plan.

app.go
alsamylkin: I think the intent here was to combine errors, can we still do this?

abhinav: See note. Combination path is never exercised because the logic above guarantees that optionErr is nil.

alsamylkin: I need more coffee :)

app.go
```diff
@@ -220,8 +258,14 @@ func (app *App) provide(constructor interface{}) {
 		return
 	}
 	app.logger.PrintProvide(constructor)
+
+	if _, ok := constructor.(Option); ok {
+		app.optionErr = fmt.Errorf("cannot Provide fx.Option: %v", constructor)
```
breerly: Instead of just saying "cannot Provide fx.Option: ..." we should go the extra mile and say something like:

```
fx.Option should be passed to fx.New directly, not through fx.Provide: %v
```

breerly: Meaning the end-user should understand what they did wrong and how to correct it just by reading the error message.

breerly: Another suggestion:

```
fx.Option cannot be passed to fx.Provide, pass it to fx.New directly instead. Received %v
```

app_test.go
```diff
@@ -196,6 +196,68 @@ func TestAppStart(t *testing.T) {
 		require.Error(t, err, "expected start failure")
 		assert.Contains(t, err.Error(), "can't invoke non-function")
 	})
+
+	t.Run("ProvideProvide", func(t *testing.T) {
```
breerly: ProvideProvide?

Please name the tests after the behavior you are verifying.

app_test.go
```diff
@@ -196,6 +196,68 @@ func TestAppStart(t *testing.T) {
...
+			),
+		)
+
+		err := app.Start(context.Background())
+		require.Error(t, err, "expected start failure")
+		assert.Contains(t, err.Error(), "cannot Provide fx.Option:")
+		assert.Contains(t, err.Error(), "fx.Provide(go.uber.org/fx_test.TestAppStart")
+	})
+
+	t.Run("InvokeInvoke", func(t *testing.T) {
```
breerly: InvokeInvoke?

Please name the tests after the behavior you are verifying.

app_test.go
abhinav: I'll try to come up with the better name but that's literally what is being verified: fx.Providing an fx.Provide.

breerly: This should be in sentence form

app.go
```diff
@@ -81,9 +83,21 @@ func (f optionFunc) apply(app *App) { f(app) }
...
+func (po provideOption) apply(app *App) {
+	app.provides = append(app.provides, po...)
+}
+
+func (po provideOption) String() string {
+	items := make([]string, len(po))
+	for i, c := range po {
+		items[i] = fxreflect.FuncName(c)
+	}
+	return fmt.Sprintf("fx.Provide(%v)", strings.Join(items, ", "))
```
akshayjshah: nit: can use `%s` here, since we know it's a string. Same in the other `String` methods.

app.go
akshayjshah: Let's keep this PR as-is. We're already trying to figure out how to run invokes in `New` in the PR you linked, which will solve the error-detection problem one way or another.

breerly:
Consider a changelog entry - this PR is useful for endusers.

abhinav:
Discussed offline. Changelog not necessary. From the point of view of end-users, the only change is that a certain corner case has better error messages.

PR #566, opened 2017-07-13
Add app.Err() to expose the underlying error from New

glibsm:
PR against #551 to try out what `app.Err()` would look and feel like.

### DIFF
--- a/app.go
+++ b/app.go
@@ -227,21 +227,24 @@ func (app *App) Err() error {
 //
 // Returns the first error encountered. Maybe shoul multiErr them together.
 func (app *App) executeInvokes() error {
+	var err error
+
 	for _, fn := range app.invokes {
 		fname := fxreflect.FuncName(fn)
 		app.logger.Printf("INVOKE\t\t%s", fname)
 
 		if _, ok := fn.(Option); ok {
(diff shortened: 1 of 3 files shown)


app.go
```diff
@@ -211,6 +211,15 @@ func New(opts ...Option) *App {
 	return app
 }
 
+// Error returns an error that may have been encountered during the
+// graph resolution.
+//
+// This includes things like incomplete graphs, circular dependencies,
+// missing dependencies, invalid constructors, etc.
+func (app *App) Error() error {
```
alsamylkin: I'd really prefer Err(), because Error() usually returns a string, implementing the error interface..

app_test.go
```diff
@@ -240,8 +252,13 @@ func TestAppStart(t *testing.T) {
 			Invoke(Invoke(func(type1) {
 			})),
 		)
+		newErr := app.Err()
+		require.Error(t, newErr)
+
 		err := app.Start(context.Background())
 		require.Error(t, err, "expected start failure")
+		assert.Equal(t, err, newErr, "start should return the same error fx.New encountered")
```
alsamylkin: should run return the same error too?

abhinav: Run just panics/exits, right? I think that's preferable. Keeps the default `main()` the same.

alsamylkin: Yeah, makes sense.

PR #567, opened 2017-07-18
Trim timeout function

alsamylkin:
It is over engineered.

### DIFF
--- a/app.go
+++ b/app.go
@@ -342,16 +342,8 @@ func (app *App) start(ctx context.Context) error {
 }
 
 func withTimeout(ctx context.Context, f func(context.Context) error) error {
-	stop := make(chan struct{})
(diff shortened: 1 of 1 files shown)


alsamylkin:
Yep: https://github.com/uber-go/fx/blob/master/app_test.go#L175

glibsm:
where is the breaking regression test that didn't work for the old solution?

abhinav:
+1 on @glibsm's comment. Previous solution was broken. If fixing a bug, prefer to add a breaking test case first.

alsamylkin:
It's not really broken, we were leaking a go routine. Sure, I'll add a test.

alsamylkin:
Actually it is fine, we simply don't need a stop channel.

PR #568, opened 2017-07-19
Logging now expands types provided in fx.Out

glibsm:
This PR expands the dig.Out types in the logs to provide a much better
logging experience.

:feelsbadman: Unfortunately, I had to re-implement a bunch of things
about the dig internals: naming conventions for optionals and named
types, and how to parse struct tags. I added a TODO that expands on
another TODO talking about exporting some of the dig internals. Probably
a topic for a conversation. I'd like to land this as is and then think
through the clean up and how dig APIs can make this experience nicer.
Any change would be pretty radical and affect `Provide` and possibly
even `Invoke`.

On the plus side, the logging is now much nicer.

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -3,7 +3,7 @@ import:
 - package: go.uber.org/multierr
   version: ^1
 - package: go.uber.org/dig
-  version: v1.0.0-rc2
+  version: 1.0.0-rc2
 testImport:
 - package: github.com/stretchr/testify
   version: ~1.1.4
--- a/glide.yaml
+++ b/glide.yaml
@@ -3,7 +3,7 @@ import:
 - package: go.uber.org/multierr
   version: ^1
 - package: go.uber.org/dig
-  version: master # TODO: pin to v1.0.0-rc2 when cut
+  version: v1.0.0-rc2
 testImport:
 - package: github.com/stretchr/testify
   version: ~1.1.4
--- a/glide.yaml
+++ b/glide.yaml
@@ -3,7 +3,7 @@ import:
 - package: go.uber.org/multierr
   version: ^1
 - package: go.uber.org/dig
-  version: v1.0.0-rc1
+  version: master # TODO: pin to v1.0.0-rc2 when cut
 testImport:
 - package: github.com/stretchr/testify
   version: ~1.1.4
--- a/internal/fxlog/fxlog_test.go
+++ b/internal/fxlog/fxlog_test.go
@@ -75,7 +75,7 @@ func TestPrint(t *testing.T) {
 		assert.Equal(t, "[Fx] PROVIDE\t*bytes.Buffer <= bytes.NewBuffer()\n", sink.String())
 	})
 
-	t.Run("printOutProvideTypesSeparately", func(t *testing.T) {
+	t.Run("printExpandsTypesInOut", func(t *testing.T) {
 		sink.Reset()
 
 		type A struct{}
--- a/internal/fxlog/fxlog_test.go
+++ b/internal/fxlog/fxlog_test.go
@@ -28,6 +28,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"go.uber.org/dig"
 )
 
 type spy struct {
@@ -72,6 +73,45 @@ func TestPrint(t *testing.T) {
 		sink.Reset()
 		logger.PrintProvide(bytes.NewBuffer)
 		assert.Equal(t, "[Fx] PROVIDE\t*bytes.Buffer <= bytes.NewBuffer()\n", sink.String())
+	})
+
+	t.Run("printOutProvideTypesSeparately", func(t *testing.T) {
+		sink.Reset()
+
+		type A struct{}
+		type B struct{}
+		type C struct{}
+		type Ret struct {
+			dig.Out
+			*A
+			B `optional:"true"`
+			C `name:"foo"`
+		}
+		logger.PrintProvide(func() Ret { return Ret{} })
+
+		s := sink.String()
+		assert.Contains(t, s, "[Fx] PROVIDE\t*fxlog.A <=")
+		assert.Contains(t, s, "[Fx] PROVIDE\t~fxlog.B <=")
+	})
+
+	t.Run("printOutNamedTypes", func(t *testing.T) {
+		sink.Reset()
+
+		type A struct{}
+		type B struct{}
+		type Ret struct {
+			dig.Out
+			*B `name:"foo"`
+
+			A1 *A `name:"primary"`
+			A2 *A `name:"secondary" optional:"true"`
+		}
+		logger.PrintProvide(func() Ret { return Ret{} })
+
+		s := sink.String()
+		assert.Contains(t, s, "[Fx] PROVIDE\t*fxlog.A:primary <=")
+		assert.Contains(t, s, "[Fx] PROVIDE\t~*fxlog.A:secondary <=")
+		assert.Contains(t, s, "[Fx] PROVIDE\t*fxlog.B:foo <=")
 	})
 
 	t.Run("printProvideInvalid", func(t *testing.T) {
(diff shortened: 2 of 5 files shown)


internal/fxreflect/fxreflect.go
```diff
@@ -21,29 +21,101 @@
...
+		for i := 0; i < e.t.NumField(); i++ {
+			field := e.t.Field(i)
+			ft := field.Type
+
+			if field.PkgPath != "" {
+				continue // skip private fields
+			}
+
+			// keep recursing to traverse all the embedded objects
+			opt, _ := strconv.ParseBool(field.Tag.Get("optional"))
```
glibsm: Not sure if this is cool or not. I don't really case if the parsing failed, I just want it to default to false in that case

internal/fxreflect/fxreflect.go
```diff
@@ -25,25 +25,84 @@ import (
...
+			if field.PkgPath != "" {
+				continue // skip private fields
+			}
+
+			// keep recursing to traverse all the embedded objects
+			k := key{
+				t:    ft,
+				name: field.Tag.Get("name"),
+			}
+			traverseOuts(k, f)
```
alsamylkin: do we need a cycle detection logic from dig? :)

internal/fxreflect/fxreflect.go
```diff
@@ -25,25 +25,84 @@ import (
...
+
+func traverseOuts(k key, f func(s string)) {
+	// skip errors
+	if isErr(k.t) {
+		return
+	}
+
+	// call funtion on non-Out types
+	if dig.IsOut(k.t) {
+		// keep recursing down on field memebers in case they are ins
```
alsamylkin: meme-beers!

internal/fxreflect/fxreflect.go
```diff
@@ -25,25 +25,84 @@ import (
...
+		traverseOuts(key{t: t}, func(s string) {
+			rtypes = append(rtypes, s)
+		})
 	}
 
 	return rtypes
 }
 
+// this type is basically straight out of dig, which is a strong signal
+// that exporting it could really DRY up some things for fx-dig relationship.
```
alsamylkin: hmm, why aren't we doing this then?

breerly:
@glibsm can you update the PR to include some sample output?

internal/fxreflect/fxreflect.go
breerly: @glibsm what would we have to expose in dig for Fx to be able to benefit?

breerly: Seems like we should prioritize making the changes in dig so that it can 1.0 without problem.

glibsm:
## Output sample

### Before

```
2017/07/20 12:13:37 [Fx] PROVIDE        zapfx.Result <= redacted/go/zapfx.New()
2017/07/20 12:13:37 [Fx] PROVIDE        sentryfx.Result <= redacted/go/sentryfx.New()
2017/07/20 12:13:37 [Fx] PROVIDE        fx.Lifecycle <= go.uber.org/fx.New.func1()
```

### After

```
2017/07/20 12:11:25 [Fx] PROVIDE        zap.AtomicLevel <= redacted/go/zapfx.New()
2017/07/20 12:11:25 [Fx] PROVIDE        *zap.Logger <= redacted/go/zapfx.New()
2017/07/20 12:11:25 [Fx] PROVIDE        *zap.SugaredLogger <= redacted/go/zapfx.New()
2017/07/20 12:11:25 [Fx] PROVIDE        zapcore.Core:sentry <= redacted/go/sentryfx.New()
2017/07/20 12:11:25 [Fx] PROVIDE        fx.Lifecycle <= go.uber.org/fx.New.func1()
```

internal/fxreflect/fxreflect.go
glibsm: Yes, that's what I'm working on prototyping now. There are several ways we can go about this, but at the very least I think exposing the `type key struct` would be a good start, as that would remove a lot of the duplicated logic here.

Communicating the keys back is going to be the tricky bit. Modifying a `Provide` API to return `([]Key, error)` is not really an option. That would break far to many places and not for that good of a reason, I think.

I think I may try to add a `Result() []Key` *method* on the container as a way to grab at the last `Provide` call.

If that's the case then we can end up with something like:

```go
c := dig.New()
err := c.Provide(func() (A,B) { ... })
// check err
for k := c.Result {
  fmt.Println("Inserted", k, "into the graph")
}
```

internal/fxreflect/fxreflect.go
glibsm: my most common typo :/

internal/fxreflect/fxreflect.go
glibsm: I looked into this. Since you can't embed pointers to `In/Out` structs you can't cause cycles =/

It's not possible to declare a recursive value struct, only through the use of pointers.

I filed the issue on the subject https://github.com/uber-go/dig/issues/131

internal/fxreflect/fxreflect.go
breerly: Perhaps a `ProvideOption` where the key gets written to a pointer...

glibsm: Hm.

image

Some prototyping to do!

akshayjshah: Can we do the opposite, exposing less instead of more?

Instead of having Fx get into the reflection game at all, we could create an option to pass a logger to dig containers. Then dig can handle logging, just like it'll eventually need to handle more structured introspection. (Of course, to dig a logger is just something that satisfies a simple interface, like `github.com/uber-go/fx/internal/fxlog.Printer`.)

Dig would also, of course, need to also accept a stack frame identifying the provider or invoker.

glibsm: I think that bringing logging into dig would do *more* rather than *less*. 

Dig is not meant for direct consumption, at least that's how we've been thinking about it and operating it. Fx is the user-facing product and should care about things like exposing a sane logger, or the ability to override it.

What I think we all agree on is that dig should allow for more introspection, without necessarily tying it to a particular use-case (logging, instrumentation, etc.)

If we want to try something something really simple, we could start by exposing an "action" channel as a `New` option, which will send notifications when a type is added, or invoked.

(Lets not focus on the names too much right now, rather than the concept)

```go
const (
  Provided = iota
  Invoked
)

type Key struct { // export existing structure
  t reflect.Type
  name string

type Action struct {
  Key

  ActionType int
}
```

Consumption side would look something like:

```go
actions := make(chan dig.Action)
c := dig.New(dig.Listener(actions))
// spin up goroutine to log actions
```

Happy to move this out of the PR into an issue discussion on dig, or to sync up today.

akshayjshah: It *does* more, but *exposes* less.

If you have the cycles for an event system, that works too.

glibsm: Filed https://github.com/uber-go/dig/issues/132 to get into the next release

glide.yaml
```diff
@@ -3,7 +3,7 @@ import:
 - package: go.uber.org/multierr
   version: ^1
 - package: go.uber.org/dig
-  version: 7a1b19ea9358a95d389397b4e494e891db3adee6 # TODO 1.0.0-rc2
+  version: v1.0.0-rc2
```
breerly: Don't think you need `v` in here.

internal/fxlog/fxlog_test.go
```diff
@@ -74,6 +75,46 @@ func TestPrint(t *testing.T) {
 		assert.Equal(t, "[Fx] PROVIDE\t*bytes.Buffer <= bytes.NewBuffer()\n", sink.String())
 	})
 
+	t.Run("printOutProvideTypesSeparately", func(t *testing.T) {
```
breerly: ExpandTypesInOut*

glibsm:
codecov is lying, diff coverage is definitely not 47%

PR #569, opened 2017-07-20
Add fx.In and fx.Out

### LINKED ISSUE #558: fx.In and fx.Out
I've came to the realization that we *should* actually re-export these types.

Users will be perplexed when they come across a `dig.In` or `dig.Out`, and it will strain our ability to write good docs (because we'll be jumping between the libs).

With `fx.In` and `fx.Out`, we'll be able to write about these features from Fx's docs, and users will draw less ?Qs? when they come across the types in the wild.

breerly:
Modules should use this directly instead of relying on dig.In and dig.Out.

Resolves #558.

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -3,7 +3,7 @@ import:
 - package: go.uber.org/multierr
   version: ^1
 - package: go.uber.org/dig
-  version: v1.0.0-rc1
+  version: f407770ddd130b106a6c615fca31f4b040ed7b1a # TODO 1.0.0-rc2
 testImport:
 - package: github.com/stretchr/testify
   version: ~1.1.4
--- a/glide.yaml
+++ b/glide.yaml
@@ -3,7 +3,7 @@ import:
 - package: go.uber.org/multierr
   version: ^1
 - package: go.uber.org/dig
-  version: f407770ddd130b106a6c615fca31f4b040ed7b1a # TODO 1.0.0-rc2
+  version: 7a1b19ea9358a95d389397b4e494e891db3adee6 # TODO 1.0.0-rc2
 testImport:
 - package: github.com/stretchr/testify
   version: ~1.1.4
--- a/inout.go
+++ b/inout.go
@@ -8,9 +8,7 @@ import "go.uber.org/dig"
 // Modules should take a single param struct that embeds an In
 // in order to provide a forwards-compatible API where additional
 // optional properties can be added without breaking.
-type In struct {
-	dig.In
-}
+type In struct{ dig.In }
 
 // Out can be embedded in return structs in order to
 // name types and provide
@@ -18,6 +16,4 @@ type In struct {
 // Modules should return a single results struct that embeds
 // an Out in order to provide a forwards-compatible API where
 // additional types can be provided over time without breaking.
-type Out struct {
-	dig.Out
-}
+type Out struct{ dig.Out }
--- a/inout.go
+++ b/inout.go
@@ -11,7 +11,7 @@ import "go.uber.org/dig"
 type In struct{ dig.In }
 
 // Out can be embedded in return structs in order to
-// name types and provide
+// provide named types.
 //
 // Modules should return a single results struct that embeds
 // an Out in order to provide a forwards-compatible API where
--- a/inout.go
+++ b/inout.go
@@ -1,3 +1,23 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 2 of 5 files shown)


breerly:
Added some integration tests so that we can be assured Fx's integration with Dig works as expected.

inout.go
```diff
@@ -0,0 +1,23 @@
...
+
+import "go.uber.org/dig"
+
+// In can be embedded in a constructor's param struct
+// in order to take advantage of named and optional types.
+//
+// Modules should take a single param struct that embeds an In
+// in order to provide a forwards-compatible API where additional
+// optional properties can be added without breaking.
+type In struct {
```
glibsm: nit: one-liner it? `type In struct { dig.In }`

breerly: sure

inout_test.go
```diff
@@ -0,0 +1,140 @@
...
+func TestOut(t *testing.T) {
+	t.Parallel()
+
+	type out struct {
+		fx.Out
+	}
+	assert.True(t, dig.IsOut(reflect.TypeOf(out{})), "expected dig.Out to work with fx.Out")
+}
+
+func TestOptionalTypes(t *testing.T) {
```
alsamylkin: Why are we testing DIG here?

breerly: For two reasons:

* So that we can see how this is supposed to work from Fx's POV
* So that we can be assured Fx's integration with Dig works as expected

breerly:
@glibsm re: "add a test for fx.In but dig.Out and vice versa?" 

Do you mean I should write tests that fx handles `dig.In` and `dig.Out` as well? I would rather leave that up to coincidence and not explicitly test it.

glibsm:
@breerly No I meant someone provides `dig.Out` with a bunch of types, but they are accepted into a constructor of `fx.Inject` through `fx.In`. Also the other way around - someone providing types through `fx.Out` but consuming through `fx.In`.

You might be able to modify an existing test to fit that scenario.

breerly:
Meh - I don't care to add that - we should expect `fx.In` and `fx.Out` in all situations once we 1.0.

inout_test.go
alsamylkin: Can we have it as an example test instead?

breerly: For now I'd rather not worry about that - I want this PR in for RC2

alsamylkin: // TODO(breerly) ?:)

glide.yaml
```diff
@@ -3,7 +3,7 @@ import:
 - package: go.uber.org/multierr
   version: ^1
 - package: go.uber.org/dig
-  version: v1.0.0-rc1
+  version: f407770ddd130b106a6c615fca31f4b040ed7b1a # TODO 1.0.0-rc2
```
glibsm: You can upgrade to `7a1b19e` which now accepts interfaces if you want.

breerly: Thanks - done!

inout_test.go
```diff
@@ -0,0 +1,158 @@
...
+	t.Parallel()
+
+	type a struct {
+		name string
+	}
+
+	// a constructor that returns the type a with name "foo"
+	type fooOut struct {
+		fx.Out
+		A *a `name:"foo"`
```
akshayjshah: For consistency, need a blank line between embedded and named fields. Same in the subtests below.

inout.go
```diff
@@ -0,0 +1,39 @@
...
+package fx
+
+import "go.uber.org/dig"
+
+// In can be embedded in a constructor's param struct
+// in order to take advantage of named and optional types.
+//
+// Modules should take a single param struct that embeds an In
+// in order to provide a forwards-compatible API where additional
+// optional properties can be added without breaking.
```
akshayjshah: In Go, they're "fields", not "properties." https://golang.org/ref/spec#Struct_types

inout_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+		Foo *foo
+		Bar *bar `optional:"true"`
+	}
+
+	t.Run("NotProvided", func(t *testing.T) {
+		t.Parallel()
+
+		ran := false
+		app := fxtest.New(t, fx.Provide(newFoo), fx.Invoke(func(in in) {
```
akshayjshah: This would be clearer if the parameter name didn't shadow the type. Same in the subtest below.

inout_test.go
```diff
@@ -0,0 +1,158 @@
...
+	newBar := func() barOut {
+		return barOut{
+			A: &a{name: "bar"},
+		}
+	}
+
+	t.Run("ResolveFoo", func(t *testing.T) {
+		t.Parallel()
+
+		// an invoke that resolves type a of name "foo"
```
akshayjshah: Is this comment stranded after some refactoring? `fooIn` is just a struct, not an option.

inout_test.go
```diff
@@ -0,0 +1,158 @@
...
+		assert.True(t, ran, "expected invoke to run")
+	})
+}
+
+func TestNamedTypes(t *testing.T) {
+	t.Parallel()
+
+	type a struct {
+		name string
+	}
```
akshayjshah: I found this test quite confusing. We're testing named fields, but the fact that the field is named `name` is irrelevant. However, the `name:"..."` tag *is* relevant. Can we give the field a different name, so that it's clearer that only the tag matters?

inout_test.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+		ran := false
+		app := fxtest.New(t, fx.Provide(newFoo, newBar), fx.Invoke(func(in barIn) {
+			assert.NotNil(t, in.A, "expected in.a to be injected")
+			assert.Equal(t, "bar", in.A.name, "expected to get type a of name bar")
+			ran = true
+		}))
+		app.MustStart().MustStop()
+		assert.True(t, ran, "expected invoke to run")
+	})
```
akshayjshah: What's the value in making these separate subtests? Are we exercising anything that we wouldn't if we wrote a single test with a param struct that expects named instances foo and bar?

breerly: *doh* 👍

PR #571, opened 2017-07-25
Inject: Support dig.In tags

### LINKED ISSUE #564: Inject doesn't respect named types
fx.Inject silently strips named and optional dig tags from structs you try to inject with it. This is surprising behavior.
If you know how it works by generating a series of invokes than it makes sense but users are likely to get confused by this.

# Option 1: Better error message but no functional change
We should at the very least warn users if they have named or optional tags on their inject

# Option 2: If Inject is called with struct implementing dig.In then respect the tags
If the input struct is a dig.In type than rather than generating a 'fanned out' constructor with all members of the struct as params, just generate a constructor with the input struct directly. That should fix both optionals and named types in one go.

abhinav:
This changes fx.Inject to respect dig.In struct tags on target structs.
That is, the following code is now valid.

    var x struct{ Foo io.Reader `optional:"true"` }
    fx.New(fx.Inject(&x))

Resolves #564.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## v1.0.0-rc3 (unreleased)
 
-- `fx.Inject` now supports `fx.In` tags on destination structs.
+- `fx.Inject` now supports `fx.In` tags on target structs.
 
 ## v1.0.0-rc2 (21 Jul 2017)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## v1.0.0-rc3 (unreleased)
 
-- No changes yet.
+- `fx.Inject` now supports `fx.In` tags on destination structs.
 
 ## v1.0.0-rc2 (21 Jul 2017)
 
--- a/inject.go
+++ b/inject.go
@@ -25,6 +25,8 @@ import (
 	"reflect"
 )
 
+var _typeOfIn = reflect.TypeOf(In{})
+
 // Inject fills the given struct with values from the dependency injection
 // container on application start.
 //
@@ -42,13 +44,23 @@ func Inject(target interface{}) Option {
 	v = v.Elem()
 	t := v.Type()
 
-	// We generate a function with one argument for each field in the target
+	// We generate a function which accepts a single dig.In struct as an
+	// argument. This struct contains all exported fields of the target
 	// struct.
 
-	argTypes := make([]reflect.Type, 0, t.NumField())
+	// Fields of the generated dig.In struct.
+	fields := make([]reflect.StructField, 0, t.NumField()+1)
 
-	// List of values in the target struct aligned with the arguments of the
-	// generated function.
+	// The fix for https://github.com/golang/go/issues/18780 requires that
+	// StructField.Name is always set but versions of Go older than 1.9 expect
+	// Name to be empty for embedded fields.
+	//
+	// We use reflect_go19 and reflect_pre_go19 with build tags to implement
+	// this behavior differently in the two Go versions.
+	fields = append(fields, anonymousField(_typeOfIn))
+
+	// List of values in the target struct aligned with the fields of the
+	// generated struct.
 	//
 	// So for example, if the target is,
 	//
@@ -58,9 +70,13 @@ func Inject(target interface{}) Option {
 	// 		Baz io.Writer
 	// 	}
 	//
-	// The generated function has the shape,
+	// The generated struct has the shape,
 	//
-	// 	func(io.Reader, io.Writer)
+	// 	struct {
+	// 		dig.In
+	// 		F0 io.Reader
+	// 		F2 io.Writer
+	// 	}
 	//
 	// And `targets` is,
 	//
@@ -69,33 +85,44 @@ func Inject(target interface{}) Option {
 	// 		target.Field(2),  // Baz io.Writer
 	// 	]
 	//
-	// As we iterate through the arguments received by the function, we can
-	// simply copy the value into the corresponding value in the targets list.
+	// As we iterate through the fields of the generated struct, we can simply
+	// copy the value into the corresponding value in the targets list.
 	targets := make([]reflect.Value, 0, t.NumField())
 
 	for i := 0; i < t.NumField(); i++ {
 		f := t.Field(i)
-		// Skip private fields.
-		if f.PkgPath != "" {
+
+		// Skip private fields
+		if f.PkgPath != "" && !f.Anonymous {
 			continue
 		}
 
-		argTypes = append(argTypes, f.Type)
+		// We don't copy over names or embedded semantics.
+		fields = append(fields, reflect.StructField{
+			Name: fmt.Sprintf("F%d", i),
+			Type: f.Type,
+			Tag:  f.Tag,
+		})
 		targets = append(targets, v.Field(i))
 	}
 
 	// Equivalent to,
 	//
-	// 	func(foo Foo, bar Bar) {
-	// 		target.Foo = foo
-	// 		target.Bar = bar
+	// 	func(r struct{dig.In; F1 Foo; F2 Bar}) {
+	// 		target.Foo = r.F1
+	// 		target.Bar = r.F2
 	// 	}
 
 	fn := reflect.MakeFunc(
-		reflect.FuncOf(argTypes, nil /* results */, false /* variadic */),
+		reflect.FuncOf(
+			[]reflect.Type{reflect.StructOf(fields)},
+			nil,   /* results */
+			false, /* variadic */
+		),
 		func(args []reflect.Value) []reflect.Value {
-			for i, arg := range args {
-				targets[i].Set(arg)
+			result := args[0]
+			for i := 1; i < result.NumField(); i++ {
+				targets[i-1].Set(result.Field(i))
 			}
 			return nil
 		},
--- a/inject.go
+++ b/inject.go
@@ -44,11 +44,11 @@ func Inject(target interface{}) Option {
 	v = v.Elem()
 	t := v.Type()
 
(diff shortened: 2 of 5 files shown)


inject.go
```diff
@@ -58,9 +70,13 @@ func Inject(target interface{}) Option {
 	// 		Baz io.Writer
 	// 	}
 	//
-	// The generated function has the shape,
+	// The generated struct has the shape,
 	//
-	// 	func(io.Reader, io.Writer)
+	// 	struct {
+	// 		dig.In
```
alsamylkin: Should we use `fx.In` instead of `dig.In`?

inject.go
```diff
@@ -69,33 +85,44 @@ func Inject(target interface{}) Option {
...
+	// copy the value into the corresponding value in the targets list.
 	targets := make([]reflect.Value, 0, t.NumField())
 
 	for i := 0; i < t.NumField(); i++ {
 		f := t.Field(i)
-		// Skip private fields.
-		if f.PkgPath != "" {
+
+		// Skip private fields
+		if f.PkgPath != "" && !f.Anonymous {
```
alsamylkin: I think we should use || to skip both embedded and anonymous fields..
Otherwise this function will panic.

inject_test.go
```diff
@@ -257,6 +270,30 @@ func TestInject(t *testing.T) {
 		assert.Nil(t, out.Result.T2, "T2 must be nil")
 		assert.True(t, gave1 == out.Result.T1, "T1 must match")
 	})
+
+	t.Run("FieldsCanBeOptional", func(t *testing.T) {
```
alsamylkin: Could you add 2 more tests:
1. For private embedded fields.
2. For a recursive dig.In struct(for my curiousity):
```golang
type A struct{
  fx.In
  B struct {
  fx.In
  
  C  int
  }
}
```

inject.go
abhinav: We are using fx.In in the code but the doc was wrong. Fixed.

inject.go
abhinav: We can't skip anonymous fields. `fx.In` totally allows for this,

```
type Foo struct {
  fx.In

  io.Reader
}
```

inject_test.go
abhinav: There's already a test for private embedded fields: EmbeddedUnexportedField.

There's also a test for a nested fx.In struct, but not the exact case you mentioned. I'll add one.

inject.go
alsamylkin: I was thinking about this case:
```golang
        var gave1 *type1
	new1 := func() *type1 {
		gave1 = &type1{}
		return gave1
	}

	var out struct {
		*type1
		T2 *type2 `optional:"true"`
	}

	app := fxtest.New(t,
		Provide(new1),
		Inject(&out),
	)

	defer app.MustStart().MustStop()
	assert.Nil(t, out.type1, "out.type1 should not be set, because it is a private field")
```

Should `out.type1` be set or not? I think it shouldn't, but may be I am missing something.

alsamylkin: This will panic:
```golang
package main

import (
	"fmt"

	"go.uber.org/fx"
)

type x struct {
	field string
}

type y struct {
	x
}

func main() {
	var Y y
	err := fx.New(
		fx.Provide(func() x {
			return x{"alex"}
		}),
		fx.Inject(&Y),
	)
	fmt.Println(err)
} 
```
Here is the output:
```
tmp ❯ go run main.go
2017/07/25 19:06:20 [Fx] PROVIDE	main.x <= main.main.func1()
2017/07/25 19:06:20 [Fx] PROVIDE	fx.Lifecycle <= go.uber.org/fx.New.func1()
2017/07/25 19:06:20 [Fx] INVOKE		reflect.makeFuncStub()
panic: reflect: reflect.Value.Set using value obtained using unexported field

goroutine 1 [running]:
reflect.flag.mustBeAssignable(0x1d9)
	/usr/local/Cellar/go/1.8/libexec/src/reflect/value.go:225 +0x190
reflect.Value.Set(0x10ec140, 0xc42000e2d0, 0x1d9, 0x10ec140, 0xc42000a830, 0x99)
	/usr/local/Cellar/go/1.8/libexec/src/reflect/value.go:1339 +0x2f
go.uber.org/fx.Inject.func2(0xc42000a800, 0x1, 0x1, 0xc42000a800, 0x0, 0x1)
	/Users/alexsamylkin/gocode/src/go.uber.org/fx/inject.go:139 +0x10b
reflect.callReflect(0xc42000a540, 0xc42003d868)
	/usr/local/Cellar/go/1.8/libexec/src/reflect/value.go:508 +0x38f
reflect.makeFuncStub(0x0, 0x0, 0x110769e, 0x4, 0xc42003dbd0, 0x108dc3f, 0xc420016510, 0xc42000a540, 0xc42000a7e0, 0x2000000020, ...)
	/usr/local/Cellar/go/1.8/libexec/src/reflect/asm_amd64.s:17 +0x40
reflect.Value.call(0xc42006c0c0, 0xc42000a540, 0x13, 0x1107636, 0x4, 0xc42000a6e0, 0x1, 0x1, 0x0, 0xc42006c0c0, ...)
	/usr/local/Cellar/go/1.8/libexec/src/reflect/value.go:434 +0x91f
reflect.Value.Call(0xc42006c0c0, 0xc42000a540, 0x13, 0xc42000a6e0, 0x1, 0x1, 0x0, 0x0, 0x80)
	/usr/local/Cellar/go/1.8/libexec/src/reflect/value.go:302 +0xa4
go.uber.org/fx/vendor/go.uber.org/dig.(*Container).Invoke(0xc42000e3a0, 0xc42006c0c0, 0xc42000a540, 0x0, 0x0, 0x0, 0x1, 0x1)
	/Users/alexsamylkin/gocode/src/go.uber.org/fx/vendor/go.uber.org/dig/dig.go:128 +0x31d
go.uber.org/fx.(*App).executeInvokes(0xc42006c120, 0x10dda40, 0xc42000e480)
	/Users/alexsamylkin/gocode/src/go.uber.org/fx/app.go:239 +0x3f3
go.uber.org/fx.New(0xc42003df58, 0x2, 0x2, 0xc42000a560)
	/Users/alexsamylkin/gocode/src/go.uber.org/fx/app.go:207 +0x40a
main.main()
	/Users/alexsamylkin/gocode/src/tmp/main.go:24 +0x150
```

breerly: Great catch Alex.

abhinav: Interesting. There appears to be a slight difference between pointer and non-pointer embedding.

Given,

```go
type foo struct{ type1 }
```

`foo.type1` is an unexported field with `PkgPath != ""` and `Name == "type1"`.

However, given,

```go
type foo struct{ *type1 }
```

`foo.type1` is a field with `PkgPath == ""` and `Name == "type1"`. You *can*
use `reflect` to `Set` it.

This may be a bug in the `reflect` package.

abhinav: https://github.com/golang/go/issues/21122

abhinav: Looks like encoding/json will manually check if `Name` starts with uppercase to decide if it should be skipped. We'll do the same.

https://golang.org/cl/50711

abhinav:
This is ready for another look if @breerly @glibsm @alsamylkin want to review it.

abhinav:
Going to land this later today if no objections.

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## v1.0.0-rc3 (unreleased)
 
-- No changes yet.
+- `fx.Inject` now supports `fx.In` tags on destination structs.
```
breerly: Use the terminology defined in the API: "...on target structs."

inject.go
```diff
@@ -42,13 +46,23 @@ func Inject(target interface{}) Option {
...
-	// We generate a function with one argument for each field in the target
+	// We generate a function which accepts a single fx.In struct as an
+	// argument. This struct contains all exported fields of the target
 	// struct.
 
-	argTypes := make([]reflect.Type, 0, t.NumField())
+	// Fields of the generated fx.In struct.
+	fields := make([]reflect.StructField, 0, t.NumField()+1)
+
+	// The fix for https://github.com/golang/go/issues/18780 requires that
```
breerly: Thanks, this is an excellent comment.

inject.go
```diff
@@ -58,9 +72,13 @@ func Inject(target interface{}) Option {
 	// 		Baz io.Writer
 	// 	}
 	//
-	// The generated function has the shape,
+	// The generated struct has the shape,
 	//
-	// 	func(io.Reader, io.Writer)
+	// 	struct {
+	// 		fx.In
```
breerly: Let's be consistent everywhere and add a space after `fx.In` and `fx.Out` embeds.

inject.go
```diff
@@ -69,37 +87,67 @@ func Inject(target interface{}) Option {
 	// 		target.Field(2),  // Baz io.Writer
 	// 	]
 	//
-	// As we iterate through the arguments received by the function, we can
-	// simply copy the value into the corresponding value in the targets list.
+	// As we iterate through the fields of the generated struct, we can simply
```
breerly: No need to put "simply". Suggest: "Iterate through the fields of the generated struct and copy the value..."

inject.go
```diff
@@ -69,37 +87,67 @@ func Inject(target interface{}) Option {
...
 	for i := 0; i < t.NumField(); i++ {
 		f := t.Field(i)
-		// Skip private fields.
-		if f.PkgPath != "" {
+
+		// Skip unexported fields.
+		if f.Anonymous {
+			// If embedded, StructField.PkgPath is not a reliable indicator of
+			// whether the field is exported. See
+			// https://github.com/golang/go/issues/21122
```
breerly: This can go on the previous line.

abhinav: Can't. Takes it to 91 characters. We usually wrap comments at 80.

PR #573, opened 2017-07-26
Rename fx.Inject to fx.Extract

### LINKED ISSUE #572: Discuss: renaming fx.Inject to fx.Extract
In thinking through how we're going to teach users about breaking glass, I've started to be convinced that `fx.Inject` should be renamed to `fx.Extract`.

I find `fx.Inject` problematic because every dependency in Fx gets injected, not just those passed via types to the inject option. From our and the end users' view, they are performing a non-ideal but very realistic "extraction" from the container. The unique part is not the "inject" aspect, that's perhaps the most common part of Fx- the unique part is extracting the deps directly into a struct in main.

I feel medium/strong about this.

breerly:
Resolves #572 

I'd like to limit the use of the term "inject" and "injection" to pure-DI only, of which enabling is Fx's primary goal. 

Bringing the "inject" terminology to this usecase is harmful because we want users to build an association with "inject" being the golden path, whereas `fx.Inject` is actually a necessary evil for large applications that need to migrate to pure-DI.

Ideally, you would never need to use `fx.Inject`; but you would and do however *inject them* everywhere in an ideal Fx app. Drawing the clear distinction between these things in naming will make teaching and learning Fx much easier.

## Sample Writing: Breaking Glass

Fx is built to enable pure-dependency injection using primarily constructor injection.

When a large application wants to migrate to Fx, it may not be practicle to do so in a single diff.

To support this usecase, we offer a way to "break glass": `fx.Extract`.

`fx.Extract` allows dependencies to be extracted into main, by pulling them out of `fx.App` directly. This lets applications adopt Fx piece by piece, by allowing them to pull instances out of `fx.App` and push them into parts of the app that have not been migrated to Fx.


CHANGELOG.md
```diff
@@ -2,7 +2,9 @@
 
 ## v1.0.0-rc3 (unreleased)
 
-- No changes yet.
+- **[Breaking]** Rename `fx.Inject` to `fx.Extract` to avoid using the
+  term "inject" outside of the ideal usecase. `fx.Extract` is a hammer to
+  break glass, not a way to achieve pure dependency injection.
```
glibsm: Don't think we should include justifications for change in a changelog, as they can get pretty wordy. We should just summarize the changes right?

Or what's the policy here

extract.go
```diff
@@ -25,17 +25,17 @@ import (
...
 // The target MUST be a pointer to a struct. Only exported fields will be
 // filled.
-func Inject(target interface{}) Option {
+func Extract(target interface{}) Option {
 	v := reflect.ValueOf(target)
 
 	if t := v.Type(); t.Kind() != reflect.Ptr || t.Elem().Kind() != reflect.Struct {
 		return Invoke(func() error {
-			return fmt.Errorf("Inject expected a pointer to a struct, got a %v", t)
+			return fmt.Errorf("Extract expected a pointer to a struct, got a %v", t)
```
glibsm: "errors start with lower case", does that apply to when it starts with a method?

glibsm:
Going to stare at this for a few minutes to let it simmer

glibsm:
...still simmering...

Had a quick chat with @bayesianmind. 

What about parity with our config framework and call this `fx.Populate` to mirror `Value.Populate`?
Or the other way around - rename config to `Value.Extract` :man_shrugging:

abhinav:
lol that's the original name this API had in #504

breerly:
I like extract best here, and populate on config.

Extract here because the naming should emphasize the "pulling out" of the fx.App whereas populate emphasize the "filling in" of a target struct.

I know this is a subtle difference, but I feel it's important for teaching and remember what the feature is for and in what circumstances it should be used.

glibsm:
Full circle!

glibsm:
I'm actually convinced with `Extract`. It should allow us to write better docs, even though `Extract` really just means `property injection` as opposed to our constructor injection.

I do see the difference between `Populate` and `Extract`. Populate already has the value on hand to serialize into something. Extract gets access to the whole graph and actually causes a lot of things to happen, like graph resolution.

Lets rebase, fix comments and approvals will follow shortly.

breerly:
Cool - I'll get this PR passing then.

One more thing that's sort of nice with this:

* `fx.Provide` emphasis on *putting types in* an `fx.App`
* `fx.Extract` emphasis on *extracting types out* of an `fx.App`

This fits nicely with how I want people to be thinking about these things. `fx.Extract` can never put types in an `fx.App`. In a way, `fx.Extract` is more similar to `fx.Invoke` in this way.

breerly:
This graph starts making lots of sense:

| Primitive  | Role |Puts Types In | Takes Types out | Execution |
| ----------| ----------- | ------------- | ---------------- | -------- |
| `fx.Provide` | Types Registration, Constructor Injection | X  | X | Will execute only if type produced is relied on by `fx.Extract` or `fx.Invoke` | 
| `fx.Extract` |  Breaking Glass, Property Injection | | X | Will execute in order provided |
| `fx.Invoke` | Entrypoints,  Parameter Injection | | X | Will execute in order provided |

breerly:
Ah! That table helped me finally be able to clearly articulate why the naming of `fx.Inject` is problematic.

Injection is about taking types out, registration is about putting types in.

But developers often confused injection for "injecting types into the container", which is the wrong - injection is only about "taking types out of the container". But still, this is a super common misconception that makes it difficult to talk about and teach fundamentals.

By keeping `fx.Inject` around, we're going to subject our end-users to the same trap: confusing `fx.Inject` for putting types into the container, instead of merely extracting them.

`fx.Extract` will help end-users avoid this misconception, since the emphasis is clearly on taking types out of the container. No one will think that the structs passed to `fx.Extract` actually end up available.

extract.go
abhinav: :man_shrugging:

Issue #577, opened 2017-07-31
Red Team Feedback


breerly:
Did a Red Team exercise with @peter-edge on the latest RC - the overall experience was super positive 🎉 🎈 🍾 , and I was able to capture lots of good feedback, mostly relating to hardening the API before release. 

All of these items should be considered "take it or leave it" - obviously the release is closing in fast, so if we reject all of these or just take on a few, that is totally fine 👍  I've lined up PRs so that if we *do* want a few, we can take them on quickly.

# Directions

1. Read this PR description first
2. VOTE on each PR
3. Comment/Discuss at the PR level

I will create a follow-up meeting where we can make a decision across the board here, either rejecting and closing the PRs or accepting. This this issue will be closed when we cut `v1.0.0` today.

# Feedback

Here are my notes from the session:

## Further Reduce API Surface

One of the things that was nice to see in the exercise was how someone new absorbed the API - we spent alot of time in the GoDoc looking at the public API. A big takeaway is that we should consider aggressively reducing the API surface, mainly providing less convenience so that the API would be even tinier. 

For example, we found ourselves wondering how nice it would be to reduce the API surface to just `fx.New`, `fx.Lifecycle/Hook`, and the options for `fx.New`. Additional APIs can always be added later, and just like with `go.uber.org/config`, we should be asking what things we can survive without for a `v1.0.0`.

### Consider deleting fx.Timeout #579 (accepted)

This helper is useful in testing, but it looks like we can ship without it. We should consider killing this func for that reason alone. The `fx.Start` and `fx.Stop` methods could even set the default timeout on a `context.Background()` if one wasn't attached already.

### Consider deleting fx.App.Done #580 (rejected)

This is a convenience function we could survive without. Most customers will call `fx.Run`, which should be good enough for now. If signal handling becomes something many apps need to do manually, then we can consider adding this back. Until then - it represents a function we'll need to explain/teach, adds extra weight to godoc, and we won't be able to delete it.

### Consider deleting fx.Extract #578 (rejected)

As far as we could tell, `fx.Extract` can be replaced with a single func passed to `fx.Invoke`, where the users treats that func as their original main. This lets them "extract" as many types as they need, use `fx.Run` instead of `fx.Start/Stop`, and gives the `fx.In` support by-way of `fx.Invoke`.

This is an interesting one - if we *can* somehow survive this way, then that's a big win because `fx.Extract` adds an alternate "invoke-like" pathway that we have to explain and teach.

## Other Suggestions

[...]

PR #578, opened 2017-07-31
Remove fx.Extract

breerly:

As far as we could tell, `fx.Extract` can be replaced with a single func passed to `fx.Invoke`, where the users treats that func as their original main. This lets them "extract" as many types as they need, use `fx.Run` instead of `fx.Start/Stop`, and gives the `fx.In` support by-way of `fx.Invoke`.

In other words, there would be no reason to close over a variable in `func main`, you could just treat the func passed to `fx.Invoke` as your main, and then use the variables from there.

This is an interesting one - if we can somehow survive this way, then that's a big win because fx.Extract adds an alternate "invoke-like" pathway that we have to explain and teach.

### Directions 

1. Read #577 first
2. Vote 👍 or 👎 on this PR


akshayjshah:
`Extract` can certainly be replaced by `Invoke`; under the hood, that's all it's doing.

I don't mind removing this, but it makes breaking glass notably more difficult - users who otherwise don't need to learn about `fx.In` and `fx.Out` now need to get all the way into the weeds.

I'd strongly prefer not to break this into a subpackage. Introducing tons of tiny packages isn't idiomatic - if `net/http` can be a single package, `fx` can too.

breerly:
I was actually arguing that it made breaking glass more difficult... then I realized you don't actually need to close over variables in main if you write 1 invoke that acts like main instead.

akshayjshah:
Of course they can - most services *can* be fully modeled as an `fx.App`. The purpose of `Extract` was to accommodate the vocal minority of internal gophers who don't *want* to use DI or fx's notion of lifecycle.

`Extract` provides an easy way to use fx only to manage the standard objects, letting the service owner manage lifecycle and most of main.

If there's debate about this, let's just delete it. It's purely additive and can be brought back later.

breerly:
Again, just reporting the feedback. 

My take is that I think I'm ok with sacrificing some convenience to get an ultra-tight API up-front. This is 500 lines of code that makes something easier, but that something can be accomplished already with `fx.Invoke`.

breerly:
From offline discussion:

> fx.Invoke can't actually be used in place of fx.Extract because the types an invoke func relies on are not always ready until lifecycle start. This makes using an invoke func in place of extract extra laborious because they would have to append an OnStart in order to ensure all types are ready. Extract on the other hand can pull types out of fx.App, and then app.Start can be called to ready eveything.

PR #579, opened 2017-07-31
Remove fx.Timeout and fx.DefaultTimeout

breerly:

This helper is useful in testing, but it looks like we can ship without it. We should consider killing this func for that reason alone. The `fx.Start` and `fx.Stop` methods could even set the default timeout on a `context.Background()` if one wasn't attached already.

### Directions 

1. Read #577 first
2. Vote 👍 or 👎 on this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## v1.0.0-rc3 (unreleased)
 
+- **[Breaking]** Remove `fx.Timeout` and `fx.DefaultTimeout`.
 - **[Breaking]** Rename `fx.Inject` to `fx.Extract`.
 - `fx.Extract` now supports `fx.In` tags on target structs.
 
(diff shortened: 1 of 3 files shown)


akshayjshah:
I certainly don't mind removing this from the main `fx` package, but we should retain it in `fxtest`.

Keep in mind that we had several hour-long meetings about this API; as far as I recall, I was the only person who felt that the standard context-creation API was sufficiently ergonomic. Has everyone else changed their mind?

breerly:
I recorded this diff as a "is it possible to survive without it?" more than a "whats the most ergonomic".

I'm flexy - moving down into `fxtest` seems really reasonable, as does not shipping it day and waiting to see how people use Fx.

abhinav:
Removing this API makes sense to me. The default context creation API is good enough.

I'm not sure about hiding `DefaultTimeout` but we can expose it later.

breerly:
From offline discussion:

> This looks good, lets move them down to `fxtest` since that's the context in which it's used.

breerly:
We determined in-person that removing these actually is desirable - proceeding in this form.

PR #580, opened 2017-07-31
Remove fx.App.Done

breerly:

This is a convenience function we could survive without. Most customers will call `fx.Run`, which should be good enough for now. If signal handling becomes something many apps need to do manually, then we can consider adding this back. Until then - it represents a function we'll need to explain/teach, adds extra weight to godoc, and we won't be able to delete it.

### Directions 

1. Read #577 first
2. Vote 👍 or 👎 on this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## v1.0.0-rc3 (unreleased)
 
+- **[Breaking]** Remove `fx.App.Done`.
 - **[Breaking]** Rename `fx.Inject` to `fx.Extract`.
 - `fx.Extract` now supports `fx.In` tags on target structs.
 
--- a/app.go
+++ b/app.go
@@ -258,7 +258,9 @@ func (app *App) Run() {
 		app.logger.Fatalf("ERROR\t\tFailed to start: %v", err)
 	}
 
-	app.logger.PrintSignal(<-app.Done())
+	signals := make(chan os.Signal, 1)
+	signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
+	app.logger.PrintSignal(<-signals)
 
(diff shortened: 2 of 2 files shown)


akshayjshah:
This makes it impossible to use `Start` and `Stop` without templating signal-handling logic into every application.

If we want to remove this, we should also remove `Start` and `Stop`.

breerly:
Start and stop is needed to test modules and Fx itself... is there a use-case outside of that?

With #578 it might be that outside of testing there is actually no real need.

peter-edge:
This method has no business whatsoever being a helper method on App, this is simple

peter-edge:
It's not an instance method, this is basic

peter-edge:
It accesses nothing on App, I really question how this ever was made an instance method on App, this is super simple stuff

glibsm:
@peter-edge You're right, we hear ya. In the *current state* it can be moved from a method to be a helper, or package-level function. 

This has been discussed at length, several times, and changed over even more times.

The reason it finally rested on this is to allow future proofing. We're trying to live with Fx `1.*` for a long time and this being a method will allow to inspect internal state in the future, possibly prefaced by logging or error handling.

Hope that clears it up.

peter-edge:
Future proofing is not shipping this method at all, this doesn't belong in the Fx API.

glibsm:
@peter-edge I agree that purely from the API perspective of Fx it seems very black and white. However, considering all the involved parties (production deployments, local development, etc), this was the only solution that seemed to cover all the areas, even if it seems odd at first look.

I can't over-state the number of discussions, and also the length of them. There were pros and cons to all the solutions. For instance, as @akshayjshah mentioned removing this from the API would require us to correctly scaffold out all the services to handle the same os.Signals over and over.

We believe that it's the best compromise we could have made.

PR #581, opened 2017-07-31
Rename fx.Provide to fx.WithConstructors

breerly:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## v1.0.0-rc3 (unreleased)
 
+- **[Breaking]** Rename `fx.Provide` to `fx.WithConstructors.
 - **[Breaking]** Rename `fx.Inject` to `fx.Extract`.
 - `fx.Extract` now supports `fx.In` tags on target structs.
 
--- a/app.go
+++ b/app.go
@@ -59,7 +59,7 @@ type optionFunc func(*App)
 
 func (f optionFunc) apply(app *App) { f(app) }
 
-// Provide registers constructors with the application's dependency injection
+// WithConstructors registers constructors with the application's dependency injection
 // container. Constructors provide one or more types, can depend on other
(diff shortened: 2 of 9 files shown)


akshayjshah:
Neither [Rob Pike's original post on this pattern](https://commandcenter.blogspot.com/2014/01/self-referential-functions-and-design.html) nor [Dave Cheney's follow-up](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) suggest this pattern: it's overly wordy and more appropriate for dynamically-typed languages, where name of a function must also indicate the type of the returned value.

Internally, some teams have adopted a `With*` style for functional options, so this may seem natural to them.

breerly:
Yeah, I could see that.

Did you find that `fx.Provide` -> `fx.Constructors` a bit better?

I actually thought that made things more understandable.

akshayjshah:
`fx.Constructors` sounds good to me, though I'd like it accompanied by `fx.Invocations` so that they're both nouns.

breerly:
Another discussion I had around `fx.Invoke` was that its not obvious how it could be used... in the discussion I kept referencing func main: "it's alot like func main, except it has access to all the types in the fx.App object`. A sort of "container-aware" func main.

I'm cool with `fx.Invocations` and I also somehow think `fx.MainFuncs` is not the worst thing ever because I think we could document it similarly to what I said above. It also is nice in that I would stop trying to document them as "entry points".  The key thing here is it would be nice to not only capture what it is, but also capture how it is supposed to be used.

breerly:
Similar with `fx.Provide` - they thought, and then myself as well, that `fx.Constructors` captured both of those things really well.

akshayjshah:
Let's discuss in person. From migrating services, it's not like `func main` at all - the lifecycle methods are the analog of main, and moving logic from `main` to `Invoke` is unlikely to work correctly.

breerly:
From discussion offline:

PR #582, opened 2017-07-31
Rename fx.Invoke to fx.WithInvokes

breerly:

One of the things that was surprising was that options didn't begin with `With*`. 

* `fx.Provide` -> `fx.WithConstructors`
* `fx.Invoke` -> `fx.WithInvokes`

I was pleasantly surprised how refreshing I found `fx.WithConstructors`, which is more verbose but much much easier to understand, and so will be better for end-users. I also found that the `With*` convention better matched the idea that these things aren't necessarily happening immediately, that instead you are instructing the `fx.App` to use these when it does it's thing.

### Directions 

1. Read #577 first
2. Vote 👍 or 👎 on this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## v1.0.0-rc3 (unreleased)
 
+- **[Breaking]** Rename `fx.Invoke` to `fx.WithInvokes`.
 - **[Breaking]** Rename `fx.Inject` to `fx.Extract`.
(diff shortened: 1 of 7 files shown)


akshayjshah:
Same as #581, with the additional point that "Invoke" isn't a noun. Neologisms should serve a purpose, and we already have the perfectly serviceable "invocation" for this same concept.

breerly:
Yeah, I found the noun part cringey. Personally I'd keep this one the same.

breerly:
From discussion offline:

PR #583, opened 2017-07-31
Rename fx.Options to fx.WithModule

breerly:

If the primary use for fx.Options is to define a module, then we should consider renaming it to fx.WithModule. In other words, it's easier to teach and recall concepts when we name things after how they are used, instead of what they are. I argued this point with the fx.Inject to fx.Extract rename.

### Directions 

1. Read #577 first
2. Vote 👍 or 👎 on this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## v1.0.0-rc3 (unreleased)
 
+- **[Breaking]** Rename `fx.Options` to `fx.WithModule`.
 - **[Breaking]** Rename `fx.Inject` to `fx.Extract`.
 - `fx.Extract` now supports `fx.In` tags on target structs.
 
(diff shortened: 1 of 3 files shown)


akshayjshah:
Thus far, we've avoided using the term "module" in code; this is intentional, since Go has "packages" instead.

I'm fine with renaming this, but the name should capture what's actually happening: for example, is composing an `fx.Logger` and some future `fx.Signals` option a module? If so, should we just rename `Option` to `Module`?

breerly:
I'm flexible, and don't feel too strongly.

I could see the argument that `fx.Module` would make alot of sense to end-users and in documentation because it describes what it is *and* when it should be used.

`fx.Options` on the other hand is a teensie bit confusing, because it describes what it is but when you come across it in the API its not actually obvious when it should be used.

I could go either way honestly - let's jam in person :)

breerly:
From discussion offline:

PR #585, opened 2017-07-31
fx.Run returns error

breerly:

It was surprising during the exercise that Run didn't return anything, instead panicking. While I view this as a nice convenience, the following might be more regular to users:

```
app := fx.New()
if err := fx.Run(); err != nil {
    log.Fatalf(err)
}
```

I could go either way on this one - since Fx is the very top-level thing, it doesn't personally bother me that we panic in `fx.Run` and I like the brevity of the current signature.

### Directions 

1. Read #577 first
2. Vote 👍 or 👎 on this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## v1.0.0-rc3 (unreleased)
 
+- **[Breaking]** `fx.Run` now returns an `error`.
 - **[Breaking]** Rename `fx.Inject` to `fx.Extract`.
 - `fx.Extract` now supports `fx.In` tags on target structs.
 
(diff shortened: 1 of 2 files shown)


breerly:
Note for discussion: 

Does this actually kill the need for `fx.App.Err()` completely?

akshayjshah:
Agreed. *Nothing* should panic, ever, and all panics should terminate the process.

Again, though, we had many hours of meetings about this. Has everyone changed their mind?

breerly:
This is merely a recording of the feedback - don't assume that I've changed my mind either.

Let's jam in person in the meeting I scheduled :)

breerly:
We weren't able to reach a consensus, keeping the current API since `Start` can still be called when a higher-level of control is needed.

PR #586, opened 2017-07-31
Move fx.Extract to fxmigrate.WithExtract

breerly:

If we decide that we shouldn't delete `fx.Extract` #578 , then we should consider moving it to it's own package so that the top-level APIs contain only the "golden path" usage. Perhaps consider `fxmigrate.Extract` or something similar.

Also, we should rename `Extract` with `WithExtract` to match other options.

### Directions 

1. Read #577 first
2. Vote 👍 or 👎 on this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## v1.0.0-rc3 (unreleased)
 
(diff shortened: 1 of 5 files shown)


breerly:
From offline discussion:

> The top-level example in the GoDoc and the documentation should itself lay out the "golden path" - that doesn't warrant stuffing this option in it's own package.

PR #587, opened 2017-07-31
Rename fx.Logger to fx.WithLogger

breerly:

One of the things that was surprising was that options didn't begin with `With*`. 

* `fx.Provide` -> `fx.WithConstructors`
* `fx.Invoke` -> `fx.WithInvokes`
* `fx.Options` -> `fx.WithModule`
* `fx.Logger` -> `fx.WithLogger`

### Directions 

1. Read #577 first
2. Vote 👍 or 👎 on this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## v1.0.0-rc3 (unreleased)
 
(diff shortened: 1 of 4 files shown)


akshayjshah:
Same as #581.

breerly:
From discussion offline:

PR #590, opened 2017-08-09
Fix the url-encoded dot in the package names

### LINKED ISSUE #538: Logging prints %2e in constructor path
Looks like Fx's output contains `%2egit` instead of `.git`:
  
```
2017/06/29 09:01:39 [Fx] PROVIDE        fx.Lifecycle <= code.uber.internal/devexp/learnfx/vendor/go.uber.org/fx.New.func1()
2017/06/29 09:01:39 [Fx] PROVIDE        envfx.Result <= code.uber.internal/devexp/learnfx/vendor/code.uber.internal/go/envfx%2egit.module()
2017/06/29 09:01:39 [Fx] PROVIDE        configfx.Result <= code.uber.internal/devexp/learnfx/vendor/code.uber.internal/go/configfx%2egit.Module()
```

This is likely due to some escaping.

glibsm:
Seems more like working around a bug in the standard library more than
anything else. The fully qualified names containing dots (github.com
etc) are represented correctly, but if a package is housed in a got-git
directory, the dot gets url encoded into %2e.

The super simple solution would just be `s/%2e/.`, but seems safer
to pull in a standard package to handle that.

Fixes #538

### DIFF
--- a/internal/fxreflect/fxreflect.go
+++ b/internal/fxreflect/fxreflect.go
@@ -136,8 +136,7 @@ func FuncName(fn interface{}) string {
 
 	// Use the stdlib to un-escape any package import paths which can happen
 	// in the case of the "dot-git" postfix. Seems like a bug in stdlib =/
-	unescaped, err := url.QueryUnescape(fnName)
-	if err == nil {
+	if unescaped, err := url.QueryUnescape(fnName); err != nil {
 		fnName = unescaped
 	}
 
--- a/internal/fxreflect/fxreflect.go
+++ b/internal/fxreflect/fxreflect.go
@@ -22,6 +22,7 @@ package fxreflect
 
 import (
 	"fmt"
+	"net/url"
 	"reflect"
 	"runtime"
 	"strings"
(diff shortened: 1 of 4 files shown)


internal/fxreflect/fxreflect.go
```diff
@@ -132,6 +133,14 @@ func FuncName(fn interface{}) string {
 	}
 
 	fnName := runtime.FuncForPC(fnV.Pointer()).Name()
+
+	// Use the stdlib to un-escape any package import paths which can happen
+	// in the case of the "dot-git" postfix. Seems like a bug in stdlib =/
+	unescaped, err := url.QueryUnescape(fnName)
```
alsamylkin: Coalesce?
```golang
if unescaped, err := url.QueryUnescape(fnName); err == nil {
  fnName = unescaped
}
```

glibsm:
Question: is this changelog worthy?

breerly:
Sure - the changelog should be things that make end-users excited to update!

Suggest: "Improve readability of startup logging"

PR #591, opened 2017-08-09
Shorten vendored paths

### LINKED ISSUE #539: Better logging for constructors in vendor/
When Fx's output encounters constructors in `vendor/` it should shorten:

```
github.com/foo/bar/vendor/go.uber.org/fx.New.func1()
```

to:

```
vendor/go.uber.org/fx.New.func1()
```

glibsm:
The vendor experience for printing out the constructor paths it not
great, since it includes the parent package as well, making it rather
long. The important bit is what happens after the vendor, not the
duplication of the current repo.

Fixes #539


internal/fxlog/fxlog_test.go
```diff
@@ -123,6 +123,15 @@ func TestPrint(t *testing.T) {
 		assert.Equal(t, "", sink.String())
 	})
 
+	t.Run("printStripsVendorPath", func(t *testing.T) {
+		sink.Reset()
+		// assert is vendored within fx and is a good test case
+		logger.PrintProvide(assert.New)
+		assert.Contains(
+			t, sink.String(),
+			"*assert.Assertions <= vendor/github.com/stretchr/testify/assert.New()")
```
alsamylkin: I think it is better to have vendor in the path, because one may accidentally use a local package from GOPATH.

glibsm: vendor *is* in the path? I don't understand the comment.

alsamylkin: For example, the `dig` package can be present in the vendor folder, when it is installed by e.g. glide or it can be taken from `%GOPATH%/src/go.uber.org/dig` folder if it is fetched by `go get` before. In both cases project will be compiled successfully.

glibsm: the gopath usage is not affected by this regex, it only strips the *current package* prefixes from the vendored location

alsamylkin: How do you know if it is a local project vendor or some other's package vendor?

glibsm: TIL go supports multiple nested vendors. Changed the regexp to be non-greedy.

breerly:
Akshay had reservations about this, let's give him a chance to respond :)

glibsm:
Sure, but lets make sure the reservations or additional discussions happen on the issue: https://github.com/uber-go/fx/issues/539.

The discussion here should be specifically about the particular implementation, i.e. "we shouldn't use regex" or something.

internal/fxreflect/fxreflect.go
```diff
@@ -23,12 +23,15 @@ package fxreflect
...
 	"fmt"
 	"reflect"
+	"regexp"
 	"runtime"
 	"strings"
 
 	"go.uber.org/dig"
 )
 
+var vendorRe = regexp.MustCompile("^.*?vendor/")
```
breerly: A comment here for regexp noobs would be useful, e.g. "match the first occurring vendor/ in  the file path".

internal/fxreflect/fxreflect.go
```diff
@@ -23,12 +23,16 @@ package fxreflect
...
 	"reflect"
+	"regexp"
 	"runtime"
 	"strings"
 
 	"go.uber.org/dig"
 )
 
+// Match from beginning of the line until the first `vendor/` (non-greedy)
+var vendorRe = regexp.MustCompile("^.*?vendor/")
```
abhinav: what if i have a package called "foovendor"

breerly:
This is going to make the output so much more readable - let's do it!

internal/fxlog/fxlog_test.go
```diff
@@ -131,6 +132,24 @@ func TestPrint(t *testing.T) {
...
+	})
+
+	t.Run("printFooVendorPath", func(t *testing.T) {
```
breerly: ty sir

PR #596, opened 2017-09-05
Add fx.NopLogger

amckinney:
When developing an Fx CLI, I found it useful to disable the standard Fx PROVIDE/INVOKE logs. By defining a fx.NoopLogger, users won't need to implement their own no-op logger which satisfies the `fx.Printer` interface.

### DIFF
--- a/app.go
+++ b/app.go
@@ -152,9 +152,7 @@ func Logger(p Printer) Option {
 }
 
 // NopLogger disables the application's log output.
-func NopLogger() Option {
-	return Logger(nopLogger{})
-}
+var NopLogger Option = Logger(nopLogger{})
 
 type nopLogger struct{}
 
--- a/app.go
+++ b/app.go
@@ -151,15 +151,14 @@ func Logger(p Printer) Option {
 	})
 }
 
-// NoopLogger is a no-op logger that implements the Printer interface.
-// When used with the Logger option, all Fx logs will be dropped.
-//
-// An application that wants to disable Fx logs should be created like so:
-//
-// app := fx.New(fx.Logger(fx.NoopLogger{}))
-type NoopLogger struct{}
+// NopLogger disables the application's log output.
+func NopLogger() Option {
+	return Logger(nopLogger{})
+}
 
-func (l NoopLogger) Printf(string, ...interface{}) {
+type nopLogger struct{}
+
+func (l nopLogger) Printf(string, ...interface{}) {
 	return
 }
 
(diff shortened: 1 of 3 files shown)


app.go
```diff
@@ -151,6 +151,18 @@ func Logger(p Printer) Option {
 	})
 }
 
+// NoopLogger is a no-op logger that implements the Printer interface.
+// When used with the Logger option, all Fx logs will be dropped.
+//
+// An application that wants to disable Fx logs should be created like so:
+//
+// app := fx.New(fx.Logger(fx.NoopLogger{}))
+type NoopLogger struct{}
```
breerly: I believe the convention is Nop* - https://golang.org/pkg/io/ioutil/#NopCloser

Let's rename to `NopLogger`.

app.go
```diff
@@ -151,6 +151,18 @@ func Logger(p Printer) Option {
 	})
 }
 
+// NoopLogger is a no-op logger that implements the Printer interface.
+// When used with the Logger option, all Fx logs will be dropped.
+//
+// An application that wants to disable Fx logs should be created like so:
+//
+// app := fx.New(fx.Logger(fx.NoopLogger{}))
```
breerly: Would a better experience be:

```
fx.New(fx.NopLogger)
```

In other words, we can make a global option called `NopLogger` instead of having the user instantiate an empty struct.

breerly:
Let's jam in person w/ @abhinav - I *think* `fx.LogLevel` might be a better abstraction, as it would solve the same problem as this and give us room to have different levels.

breerly:
Perhaps starting w/ this and adding LogLevel later wouldn't be the worst.

app.go
```diff
@@ -151,6 +151,17 @@ func Logger(p Printer) Option {
 	})
 }
 
+// NopLogger disables the application's log output.
+func NopLogger() Option {
+	return Logger(nopLogger{})
```
abhinav: I don't feel strongly about this but: I'm not sure this needs to be a function. It can just be a constant option.

```
app := fx.New(fx.NopLogger, ....)
```

app_test.go
```diff
@@ -347,3 +347,8 @@ func TestReplaceLogger(t *testing.T) {
 	app.RequireStart().RequireStop()
 	assert.Contains(t, spy.String(), "RUNNING")
 }
+
+func TestNoopLogger(t *testing.T) {
```
breerly: `TestNopLogger`*

PR #599, opened 2017-10-03
Add App.Execute for CLI use case

gonix:
This enables `fx.App` to be used more cleanly for CLI applications, e.g.
```go
func main() {
	if os.Args[1] == "foo" {
		newApp().Execute(func(foo Foo) {
			foo.Thing()
		})
	} else {
		newApp().Execute(func(bar Bar) {
			bar.Thing()
		})
	}
}
```

Instead of resorting to workarounds like:
```go
func main() {
	if os.Args[1] == "foo" {
		var foo Foo
		app := fx.App(
			// ... options
			fx.Invoke(func(f Foo) {
				foo = f
			}),
		)
		app.Start()
		foo.Thing()
		app.Stop()
	} else {
		var bar Bar
		app := fx.App(
			// ... options
			fx.Invoke(func(b Bar) {
				bar = b
			}),
		)
		app.Start()
		bar.Thing()
		app.Stop()
	}
}
```

### DIFF
--- a/app.go
+++ b/app.go
@@ -251,21 +251,61 @@ func (app *App) executeInvokes() error {
 //
 // See Start and Stop for application lifecycle details.
 func (app *App) Run() {
+	err := app.Execute(func() {
+		app.logger.Printf("RUNNING")
+		app.logger.PrintSignal(<-app.Done())
+	})
+	if err != nil {
+		// errors already logged in app.Execute()
+		os.Exit(1)
+	}
+}
+
+// Execute starts the application, invokes provided functions, and then
+// gracefully shuts the application down.
+//
+// Arguments for these functions are provided from the application's dependency
+// injection container.
+//
+// Invoked functions may have any number of returned values. If the final
+// returned object is an error, it's assumed to be a success indicator. All
+// other returned values are discarded. If one of invocations results in error,
+// the rest of the functions will not be invoked.
+//
+// Execute uses DefaultTimeout for the start and stop timeouts.
+//
+// See Start and Stop for application lifecycle details.
+func (app *App) Execute(funcs ...interface{}) (err error) {
 	startCtx, cancel := context.WithTimeout(context.Background(), defaultTimeout)
 	defer cancel()
 
-	if err := app.Start(startCtx); err != nil {
-		app.logger.Fatalf("ERROR\t\tFailed to start: %v", err)
+	if err = app.Start(startCtx); err != nil {
+		return
 	}
 
-	app.logger.PrintSignal(<-app.Done())
+	defer func() {
+		stopCtx, cancel := context.WithTimeout(context.Background(), defaultTimeout)
+		defer cancel()
 
-	stopCtx, cancel := context.WithTimeout(context.Background(), defaultTimeout)
-	defer cancel()
+		if stopErr := app.Stop(stopCtx); stopErr != nil {
+			app.logger.Printf("ERROR\t\tFailed to stop cleanly: %v", stopErr)
+			err = multierr.Append(err, stopErr)
+		}
+	}()
 
-	if err := app.Stop(stopCtx); err != nil {
-		app.logger.Fatalf("ERROR\t\tFailed to stop cleanly: %v", err)
+	for _, fn := range funcs {
+		fname := fxreflect.FuncName(fn)
+		app.logger.Printf("EXECUTE\t%s", fname)
+
+		err = app.container.Invoke(fn)
+
+		if err != nil {
+			app.logger.Printf("Error executing %q: %v", fname, err)
+			return
+		}
 	}
+
+	return
 }
 
 // Start executes all the OnStart hooks of the resolved object graph
@@ -340,7 +380,7 @@ func (app *App) start(ctx context.Context) error {
 		return err
 	}
 
-	app.logger.Printf("RUNNING")
+	app.logger.Printf("STARTED")
 	return nil
 }
 
--- a/app_test.go
+++ b/app_test.go
@@ -341,11 +341,191 @@ func TestAppStop(t *testing.T) {
 	})
 }
 
+func TestExecute(t *testing.T) {
+	type type1 struct{}
+	type type2 struct{}
+
+	t.Run("HappyPath", func(t *testing.T) {
+		var n int
+		hook := func(lc Lifecycle) {
+			lc.Append(Hook{
+				OnStart: func(context.Context) error {
+					n++
+					return nil
+				},
+				OnStop: func(context.Context) error {
+					n++
+					return nil
+				},
+			})
+		}
+		app := fxtest.New(t,
+			Provide(func() type1 { return type1{} }),
+			Provide(func() type2 { return type2{} }),
+			Invoke(hook),
+		)
+		err := app.Execute(
+			func(type1) { n++ },
+			func(type2) { n++ },
+		)
+		assert.NoError(t, err)
+		assert.Equal(t, 4, n)
+	})
+
+	t.Run("StartError", func(t *testing.T) {
+		var n int
+		failStart := func(lc Lifecycle) {
+			lc.Append(Hook{
+				OnStart: func(context.Context) error {
+					n++
+					return errors.New("OnStart fail")
+				},
+				OnStop: func(context.Context) error {
+					n++
+					return nil
+				},
+			})
+		}
+		app := fxtest.New(t,
+			Provide(func() type1 { return type1{} }),
+			Invoke(failStart),
+		)
+		err := app.Execute(
+			func(type1) {
+				require.FailNow(t, "Execute action must not be called")
+			},
+		)
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "OnStart fail")
+		assert.Equal(t, 2, n)
+	})
+
+	t.Run("ExecuteError", func(t *testing.T) {
+		var n int
+		hook := func(lc Lifecycle) {
+			lc.Append(Hook{
+				OnStart: func(context.Context) error {
+					n++
+					return nil
+				},
+				OnStop: func(context.Context) error {
+					n++
+					return nil
+				},
+			})
+		}
+		app := fxtest.New(t,
+			Provide(func() type1 { return type1{} }),
+			Provide(func() type2 { return type2{} }),
+			Invoke(hook),
+		)
+		execErr := errors.New("Execute fail")
+		err := app.Execute(
+			func(type1) error {
+				n++
+				return execErr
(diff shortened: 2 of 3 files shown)


gonix:
`fxreflect_test` failure on Go:tip seems to be not related to this diff...

marpetr:
The main benefit of this change is that it opens the possibility of invoking a function after the app is started. For us, this would be most useful in tests.

Consider a test body function that describes its dependencies via its arguments:
```go
func someTestFunction(a TypeA, b TypeB) {
  // assume that lifecycle is started
  // perform some tests on a and b
}
```

### Attempt 1
If we want Fx to inject the dependencies through reflection, we can do the following:
```go
app := fx.New(..., fx.Invoke(someTestFunction))
```

The problem is that `someTestFunction` will be invoked before the lifecycle is started.

### Attempt 2
As a workaround, `someTestFunction` could be implemented as:
```go
func someTestFunction(lifecycle fx.Lifecycle, a TypeA, b TypeB) {
  lifecycle.Append(Hook{
      // all previous lifecycle hooks have been started
      // here goes the actual code
    },
  })
}
``` 
For this to work, `fx.Invoke(someTestFunction)` must be the last option in `fx.New`.

### Attempt 3
In our project, we opted for the following solution. The dependencies are extracted through a temporary struct and are used after starting the app:
```go
var deps struct { DepA TypeA, DepB TypeB } // Field names must be exported!
app := fx.New(..., fx.Extract(&deps))
app.Start()
defer app.Stop()

// here goes the code that depends on deps.DepA and deps.DepB
```

The cons of this example is that this logic cannot be encapsulated and hidden away from users.

I feel that our use case could be solved with a smaller code change, perhaps just by allowing to invoke a function after the app is started. If `fx.App` exposes such a capability, the CLI/Execute use case could be implemented in a separate package ("go.uber.org/fx/fxcli") so as not to pollute the main package.

abhinav:
Hello. Thanks for your patience with the review latency, and thanks for
explaining the use case.

Minor clarification about Attempt 2: Everything that function depends on will
have been started by the time that hook is called so you don't necessarily
need to put it last. The reason you probably need to put it last is that
you're putting the actual logic in there. The expectation is that OnStart
starts the work without blocking and OnStop (if any) stops that work. This
assumption works for services but is problematic for command line
applications. I'll talk about this in a moment but first, I want to explain
why we can't accept this PR.

We intentionally chose to make the dependency container inaccessible once the
application had started. We want to discourage reliance on the container as
part of the request path. It will be very easy for someone to decide to do
this,

```go
func (h *myHandler) Handle(ctx context.Context, req *Request) (*Response, error) {
    h.app.Execute(func(...) {
    })
    // ..
}
```

Besides code hygiene concerns, this will negatively impact performance pretty
significantly. Documenting that this is a discouraged pattern isn't enough
because in our experience, most people skip over that part of the
documentation or choose to ignore it if their use case is satisfied.

Even if the API was accepted, on the code side, none of the internals for Fx
or the underlying library Dig are thread-safe. Single-threaded execution was
assumed because of the design decision outlined above. A significant number of
changes will be required to both, Fx and Dig to make something like this
possible, and I don't think that would be worth the effort required.

That said, your use case is valid and this isn't the first we've run into it.
This is on our backlog. See T1145851 for more information.

Currently, the almost-correct way to do this will look something like the
following. (I say almost-correct because I haven't verified if this code
works.)

```go
func main() {
    // We will receive errors encountered by `run()` here.
    done := make(chan error)
    
    app := fx.New(
        ...,
        fx.Invoke(func(a A, b, B, c C, lc fx.Lifecycle) {
            // Instead of doing any work right away, we schedule `run()` to be
            // called asynchronously and fill `done` with its result.
            lc.Append(fx.Hook{
                    go func() {
                        done <- run(a, b, c)
                    }()
                    return nil
                },
            })
        }),
    )

    // This will call the OnStart defined above, which will start `run()` in
    // the background.
    ctx := ...
    if err := app.Start(ctx); err != nil {
        log.Fatal(err)
    }

    // Wait until run() is done
    err := <-done
    
    // Stop everything
    ctx := ...
    app.Stop(ctx)
    
    // Non-zero exit code if `run` failed.
    if err != nil {
        log.Fatal(err)
    }
}

func run(a A, b B, c C) error {
    // main program logic goes here
}
```

We would like to add a clean porcelain on top of this that will let you pass
just the `run` function to Fx somehow and let it handle injecting the
dependencies, waiting for the function to finish, communicating the error up,
calling Stop, etc.  We haven't been able to invest much time into this and
don't have a very good idea of how we want this API to look just yet. (We're
happy to hear suggestions if you have any!)

For now, I suggest tracking progress for this feature in T1145851. We'll try
to prioritize the feature based on customer demand.

akshayjshah:
Everything that @abhinav points out is absolutely correct. I'd further point out that acceptable support for building CLI applications requires quite a bit more than this - at the very least, we need a reasonable way to integrate sub-commands, with distinct flags, into the `Run`/`Execute` portion of the app.

Basically, decent support for this use case will take enough new API surface that I'd prefer to make it a separate sub-package and think more carefully about it.

Also, sorry for the review delay!

gonix:
@abhinav I see your point about someone figuring out that calling `Execute` from "inside" is a good idea. I thought lifecycle management in `Execute` would be enough reason for people to understand that it is not a good idea but I agree that for someone who does not read the documentation this might lead to very painful learning experience.

That said I'm not too attached to the `Execute` API and may I suggest then an alternative API: introducing additional `fx.Option` e.g. `fx.Run` with usage very similar to `fx.Invoke` with the difference that `fx.Run` functions are invoked by `fx.App.Run` instead of default `app.logger.PrintSignal(<-app.Done())`. If no `fx.Run` options were provided `fx.App.Run` would fall back to the default behavior. Hope this makes sense.

What do you guys think?

@akshayjshah I don't see much reason for `fx.App` to handle or care about sub-commands and or flags. I think there are enough good packages for that purpose in the wild already. For example using `github.com/spf13/cobra` I imagine the use case would look something like this (ignoring `App.Execute` API issues for a moment):
```go
var rootCmd = &cobra.Command{
    // ...
    RunE: func(cmd *cobra.Command, args []string) error {
        return newApp().Execute(func(a A, b B) error {
            return a.Foo(b.Bar(args))
        }
    }
}

var subCmd = &cobra.Command{
    // ...
    RunE: func(cmd *cobra.Command, args []string) error {
        return newApp().Execute(func(b B) error {
            return b.Sub(cmd.Flags().GetString("foo"))
        }
    }
}
```
Or even less boilerplate with something like this:
```go
var rootCmd = &cobra.Command{
    // ...
    RunE: execute(func(a A, b B) error {
        return a.Foo(b.Bar(args))
    })
}

var subCmd = &cobra.Command{
    // ...
    RunE: execute(func(b B) error {
        return b.Sub(cmd.Flags().GetString("foo"))
    })
}

func execute(f interface{}) func(cmd *cobra.Command, args []string) error {
    return func(cmd *cobra.Command, args []string) error {
        // one could check for global cmd flags here as well
        return newApp().Execute(f)
    }
}
```
Do you have something different in mind?

marpetr:
It's possible to implement App.Execute through some use of reflection:

```go
package main

import (
  "context"
  "errors"
  "fmt"
  "os"
  "reflect"

  "go.uber.org/fx"
)

func fxExecute(f interface{}, opts ...fx.Option) (err error) {
  // Validate `f` argument
  v := reflect.ValueOf(f)
  if !v.IsValid() || v.IsNil() {
    err = errors.New("f must not be nil")
    return
  }
  t := v.Type()
  if t.Kind() != reflect.Func || t.NumOut() > 0 {
    err = errors.New("f must be a function with no return value")
    return
  }

  // Create `extractor` function of the same type.
  // It stores the invoke-time arguments into `inputs` slice.
  var inputs []reflect.Value
  extractor := reflect.MakeFunc(t, func(args []reflect.Value) (result []reflect.Value) {
    inputs = args
    return nil
  })

  // Create the app
  opts = append([]fx.Option{
    fx.Invoke(extractor.Interface()),
  }, opts...)
  app := fx.New(opts...)

  // Start the application, execute `f`, and then stop the application
  if err = app.Start(context.Background()); err != nil {
    return
  }
  defer func() {
    err = app.Stop(context.Background())
  }()

  v.Call(inputs)
  return
}

// Example code

type lifecycleStarted *bool

func sampleLifecycle(lifecycle fx.Lifecycle) lifecycleStarted {
  var started bool
  lifecycle.Append(fx.Hook{
    OnStart: func(context.Context) error { started = true; return nil },
    OnStop:  func(context.Context) error { started = false; return nil },
  })
  return &started
}

func main() {
  var module = fx.Option(
    fx.Provide(sampleLifecycle),
  )

  fxExecute(func(s lifecycleStarted) {
    fmt.Fprintln(os.Stderr, "Is lifecycle started?", *s)
  }, module)
}
```

The code produces the following output:
```
2017/11/16 13:54:42 [Fx] PROVIDE	main.lifecycleStarted <= main.sampleLifecycle()
2017/11/16 13:54:42 [Fx] PROVIDE	fx.Lifecycle <= vendor/go.uber.org/fx.New.func1()
2017/11/16 13:54:42 [Fx] INVOKE		reflect.makeFuncStub()
2017/11/16 13:54:42 [Fx] START		main.sampleLifecycle()
2017/11/16 13:54:42 [Fx] RUNNING
Is lifecycle started? true
2017/11/16 13:54:42 [Fx] STOP		main.sampleLifecycle()
```

The `fxExecute` function could be part of a separate package, maybe even fxtest?

marpetr:
@abhinav your almost-correct example should only call `run` only after the lifecycle is started in order to avoid races.

abhinav:
I agree that it's possible to do this without implementing a full CLI package
which knows/cares about the CLI flags or subcommands. I'm also aware that this
is possible to do with reflection. The implementation itself isn't a big
concern; it's the user-facing API. We haven't had time to invest into figuring
out the right user-facing API for something like fxExecute.

I'm open to adding something like fxExecute in just fxtest for now to make
testing slightly easier. It could be a new `fx.Option` implemented only in
`fxtest`, or a `Run` method on `fxtest.App`. This would need further
discussion from the rest of the team. Perhaps the comments section of this PR
isn't the right place to discuss that API. Can you create an issue here or on
Phab with your proposed strawman fxtest API? Thanks.

RE: calling `run` after lifecycle has started: All the dependencies declared
for `run` (as parameters) have already been started by this point. You don't
need to wait for the other components to start because everything you care
about is already running.

akshayjshah:
😶  Closing due to inactivity. 😶 

This is just housekeeping - leave a comment and re-open if necessary!

PR #605, opened 2017-12-05
Populate: Support populating any value without requiring structs

prashantv:
Extract requires users to create a struct, but this struct does not
match the standard fx.In struct that users use when using parameters.

Populate will allow users to get values from the dependency injection
container without creating an additional struct, and expects users to
use fx.In for getting multiple values in a struct or if they want to
get named values.

### DIFF
--- a/populate.go
+++ b/populate.go
@@ -25,8 +25,8 @@ import (
 	"reflect"
 )
 
-// Populate sets values with values from the dependency injection container
-// before application start. All arguments must be pointers to the values
+// Populate sets targets with values from the dependency injection container
+// before application start. All targets must be pointers to the values
 // that must be populated. Pointers to structs that embed In are supported,
 // which can be used to populate multiple values in a struct.
 func Populate(targets ...interface{}) Option {
@@ -34,11 +34,11 @@ func Populate(targets ...interface{}) Option {
 	targetTypes := make([]reflect.Type, len(targets))
 	for i, t := range targets {
 		if t == nil {
-			return invokeErr(fmt.Errorf("argument %v is nil", i+1))
+			return invokeErr(fmt.Errorf("failed to Populate: target %v is nil", i+1))
 		}
 		rt := reflect.TypeOf(t)
 		if rt.Kind() != reflect.Ptr {
-			return invokeErr(fmt.Errorf("argument %v is not a pointer type, got %T", i+1, t))
+			return invokeErr(fmt.Errorf("failed to Populate: target %v is not a pointer type, got %T", i+1, t))
 		}
 
 		targetTypes[i] = reflect.TypeOf(t).Elem()
--- a/populate.go
+++ b/populate.go
@@ -1,0 +1,69 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package fx
+
+import (
+	"fmt"
+	"reflect"
+)
+
+// Populate sets values with values from the dependency injection container
+// before application start. All arguments must be pointers to the values
+// that must be populated. Pointers to structs that embed In are supported,
+// which can be used to populate multiple values in a struct.
+func Populate(targets ...interface{}) Option {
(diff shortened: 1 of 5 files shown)


populate.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"fmt"
+	"reflect"
+)
+
+// Populate sets values with values from the dependency injection container
+// before application start. All arguments must be pointers to the values
+// that must be populated. Pointers to structs that embed In are supported,
+// which can be used to populate multiple values in a struct.
+func Populate(targets ...interface{}) Option {
```
abhinav: Pleas add usage samples or example tests before merging. We need to
demonstrate basic usage plus usage with `fx.In`s.

populate_test.go
```diff
@@ -0,0 +1,172 @@
...
+func TestPopulate(t *testing.T) {
+	type t1 struct{}
+	type t2 struct{}
+
+	t.Run("populate nothing", func(t *testing.T) {
+		app := fxtest.New(t,
+			Provide(func() *t1 { panic("should not be called ") }),
+			Populate(),
+		)
+		require.NoError(t, app.Err())
```
abhinav: would `app.RequireStart().RequireStop()` be preferable here?

populate_test.go
```diff
@@ -0,0 +1,172 @@
...
+		app := fxtest.New(t,
+			Provide(func() *t1 { return &t1{} }),
+			Provide(func() *t2 { return &t2{} }),
+
+			Populate(&targets),
+		)
+		require.NoError(t, app.Err())
+		require.NotNil(t, targets.V1, "did not populate field 1")
+		require.NotNil(t, targets.V2, "did not populate field 2")
+	})
```
abhinav: can we have one for populate value group too?

abhinav:
For other reviewers: This API will eventually deprecate `fx.Extract`.

populate.go
glibsm: worth mentioning that it mirrors `go.uber.org/config.Provider.Populate`? or nah?

populate.go
```diff
@@ -0,0 +1,69 @@
...
+	// Validate all targets are non-nil pointers.
+	targetTypes := make([]reflect.Type, len(targets))
+	for i, t := range targets {
+		if t == nil {
+			return invokeErr(fmt.Errorf("argument %v is nil", i+1))
```
glibsm: i think the error should start with `failed to Populate...`, no?

populate.go
```diff
@@ -0,0 +1,69 @@
...
+// THE SOFTWARE.
+
+package fx
+
+import (
+	"fmt"
+	"reflect"
+)
+
+// Populate sets values with values from the dependency injection container
```
amckinney: Populate sets targets ...*

populate.go
```diff
@@ -0,0 +1,69 @@
...
+// before application start. All arguments must be pointers to the values
```
amckinney: nit / not necessary: Consider changing "All arguments must be pointers..." to "All targets must be pointers..." to be consistent with the selected argument name.

populate.go
prashantv: 👍  good catch

populate.go
prashantv: 👍

populate.go
prashantv: Good idea, adding a simple example test, let me know if you want to make it more complex or if it's fine.

The intention was to keep this similar to `Populate`, but not sure it's worth calling that out (wanted a consistent experience, but they don't really need to be cross-referenced).

populate.go
prashantv: good idea, changed to "failed to Populate: %v", and also changed the messages to reference "target %v" rather than "argument %v"

populate_test.go
prashantv: Sure, can update all of them

populate_test.go
prashantv: Added tests for named and group.

glibsm:
Please update the changelog

populate_example_test.go
```diff
@@ -0,0 +1,45 @@
...
+	// Some external module that provides a user name.
+	type Username string
+	UserModule := fx.Provide(func() Username { return "john" })
+
+	// We want to get the user out of the dependency injection container.
+	var user Username
+	fx.New(
+		UserModule,
+
+		fx.Populate(&user),
```
abhinav: If we're showing a working example, we should have Start/Stop in there, but
doing that will make the example less readable IMO.

It may be preferable to have usage in the comments for Populate instead
because it lets us make assumptions about the rest of the code.

```
//   var conn *sql.DB
//   app := fx.New(
//     database.Module,
//     fx.Populate(&conn),
//   )
//   if err := app.Start(ctx); err !
//     return err
//   }
```

Thoughts?

prashantv: Hmm I like the idea of an example showing up in the godoc nicely, I can add start/stop at the end with comments.

prashantv:
Updated the changelog and mentioned `Start` in the example.

populate.go
```diff
@@ -0,0 +1,69 @@
...
+	//   *targets[1] = t2
+	//   [...]
+	// }
+	//
+	fnType := reflect.FuncOf(targetTypes, nil, false /* variadic */)
+	fn := reflect.MakeFunc(fnType, func(args []reflect.Value) []reflect.Value {
+		for i, arg := range args {
+			reflect.ValueOf(targets[i]).Elem().Set(arg)
+		}
+		return nil
```
vkuzmin-uber: "return nil" is not obvious here.  Doc says https://golang.org/pkg/reflect/#MakeFunc

- returns the results as a slice of Values, one per formal result.

Can you add some comment why "nil" is fine here or just explain it to me. Thanks.

akshayjshah: `nil` is a valid zero-length slice.

populate.go
```diff
@@ -0,0 +1,69 @@
...
+// Populate sets targets with values from the dependency injection container
+// before application start. All targets must be pointers to the values
+// that must be populated. Pointers to structs that embed In are supported,
+// which can be used to populate multiple values in a struct.
+func Populate(targets ...interface{}) Option {
+	// Validate all targets are non-nil pointers.
+	targetTypes := make([]reflect.Type, len(targets))
+	for i, t := range targets {
+		if t == nil {
+			return invokeErr(fmt.Errorf("failed to Populate: target %v is nil", i+1))
```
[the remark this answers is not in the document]
prashantv: "index" is also confusing since this is a method, not a slice. I originally had "argument", but considering we call them `targets` as the name, I thought this would be clearer.

I have no strong opinions either way, feel free to send a PR if you think this could be clearer.

PR #606, opened 2017-12-09
Improve documentation

akshayjshah:
Make a variety of improvements to Fx's documentation:

* Clarify and expand the docs for most types and methods, removing
  references to "dependency injection containers" from all types except
  `In` and `Out`.
* Migrate our internal getting started example to a package-level
  example. This sacrifices the step-by-step walkthrough, but makes the
  example much easier to find and validates its correctness in unit tests.
* Remove the example for `Extract`, since users should now prefer
  `Populate`. I've also added a note that we plan to deprecate Extract,
  but haven't actually added an official `staticcheck`-compatible
  deprecation.
* Fully document the advanced features offered by the `In` and `Out`
  types. This moves the docs for an important set of features to the
  our users' entry point. In a follow-up diff, I plan to replace the
  now-duplicate dig docs with a pointer to Fx's documentation.

### DIFF
--- a/app.go
+++ b/app.go
@@ -51,8 +51,8 @@ func (f optionFunc) apply(app *App) { f(app) }
 
 // Provide registers any number of constructor functions, teaching the
 // application how to instantiate various types. The supplied constructor
-// functions must return one or more objects, may depend on other types
-// available in the application, and may return an error. For example:
+// function(s) may depend on other types available in the application, must
+// return one or more objects, and may return an error. For example:
 //
 //  // Constructs type *C, depends on *A and *B.
 //  func(*A, *B) *C
@@ -66,11 +66,11 @@ func (f optionFunc) apply(app *App) { f(app) }
 //
 // The order in which constructors are provided doesn't matter, and passing
 // multiple Provide options appends to the application's collection of
-// constructors. Constructors are called lazily and their results are cached
-// for reuse (so instances of a type are effectively singletons within an
-// application). Taken together, these properties make it perfectly reasonable
-// to Provide a large number of constructors even if only a fraction of them
-// are used.
+// constructors. Constructors are called only if one or more of their returned
+// types are needed, and their results are cached for reuse (so instances of a
+// type are effectively singletons within an application). Taken together,
+// these properties make it perfectly reasonable to Provide a large number of
+// constructors even if only a fraction of them are used.
 //
 // See the documentation of the In and Out types for advanced features,
 // including optional parameters and named instances.
@@ -128,7 +128,7 @@ func (io invokeOption) String() string {
 	return fmt.Sprintf("fx.Invoke(%s)", strings.Join(items, ", "))
 }
 
-// Options composes a collection of Options into a single Option. This allows
+// Options converts a collection of Options into a single Option. This allows
 // packages to bundle sophisticated functionality into easy-to-use Fx modules.
 // For example, a logging package might export a simple option like this:
 //
@@ -156,6 +156,9 @@ func (io invokeOption) String() string {
 // line of code:
 //
 //  app := fx.New(server.Module)
+//
+// Use this pattern sparingly, since it limits the user's ability to customize
+// their application.
 func Options(opts ...Option) Option {
 	return optionGroup(opts)
 }
@@ -177,7 +180,7 @@ func (og optionGroup) String() string {
 }
 
 // Printer is the interface required by Fx's logging backend. It's implemented
-// by most loggers, including the standard library's.
+// by most loggers, including the one bundled with the standard library.
 type Printer interface {
 	Printf(string, ...interface{})
 }
@@ -208,10 +211,10 @@ func (l nopLogger) Printf(string, ...interface{}) {
 // New creates and initializes an App. All applications begin with a
 // constructor for the Lifecycle type already registered.
 //
-// In addition to that built-in functionality, users typically pass New a
-// handful of Provide options and one or more Invoke options. The Provide
-// options teach the application how to instantiate a variety of types, and
-// the Invoke options describe how to initialize the application.
+// In addition to that built-in functionality, users typically pass a handful
+// of Provide options and one or more Invoke options. The Provide options
+// teach the application how to instantiate a variety of types, and the Invoke
+// options describe how to initialize the application.
 //
 // When created, the application immediately executes all the functions passed
 // via Invoke options. To supply these functions with the parameters they
@@ -224,14 +227,16 @@ func (l nopLogger) Printf(string, ...interface{}) {
 // New returns and the application is ready to be started using Run or Start.
 // On startup, it executes any OnStart hooks registered with its Lifecycle.
 // OnStart hooks are executed one at a time, in order, and must all complete
-// within a configurable deadline (by default, 30 seconds).
+// within a configurable deadline (by default, 15 seconds). For details on the
+// order in which OnStart hooks are executed, see the documentation for the
+// Start method.
 //
 // At this point, the application has successfully started up. If started via
 // Run, it will continue operating until it receives a shutdown signal from
-// Done; if started explicitly via Start, it will operate until the user calls
-// Stop. On shutdown, OnStop hooks execute one at a time, in reverse order,
-// and must all complete within a configurable deadline (again, 30 seconds by
-// default).
+// Done (see the Done documentation for details); if started explicitly via
+// Start, it will operate until the user calls Stop. On shutdown, OnStop hooks
+// execute one at a time, in reverse order, and must all complete within a
+// configurable deadline (again, 15 seconds by default).
 type App struct {
 	err       error
 	container *dig.Container
@@ -276,8 +281,9 @@ func New(opts ...Option) *App {
 }
 
 // Run starts the application, blocks on the signals channel, and then
-// gracefully shuts the application down. It uses DefaultTimeout for the start
-// and stop timeouts. It's designed to make typical applications simple to run.
+// gracefully shuts the application down. It uses the default value of 15
+// seconds for the start and stop timeouts. It's designed to make typical
+// applications simple to run.
 //
 // However, all of Run's functionality is implemented in terms of the exported
 // Start, Done, and Stop methods. Applications with more specialized needs
@@ -335,7 +341,7 @@ func (app *App) Start(ctx context.Context) error {
 
 // Stop gracefully stops the application. It executes any registered OnStop
 // hooks in reverse order, so that each constructor's stop hooks are called
-// before its dependencies'.
+// before its dependencies' stop hooks.
 //
 // If the application didn't start cleanly, only hooks whose OnStart phase was
 // called are executed. However, all those hooks are executed, even if some
@@ -345,7 +351,9 @@ func (app *App) Stop(ctx context.Context) error {
 }
 
 // Done returns a channel of signals to block on after starting the
-// application.
+// application. Applications listen for the SIGINT and SIGTERM signals; during
+// development, users can send the application SIGTERM by pressing Ctrl-C in
+// the same terminal as the running process.
 func (app *App) Done() <-chan os.Signal {
 	c := make(chan os.Signal, 1)
 	signal.Notify(c, syscall.SIGINT, syscall.SIGTERM)
--- a/app.go
+++ b/app.go
@@ -38,7 +38,9 @@ import (
 
 const defaultTimeout = 15 * time.Second
 
-// An Option configures an App.
+// An Option configures an App using the functional options paradigm
+// popularized by Rob Pike. If you're unfamiliar with this style, see
+// https://commandcenter.blogspot.com/2014/01/self-referential-functions-and-design.html.
 type Option interface {
 	apply(*App)
 }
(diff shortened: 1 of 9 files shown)


app.go
```diff
@@ -47,29 +49,31 @@ type optionFunc func(*App)
 
 func (f optionFunc) apply(app *App) { f(app) }
 
-// Provide registers constructors with the application's dependency injection
-// container. Constructors provide one or more types, can depend on other
-// types available in the container, and may optionally return an error. For
-// example:
+// Provide registers any number of constructor functions, teaching the
+// application how to instantiate various types. The supplied constructor
+// functions must return one or more objects, may depend on other types
```
amckinney: nit: The supplied constructor function(s) ...

app.go
```diff
@@ -47,29 +49,31 @@ type optionFunc func(*App)
...
+// available in the application, and may return an error. For example:
```
amckinney: consider: might be best to lead in with "may depend on other types..." so that the notes concerning the return types are consecutive. i.e:

"The supplied constructor functions may depend on other types available in the application, must return one or more objects, and may return an error."

I realize you may have thought of this and ordered with "must" followed by "may" notes, but felt it was worth suggesting anyways. I'm fine with either.

app.go
```diff
@@ -160,7 +200,38 @@ func (l nopLogger) Printf(string, ...interface{}) {
...
-// An App is a modular application built around dependency injection.
+// An App is a modular application built around dependency injection. Most
+// users will only need to use the New constructor and the all-in-one Run
+// convenience method. In more unusual cases, users may need to use the Err,
+// Start, Done, and Stop methods by hand instead of relying on Run.
+//
+// New creates and initializes an App. All applications begin with a
+// constructor for the Lifecycle type already registered.
+//
+// In addition to that built-in functionality, users typically pass New a
```
amckinney: remove New

app.go
```diff
@@ -160,7 +200,38 @@ func (l nopLogger) Printf(string, ...interface{}) {
...
+// via Invoke options. To supply these functions with the parameters they
+// need, the application looks for constructors that return the appropriate
+// types; if constructors for any required types are missing or any
+// invocations return an error, the application will fail to start (and Err
+// will return a descriptive error message).
+//
+// Once all the invocations (and any required constructors) have been called,
+// New returns and the application is ready to be started using Run or Start.
+// On startup, it executes any OnStart hooks registered with its Lifecycle.
+// OnStart hooks are executed one at a time, in order, and must all complete
```
amckinney: it might be overkill, but we could reference what we mean by "in order" by referring to the docs for the `Start` method. I'll leave this up to you.

example_test.go
```diff
@@ -67,42 +155,69 @@ func NewMux(lc fx.Lifecycle, logger *log.Logger) *http.ServeMux {
...
-		// Provide all the constructors we need.
-		fx.Provide(NewLogger, NewHandler, NewMux),
-		// Before starting, register the handler. This forces resolution of all
-		// the types Register function depends on: *http.ServeMux and http.Handler.
-		// Since the mux is now being used, its startup hook gets registered
-		// and the application includes an HTTP server.
+		// Provide all the constructors we need, which teaches Fx how we'd like to
+		// construct the *log.Logger, http.Handler, and *http.ServeMux types.
+		// Remember that constructors are called lazily, so this block doesn't do
+		// anything on its own.
```
amckinney: optional: "anything on its own (aside from providing the constructors into the application)"

extract.go
```diff
@@ -30,10 +30,11 @@ import (
 var _typeOfIn = reflect.TypeOf(In{})
 
 // Extract fills the given struct with values from the dependency injection
-// container on application start.
+// container on application initialization. The target MUST be a pointer to a
+// struct. Only exported fields will be filled.
 //
-// The target MUST be a pointer to a struct. Only exported fields will be
-// filled.
+// Extract will be deprecated soon: use Populate instead, which doesn't
```
amckinney: Should we make the change necessary to deprecate Extract before adding these docs in? That way this can simply read: "Extract is deprecated: ..."

inout.go
```diff
@@ -22,18 +22,216 @@ package fx
...
+//     Conn  *sql.DB
+//     Cache *redis.Client `optional:"true"`
+//   }
+//
+// If an optional field isn't available in the container, the constructor
+// receives the field's zero value.
+//
+//   func NewUserGateway(p UserGatewayParams, log *log.Logger) (*UserGateway, error) {
+//     if p.Cache != nil {
+//       log.Print("Logging disabled")
```
amckinney: log.Print("Caching disabled")

amckinney:
Looks great so far! Thanks for this!

app.go
```diff
@@ -268,33 +301,45 @@ func (app *App) Run() {
...
+// encountered any errors in application initialization.
 func (app *App) Start(ctx context.Context) error {
 	return withTimeout(ctx, app.start)
 }
 
 // Stop gracefully stops the application. It executes any registered OnStop
-// hooks in reverse order (from the leaves of the dependency tree to the
-// roots), so that types are stopped before their dependencies.
+// hooks in reverse order, so that each constructor's stop hooks are called
+// before its dependencies'.
```
abhinav: > [..] dependencies' stop hooks.

inout.go
```diff
@@ -22,18 +22,216 @@ package fx
...
+// Some use cases call for multiple values of the same type. Fx will allow
+// multiple values of the same type in an application's dependency injection
+// container via `name:".."` tags parameter and result structs.
+//
+// A constructor that produces a result struct can tag any field with
+// `name:".."` to have the corresponding value added to the graph under the
+// specified name.
+//
+//   type ConnectionResult struct {
+//     fx.Out
```
abhinav: The docs for fx.In talking about result structs and fx.Out before that is ever mentioned doesn't seem right. 

Perhaps advanced usage should be documented in the package docs like dig, or this should mention "see [..] for more information on how to produce named values from constructors".

inout.go
```diff
@@ -22,18 +22,216 @@ package fx
...
+// To make it easier to produce and consume many values of the same type, Fx
+// supports value groups. Value groups allow constructors to send values to a
+// named, unordered collection in the container. Other constructors can
+// request all values in this collection as a slice.
+//
+// Constructors can send values into value groups by returning a result struct
+// tagged with `group:".."`.
+//
+//   type HandlerResult struct {
+//     fx.Out
```
abhinav: same concern as above

inout.go
akshayjshah: Makes sense. Discussed offline - plan it to separate the producer side (e.g., named value, value group) docs from the consume docs.

app.go
```diff
@@ -89,15 +93,23 @@ func (po provideOption) String() string {
...
+// run in order. Invocations may have any number of returned values. If the
+// final returned object is an error, it's assumed to be a success indicator.
+// All other returned values are discarded.
 //
-// Unlike constructors, invoked functions are always executed, and they're
-// always run in order. Invoked functions may have any number of returned
-// values. If the final returned object is an error, it's assumed to be a
-// success indicator. All other returned values are discarded.
+// Typically, invoked functions take a handful of high-level objects (whose
+// constructors depend on lower-level objects) and introduce them to each
```
glibsm: `and introduce them to each other` not sure what is meant by this

app.go
```diff
@@ -47,29 +49,31 @@ type optionFunc func(*App)
...
 //
-//  // Provides type *B and *C, depends on *A, and can fail.
+//  // Constructs types *B and *C, depends on *A, and can fail.
 //  func(*A) (*B, *C, error)
 //
-// The order in which constructors are provided doesn't matter. Constructors
-// are called lazily and their results are cached for reuse.
+// The order in which constructors are provided doesn't matter, and passing
+// multiple Provide options appends to the application's collection of
+// constructors. Constructors are called lazily and their results are cached
```
glibsm: `Constructors are called lazily`

I think it's worth to dive into something like `constructors are called when one or more of their return types are needed ...`

app.go
```diff
@@ -116,7 +128,34 @@ func (io invokeOption) String() string {
 	return fmt.Sprintf("fx.Invoke(%s)", strings.Join(items, ", "))
 }
 
-// Options composes a collection of Options into a single Option.
+// Options composes a collection of Options into a single Option. This allows
```
glibsm: `Options composes` reads weird

app.go
```diff
@@ -116,7 +128,34 @@ func (io invokeOption) String() string {
...
+// packages to bundle sophisticated functionality into easy-to-use Fx modules.
+// For example, a logging package might export a simple option like this:
+//
+//  package logging
+//
+//  var Module = fx.Provide(func() *log.Logger {
+//    return log.New(os.Stdout, "", 0)
+//  })
+//
+// A shared all-in-one microservice package could then use Options to bundle
```
glibsm: @abhinav just wrote a section on how not to do this :(

Is there a better example we can use for composition that doesn't cross library bounds?

akshayjshah: Doing this is basically the whole point - that's what `uberfx` is internally. Why not mention this?

app.go
akshayjshah: Sounds good, added a sentence pointing users there.

example_test.go
akshayjshah: Reworded to "doesn't do much".

extract.go
akshayjshah: I'd prefer not to. Internally, deprecating Extract will immediately break any build that's using the `staticcheck` linter, and I'd like to think more carefully about how to gracefully migrate those users without affecting stability over the holidays.

Basically, this is going to take some legwork, and I'd like to land these doc changes without blocking on that.

inout.go
akshayjshah: good catch!

app.go
```diff
@@ -137,7 +176,7 @@ func (og optionGroup) String() string {
 	return fmt.Sprintf("fx.Options(%s)", strings.Join(items, ", "))
 }
 
-// Printer is the interface required by fx's logging backend. It's implemented
+// Printer is the interface required by Fx's logging backend. It's implemented
 // by most loggers, including the standard library's.
```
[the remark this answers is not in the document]
akshayjshah: Sounds good. Reworded to "including the one bundled with the standard library" to avoid the now-overloaded "provide."

app.go
```diff
@@ -160,7 +200,38 @@ func (l nopLogger) Printf(string, ...interface{}) {
...
+// need, the application looks for constructors that return the appropriate
+// types; if constructors for any required types are missing or any
+// invocations return an error, the application will fail to start (and Err
+// will return a descriptive error message).
+//
+// Once all the invocations (and any required constructors) have been called,
+// New returns and the application is ready to be started using Run or Start.
+// On startup, it executes any OnStart hooks registered with its Lifecycle.
+// OnStart hooks are executed one at a time, in order, and must all complete
+// within a configurable deadline (by default, 30 seconds).
```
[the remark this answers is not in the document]
akshayjshah: :/ I've been so wrong for so long.

app.go
```diff
@@ -160,7 +200,38 @@ func (l nopLogger) Printf(string, ...interface{}) {
...
+//
+// At this point, the application has successfully started up. If started via
+// Run, it will continue operating until it receives a shutdown signal from
```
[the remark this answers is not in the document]
akshayjshah: Added those notes to the `Done` docs and added a "see the Done docs for details" note here.

app.go
```diff
@@ -160,7 +200,38 @@ func (l nopLogger) Printf(string, ...interface{}) {
...
+// Done; if started explicitly via Start, it will operate until the user calls
+// Stop. On shutdown, OnStop hooks execute one at a time, in reverse order,
+// and must all complete within a configurable deadline (again, 30 seconds by
+// default).
```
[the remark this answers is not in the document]
akshayjshah: ack.

app.go
akshayjshah: Sure, can do.

example_test.go
```diff
@@ -30,31 +30,119 @@ import (
...
+// know how to construct a logger and needs an HTTP handler, it will fail to
+// start.
+//
+// Functions may also return multiple objects. For example, we could combine
+// NewHandler and NewLogger into a single function:
+//
+//   func NewHandlerAndLogger() (*log.Logger, http.Handler, error)
+//
+// Fx also understands this idiom, and would treat NewHandlerAndLogger as the
+// constructor for both the *log.Logger and http.Handler types.
```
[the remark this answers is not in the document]
akshayjshah: sounds good.

app.go
akshayjshah: Happy to come back to the wording here as necessary; this proved much clearer in our internal "intro to Fx" documentation. This means, very broadly, that `Invoke` is typically used to take two types and make them interact with each other somehow. Often, they're taking the top-level user-defined objects (like HTTP handlers) and making them interact with the top-level type provided by some shared library (like a mux that's already bound to a port, with some common middleware applied).

app.go
akshayjshah: Reworded to "Options converts"/

app.go
akshayjshah: As always, this is nuanced - the pattern referenced here is fine, as long as it preserves the user's ability to break glass and consume each module separately. Abhinav's point is that some compositions of options make this quite a bit harder.

Rather than rewording this section, which is IMO the strongest selling point of Fx overall, I've added a more general "be cautious" note to the end.

app.go
akshayjshah: Done!

akshayjshah:
All comments should be addressed, though I expect us to continue iterating on these docs. Reviewers, please take another look 🙏

PR #611, opened 2018-01-04
CHANGELOG: Standardize format

abhinav:
This switches the CHANGELOG to the format defined at
http://keepachangelog.com/en/1.0.0/.

See also uber-go/dig#180. Unlike the Dig PR, only v1.0 onwards were
updated with Added/Removed/etc. information.

Preview at https://github.com/uber-go/fx/blob/standardize-changelog/CHANGELOG.md


abhinav:
CC @uber-go/fx

PR #612, opened 2018-01-04
Update LICENSE file and headers

abhinav:


LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2017 Uber Technologies, Inc.
+Copyright (c) 2017-2018 Uber Technologies, Inc.
```
breerly: Is this a standard practice?

glibsm: If we're going full coverage, first commits were in 2016.

`2016-2018`

abhinav: My understanding is that this is the more correct form. Our tooling doesn't support this for comments but we can at least manually do it for this file.

abhinav: Perhaps someone better informed would care to correct me on this. @kriskowal and @peter-edge did some research into this a while ago.

kriskowal: The correct practice is for the copyright to cover the entire range of active changes. Failing that, the strongest copyright is the oldest copyright. Updating it to the most recent year weakens our case, but we at least have version control to dig this up on a file-by-file basis.

A couple years ago, I updated uber-licence to strictly extend copyright ranges. It took about a week, and involved normalizing license comments to establish whether the existing comment was equivalent. https://github.com/uber/uber-licence

PR #616, opened 2018-02-18
Make it easier to set custom Start and Stop timeouts

vkuzmin-uber:
(Amended from @vkuzmin-uber's original summary.)

@vkuzmin-uber points out that Fx offers only two options for setting app start/stop timeouts: use the defaults or drop down to the `Start`/`Stop`/`Done` APIs directly. This PR offers a middle ground by introducing `StartTimeout` and `StopTimeout` options.

We discussed this quite a bit while designing the Fx 1.x APIs, and agreed to defer making a decision - we wanted a low-level API that let applications use contexts partially consumed by other libraries, but also reasonable behavior by default. Since it turns out that many internal applications need longer to start up but are otherwise happy with `Run`, introducing these options makes things a bit simpler for them.

As a nice bonus, this makes it possible for `fxtest` to respect the configured timeouts.

### DIFF
--- a/app.go
+++ b/app.go
@@ -36,7 +36,10 @@ import (
 	"go.uber.org/multierr"
 )
 
-const defaultTimeout = 15 * time.Second
+// DefaultTimeout is the default timeout for starting or stopping an
+// application. It can be configured with the StartTimeout and StopTimeout
+// options.
+const DefaultTimeout = 15 * time.Second
 
 // An Option configures an App using the functional options paradigm
 // popularized by Rob Pike. If you're unfamiliar with this style, see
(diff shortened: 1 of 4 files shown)


app.go
```diff
@@ -179,6 +179,36 @@ func (og optionGroup) String() string {
 	return fmt.Sprintf("fx.Options(%s)", strings.Join(items, ", "))
 }
 
+// OnStartTimeout changes the application's start timeout.
+func OnStartTimeout(v time.Duration) Option {
```
akshayjshah: I'd prefer `StartTimeout` and `StopTimeout`.

app.go
```diff
@@ -179,6 +179,36 @@ func (og optionGroup) String() string {
...
+	return app.onStartTimeout
+}
+
+// GetStopTimeout returns on start timeout.
+//
+// Most users won't need to use this method, because they may set timeout if necessary.
+// This can be useful for unit tests when need to read it to pass then to Stop method.
+func (app *App) GetStopTimeout() time.Duration {
+	return app.onStopTimeout
+}
```
akshayjshah: I'd prefer to remove these. They don't help users write meaningful tests - we're basically calling `SetFoo` and testing by calling `GetFoo`. Instead, we should test by setting the start timeout and verifying that we can actually block for however long is necessary.

app.go
```diff
@@ -238,12 +268,14 @@ func (l nopLogger) Printf(string, ...interface{}) {
...
-	invokes   []interface{}
-	logger    *fxlog.Logger
+	err            error
+	container      *dig.Container
+	lifecycle      *lifecycleWrapper
+	provides       []interface{}
+	invokes        []interface{}
+	logger         *fxlog.Logger
+	onStartTimeout time.Duration
+	onStopTimeout  time.Duration
```
akshayjshah: `startTimeout` and `stopTimeout`.

app.go
akshayjshah: Ah, I see - we need to export *some* way to get this information from `fxtest`.

app_test.go
```diff
@@ -172,6 +172,44 @@ func TestOptions(t *testing.T) {
...
 
+func TestTimeoutOptions(t *testing.T) {
+	const timeout = time.Minute
+	// Further assertions can't succeed unless the test timeout is greater than the default.
+	require.True(t, timeout > DefaultTimeout, "test assertions require timeout greater than default")
+
+	var started, stopped bool
+	assertCustomContext := func(ctx context.Context, phase string) {
+		deadline, ok := ctx.Deadline()
+		require.True(t, ok, "no %s deadline", phase)
```
abhinav: Using `require` here seems unsafe. It will call `t.Fatal` in case of failure
to kill the current goroutine, and with `app.Start`, that's not the test
goroutine.

fxtest/fxtest.go
```diff
@@ -64,7 +64,10 @@ func New(tb TB, opts ...fx.Option) *App {
 
 // RequireStart calls Start, failing the test if an error is encountered.
 func (app *App) RequireStart() *App {
-	if err := app.Start(context.Background()); err != nil {
+	startCtx, cancel := context.WithTimeout(context.Background(), app.StartTimeout())
```
abhinav: This bugfix will need its own changelog entry

app_test.go
akshayjshah: Good point, fixed.

akshayjshah:
Landing this. Perhaps because the PR is opened off a fork, the Travis push builds are failing. All the builds on this branch pass, though.

vkuzmin-uber:
Thank you. I appreciate it!

PR #622, opened 2018-04-24
Fix golint import

abhinav:
This fixes the import path for golint in the glide.yaml and Makefile,
and runs `glide up` on the repository.

The `glide up` required error message assertions in certain tests to be
changed. See corresponding commit for details.

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -16,6 +16,8 @@ testImport:
 - package: golang.org/x/tools
   subpackages:
   - cover
-- package: github.com/golang/lint
+- package: golang.org/x/lint
+  repo: https://github.com/golang/lint
+  vcs: git
   subpackages:
   - golint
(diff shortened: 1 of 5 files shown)


glide.yaml
```diff
@@ -16,6 +16,8 @@ testImport:
 - package: golang.org/x/tools
   subpackages:
   - cover
-- package: github.com/golang/lint
+- package: golang.org/x/lint
```
glibsm: i'm not clear on why this change is necessary, especially if we're under the hood doing the rewriting through glide

abhinav: We ran into this in YARPC and ThriftRW: With golang/lint@c363707,
golint imports another package from its repo using the golang.org/x/lint
import path.

If you use github.com/golang/lint, and try to build 
github.com/golang/lint/golint, it will fail because there's nothing
under vendor/golang.org/x/lint, which is what it's importing.

PR #623, opened 2018-04-24
Drop Go 1.8 support

abhinav:
This drops Go 1.8 support from Fx, allowing us to delete a hack that was
in place to make fx.Extract work with both, Go 1.8 and 1.9.

Note: We're quoting the Go version numbers in .travis.yml because 1.10
without quotes is interpreted as a float by YAML, giving us Go 1.1.


.travis.yml
```diff
@@ -3,8 +3,8 @@ sudo: false
 
 go_import_path: go.uber.org/fx
 go:
-  - 1.8
-  - 1.9
+  - "1.9"
+  - "1.10"
```
glibsm: strings necessary all of the sudden?

abhinav: Yeah, see PR description. Just `1.10` is interpreted as a float in YAML, so it becomes `1.1`.

PR #625, opened 2018-06-07
app: Add fx.Error option

amckinney:
It's often useful for libraries to create an `Option` that is added to user `App`s. If creating an `Option` depends on any sort of error-prone logic, there currently isn't a way to signal to the `App` that we failed to create an `Option`. In these cases, the library can ungracefully panic to prevent application startup, but this is less than ideal. 

For example,

```
type Foo struct{
  Bar string
}

func newOption(f Foo) Option {
  if f.Bar == "" {
    panic("%T does not have Bar set", f)
  }
  ...
}
```

It'd be useful to support an `Error` `Option` for this purpose specifically. The above example would change to:

```
func newOption(f Foo) Option {
  if f.Bar == "" {
    return Error(fmt.Errorf("%T does not have Bar set", f)
  }
  ...
}
```

### DIFF
--- a/app.go
+++ b/app.go
@@ -133,23 +133,14 @@ func (io invokeOption) String() string {
 
 // Error registers any number of errors with the application to short-circuit
 // startup. If more than one error is given, the errors are combined into a
-// single error, where each of the error messages are delimited by a newline.
+// single error.
 //
-// Similar to invocations, errors are applied in order. All Provide options
-// registered after an Error option will not be applied. Similarly, no
-// invokes will be run, including those that were registered before an Error.
+// Similar to invocations, errors are applied in order. All Provide and Invoke
+// options registered before or after an Error option will not be applied.
 func Error(errs ...error) Option {
-	return errOption(errs)
-}
-
-type errOption []error
-
-func (eo errOption) apply(app *App) {
-	app.err = multierr.Append(app.err, multierr.Combine(eo...))
-}
-
-func (eo errOption) String() string {
-	return fmt.Sprintf("fx.Error(%s)", multierr.Combine(eo...).Error())
+	return optionFunc(func(app *App) {
+		app.err = multierr.Append(app.err, multierr.Combine(errs...))
+	})
 }
 
 // Options converts a collection of Options into a single Option. This allows
--- a/app.go
+++ b/app.go
@@ -131,6 +131,27 @@ func (io invokeOption) String() string {
 	return fmt.Sprintf("fx.Invoke(%s)", strings.Join(items, ", "))
 }
 
+// Error registers any number of errors with the application to short-circuit
+// startup. If more than one error is given, the errors are combined into a
+// single error, where each of the error messages are delimited by a newline.
+//
+// Similar to invocations, errors are applied in order. All Provide options
+// registered after an Error option will not be applied. Similarly, no
+// invokes will be run, including those that were registered before an Error.
+func Error(errs ...error) Option {
+	return errOption(errs)
+}
+
+type errOption []error
+
+func (eo errOption) apply(app *App) {
+	app.err = multierr.Append(app.err, multierr.Combine(eo...))
+}
+
+func (eo errOption) String() string {
+	return fmt.Sprintf("fx.Error(%s)", multierr.Combine(eo...).Error())
+}
+
 // Options converts a collection of Options into a single Option. This allows
 // packages to bundle sophisticated functionality into easy-to-use Fx modules.
 // For example, a logging package might export a simple option like this:
--- a/app_test.go
+++ b/app_test.go
@@ -106,7 +106,7 @@ func TestError(t *testing.T) {
 
 		app := fxtest.New(t,
 			Error(nil),
-			Invoke(func() { invoked = true }),
+			Invoke(func() { invoked = true }), // should not be called
 		)
 		err := app.Err()
 		require.NoError(t, err)
@@ -118,7 +118,7 @@ func TestError(t *testing.T) {
 
 		app := fxtest.New(t,
 			Error(errors.New("module failure")),
-			Invoke(func() { invoked = true }),
+			Invoke(func() { invoked = true }), // should not be called
 		)
 		err := app.Err()
 		require.Error(t, err)
@@ -128,21 +128,27 @@ func TestError(t *testing.T) {
 
 	t.Run("MultipleErrorOption", func(t *testing.T) {
 		type A struct{}
-		panicFn := func(A) { panic("do not call me") }
+		var provided, invoked bool
 
 		app := fxtest.New(t,
-			Provide(func() A { return A{} }),
-			Invoke(panicFn),
+			Provide(
+				func() A { // should not be called
+					provided = true
+					return A{}
+				},
+			),
+			Invoke(func(A) { invoked = true }), // should not be called
 			Error(
 				errors.New("module A failure"),
 				errors.New("module B failure"),
 			),
-			Provide(panicFn),
 		)
 		err := app.Err()
 		require.Error(t, err)
 		assert.Contains(t, err.Error(), "module A failure")
 		assert.Contains(t, err.Error(), "module B failure")
+		assert.False(t, provided)
+		assert.False(t, invoked)
 	})
 }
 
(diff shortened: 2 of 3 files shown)


app.go
```diff
@@ -131,6 +131,27 @@ func (io invokeOption) String() string {
...
+
+type errOption []error
+
+func (eo errOption) apply(app *App) {
+	app.err = multierr.Append(app.err, multierr.Combine(eo...))
+}
+
+func (eo errOption) String() string {
+	return fmt.Sprintf("fx.Error(%s)", multierr.Combine(eo...).Error())
+}
```
akshayjshah: Do we need to implement `fmt.Stringer`? Most of the other options don't.

app.go
```diff
@@ -131,6 +131,27 @@ func (io invokeOption) String() string {
...
+// Error registers any number of errors with the application to short-circuit
+// startup. If more than one error is given, the errors are combined into a
+// single error, where each of the error messages are delimited by a newline.
+//
+// Similar to invocations, errors are applied in order. All Provide options
+// registered after an Error option will not be applied. Similarly, no
+// invokes will be run, including those that were registered before an Error.
+func Error(errs ...error) Option {
+	return errOption(errs)
+}
```
akshayjshah: I think this is simple enough that we can just use `optionFunc` instead of creating a new type.

app_test.go
```diff
@@ -100,6 +100,52 @@ func TestInvokes(t *testing.T) {
...
+		assert.False(t, invoked)
+	})
+
+	t.Run("MultipleErrorOption", func(t *testing.T) {
+		type A struct{}
+		panicFn := func(A) { panic("do not call me") }
+
+		app := fxtest.New(t,
+			Provide(func() A { return A{} }),
+			Invoke(panicFn),
```
akshayjshah: Can we leave a comment here explaining that this invocation shouldn't be executed?

app_test.go
```diff
@@ -100,6 +100,52 @@ func TestInvokes(t *testing.T) {
...
+			Error(
+				errors.New("module A failure"),
+				errors.New("module B failure"),
+			),
+			Provide(panicFn),
```
akshayjshah: If we're asserting that this option shouldn't have any effect, let's make an assertion to that effect.

app.go
```diff
@@ -131,6 +131,27 @@ func (io invokeOption) String() string {
 	return fmt.Sprintf("fx.Invoke(%s)", strings.Join(items, ", "))
 }
 
+// Error registers any number of errors with the application to short-circuit
+// startup. If more than one error is given, the errors are combined into a
+// single error, where each of the error messages are delimited by a newline.
```
akshayjshah: Don't mind taking varargs here, but let's not document the details of the error formatting - that's not part of the API contract.

This option is a little less self-explanatory than the others, so a small testable example would be nice.

app.go
amckinney: Works for me, I'm fine with leaving this out.

app.go
amckinney: Yea, now that we aren't using `String`, that sounds good.

app.go
amckinney: :+1:

app_test.go
```diff
@@ -100,6 +100,58 @@ func TestInvokes(t *testing.T) {
...
+		type A struct{}
+		var provided, invoked bool
+
+		app := fxtest.New(t,
+			Provide(
+				func() A { // should not be called
+					provided = true
+					return A{}
+				},
+			),
```
akshayjshah: Whitespace is a little odd here - can we put the `func() A` on the same line as the `Provide(`?

error_example_test.go
```diff
@@ -0,0 +1,48 @@
...
+		// so none of the Provide or Invoke options will be run.
+		fx.Error(errors.New("failed to build module")),
+		fx.Invoke(panicFn),
+	)
+
+	fmt.Println(app.Err())
+
+	// Output:
+	// failed to build module
+}
```
akshayjshah: Discussed offline - this example feels a little odd because there's no reason for the owner of `func main` to copy it. It might be better to write from the perspective of a module author. One possible example is starting an HTTP server and reading the port from the environment - if `$PORT` isn't set, we return an error instead of a `Provide`.

app_test.go
```diff
@@ -100,6 +100,57 @@ func TestInvokes(t *testing.T) {
 	})
 }
 
+func TestError(t *testing.T) {
+	t.Run("NilErrorOption", func(t *testing.T) {
+		var invoked bool
+
+		app := fxtest.New(t,
+			Error(nil),
+			Invoke(func() { invoked = true }), // should not be called
```
abhinav: nit: instead of tracking `invoked` like this, you can mark the test as failed.

```
Invoke(func() { t.Errorf("Invoke should not be called") }),
```

abhinav: Also, in this case, the comment appears to be incorrect. You *do* expect
Invoke to be called here. Tracking the bool for that case is fine.

error_example_test.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+func ExampleError() {
+	// A module that provides a HTTP server depends on
+	// the $PORT environment variable. If the variable
+	// is unset, the module returns an fx.Error option.
+	newHTTPServer := func() fx.Option {
+		port := os.Getenv("PORT")
+		if port == "" {
+			return fx.Error(errors.New("failed to build module: $PORT is not set"))
```
abhinav: nit: I would drop the "failed to build module" prefix.

app.go
```diff
@@ -131,6 +131,18 @@ func (io invokeOption) String() string {
...
 
+// Error registers any number of errors with the application to short-circuit
+// startup. If more than one error is given, the errors are combined into a
+// single error.
+//
+// Similar to invocations, errors are applied in order. All Provide and Invoke
+// options registered before or after an Error option will not be applied.
+func Error(errs ...error) Option {
+	return optionFunc(func(app *App) {
+		app.err = multierr.Append(app.err, multierr.Combine(errs...))
```
prashantv: When `Error` is applied, does it end up stopping the rest of the constructors/invokes? Want to make sure we don't see errors about missing types (because there was some other error instead)

can we verify that through unit tests for both `Provide` and `Invoke` relying on a missing type

error_example_test.go
```diff
@@ -0,0 +1,55 @@
...
+		}
+		return fx.Provide(&http.Server{
```
prashantv: should be `fx.Provide(func() *http.Server { ... })`

app_test.go
amckinney: Good catch, updated.

error_example_test.go
amckinney: Done

app.go
amckinney: As the comment describes, all `Provide` and `Invoke` options registered before/after an `Error` option will not be applied. The `SingleErrorOption` and `MultipleErrorOption` sub-tests in `TestError` both test this behavior. 

I will add tests focused around a `Provide` and `Invoke` missing dependencies to further verify this.

app_test.go
```diff
@@ -100,6 +100,70 @@ func TestInvokes(t *testing.T) {
...
+			Error(
+				errors.New("module A failure"),
+				errors.New("module B failure"),
+			),
+		)
+		err := app.Err()
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "module A failure")
+		assert.Contains(t, err.Error(), "module B failure")
+		assert.NotContains(t, err.Error(), "fx_test.A is not in the container")
```
prashantv: just `not in the container` is probably enough

app_test.go
```diff
@@ -100,6 +100,70 @@ func TestInvokes(t *testing.T) {
...
+				t.Errorf("B is missing from the container; Provide should not be called")
+				return A{}
+			},
+			),
+			Error(errors.New("module failure")),
+			Invoke(func(A) { t.Errorf("A was not provided; Invoke should not be called") }),
+		)
+		err := app.Err()
+		require.Error(t, err)
+		assert.Equal(t, err.Error(), "module failure")
```
prashantv: can combine require + assert.Equal using `assert.EqualError`

PR #630, opened 2018-07-18
dot graph: VisualizeError on invoke failure using ErrorHook option

leyao:
This PR includes changes to fx to use the new features in the `feature/graphviz` branch in `uber-go/dig`. On `Invoke` failures, we pass the error to the registered error handlers provided by users through `fx.ErrorHook`. This allows users to get a graph representation on what's causing the invoke failure. The error handlers can then process the information accordingly (eg. print to log or write to file).
* Updated `glide.yaml` and `glide.lock` to use `go.uber.org/dig` feature branch.
* Added `ErrorHook` option to `App` to register error handlers.
* `ErrorHandler`s handle the error produced on invoke failure. The error is wrapped in a wrapper that also contains the DOT format graph generated by `dig.Visualize`.

### DIFF
--- a/app.go
+++ b/app.go
@@ -294,13 +294,13 @@ type ErrorHandler interface {
 
 type errorHookOption []ErrorHandler
 
-type errorHandlerList []ErrorHandler
-
 func (eho errorHookOption) apply(app *App) {
 	app.errorHooks = append(app.errorHooks, eho...)
 }
 
-func (ehl errorHandlerList) onError(err error) {
+type errorHandlerList []ErrorHandler
+
+func (ehl errorHandlerList) HandleError(err error) {
 	for _, eh := range ehl {
 		eh.HandleError(err)
 	}
@@ -337,16 +337,17 @@ func New(opts ...Option) *App {
 
 	if err := app.executeInvokes(); err != nil {
 		app.err = err
-		var b bytes.Buffer
 
 		if dig.CanVisualizeError(err) {
+			var b bytes.Buffer
 			dig.Visualize(app.container, &b, dig.VisualizeError(err))
+			err = errorWithGraph{
+				graph: b.String(),
+				err:   err,
+			}
 		}
 
-		errorHandlerList(app.errorHooks).onError(errorWithGraph{
-			graph: b.String(),
-			err:   err,
-		})
+		errorHandlerList(app.errorHooks).HandleError(err)
 	}
 
 	return app
@@ -374,10 +375,8 @@ func (err errorWithGraph) Error() string {
 
 // VisualizeError returns the visualization of the error if available.
 func VisualizeError(err error) (string, error) {
-	if e, ok := err.(errWithGraph); ok {
-		if e.Graph() != "" {
-			return string(e.Graph()), nil
-		}
+	if e, ok := err.(errWithGraph); ok && e.Graph() != "" {
+		return string(e.Graph()), nil
 	}
 	return "", errors.New("unable to visualize error")
 }
--- a/app.go
+++ b/app.go
@@ -352,6 +352,7 @@ func New(opts ...Option) *App {
 	return app
 }
 
+// DotGraph is a Graphviz DOT format graph.
 type DotGraph string
 
 type errWithGraph interface {
--- a/app.go
+++ b/app.go
@@ -21,6 +21,7 @@
 package fx
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"os"
@@ -275,6 +276,32 @@ type App struct {
 	logger       *fxlog.Logger
 	startTimeout time.Duration
 	stopTimeout  time.Duration
+	errorHooks   errorHookOption
+}
+
+// ErrorHook registers error handlers that implement error handling functions.
+// They are executed on invoke failures. Passing multiple ErrorHandlers appends
+// the new handlers to the application's existing list.
+func ErrorHook(funcs ...ErrorHandler) Option {
+	return errorHookOption(funcs)
+}
+
+// ErrorHandler implements HandleError. They are used as error hooks and are
+// called on invoke errors.
+type ErrorHandler interface {
+	HandleError(error)
+}
+
+type errorHookOption []ErrorHandler
+
+func (eho errorHookOption) apply(app *App) {
+	app.errorHooks = append(app.errorHooks, eho...)
+}
+
+func (eho errorHookOption) onError(err error) {
+	for _, eh := range eho {
+		eh.HandleError(err)
+	}
 }
 
 // New creates and initializes an App, immediately executing any functions
@@ -308,9 +335,44 @@ func New(opts ...Option) *App {
 
 	if err := app.executeInvokes(); err != nil {
 		app.err = err
+		var b bytes.Buffer
+
+		if dig.CanVisualizeError(err) {
+			dig.Visualize(app.container, &b, dig.VisualizeError(err))
+		}
+
+		app.errorHooks.onError(wrappedErr{
+			graph: b.String(),
+			err:   err,
+		})
 	}
 
 	return app
+}
+
+type wrappedErr struct {
+	graph string
+	err   error
+}
+
+type isWrapped interface {
+	Graph() string
+}
+
+func (we wrappedErr) Graph() string {
+	return we.graph
+}
+
+func (we wrappedErr) Error() string {
+	return we.err.Error()
+}
+
+// VisualizeError returns the visualization of the error if available.
+func VisualizeError(err error) string {
+	if e, ok := err.(isWrapped); ok {
+		return e.Graph()
+	}
+	return ""
 }
 
 // Run starts the application, blocks on the signals channel, and then
--- a/app.go
+++ b/app.go
@@ -23,6 +23,7 @@ package fx
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"os/signal"
@@ -286,10 +287,10 @@ func ErrorHook(funcs ...ErrorHandler) Option {
 	return errorHookOption(funcs)
 }
 
-// ErrorHandler implements HandleError. They are used as error hooks and are
-// called on invoke errors.
+// ErrorHandler implements Handle. They are used as error hooks and are called
+// on invoke errors.
 type ErrorHandler interface {
-	HandleError(error)
+	Handle(error)
 }
 
 type errorHookOption []ErrorHandler
@@ -300,7 +301,7 @@ func (eho errorHookOption) apply(app *App) {
 
 func (eho errorHookOption) onError(err error) {
 	for _, eh := range eho {
-		eh.HandleError(err)
+		eh.Handle(err)
 	}
 }
 
@@ -368,11 +369,13 @@ func (we wrappedErr) Error() string {
 }
 
 // VisualizeError returns the visualization of the error if available.
-func VisualizeError(err error) string {
+func VisualizeError(err error) (string, error) {
 	if e, ok := err.(isWrapped); ok {
-		return e.Graph()
+		if e.Graph() != "" {
+			return e.Graph(), nil
+		}
 	}
-	return ""
+	return "", errors.New("unable to visualize error")
 }
 
 // Run starts the application, blocks on the signals channel, and then
--- a/app.go
+++ b/app.go
@@ -277,7 +277,7 @@ type App struct {
 	logger       *fxlog.Logger
 	startTimeout time.Duration
 	stopTimeout  time.Duration
-	errorHooks   errorHookOption
+	errorHooks   []ErrorHandler
 }
 
 // ErrorHook registers error handlers that implement error handling functions.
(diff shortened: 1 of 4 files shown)


app.go
```diff
@@ -308,11 +335,46 @@ func New(opts ...Option) *App {
...
+func (we wrappedErr) Error() string {
+	return we.err.Error()
+}
+
+// VisualizeError returns the visualization of the error if available.
+func VisualizeError(err error) string {
+	if e, ok := err.(isWrapped); ok {
+		return e.Graph()
+	}
+	return ""
```
AlexAPB: I'd think that as a user of this library, if the error is not visualizable, I'd like some indication rather than an empty string -- however, I'm not entirely sure of the precedents set by the rest of `Fx`.

Maybe this function should also return an error, or at least a standard message like:
`"cannot visualize error: ..."`

app_test.go
```diff
@@ -84,6 +84,16 @@ func TestNewApp(t *testing.T) {
...
 }
 
+var errCount = 0
+
+type testErrHandler struct {
+	count int
+}
+
+func (eh testErrHandler) HandleError(err error) {
+	errCount++
```
AlexAPB: I believe that you intended to have `eh.count++` and remove the global `errCount` var.
Then your test below can modify the assertion to be `assert.Equal(t, 1, errHandler.count)`

app_test.go
```diff
@@ -473,3 +493,66 @@ func TestNopLogger(t *testing.T) {
...
+		assert.Equal(t, "", VisualizeError(err))
+	})
+
+	t.Run("WrappedError", func(t *testing.T) {
+		err := testWrappedErr{}
+		assert.Equal(t, "graph", VisualizeError(err))
+	})
+}
+
+var errStr, graphStr string
```
AlexAPB: I'm never a fan of globals, even in tests. Leaving globals around may cause future tests to rely on previous behviour which is tricky to debug.

What if we made these fields on `graphErrHandler` and do the assertions directly on the `errHandler` you define?

app_test.go
leyao: I changed both the counter and the strings to fields in the handlers 😬

app.go
leyao: Good call. I've changed the function to return an error if there is no graph or if the graph is empty. That actually eliminates some of the checks needed when I'm calling it elsewhere.

breerly:
Mind updating the PR description to include the rationale, as well as a screenshot showing what users will see in the console?

app.go
```diff
@@ -275,6 +277,32 @@ type App struct {
...
+// They are executed on invoke failures. Passing multiple ErrorHandlers appends
+// the new handlers to the application's existing list.
+func ErrorHook(funcs ...ErrorHandler) Option {
+	return errorHookOption(funcs)
+}
+
+// ErrorHandler implements Handle. They are used as error hooks and are called
+// on invoke errors.
+type ErrorHandler interface {
+	Handle(error)
```
AlexAPB: So I know I suggested that you rename this from `HandleError` to `Handle`, but an offline discussion with @abhinav made me realize that you had it right the first time :p

Naming this `Handle(error)` means that if you ever intend to have additional `Handle` methods, you'll lose naming consistency since the 'default' `Handle()` method is only for errors. Since structs implicitly implement interfaces and we don't have function overloading, calling this `Handle(error)` is likely undesirable.

sorry :/

leyao:
Hey @breerly, I've updated the description to include more information. There is no update to what users will see in the console in this PR. The graph will be passed to the error handlers provided by users, so they decide what to print to the console.

app.go
```diff
@@ -275,6 +277,32 @@ type App struct {
...
+
+// ErrorHook registers error handlers that implement error handling functions.
+// They are executed on invoke failures. Passing multiple ErrorHandlers appends
+// the new handlers to the application's existing list.
+func ErrorHook(funcs ...ErrorHandler) Option {
+	return errorHookOption(funcs)
+}
+
+// ErrorHandler implements Handle. They are used as error hooks and are called
+// on invoke errors.
```
abhinav: outdated comment (handle vs handle error). How about just,

    // ErrorHandler handles Fx application startup errors.

app.go
```diff
@@ -275,6 +277,32 @@ type App struct {
...
+	HandleError(error)
+}
+
+type errorHookOption []ErrorHandler
+
+func (eho errorHookOption) apply(app *App) {
+	app.errorHooks = append(app.errorHooks, eho...)
+}
+
+func (eho errorHookOption) onError(err error) {
```
abhinav: if possible, i'd like to avoid putting the actual business logic on the Option object. We should just fill the list of handlers on the app and let it handle it. 

```
type app struct {
  ...
  errorHooks []ErrorHandler
}
```

Or better, declare a `errorHandlerList` which has this method.

```
type errorHandlerList []ErrorHandler

func (eh errorHandlerList) HandleError(err error) {
  ...
}

// In app, do errorHandlerList(app.errorHandlers).Handle(err).
```

This is a little extra code but it's preferable to keep the options separated.

app.go
```diff
@@ -308,11 +336,48 @@ func New(opts ...Option) *App {
...
+		app.errorHooks.onError(wrappedErr{
+			graph: b.String(),
+			err:   err,
+		})
 	}
 
 	return app
 }
 
+type wrappedErr struct {
```
abhinav: how about visualizedError or errorWithGraph?

app.go
```diff
@@ -308,11 +336,48 @@ func New(opts ...Option) *App {
...
+	graph string
+	err   error
+}
+
+type isWrapped interface {
+	Graph() string
+}
```
abhinav: We can drop this interface and cast the `err` directly to `wrappedErr` (whatever its name ends up being).

```
if e, ok := err.(wrappedErr); ok {
  return e.graph
}
```

app_test.go
```diff
@@ -84,6 +84,14 @@ func TestNewApp(t *testing.T) {
 	})
 }
 
+type testErrHandler struct {
+	count int
+}
+
+func (eh *testErrHandler) HandleError(err error) {
+	eh.count++
```
abhinav: probably easier to test the error handler by building a func-based implementation.

```
type errorHandlerFunc func(error)

func (f errorHandlerFunc) HandleError(err error) { f(err) }
```

Then in the test where you're looking at the count, you can do,

```
count := 0
h := errHandlerFunc(func(err error) {
  count++
  assert stuff about the error here
})
assert.Equal(t, 1, count)
```

app_test.go
```diff
@@ -473,3 +491,77 @@ func TestNopLogger(t *testing.T) {
 	app := fxtest.New(t, NopLogger)
 	app.RequireStart().RequireStop()
 }
+
+type testWrappedErr struct {
+	graph string
+}
+
```
abhinav: We don't need to support custom implementations of wrappedError IMO

app_test.go
```diff
@@ -473,3 +491,77 @@ func TestNopLogger(t *testing.T) {
...
+func (we testWrappedErr) Error() string {
+	return "great sadness"
+}
+
+func TestVisualizeError(t *testing.T) {
+	t.Run("NotWrappedError", func(t *testing.T) {
+		graph, err := VisualizeError(errors.New("great sadness"))
+		assert.Equal(t, "", graph)
```
abhinav: optional: can use assert.Empty

app_test.go
```diff
@@ -473,3 +491,77 @@ func TestNopLogger(t *testing.T) {
...
+}
+
+type graphErrHandler struct {
+	errStr   string
+	graphStr string
+}
+
+func (eh *graphErrHandler) HandleError(err error) {
+	eh.errStr = err.Error()
+	eh.graphStr, _ = VisualizeError(err)
```
abhinav: if you implement errorHandlerFunc from above, this becomes unnecessary

app_test.go
```diff
@@ -473,3 +491,77 @@ func TestNopLogger(t *testing.T) {
...
+
+		errHandler := graphErrHandler{}
+		fxtest.New(t,
+			Provide(func() (B, error) { return B{}, fmt.Errorf("great sadness") }),
+			Provide(func(B) A { return A{} }),
+			Invoke(func(A) {}),
+			ErrorHook(&errHandler),
+		)
+		assert.Contains(t, errHandler.graphStr, "\"fx_test.B\" [color=red];")
+		assert.Contains(t, errHandler.graphStr, "\"fx_test.A\" [color=orange];")
```
abhinav: instead of escaping double quotes, you can use backticks to let you write double quotes as-is

    assert.Contains(t, ..., `"fx_test.B" [color=red];`)

app_test.go
abhinav: Actually, it looks like testing is easier like this. In that case, can we rename the interface to `errorWithGraph` or something similar. "Wrapped" is too generic.

Also, in that case, we should maybe rename `Graph()` to something more verbose to avoid conflicts. Alternatively, it can return a `type GraphvizGraph string` instead of `string`, which guarantees that there are no conflicts with any other `error` implementation that happens to have a `Graph() string` method.

app.go
leyao: Added an `errorHandlerList` type that implements the `onError` function.

app.go
leyao: `errorWithGraph` sounds good, I'll rename it to that 😄

app.go
leyao: The cast to interface is needed for the tests to work, since I can't create a `wrappedErr` in the tests (they're in the `fx_test` package)

app_test.go
leyao: Cool, never thought that it can be done like that 👍

app_test.go
leyao: I created `type DotGraph string` and changed the function to be `func Graph() DotGraph`

app_test.go
leyao: Sure that's a lot neater.

app_test.go
leyao: That makes it a lot more readable 😄

app.go
```diff
@@ -287,20 +287,21 @@ func ErrorHook(funcs ...ErrorHandler) Option {
...
-// on invoke errors.
+// ErrorHandler handles Fx application startup errors.
 type ErrorHandler interface {
 	HandleError(error)
 }
 
 type errorHookOption []ErrorHandler
 
+type errorHandlerList []ErrorHandler
+
```
abhinav: nit: move errorHandlerList next to its method, and errorHookOption next to its method.

We generally go for, type, constructor (if any), and methods when ordering them.

app.go
```diff
@@ -308,11 +337,51 @@ func New(opts ...Option) *App {
...
 		app.err = err
+		var b bytes.Buffer
+
+		if dig.CanVisualizeError(err) {
+			dig.Visualize(app.container, &b, dig.VisualizeError(err))
+		}
+
+		errorHandlerList(app.errorHooks).onError(errorWithGraph{
+			graph: b.String(),
+			err:   err,
```
abhinav: We probably want to call `onError` with the original error if no wrapping was
done instead of wrapping with an empty buffer.

Let's change this to,

```
if dig.CanVisualizeError(err) {
    var b bytes.Buffer
    dig.Visualize(..)
    err = errorWithGraph{...}
}

errorHandlerList(app.errorHooks).onError(err)
```

app.go
```diff
@@ -308,11 +337,51 @@ func New(opts ...Option) *App {
...
+}
+
+func (err errorWithGraph) Error() string {
+	return err.err.Error()
+}
+
+// VisualizeError returns the visualization of the error if available.
+func VisualizeError(err error) (string, error) {
+	if e, ok := err.(errWithGraph); ok {
+		if e.Graph() != "" {
```
abhinav: optional nit: can collapse if

```
if e, ok := err.(errWithGraph); ok && e.Graph() != "" {
  ..
}
```

app_test.go
```diff
@@ -473,3 +491,76 @@ func TestNopLogger(t *testing.T) {
...
+}
+
+func (we testErrorWithGraph) Error() string {
+	return "great sadness"
+}
+
+func TestVisualizeError(t *testing.T) {
+	t.Run("NotWrappedError", func(t *testing.T) {
+		graph, err := VisualizeError(errors.New("great sadness"))
+		assert.Empty(t, graph)
```
abhinav: we don't need to assert that the graph is empty as long as error is not nil.

The contract for functions that return an error is that if err is not nil, the
other value is meaningless/can be anything.

app_test.go
```diff
@@ -473,3 +491,76 @@ func TestNopLogger(t *testing.T) {
...
+			errStr = err.Error()
+			graphStr, _ = VisualizeError(err)
+		})
+		fxtest.New(t,
+			Provide(func() A { return A{} }),
+			Invoke(func(A) error { return fmt.Errorf("great sadness") }),
+			ErrorHook(h),
+		)
+		assert.Equal(t, "great sadness", errStr)
+		assert.Empty(t, graphStr)
```
abhinav: In this case we should get the original error back (after you make the change
suggested above) so we can just do `assert.Equal` on the original error object.

Issue #610, opened 2017-12-21
Injecting Named Values without rewriting construtors


abhinav:
Based on offline discussions, we want to support the ability to inject a named
value into an Fx application without rewriting or wrapping an existing
constructor to return an `fx.Out` struct.

Specifically, given the constructor,

```go
func NewThing(...) (*Thing, error)
```

We want users to be able to `fx.Provide` a `*Thing` with the name "foo"
without having to write the following.

```go
type thingResult struct {
    fx.Out
    
    Thing *Thing `name:"foo"`
}

func newThing(...) (thingResult, error) {
    thing, err := NewThing(...)
    return thingResult{Thing: thing}, err
}
```

Implementing support for this will become significantly easier once uber-go/dig#175
lands but we still need to decide on the user-facing API for this.

Based on an offline discussion with @prashantv, these are the options we are
considering for the Fx-level API for this functionality.

-   ### `fx.Annotated` struct

    Users will `fx.Provide` this struct instead of their constructor.

    ```go
    fx.Provide(
        ...,
        fx.Annotated{
            Name: "foo",
            Fn: NewThing,
        },
    ),
    ```

    **Pros:**

    -   Idiomatic in Go
    -   Easy to add new options like `Group: "foo"`

    **Cons:**

    -   If we ever decide to support providing things besides constructors,
        we're stuck with the name `Fn`.

-   ### Builder Pattern
 
    Users will call `fx.Annotate` on a constructor and receive an opaque
    `fx.Annotated` object which they can call `Name` on.

    ```go
    fx.Provide(
        ...,
        fx.Annotate(fn).Name("foo"),
    )
    ```

    **Pros:**

    -   Easy to add new options: `fx.Annotate(fn).Name("foo").Group("bar")`
    -   No naming commitments for things that can be `Provide`d

    **Cons:**

    -   Builder Pattern is non-standard in Go: we don't know of any production
        APIs from the Go team that use the Builder Pattern

-   ### Decorator Functions

    Users will call `fx.Named`, providing a name and a constructor.

    ```go
    fx.Provide(
        ...,
        fx.Named("foo", fn),
    )
    ```

    **Pros:**

    -   Idiomatic in Go
    -   No naming commitments for things that can be `Provide`d

    **Cons:**

    -   Namespace pollution: adding a Group option requires taking up
        `fx.Grouped`
    -   Chaining options is annoying: `fx.Grouped("foo", fx.Named("bar", fn))`

[...]

PR #633, opened 2018-07-20
Implement struct fx.Annotated to inject named values without relying on fx.Out

dimiro1:
See: https://github.com/uber-go/fx/issues/610

This is a version for review, there are a few things I  would like to receive feedback.

- I am not familiar with the codebase, so, I am not sure if the implementation in the *app is the best possible implementation;
- I needed to update dig to use master, because of the dig.Name was not released yet;
- I still need to add documentation;
- There are a couple of questions, how to deal with Optional dependencies? and How to implement groups?

### DIFF
--- a/annotated_test.go
+++ b/annotated_test.go
@@ -1,0 +1,39 @@
+package fx_test
+
+import (
+	"testing"
+	"go.uber.org/fx/fxtest"
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/fx"
+)
+
+func TestAnnotated(t *testing.T) {
+	type a struct {
+		name string
+	}
+
+	type in struct {
+		fx.In
+
+		A *a `name:"foo"`
+	}
+	newA := func() *a {
+		return &a{name: "foo"}
+	}
+
+	t.Run("Provided", func(t *testing.T) {
+		app := fxtest.New(t,
+			fx.Provide(
+				fx.Annotated{
+					Name: "foo",
+					Fn:   newA,
+				},
+			),
+			fx.Invoke(func(in in) {
+				assert.NotNil(t, in.A, "expected in.A to be injected")
+				assert.Equal(t, "foo", in.A.name, "expected to get a type 'a' of name 'foo'")
+			}),
+		)
+		app.RequireStart().RequireStop()
+	})
+}
--- a/app.go
+++ b/app.go
@@ -404,6 +404,13 @@ func (app *App) provide(constructor interface{}) {
 		return
 	}
 
+	if a, ok := constructor.(Annotated); ok {
+		if err := app.container.Provide(a.Fn, dig.Name(a.Name)); err != nil {
+			app.err = err
+		}
+		return
+	}
+
 	if err := app.container.Provide(constructor); err != nil {
 		app.err = err
 	}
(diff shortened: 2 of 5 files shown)


abhinav:
Hello and thanks for the PR! This is a good start.

To answer your questions (in-order):

- Doing this in `*app` is fine for now. If we end up adding another type like
  `Annotated`, then we would probably move part of the logic into methods on
  those types.
- That's alright. We can tag a release when this is ready and re-pin.
- :)
- We don't need to worry about optionals with this. You can't provide
  something as optional, you can only depend on other things as optional,
  which is still possible by having you constructor consume an `fx.In`.
  
  For groups, we'll add a `dig.Grouped` or similar API to dig. IMO that's not
  a blocker for a v1 of fx.Annotated.

annotated_test.go
```diff
@@ -0,0 +1,39 @@
...
+		app := fxtest.New(t,
+			fx.Provide(
+				fx.Annotated{
+					Name: "foo",
+					Fn:   newA,
+				},
+			),
+			fx.Invoke(func(in in) {
+				assert.NotNil(t, in.A, "expected in.A to be injected")
+				assert.Equal(t, "foo", in.A.name, "expected to get a type 'a' of name 'foo'")
```
abhinav: Instead of putting assertions in the `Invoke`, consider moving them out with
the use of an [`fx.Populate`](https://godoc.org/go.uber.org/fx#Populate).

    var result struct {
        fx.In
        
        A *a `name:"foo"`
    }
    
    app := fxtest.New(t,
        ...,
        fx.Populate(&result),
    )
    
    assert.[..]
    
Otherwise, if we have a bug that neglects to call that `fx.Invoke`, the
assertion will be ignored and the test will still pass.

app.go
```diff
@@ -404,6 +404,13 @@ func (app *App) provide(constructor interface{}) {
 		return
 	}
 
+	if a, ok := constructor.(Annotated); ok {
+		if err := app.container.Provide(a.Fn, dig.Name(a.Name)); err != nil {
```
abhinav: For the case where the constructor is just a function, we should probably
verify that we're not producing `fx.Annotated`s into the container. It would
be easy for someone to assume that they can do `func() fx.Annotated` to get
the same behavior so we can catch that now and give them a better error
message.

You should be able to adapt some of the code in `"internal/fxreflect".ReturnTypes`
to get a list of `reflect.Type`s being provided by the constructor and check
that one of them isn't `reflect.TypeOf(Annotated{})`. (The code in that
package has some duplication with dig; we plan on exposing APIs in dig to
reduce that in the future.)

dimiro1:
Thanks for your feedback @abhinav. Next week I will have time to work on it again.

glibsm:
@dimiro1 Looks like this PR has stalled. 

Are you able to pick this back up at some point, or this kind of functionality is no longer desired for your use case?

dimiro1:
Hello @glibsm, sorry for that. I was planning to work on the PR but I could not find time to work on it. Feel free to close the PR.

abhinav:
@dimiro1 Thanks for the contribution nonetheless. We have a good idea of the
direction for this feature now and this PR will be helpful as soon as someone
has a chance to work on it.

glibsm:
Superseeded by #656

PR #643, opened 2018-09-19
Update dig version and leverage deference of validation

dnathe4th:
dig@1.5 introduces two significant performance improvements for the container underlying FX. The underlying cycle detection algorithm improvements are most significant for deep, highly-connected dependency graphs. Developers using FX for DI with a significant number of shared components (i.e. a configuration library depended on by most other packages) will see the biggest improvement in startup time.

Additionally, the new `dig.DeferAcyclicVerification()` container constructor option was designed specifically to alleviate redundant safety checks as dependencies as `Provide`d in a tight loop a la FX. 

See https://github.com/uber-go/dig/pull/208 and https://github.com/uber-go/dig/pull/209 for concrete details on the new `dig` features.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Make `fxtest.New` ensure that the app was created successfully. Previously,
   it would return the app (similar to `fx.New`, which expects the user to verify
   the error).
+- Update dig container to defer acyclic validation until after Invoke.
 
 ## [1.7.0] - 2018-08-16
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   it would return the app (similar to `fx.New`, which expects the user to verify
   the error).
 - Update dig container to defer acyclic validation until after Invoke. Application
-  startup time should improve proportional to the side of the dependency graph.
+  startup time should improve proportional to the size of the dependency graph.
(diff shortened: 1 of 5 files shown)


glibsm:
The failed test appears to catch a real issue

dnathe4th:
Hm. I have to disagree that the test catches a real issue. Is there any reason to expect an fx App to be created without an invoke? If so, I need a new strategy. If not, I can update the test so it is not so contrived as to be useless.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Make `fxtest.New` ensure that the app was created successfully. Previously,
   it would return the app (similar to `fx.New`, which expects the user to verify
   the error).
+- Update dig container to defer acyclic validation until after Invoke.
```
akshayjshah: I'd note that this improves performance by reducing repeated work.

glide.yaml
```diff
@@ -3,7 +3,7 @@ import:
 - package: go.uber.org/multierr
   version: ^1
 - package: go.uber.org/dig
-  version: ^1.4
+  version: ^1.5
```
akshayjshah: We've been bad about this in the past; please leave a comment noting that we need at least v1.5 for the `DeferAcyclicVerification` option.

glibsm:
Dom and I talked offline at length and we agreed that absence of `fx.Invoke` is effectively a no-op and it's ok to allow the cycles at that point (since nothing is used anyways).

We considered other approaches, but none were as simple as "the first invoke will validate".

akshayjshah:
Thanks for the fixes and responding to feedback, Dom! Sorry that it's taken a while - I'll try to merge and cut a release of this tomorrow.

Issue #626, opened 2018-06-13
fxtest.Lifecycle RequireStart and RequireStop do not close context


jeffbean:
Given the Lifecycle holds the Hooks that rely on getting a Context for OnStart and OnStop. 

With this to test out application startups the fxtest.Lifecycle has the `RequireStart` and `RequireStop` methods that create and leave open a [Background context](https://github.com/uber-go/fx/blob/master/fxtest/fxtest.go#L110). 

Using goleaks today i found leaked goroutines when using the fxtest methods described above. The below example uses the lifecycle context to handle the case when its shutting down. 

```go
// hand wave something that is configured to register these Hooks.  

p.Lifecycle.Append(fx.Hook{
    OnStart: func(ctx context.Context) error {
        go runForeverPlz(ctx)
        return nil
    },
})

func runForeverPlz(ctx context.Context) {
    t := time.NewTicker(time.Second)
    for {
        select{
        case <-t.C:
            fmt.Println("hello")
        case <-ctx.Done():
            fmt.Println("done running forever")
            return
        }
    }
}
```
This test setup will leak goroutines:
```go
func TestAppStartSop(t *testing.T) {
    ls := fxtest.NewLifecycle(t)
    
    // hand wave some struct that is using this Testing Lifecycle 
    
    ls.RequireStart().RequireStop() // LEAKS FOUND HERE
}

```
This was my simple fix to get the closed context.
```go
func TestAppStartSop(t *testing.T) {
    ls := fxtest.NewLifecycle(t)
    
    // hand wave some struct that is using this Testing Lifecycle 
    
    ctx, cancel := context.WithCancel(context.Background())

    assert.NoError(t, ls.Start(ctx))
    assert.NoError(t, ls.Stop(ctx))

    cancel()
}

```

PR #644, opened 2018-09-20
Add changelog entry for leak fix

akshayjshah:
The fix for #626 didn't include a changelog entry.

@jeffbean, I figured that you'd already put enough work into that diff ;)


jeffbean:
thanks :D appreciate it  - Approved

PR #651, opened 2018-10-10
Provide DOT graph of container by default

zmt:
This will enable users to access a DOT representation of the dependency graph in their app container for debugging successfully starting applications. Before this change, the DOT graph was only created on error.


app.go
```diff
@@ -461,6 +473,20 @@ func (app *App) StopTimeout() time.Duration {
 	return app.stopTimeout
 }
 
+// DotGraph provides a graph of dependencies in the container as a string in DOT
+// syntax.
+func (app *App) DotGraph() DotGraph {
+	// Memoize the graph
+	if app.dotgraph != nil {
```
AlexAPB: super nit: you can invert the if statement to deal with creation and then always return the dotgraph

app_test.go
```diff
@@ -89,6 +90,30 @@ func TestNewApp(t *testing.T) {
...
+	t.Run("Visualizer", func(t *testing.T) {
+		found := false
+		fixture := "fxtest/graph.dot"
+		type A struct{}
+		type B struct{}
+		var g DotGraph
+		app := fxtest.New(t,
+			Option(Visualizer()),
+			Provide(func() A { return A{} }),
+			Provide(func() B { return B{} }),
```
rhang-uber: Make a dependency that changes the graph to make it clear what you are attempting to test.

app.go
```diff
@@ -335,6 +344,11 @@ func New(opts ...Option) *App {
 		return app
 	}
 
+	if app.visualize {
```
rhang-uber: This should be moved up since, since there is error handling above this statement that should include this process.

app.go
```diff
@@ -335,6 +344,11 @@ func New(opts ...Option) *App {
...
+		app.provide(app.DotGraph)
+		app.invokes = append(app.invokes, app.DotGraph) // must be invoked last
```
rhang-uber: I understand where you are going, but I wonder if there is another way in fx to allow for controller over the order of execution of the dependencies injected. It just feels a bit strange there is an if statement at the end -- which in theory could lead to even more if statements.

app_test.go
rhang-uber: Specifically, the packages that you are Providing to the fx app should be actual dependencies. The examples you provide don't exercise the logic of dig visualize.

app.go
```diff
@@ -461,6 +473,20 @@ func (app *App) StopTimeout() time.Duration {
 	return app.stopTimeout
 }
 
+// DotGraph provides a graph of dependencies in the container as a string in DOT
+// syntax.
+func (app *App) DotGraph() DotGraph {
```
abhinav: Why export this? I would err on the side of not exposing new APIs if possible.

app.go
```diff
@@ -210,6 +210,13 @@ func StopTimeout(v time.Duration) Option {
 	})
 }
 
+// Visualizer enables providing visualization of the container dependency graph.
+func Visualizer() Option {
+	return optionFunc(func(app *App) {
```
abhinav: Why make this opt-in? There's no harm in putting it in by default.

app.go
```diff
@@ -461,6 +473,20 @@ func (app *App) StopTimeout() time.Duration {
...
 
+// DotGraph provides a graph of dependencies in the container as a string in DOT
+// syntax.
+func (app *App) DotGraph() DotGraph {
+	// Memoize the graph
+	if app.dotgraph != nil {
+		return *app.dotgraph
+	}
+	b := &bytes.Buffer{}
+	dig.Visualize(app.container, b) // TODO: handle errors gracefully
```
abhinav: RE: error handling: `fx.Provide`s can just return errors to break the container. Although in our case, we'll maybe want to provide an empty graph with an error message or something rather than fail application startup.

app.go
zmt: I think this is actually clearer logically, even if it is not optimized for fewest return statements.

app_test.go
zmt: Done.

app.go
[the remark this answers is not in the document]
zmt: Eliminated the need for explicitly calling it on initialization.

app.go
zmt: Removed the boolean implementation.

app_test.go
zmt: Done.

app.go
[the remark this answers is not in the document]
zmt: Removed.

app.go
zmt: Removed.

app.go
zmt: Removed.

app.go
zmt: Done.

app.go
zmt: I didn't want to alter the default behavior.

app.go
zmt: I like the empty graph suggestion.  I think I will go with that.

app.go
abhinav: Discussed offline: We'll do this by default.

app.go
```diff
@@ -461,6 +467,25 @@ func (app *App) StopTimeout() time.Duration {
...
+	// Memoize the graph
+	if app.dotgraph != nil {
+		return *app.dotgraph
+	}
+	b := &bytes.Buffer{}
+	if err := dig.Visualize(app.container, b); err != nil {
+		b.WriteString(`digraph {
+  graph [];
+  
+  error  [shape=plaintext label="Error graphing container"];
```
abhinav: Let's include the error message contents.

app.go
```diff
@@ -461,6 +467,25 @@ func (app *App) StopTimeout() time.Duration {
 	return app.stopTimeout
 }
 
+func (app *App) visualizeDOT() DotGraph {
+	// Memoize the graph
+	if app.dotgraph != nil {
+		return *app.dotgraph
+	}
+	b := &bytes.Buffer{}
```
abhinav: style nit: use `var` form for zero value of buffer.

    var buff bytes.Buffer

app.go
abhinav: Also, can we the generation of the error-case graph into a function rather than inlining here.

app.go
```diff
@@ -461,6 +467,25 @@ func (app *App) StopTimeout() time.Duration {
...
+	if err := dig.Visualize(app.container, b); err != nil {
+		b.WriteString(`digraph {
```
abhinav: Instead of WriteString-ing a string into the buffer and then extracting it, we can just return it directly.

```
if err := ...; err != nil {
    return DotGraph(...)
}
```

app.go
```diff
@@ -461,6 +467,25 @@ func (app *App) StopTimeout() time.Duration {
 	return app.stopTimeout
 }
 
+func (app *App) visualizeDOT() DotGraph {
+	// Memoize the graph
+	if app.dotgraph != nil {
```
abhinav: memoization not needed; fx does it for you

app.go
```diff
@@ -461,6 +460,18 @@ func (app *App) StopTimeout() time.Duration {
...
 
+func (app *App) dotGraph() DotGraph {
+	var b bytes.Buffer
+	if err := dig.Visualize(app.container, &b); err != nil {
+		return app.dotGraphErr(err)
+	}
+	return DotGraph(b.String())
+}
+
+func (app *App) dotGraphErr(err error) DotGraph {
```
zmt: I should add a test to exercise this, but wasn't immediately sure how.

app.go
```diff
@@ -461,6 +460,16 @@ func (app *App) StopTimeout() time.Duration {
 	return app.stopTimeout
 }
 
+func (app *App) dotGraph() DotGraph {
+	var b bytes.Buffer
+	msg := "digraph { graph []; error [shape=plaintext label=\"%q\"]; }"
+	if err := dig.Visualize(app.container, &b); err != nil {
```
zmt: Should I just ignore this error and comment? I can't get past the test coverage requirement unless I test this and I haven't figured out a simple way to trigger the error from `dig.Visualize`.

abhinav: I don't see an easy way to trigger this error condition but we should still have logic to handle it. It'll have to remain uncovered.

---

If we *really* wanted to cover it, we could declare a private interface for the buffer and a private field for generating buffers.

```
type buffer interface {
  io.Writer
  fmt.Stringer
}

type App struct {
  newBuffer buffer
}
```

Then in the test you would return an implementation that returns an error on write.

zmt: Thanks - I think we will need to cover this especially if we let the error trigger panic as Akshay suggested.

app.go
```diff
@@ -461,6 +460,16 @@ func (app *App) StopTimeout() time.Duration {
...
+		err = fmt.Errorf(msg, err)
+		dig.Visualize(app.container, &b, dig.VisualizeError(err))
```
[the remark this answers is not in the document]
zmt: Yeah. I realized that after I pushed this up.

app_test.go
```diff
@@ -89,6 +89,28 @@ func TestNewApp(t *testing.T) {
...
+			Provide(func() A { return A{} }),
+			Provide(func(A) B { return B{} }),
+			Provide(func(A, B) C { return C{} }),
+			Populate(&g),
+		)
+		defer app.RequireStart().RequireStop()
+		require.NoError(t, app.Err())
+		assert.Contains(t, g, `"fx_test.A" [label=<fx_test.A>];`)
+		assert.Contains(t, g, `"fx_test.B" [label=<fx_test.B>];`)
+		assert.Contains(t, g, `"fx_test.C" [label=<fx_test.C>];`)
```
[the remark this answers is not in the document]
zmt: I had created a fixture, but go1.11 and go1.10 changed the graph output. I didn't really want to get into generating 2 fixtures and parsing `runtime.Version`, so  I went for asserting substrings that were in common across go versions. I could go back to that approach.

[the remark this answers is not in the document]
abhinav: No, that's probably worse. In that case, I would just assert one of the cases -- just to make sure that we don't have an error message.

app.go
```diff
@@ -461,6 +460,16 @@ func (app *App) StopTimeout() time.Duration {
 	return app.stopTimeout
 }
 
+func (app *App) dotGraph() DotGraph {
+	var b bytes.Buffer
+	msg := "digraph { graph []; error [shape=plaintext label=\"%q\"]; }"
```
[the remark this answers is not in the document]
zmt: Moot.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+### Added
+- Provide DOT graph of dependencies in the container.
```
abhinav: fx.DotGraph is now provided to the container by default. It contains a
    visualization of the dependency graph in a running Fx application.

And let's add something similar to this in the docs for DotGraph itself
because we haven't yet documented anywhere that this is happening.

kriskowal:
Did @rhang-uber give this a final review? I see he’s still listed as blocking.

PR #655, opened 2018-11-05
Prepare v1.8.0

zmt:


zmt:
Moved to 655 as a minor release.

zmt:
@kriskowal or @akshayjshah - @abg suggested we get one more stamp before I release.

Issue #653, opened 2018-10-25
modify objects already provided in the graph


srikrsna:
First of all, thank you for this excellent library. It is idiomatic and helped us a lot in deploying our services with ease. 

One thing that will be nice to have (and I would like to contribute to) is to have a way to modify objects that are already provided in the graph. As we deployed our services such patterns were required more on more.

Example: 

We wanted to add Open Census metrics and trace to our application. This can be done by wrapping the http handler in Open Census's http handler. Currently we modify the handler provider/invoker function to check if open census is enabled and wrap the http handler to return an Open Census enable http handler. If we had a way to modify provided objects especially interfaces i.e. provide a wrapped implementation things like tracing caching auth can be separated into their own layer.

This can be implemented without any breaking changes. It will require a new API, something like, `fx.Modify(...)` which will take inputs from providers and return providers that were already in the graph. Modifiers will only be applied if types returned by them have been requested by invokers.

What do you think of this?

glibsm:
Linking in the related discussions: #531 #649

srikrsna:
Option C of #531 is more or less what I am proposing and would elegantly solve my problem. One concern seems to be that the order in which such decorators will be applied. Can't they be like Invoke, i.e. in the order provided?

srikrsna:
Any plans to implement this?

srikrsna:
I am unable to understand how named values and value groups differ from the graph's perspective they are like any other value, aren't they? Could you please explain.

And I suppose this should be supported by dig first. Should I open an issue there?

abhinav:
Yes, happy to clarify! From the point of view of `fx.Decorate`, named values
and value groups open up some questions:

- When I provide a decorator for type `F`, does that apply to all instances of
  `F` or only the unnamed ones?

    ```
        fx.Provide(
            fx.Annotated{Name: "foo", Target: func() F { ... }},
        ),
        fx.Decorate(
            func(F) F { ... },
        ),
    ```

  If it applies only to unnamed ones, that means decorators have to explicitly
  specify which named value they apply to.

- Similarly, if `F` is actually being fed into a value group, does the
  decorator for `F` apply to all `F`s provided to that group?
 
    ```
        fx.Provide(
            fx.Annotated{Group: "foo", Target: producer1},
            fx.Annotated{Group: "foo", Target: producer2},
            fx.Annotated{Group: "foo", Target: producer3},
        ),
        fx.Decorate(
            func(F) F { ... },
        ),
    ```

[...]

PR #656, opened 2018-11-07
Added Annotate Provide Option to inject named values without relying on Out

srikrsna:
Picking off where #633 was left out for #610. I believe once this is implemented #653 can be reasoned about in a better way.

I've also tried to incorporate all code review comments from #633 and also added some documentation.

### DIFF
--- a/annotated_test.go
+++ b/annotated_test.go
@@ -83,7 +83,7 @@ func TestAnnotatedWrongUsage(t *testing.T) {
 			),
 			fx.Populate(&in),
 		)
-		assert.Contains(t, app.Err().Error(), "fx.Annotated should be passed to fx.Provide directly, it should not be returned by the constructor")
+		assert.Contains(t, app.Err().Error(), "fx.Annotated should be passed to fx.Provide directly", "expected error when return types were annotated")
 	})
 
 	t.Run("Result Type", func(t *testing.T) {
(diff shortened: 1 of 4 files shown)


annotated_test.go
```diff
@@ -0,0 +1,100 @@
...
+	"go.uber.org/fx"
+	"go.uber.org/fx/fxtest"
+)
+
+func TestAnnotated(t *testing.T) {
+	type a struct {
+		name string
+	}
+	type in struct {
+		fx.In
```
abhinav: minor style nit: we've found it more readable to differentiate between embedded fields and regular fields by adding an empty line between the field groups

```suggestion
		fx.In

```

annotated_test.go
```diff
@@ -0,0 +1,100 @@
...
+		assert.Equal(t, "foo", in.A.name, "expected to get a type 'a' of name 'foo'")
+	})
+}
+
+func TestAnnotatedWrongUsage(t *testing.T) {
+	type a struct {
+		name string
+	}
+	type in struct {
+		fx.In
```
abhinav: same as before: minor nit:

```suggestion
		fx.In

```

annotated_test.go
```diff
@@ -0,0 +1,100 @@
...
+			fx.Provide(
+				fx.Annotated{
+					Name: "foo",
+					Fn: func() in {
+						return in{A: &a{name: "foo"}}
+					},
+				},
+			),
+		)
+		assert.Error(t, app.Err(), "expected error when return types were annotated")
```
glibsm: nit: assert some part of the error message to make sure it's the one you expected (and not some other error message in case the test was not set up correctly)

srikrsna:
All Done.

abhinav:
Thanks, @srikrsna!

PR #657, opened 2018-11-08
Annotated: Rename Fn to Target

abhinav:
This renames the Fn field to Target to account for the possibility of
wanting to provide things besides functions with `fx.Provide` and
`fx.Annotated`.

This also alters the docs for Annotated to account for value group
support in the future by decoupling them from the idea of named values.

### DIFF
--- a/annotated.go
+++ b/annotated.go
@@ -20,33 +20,35 @@
 
 package fx
 
-// Annotated is a special provide type that specifies that all values produced by a
-// constructor should have the given name. See also In and Out documentation
-// about Named Values.
-//
-// Given,
-//
-//   func NewReadOnlyConnection(...) (*Connection, error)
-//   func NewReadWriteConnection(...) (*Connection, error)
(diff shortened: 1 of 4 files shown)


annotated.go
```diff
@@ -20,33 +20,35 @@
...
+	//   }
+	//
+	//   fx.Provide(func(...) (Result, error) {
+	//     conn, err := NewReadOnlyConnection(...)
+	//     return Result{Connection: conn}, err
+	//   })
 	Name string
-	Fn   interface{}
+
+	// Target is the constructor being annotated with fx.Annotated.
```
glibsm: if the sole purpose of renaming Fn to Target is to make room for supporting things other than constructors then saying `Target is the constructor` is a bit confusing

abhinav: We can update the docs here if/when we add support for other portions. The
docs are mutable, the field name isn't.

PR #659, opened 2018-11-14
fx.Annotated: Support Value Groups

vyshah:
This adds support for value groups to fx.Annotated, passing the
relevant Dig ProvideOption if the `Group` field is set. The new option
was added in Dig 1.7 so this also raises the lower version constraint
for Dig.

See also #610.

### DIFF
--- a/app.go
+++ b/app.go
@@ -482,8 +482,8 @@ func (app *App) provide(constructor interface{}) {
 		return
 	}
 
-	var opts []dig.ProvideOption
 	if a, ok := constructor.(Annotated); ok {
+		var opts []dig.ProvideOption
 		switch {
 		case len(a.Group) > 0 && len(a.Name) > 0:
 			app.err = fmt.Errorf("fx.Annotate may not specify both name and group for %v", constructor)
--- a/app.go
+++ b/app.go
@@ -483,8 +483,17 @@ func (app *App) provide(constructor interface{}) {
 	}
 
 	if a, ok := constructor.(Annotated); ok {
-		if err := app.container.Provide(a.Target, dig.Name(a.Name)); err != nil {
-			app.err = err
+		switch {
+		case len(a.Group) > 0 && len(a.Name) > 0:
+			app.err = fmt.Errorf("fx.Annotate may not specify both name and group for %v", constructor)
+		case len(a.Name) > 0:
+			if err := app.container.Provide(a.Target, dig.Name(a.Name)); err != nil {
+				app.err = err
+			}
+		case len(a.Group) > 0:
+			if err := app.container.Provide(a.Target, dig.Group(a.Group)); err != nil {
+				app.err = err
+			}
 		}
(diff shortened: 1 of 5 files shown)


app.go
```diff
@@ -483,8 +483,17 @@ func (app *App) provide(constructor interface{}) {
...
-			app.err = err
+		switch {
+		case len(a.Group) > 0 && len(a.Name) > 0:
+			app.err = fmt.Errorf("fx.Annotate may not specify both name and group for %v", constructor)
+		case len(a.Name) > 0:
+			if err := app.container.Provide(a.Target, dig.Name(a.Name)); err != nil {
+				app.err = err
+			}
+		default:
+			if err := app.container.Provide(a.Target, dig.Group(a.Group)); err != nil {
```
abhinav: This should explicitly check for `len(a.Group) > 0` because using fx.Annotated without any annotations (`fx.Annotated{Target: fn}`) is valid usage/equivalent to not having fx.Annotated.

abhinav:
@vyshah The relevant Dig change has been merged and released. Did you want to pick this back up?

app.go
```diff
@@ -483,8 +483,17 @@ func (app *App) provide(constructor interface{}) {
...
+		case len(a.Group) > 0 && len(a.Name) > 0:
+			app.err = fmt.Errorf("fx.Annotate may not specify both name and group for %v", constructor)
+		case len(a.Name) > 0:
+			if err := app.container.Provide(a.Target, dig.Name(a.Name)); err != nil {
+				app.err = err
+			}
+		case len(a.Group) > 0:
+			if err := app.container.Provide(a.Target, dig.Group(a.Group)); err != nil {
+				app.err = err
+			}
```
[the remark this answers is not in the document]
vyshah: d'oh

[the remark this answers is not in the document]
abhinav: Suggest restructuring to,

```go
var opts []dig.ProvideOption
switch {
case len(a.Group) > 0 && len(a.Name) > 0:
  app.err = ...
  return
case len(a.Name) > 0:
  opts = append(opts, dig.Name(a.Name))
case len(a.Group) > 0:
  opts = append(opts, dig.Name(a.Group))
}

if err := app.container.Provide(a.Target, opts...); err != nil {
  ...
}
```

app_test.go
```diff
@@ -105,6 +105,55 @@ func TestNewApp(t *testing.T) {
...
+				func(b B) {
+					assert.NotNil(t, b.Foo)
+					assert.NotNil(t, b.Bar)
+				},
+			),
+		)
+
+		defer app.RequireStart().RequireStop()
+		require.NoError(t, app.Err())
+	})
```
abhinav: We should probably test the following two cases as well.

- Setting Group
- Leaving both, Group and Name unset

vyshah: yeah, good point

PR #660, opened 2018-11-29
[WIP] Add a Shutdowner interface to allow manual shutdown from a module

SyrieBianco:
Add a Shutdowner interface with a Shutdown function, that provides modules with the functionality to signal to the app to shutdown. Generate the Shutdowner upon construction of the App object, and make it available to the container. 

Summary of Changes:
1) Add `dones`, a slice of done channels, and `mu`, a mutex, to the App object in order to maintain references to all done channels created via the `Done` method, and to be able to concurrency-safely read from and write to the `dones`
2) Change the `Done` method to record created channels on the app as well as returning a read-only channel
3) Create a Shutdowner interface by returning a closure that has access to the app object's `dones` and `mu` 
4) During the the construction of the `App`, instantiate the `dones` field and provide the Shutdowner to the container.

Points of consideration:
1) Tests. I need them. I don't yet have them.
2) What `ShutdownOption`s should be added to the shutdown function and available to modules? Should any shutdown-related `Option`s be surfaced to the end user, such as white-listing which modules are allowed to call `Shutdown()`
3) Does the closure need pointers to the slice and mutex in order to function as intended?
4) Function signature/error handling: should we/how should we address the possibility that some channels do not receive the intended signal?


shutdown_test.go
```diff
@@ -0,0 +1,120 @@
...
+				},
+			})
+		}))
+
+		done1, done2 := app.Done(), app.Done()
+		startCtx, cancel := context.WithTimeout(context.Background(), app.StartTimeout())
+		defer cancel()
+		assert.NoError(t, app.Start(startCtx), "error in app start")
+
+		assert.NoError(t, app.shutdowner().Shutdown())
```
amckinney: consistency nit: It looks like you missed an `assert` that includes an informative message.
```
assert.NoError(t, app.shutdowner().Shutdown(), "failed to shutdown")
```

shutdown.go
```diff
@@ -0,0 +1,81 @@
...
+func (app *App) shutdowner() Shutdowner {
+	return &shutdowner{
+		broadcastSignal: app.signalBroadcaster(),
+	}
+}
+
+func (app *App) signalBroadcaster() func(os.Signal) error {
+	return func(signal os.Signal) error {
+		app.mu.RLock()
+		defer app.mu.RUnlock()
```
amckinney: nit: We should be able to avoid the `defer` if we manually unlock after we read all of the `app.dones` channels.

shutdown_test.go
```diff
@@ -0,0 +1,120 @@
...
+				OnStart: func(ctx context.Context) error {
+					started = true
+					return nil
+				},
+				OnStop: func(ctx context.Context) error {
+					stopped = true
+					return nil
+				},
+			})
+		}))
```
akshayjshah: We've got this pattern in a few places now. Can we factor it out into a reusable option?

```
type lifecycleVerifier struct {
  fx.Option

  started, stopped bool
}

func newLifecycleVerifier() *lifecycleVerifier {
  var started, stopped bool
  opt := Invoke(func(l Lifecycle) {
    l.Append(Hook{
        started = true
        return nil
      },
      OnStop: func(context.Context) error {
        stopped = true
        return nil
      },
    })
  })
  return &lifecycleVerifier{
    Option: opt,
    started: started,
    stopped: stopped,
  }
}
```

For bonus points, have the verifier struct include a `testing.TB` and add methods that assert that `started` and `stopped` are true.

(There are probably some errors in there, but hopefully you get the point.)

app_test.go
```diff
@@ -62,6 +62,16 @@ func TestNewApp(t *testing.T) {
...
 	})
 
+	t.Run("ProvidesShutdowner", func(t *testing.T) {
+		found := false
+		app := fxtest.New(t, Invoke(func(s Shutdowner) {
+			assert.NotNil(t, s)
+			found = true
+		}))
+		defer app.RequireStart().RequireStop()
+		assert.True(t, found)
```
akshayjshah: Do we need the extra assertion on `found`? The rest of the unit tests are already asserting that invoked functions are always run.

shutdown_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+		done := app.Done()
+		s := app.shutdowner().(*shutdowner)
+		s.broadcastSignal(syscall.SIGINT)
```
akshayjshah: Why do we need to use unexported API here? I think we should be able to use the `Populate` option to pull the `Shutdowner` out of the container without using the unexported `shutdowner()` method. We should also be able to fill the first channel by calling `Shutdown` and asserting that we didn't get an error back.

akshayjshah:
Overall, I like the way this came out - nice work!

I think we may want to expand the `Shutdowner` API a bit, though. In production, we'll want to know which module/package/whatever called `Shutdown`, otherwise it's going to be a complete mystery why applications are randomly shutting down. I'd prefer not to rely on each user of `Shutdown` doing the right thing and logging a good message; instead, I'd like the `Shutdowner` itself to collect structured logs and metrics about this.

Unfortunately, this will take a bit of work - the logger (or whatever other struct we're using to track this info) needs to come from inside the container, so it's not available at the Fx layer. Instead, I think we'll want applications to be able to register callbacks that execute just before we actually shut down the app. (Something like `OnShutdown(func(*ShutdownInfo))`.) We don't need to actually implement that functionality in this PR, but we do need to add a method that enables callback registration to the `Shutdowner` interface.

app.go
```diff
@@ -320,6 +324,7 @@ func New(opts ...Option) *App {
 		logger:       logger,
 		startTimeout: DefaultTimeout,
 		stopTimeout:  DefaultTimeout,
+		dones:        []chan os.Signal{},
```
abhinav: nit: nil is a valid empty slice. Don't need to do this.

app.go
```diff
@@ -330,6 +335,7 @@ func New(opts ...Option) *App {
 		app.provide(p)
 	}
 	app.provide(func() Lifecycle { return app.lifecycle })
+	app.provide(func() Shutdowner { return app.shutdowner() })
```
abhinav: This can be `app.provide(app.shutdowner)`.

shutdown.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+	"fmt"
+	"os"
+	"syscall"
+)
+
+// Shutdowner provides a method that can manually trigger the shutdown of the
+// application by sending a signal to all open Done channels. Shutdowner works
+// both on applications using Run as well as Start, Done, and Stop.
```
abhinav: nit: we should probably mention here that this is provided to all Fx
applications.

shutdown.go
```diff
@@ -0,0 +1,81 @@
...
+type Shutdowner interface {
+	Shutdown(...ShutdownOption) error
+}
+
+// Shutdown broadcasts a signal to all of the Done channels on the application
+// thereby enabling the start of the Stop process.
+func (s *shutdowner) Shutdown(opts ...ShutdownOption) error {
+	return s.broadcastSignal(syscall.SIGTERM)
```
abhinav: nit: ordering: entity generally, we go for: type declaration, constructor (if any), methods.

```
type shutdowner struct {
  ..
}

func (s *shutdowner) Shutdown(...) error {
  ..
}
```

(Except when the methods of a single type are spread across multiple files, so
this doesn't apply to `app` methods below.)

shutdown.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func (app *App) shutdowner() Shutdowner {
+	return &shutdowner{
+		broadcastSignal: app.signalBroadcaster(),
+	}
+}
+
+func (app *App) signalBroadcaster() func(os.Signal) error {
+	return func(signal os.Signal) error {
```
abhinav: You can make this a bit more readable by making the anonymous func a method of
app and not calling `signalBroadcaster`.

```
func (app *App) shutdowner() Shutdowner {
    return &shutdowner{broadcastSignal: app.broadcastSignal}
}

func (app *App) broadcastSignal(signal os.Signal) error {
  app.mu.RLock()
  defer app.mu.RUnlock()

    ...
}
```

shutdown.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// ShutdownOption provides a way to configure properties of the shutdown
+// process. Currently, no options have been implemented.
+type ShutdownOption interface {
+	apply(*shutdowner)
+}
+
+type shutdowner struct {
+	broadcastSignal func(os.Signal) error
```
abhinav: Is there a string reason to use a `func` reference here instead of just `*app`?

```
type shutdowner struct{ app *app }

func (s *shutdowner) Shutdown(...) error {
    return app.broadcastSignal(syscall.SIGTERM)
}
```

shutdown.go
```diff
@@ -0,0 +1,81 @@
...
+		defer app.mu.RUnlock()
+
+		var unsent int
+		for i, done := range app.dones {
+			select {
+			case done <- signal:
+			default:
+				// signal was never sent to channel
+				unsent++
+				app.logger.Printf("done channel %d at capacity, did not receive shutdown signal", i)
```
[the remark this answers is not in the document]
abhinav: I don't know if we need to record this at all. The contract for signal
handlers is if you're listening, you get the message, otherwise it's lost. The
same should apply here if we're emulating that.

shutdown.go
```diff
@@ -0,0 +1,82 @@
...
+// process. Currently, no options have been implemented.
+type ShutdownOption interface {
+	apply(*shutdowner)
+}
+
+type shutdownOptionFunc func(*shutdowner)
+
+// Shutdowner provides a method that can manually trigger the shutdown of the
+// application by sending a signal to all open Done channels. Shutdowner works
+// both on applications using Run as well as Start, Done, and stop.
```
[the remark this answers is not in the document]
SyrieBianco: 👌🏼👍🏼

app.go
```diff
@@ -442,12 +448,20 @@ func (app *App) Stop(ctx context.Context) error {
...
 // the same terminal as the running process.
+// Alternateively, a signal can be broadcast to all done channels manually by
+// using the Shutdown functionality (see the Shutdowner documentation for details)
+// todo(syrie): add Shutdowner documentation.
 func (app *App) Done() <-chan os.Signal {
 	c := make(chan os.Signal, 1)
 	signal.Notify(c, syscall.SIGINT, syscall.SIGTERM)
+
+	app.mu.Lock()
+	defer app.mu.Unlock()
```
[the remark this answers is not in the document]
SyrieBianco: I agree that in this simpler case it makes sense to forgo the performance cost of defer since it doesn't add any readability benefit, but as a rule I've been used to defaulting to defer per the style guide. I'll put that through.

shutdown.go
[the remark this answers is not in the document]
SyrieBianco: @amckinney lmk what you think of that comment
@abhinav I'm not committed to returning this as an error, necessarily, but I do think that it is valuable to indicate that the done channel already received a SIGTERM or SIGINT signal that has not yet been cleared from the channel (even if we can't provide a whole lot of context around what sent the blocking signal). I also agree with @akshayjshah that it makes sense to *either* log the error or return the error, and I also agree that it makes more sense to return the error than log it.

shutdown.go
SyrieBianco: I was wondering if there was a particular pattern — thanks!

shutdown.go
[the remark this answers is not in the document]
SyrieBianco: @akshayjshah if we only return the error outside of the loop and forgo logging each blocked channel as we come across it, do we think it's still worthwhile to retain the index of each blocked channel and combine them into the single error message?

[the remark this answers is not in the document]
akshayjshah: Probably not. An error message like "failed to send SIGTERM to 4 of 7 channels" is good enough.

shutdown_test.go
SyrieBianco: True; I wanted to be able to broadcast a different signal — SIGINT rather than SIGTERM — just to extra-assert that the channel received the signal from something other than `.Shutdown()`. That said, I'm generally in agreement with not using unexported methods in tests, and I don't think we gain anything really valuable from sending a SIGINT

shutdown.go
```diff
@@ -0,0 +1,78 @@
...
+		default:
+			// shutdown called when done channel has already received a
+			// termination signal that has not been cleared
+			unsent++
+		}
+	}
+
+	if unsent != 0 {
+		err := fmt.Errorf("failed to send %v signal to %v out of %v channels", signal, unsent, len(app.dones))
+		return err
```
abhinav: nit: can do `return fmt.Errorf(...)`

shutdown_test.go
```diff
@@ -0,0 +1,108 @@
...
+			t,
+			lv,
+			fx.Populate(&s),
+		)
+
+		done := app.Done()
+		assert.NoError(t, s.Shutdown(), "error returned from first shutdown call")
+		app.RequireStart()
+
+		assert.Error(t, s.Shutdown(), "no error returned when shutdown is called with a blocked channel")
```
abhinav: We should assert portion of the error message here.

```
if err := s.Shutdown(); assert.Error(t, err, "expected Shutdown to fail") {
  assert.Contains(t, err.Error(), "bla bla bla")
}
```

shutdown_test.go
```diff
@@ -0,0 +1,108 @@
...
+	})
+}
+
+type lifecycleVerifier struct {
+	fx.Option
+
+	started, stopped bool
+
+	tb testing.TB
+}
```
abhinav: Design-wise, I wouldn't embed fx.Option into lifecycleVerifier.

I would either make `newLifecycleVerifier` return two things (the verifier and
the fx.Option) or add a method on the verifier that accepts an fx.Lifecycle
and `fx.Invoke` that.

```
type lifecycleVerifier struct {
  ..
}

func newLifecycleVerifier(tb testing.TB) *lifecycleVerifier {
  ...
}

func (lv *lifecycleVerifier) Append(lc fx.Lifecycle) {
  ...
}

func (lv *lifecycleVerifier) Verify() {
  ...
}
```

And for usage,

```go
lv := newLifecycleVerifier(t)
defer lv.Verify()

fxtest.New(..., fx.Invoke(lv.Append))
```

abhinav: That said, it's unclear to me why we need this. The use of RequireStart and
RequireStop guarantees that the lifecycle is started and stopped.

Is this to protect against forgetting to call against RequireStart/RequireStop?

app_test.go
SyrieBianco: So I patterned this test off of the "ProvidesLifecycle" line for line. I don't think that the assertion is necessary, but I figured that the tests to confirm whether or not a type is provided to the container should be the same in content. lmk if you think I should remove the found assertion from both tests or leave it.

shutdown_test.go
SyrieBianco: 👌🏼👍🏼

app.go
```diff
@@ -445,9 +450,15 @@ func (app *App) Stop(ctx context.Context) error {
 // application. Applications listen for the SIGINT and SIGTERM signals; during
 // development, users can send the application SIGTERM by pressing Ctrl-C in
 // the same terminal as the running process.
+// Alternatively, a signal can be broadcast to all done channels manually by
+// using the Shutdown functionality (see the Shutdowner documentation for details)
```
akshayjshah: nit: please add a period, exclamation point, or terminal punctuation of your choice at the end of this sentence.

PR #663, opened 2019-01-11
Update license header

abhinav:


annotated.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2018 Uber Technologies, Inc.
+// Copyright (c) 2019 Uber Technologies, Inc.
```
glibsm: should technically be a range 201?-2019, or did we drop that requirement across all OSS?

PR #668, opened 2019-02-01
Add constructor function semantic guidance to fx.Provide

jcorbin:


app.go
```diff
@@ -81,6 +81,11 @@ func (f optionFunc) apply(app *App) { f(app) }
 //
 // See the documentation of the In and Out types for advanced features,
 // including optional parameters and named instances.
+//
+// Constructor functions should functions perform as little external
```
glibsm: `functions should functions`?

jcorbin: good catch, was just leftover word, removed.

PR #675, opened 2019-04-29
[WIP] Add fx.OnStart

prashantv:
fx.OnStart reduces boilerplate when adding a function to be run
when the application is started.

It reduces boilerplate like:
```go
fx.Invoke(
    func(lc fx.Lifecycle, a *Type1, b *Type2) {
        lc.Append(fx.Hook{
                doSomething(a, b)
            },
        })
    },
)
```

into

```go
fx.OnStart(
    func(a *Type1, b *Type2) {
        doSomething(a, b)
    },
)
```

This is a WIP implementation which implements the logic "outside" of fx, to start
a discussion of whether this is what we want.

There are some details to consider:
 * What happens if the graph has a context and the passed in function has a context
   Currently: If the first argument is a context, it's always the Start context. Any
   other argument as context is provided from the graph.
 * What is the source of the OnStart for errors?
   Today it'll be a generated function, but fx should probably skip these.
   We might want a `t.Helper()` to provide fx helpers that don't pollute the
   stack.

This could also be implemented "internally" in fx, which would complicate the core
framework but possibly provide better stacks/errors.

See T2133659 for more details.

### DIFF
--- a/onstart.go
+++ b/onstart.go
@@ -1,0 +1,118 @@
+// Copyright (c) 2019 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package fx
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+)
+
+var _ctxType = typeOfInterface(new(context.Context))
+var _lifecycleType = typeOfInterface(new(Lifecycle))
+var _errType = typeOfInterface(new(error))
+
+func typeOfInterface(newIfacePtr interface{}) reflect.Type {
+	return reflect.TypeOf(newIfacePtr).Elem()
+}
+
+// OnStart registers functions that are executed when the application is started.
+// Arguments for these invocations are built using the constructors provided by
+// Provide. Passing multiple OnStart options appends new functions to the applications'
+// existing list.
+//
+// Similar to Invokes, functions are always executed, and they are run in order.
+// OnStart functions may return no values, or an error. Any other return values
+// cause errors.
+//
+// If the first argument of the passed in function is a context.Context, then the
+// context from Start is passed in.
+func OnStart(funcs ...interface{}) Option {
+	var opts []Option
+	for _, f := range funcs {
+		opts = append(opts, onStart(reflect.ValueOf(f)))
+	}
+	return Options(opts...)
+}
+
+func validateOnStartFunc(ft reflect.Type) error {
+	if kind := ft.Kind(); kind != reflect.Func {
+		return fmt.Errorf("OnStart must be passed a function, got %v", kind)
+	}
+	switch ft.NumOut() {
+	case 0:
+		// No return value, nothing to validate.
+	case 1:
+		// Must be error
+		if ft.Out(0) != _errType {
+			return fmt.Errorf("OnStart functions can only return an error, found %v", ft.Out(0))
+		}
+	default:
+		return fmt.Errorf("OnStart functions must have no returns, or a single error return")
+	}
+
+	return nil
+}
+
+func onStart(f reflect.Value) Option {
+	ft := f.Type()
+
+	if err := validateOnStartFunc(ft); err != nil {
+		return invokeErr(err)
+	}
+
+	// If the first argument is a context, ignore it, but track we need it for later.
+	isFirstCtx := ft.NumIn() > 0 && ft.In(0) == _ctxType
+
+	// The invoke function will take a lifecycle, followed by any arguments to f.
+	invokeInTypes := []reflect.Type{_lifecycleType}
+	for i := 0; i < ft.NumIn(); i++ {
+		if i == 0 && isFirstCtx {
+			continue
+		}
+
+		invokeInTypes = append(invokeInTypes, ft.In(i))
+	}
+
+	invokeFuncType := reflect.FuncOf(invokeInTypes, []reflect.Type{_errType}, false /* variadic */)
+
+	invokeFunc := reflect.MakeFunc(invokeFuncType, func(args []reflect.Value) []reflect.Value {
+		// The first argument to invoke is a lifecycle, let's extract that.
+		lifecycle := args[0].Interface().(Lifecycle)
+		lifecycle.Append(Hook{
+			OnStart: func(ctx context.Context) error {
+				var startArgs []reflect.Value
+
+				if isFirstCtx {
+					startArgs = append(startArgs, reflect.ValueOf(ctx))
+				}
+
+				startArgs = append(startArgs, args[1:]...)
+				f.Call(startArgs)
+				return nil
+			},
+		})
+
+		return []reflect.Value{reflect.New(_errType).Elem()}
+	})
+
+	return Invoke(invokeFunc.Interface())
+}
--- a/onstart_test.go
+++ b/onstart_test.go
@@ -1,0 +1,80 @@
+// Copyright (c) 2019 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package fx
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestOnStart(t *testing.T) {
+	type A string
+	type B string
+
+	type Params struct {
+		In
+
+		A A
+		B B
+	}
+
+	var called []string
+
+	// No arguments
+	noArgs := func() {
+		called = append(called, "noArgs")
+	}
+	hasArgs := func(a A, b B) {
+		called = append(called, fmt.Sprintf("hasArgs(%v,%v)", a, b))
+	}
+	hasCtxAndArgs := func(ctx context.Context, a A) {
+		called = append(called, fmt.Sprintf("hasCtxAndArgs(%v,%v)", ctx != nil, a))
+	}
+	hasParams := func(p Params) {
+		called = append(called, fmt.Sprintf("hasParams(%v,%v)", p.A, p.B))
+	}
+
+	app := New(
+		Provide(
+			func() A { return "val-A" },
+			func() B { return "val-B" },
+		),
+
+		OnStart(noArgs, hasArgs),
+		OnStart(hasCtxAndArgs),
+		OnStart(hasParams),
+	)
+	require.NoError(t, app.Err(), "Failed to create app")
+	err := app.Start(context.Background())
+	require.NoError(t, err)
+
+	want := []string{
+		"noArgs",
+		"hasArgs(val-A,val-B)",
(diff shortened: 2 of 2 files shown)


glibsm:
Them main reason the start/stop hooks are so verbose has to do with how they are rolled back in the event that something goes wrong in one of them, or the application is shutting down.

In our implementation, `OnStart` hook gets correlated with an `OnStop` hook, wrapped up in an `fx.Hook` object that gets incrementally added into a slice. We move forward if things are good, and we move back if things are bad.

In your proposed design, how would one solve the problem of needing a start and stop correlation (e.g.. open/close some connections, or alike)? Or are you simply suggesting that there are enough simple use cases where only the `OnStart` hook is needed and thus we can create some sugar to simplify.

abhinav:
I wouldn't be opposed to `fx.OnStart` optionally returning the cleanup
function as a special type.

Maybe something like this?

```
package fx

type OnStop func(context.Context)
```

```
fx.OnStart(
  func(a *Type1, b *Type2) fx.OnStop {
    ...
    return func(ctx context.Context) {
        conn.Close()
    },
  },
)
```

glibsm:
> I wouldn't be opposed to `fx.OnStart` optionally returning the cleanup
> function as a special type.

That could be very elegant!

prashantv:
I really like the idea of the on-start function returning a special `fx.OnStop` type,

Visualizing the different options with a dummy "connection" type that needs a Start/Stop,

**Today**

```go
fx.Provide(newConn),
fx.Invoke(func(conn *Conn, lc fx.Lifecycle) {
  lc.Append(fx.Hook{
    OnStart: conn.Start,
    OnStop: conn.Stop,
  })
})
```

**Proposed**

```go
fx.Provide(newConn),
fx.OnStart(func(ctx context.Context, conn *Conn) (fx.OnStop, error) {
  if err := conn.Start(ctx); err != nil {
    return err
  }

  return fx.OnStop(conn.Stop), nil
})
```

prashantv:
I thought about this a little, and wonder if we should make `OnStart` and `OnStop` symmetrical, while also allowing `OnStop` functions to be returned from `OnStart`.

By symmetric, I'd like::
 * Allow `OnStop` to be passed directly into `New` for cases where no start hook is needed.
 * Optionally accept `context.Context` and return `error`, rather than always requiring them.

However, making `OnStop` return an `Option` doesn't play well with allowing `OnStart` to return return an`OnStop` (can't enforce this at the type-level).

We could have separate APIs for a stop function returned from an `OnStart` vs passing in `OnStop` hooks directly to `New`.

The cleanest API is `type OnStop func(context.Context) error`, but this would be very specific to `OnStart` and isn't very symmetric.

Thoughts?

abhinav:
The type for the stop function doesn't have to be called OnStop. We can call
it Stopper.

```
type Stopper func(context.Context) error

// OnStop [..]
//
// If any of the functions of an OnStart return a Stopper as their last non-
// error argument, the Stopper will be queued up with OnStop.
//
//  fx.New(
//    ...,
//    OnStart(func(server *http.Server) Stopper {
//      go server.ListenAndServe()
//      return server.Shutdown
//    }),
//  )
//
// The above is equivalent to,
//
//  fx.New(
//    ...,
//    OnStart(func(server *http.Server) {
//      go server.ListenAndServe()
//    }),
//    OnStop(func(ctx context.Context, server *http.Server) error {
//      return server.Shutdown(ctx)
//    }),
//  )
func OnStart(fs ...interface{}) Option

// OnStop [..]
func OnStop(fs ...interface{}) Option
```

prashantv:
> The type for the stop function doesn't have to be called OnStop. We can call
it Stopper.
>
> ```
> type Stopper func(context.Context) error
> ```

Yep, we'll always require `context.Context` and `error` for the `OnStop` function if we do that. The alternative It could also be a private interface,
```go
type Stopper interface {
  onStop()
}
```

But then we'd also need another API to build a stopper from a function.

I think just adding `Stopper` is probably good enough for now, since we can always add wrappers to create a `func(context.Context) error` from functions without the arg/return.

abhinav:
The other reason to require the returned Stopper to accept a context is to avoid users accidentally using the OnStart context.

If you added a `ToStopper(interface{}) Stopper` that accepted a `func() error` in addition to `func(context.Context) error`, it would be easy to make this mistake:

```
OnStart(func(ctx context.Context, server *Server) (Stopper, error) {
    err := server.Start(ctx)
    return func() error {
        server.Stop(ctx)
    }, err
})
```

server.Stop will get called with an expired context.

prashantv:
Updated change with support for returning a `Stopper` from `OnStart`. The implementation works fine, but I'm not happy with the original function locations being hidden due to use of `reflect`, e.g.,

```
2019/07/11 07:38:10 [Fx] PROVIDE        fx.Shutdowner <= go.uber.org/fx.(*App).shutdowner-fm()
2019/07/11 07:38:10 [Fx] PROVIDE        fx.DotGraph <= go.uber.org/fx.(*App).dotGraph-fm()
2019/07/11 07:38:10 [Fx] INVOKE         reflect.makeFuncStub()
2019/07/11 07:38:10 [Fx] START          reflect.callReflect()
2019/07/11 07:38:10 [Fx] RUNNING
2019/07/11 07:38:10 [Fx] STOP           reflect.callReflect()
```

One option I had in mind was to create a wrapper type that can contain additional information (either a function name or pc) that would be used when printing the above (and used for errors generated by dig).

glibsm:
> Updated change with support for returning a `Stopper` from `OnStart`. The implementation works fine, but I'm not happy with the original function locations being hidden due to use of `reflect`, e.g.,
> 
> ```
> 2019/07/11 07:38:10 [Fx] PROVIDE        fx.Shutdowner <= go.uber.org/fx.(*App).shutdowner-fm()
> 2019/07/11 07:38:10 [Fx] PROVIDE        fx.DotGraph <= go.uber.org/fx.(*App).dotGraph-fm()
> 2019/07/11 07:38:10 [Fx] INVOKE         reflect.makeFuncStub()
> 2019/07/11 07:38:10 [Fx] START          reflect.callReflect()
> 2019/07/11 07:38:10 [Fx] RUNNING
> 2019/07/11 07:38:10 [Fx] STOP           reflect.callReflect()
> ```
> 
> One option I had in mind was to create a wrapper type that can contain additional information (either a function name or pc) that would be used when printing the above (and used for errors generated by dig).

This all seems like a lot of work. All for what... to save a few lines for syntactic sugar purposes? What other benefits would there be if we get it all wired up as a one liner on start

prashantv:
> This all seems like a lot of work. All for what... to save a few lines for syntactic sugar purposes? What other benefits would there be if we get it all wired up as a one liner on start

The debuggability issue already exists with `fx.Extract` and `fx.Provide`, so this work would improve the user experience for other use-cases that are important for testing.
```
[snippet from running TestPopulate]
        printer.go:35: [Fx] INVOKE              reflect.makeFuncStub()
```

Some benefits:
 - It makes it more obvious how to run logic at start rather than app creation time. A lot of users are incorrectly using `fx.Invoke` to run their application logic which should actually be run on start. Compare the following:
```go
// We have a function that should run on application startup that has dependencies:
func myStartLogic(logger *zap.Logger, c *Client) error { ... }

// Most users end up using fx.Invoke since it's simpler:
fx.Invoke(myStartLogic)

// To do it correctly:
fx.Invoke(func(lc fx.Lifecycle, logger *zap.Logger, c *Client) error {
  lc.Append(Hook{
      return myStartLogic(logger, c)
    },
  })
}

// With OnStart
fx.OnStart(myStartLogic)
```

The correct way is not only verbose, but also has duplication (each additional argument is duplicated in the signature of the invoke function and in the call). This has to happen for multiple types, so it ends up adding a lot of noise to service code. It's also not obvious (most users don't interact with `fx.Lifecycle`, so they don't know to use it)

 - `fx.Lifecycle` becomes less important -- there's no need to know about an additional type. I think this will simplify modules as well.

Today, many modules created within a service have 2 separate constructors: one for the underlying type, and another for fx that handles the lifecyle. If we have on-start, we wouldn't need the second function, it could instead me something like:
```
var module = fx.Options(
  fx.Provide(constructorFn),
  fx.OnStart((*myType).Start),
)
```

This is definitely not a required change, but I've noticed the lack of this ability hurting readability and correctness (use of Invoke).

abhinav:
I definitely see the value of this change. I too found it annoying to deal
with lifecycles and OnStart just to write a simple `main` function. It's
especially annoying because your actual `start` function's dependencies have
to be replicated by hand in the `fx.Invoke` that adds the OnStart hook.

```go
func main() {
    fx.New(
        ...,
        fx.Invoke(
            func(d1 D1, d2, D2, ..., dN DN, lc fx.Lifecycle) error {
                lc.Append(fx.Lifecycle{
                    OnStart: func(ctx contxt.Context) error {
                        return run(d1, d2, ..., dN)
                    },
                })
            },
        ),
    ).Run()
}

func run(d1 D1, d2, D2, ..., dN DN) {
    ...
}
```

Versus, with this API,

```go
func main() {
    fx.New(
        ...,
        fx.OnStart(run),
    ).Run()
}

func run(d1 D1, d2, D2, ..., dN DN) error {
    ...
}
```

(BTW, we need to clarify that OnStart functions also block program startup so
long-running tasks must be executed in a goroutine.)

onstart.go
```diff
@@ -0,0 +1,188 @@
...
+				var startArgs []reflect.Value
+
+				if onStartDesc.hasCtx {
+					startArgs = append(startArgs, reflect.ValueOf(ctx))
+				}
+
+				startArgs = append(startArgs, args[1:]...)
+				res := f.Call(startArgs)
+
+				// If the OnStop returns an error, then it must be the last argument
```
abhinav: typo: s/argument/result

onstart.go
```diff
@@ -0,0 +1,188 @@
...
+				if onStartDesc.hasErr {
+					if err := res[len(res)-1].Interface(); err != nil {
+						return err.(error)
+					}
+				}
+
+				// If the OnStop returns a Stopper, it must be the first argument.
```
abhinav: typo: s/argument/result

onstart.go
```diff
@@ -0,0 +1,188 @@
...
+			OnStop: func(ctx context.Context) error {
+				if onStop == nil {
+					return nil
+				}
+
+				return onStop(ctx)
+			},
+		})
+
+		return []reflect.Value{reflect.New(_errType).Elem()}
```
abhinav: Can use reflect.Zero
```suggestion
		return []reflect.Value{reflect.Zero(_errType)}
```

glibsm:
> I definitely see the value of this change. I too found it annoying to deal
> with lifecycles and OnStart just to write a simple `main` function. It's
> especially annoying because your actual `start` function's dependencies have
> to be replicated by hand in the `fx.Invoke` that adds the OnStart hook.
> 
> ```go
> func main() {
>     fx.New(
>         ...,
>         fx.Invoke(
>             func(d1 D1, d2, D2, ..., dN DN, lc fx.Lifecycle) error {
>                 lc.Append(fx.Lifecycle{
>                     OnStart: func(ctx contxt.Context) error {
>                         return run(d1, d2, ..., dN)
>                     },
>                 })
>             },
>         ),
>     ).Run()
> }
> 
> func run(d1 D1, d2, D2, ..., dN DN) {
>     ...
> }
> ```
> 
> Versus, with this API,
> 
> ```go
> func main() {
>     fx.New(
>         ...,
>         fx.OnStart(run),
>     ).Run()
> }
> 
> func run(d1 D1, d2, D2, ..., dN DN) error {
>     ...
> }
> ```
> 
> (BTW, we need to clarify that OnStart functions also block program startup so
> long-running tasks must be executed in a goroutine.)

The main issue I have is the added incentive and shortcut to jump straight into the onstart blocks.

One of the initial suggestions was to not even have the onstart/stop hooks and treat objects as fully ready by the time the constructor exists to stop this game of where to put your code. Perhaps there was some merit to that.

A lot of this deferral into the start blocks (which I assume is forcing the need to create this sugar) is creating an additional unordered layers of how dependencies get brought up, and documentation burden (this or that not ready to use until after OnStart).

glibsm:
> `fx.Lifecycle` becomes less important -- there's no need to know about an additional type. I think this will simplify modules as well.

For the consumers, yes, it becomes less important. Module authors are likely still interested in this.

> Today, many modules created within a service have 2 separate constructors: one for the underlying type, and another for fx that handles the lifecyle. If we have on-start, we wouldn't need the second function, it could instead me something like:

I haven't seen such patterns, but you and I have access to different codebases. 

You're not really solving the duplication pattern with `fx.OnStart` for the module authors, they still have to write and duplicate their parameters (if they don't write a single constructor).

abhinav:
> One of the initial suggestions was to not even have the onstart/stop hooks and treat objects as fully ready by the time the constructor exists to stop this game of where to put your code. Perhaps there was some merit to that.

I agree that if we had that option right now we would want to kill Lifecycle
in favor of "returned = ready".  Unfortunately the system currently treats
instantiation and startup as separate things. Many of our Fx modules assume
that an object won't be used until start is called.

RE: Invoke-Start pattern: @prashantv, that's an anti-pattern for modules
because then they're unconditionally having themselves instantiated even if
the user doesn't want them. Appending to Lifecycle from the module constructor
is preferable: the HTTP server will be queued to be run only if it's actually
requested by the user.

Actually, this issue is perhaps discussing more for this feature. This feature
will make it easy for modules to accidentally make themselves unconditional
hard dependencies. Is there a way we can prevent that?

prashantv:
> RE: Invoke-Start pattern: @prashantv, that's an anti-pattern for modules
> because then they're unconditionally having themselves instantiated even if
> the user doesn't want them. Appending to Lifecycle from the module constructor
> is preferable: the HTTP server will be queued to be run only if it's actually
> requested by the user.
> 
> Actually, this issue is perhaps discussing more for this feature. This feature
> will make it easy for modules to accidentally make themselves unconditional
> hard dependencies. Is there a way we can prevent that?

That's a great point. Only solution I can think of is to only run the `OnStart` if all the other dependencies are already created. I think it would require something new like `dig.LazyInvoke`.

prashantv:
Another approach to avoid unconditionally instantiating all dependencies,

```go
  fx.Provide(fx.Annotated{
    Target: someConstructor,
    OnStart: func(...) error {},
    OnStop: func(...) {},
  }
```

The `OnStart` and `OnStop` will only be run if the target constructor needed to be run. E.g., we'd be merging the constructors to make this work. The `OnStart` and `OnStop` can take any additional types as well, and optionally return an error.

We can also support a special target, `fx.Always`, for cases where OnStart / OnStop should always be run. This could be used to provide helpers `fx.OnStart(..)` and `fx.OnStop(..)` which just do `fx.Provide(fx.Annotated{Target: fx.Always, OnStop: f}))`

abhinav:
`fx.Annotated{OnStart, OnStart}` sounds more reasonable. It's close to
`fx.OnStart(func(..))` in terms of usability without the eager evaluation
issues of an `Invoke`.

I'm guessing that the `OnStart` and `OnStop` fields will be `interface{}`s.
This adds a couple things to be aware of,

- They will need to be able to accept anything produced by the `Target` (which
  may be producing multiple things), and anything that's provided to the
  container.
- They will need to optionally accept a `context.Context` as their first
  argument.
- They will need to optionally be able to return an `error`.

Also, rather than an `OnStop` field, I suspect we still want the stopping
behavior to be returned by `OnStart` in case there's any shared state between
startup and shutdown behaviors, which would otherwise have to be captured
either on the value produced by the Target or in a closure.

```
OnStart: func(..., server *http.Server) error {
  ln, err := net.Listen(...)
  ...
  
  go server.Serve(ln)
},
OnStop: ??? // how do we ln.Close()
```

Versus,

```
OnStart: func(..., server *http.Server) (fx.Stopper, error) {
  ln, err := net.Listen(...)
  ...
  
  go server.Serve(ln)
  
  return func(ctx context.Context) error {
    return ln.Close()
  }, nil
},
```

abhinav:
Thanks for the contribution and discussion @prashantv.
We may take this over in the near future; we're considering options here.

Internal ref: GO-201

PR #679, opened 2019-08-07
Return an error when attempting to shutdown without any Done channels

SyrieBianco:
If `Shutdowner.Shutdown()` is called before any Done channels have been opened — say in an `fx.Invoke` option when using `Run()` — the Shutdowner will silently fail to initiate the `App.Stop` process. 

Change the Shutdowner to return an error if no Done channels yet exist; this will cause the app initialization to fail if `Shutdown()` is called in an Invoke and will also return a legible error to users.

### DIFF
--- a/shutdown.go
+++ b/shutdown.go
@@ -21,12 +21,14 @@
 package fx
 
 import (
+	"errors"
 	"fmt"
 	"os"
 )
 
 // Shutdowner provides a method that can manually trigger the shutdown of the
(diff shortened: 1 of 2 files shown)


abhinav:
I'm not convinced this is desirable.
If nobody has called `Done()` when we call `Shutdown()`,
perhaps we should record the shutdown and have `Done()` return a closed channel.
This will unblock the caller if there is indeed a data race in such a case.

PR #680, opened 2019-09-10
Show in the README.md how go mod can be used instead of glide

belfazt:
As `glide` is now _unmantained_ I'd like to show the alternative of using `go mod` in the `README.md`


glibsm:
Mirrors https://github.com/uber-go/dig/#installation nicely.

Glide is purely there for legacy reasons at this point

PR #685, opened 2019-11-10
Migrate to Go modules

abhinav:
This migrates Fx to Go modules. This simplifies some of the build and CI
setup in the process.

I've left the glide.yaml in place for legacy ecosystems.

A test that referred to import paths of vendored dependencies had to be
fixed.

We should wait until uber-go/dig#250 is released before we do this,
though.


abhinav:
Ready for review.

PR #686, opened 2019-11-14
Drop closure-based fx.Option implementation

abhinav:
We have a handful of `fx.Option` implementations based on `optionFunc`
which captures value using a closure. That makes them opaque and
uninspectable.

This change drops the `optionFunc` type in favor of concrete
implementations for all `fx.Option`s. For further inspection, each such
instance includes a `String()` implementation.

Note that we've intentionally kept `fmt.Stringer` out of the `fx.Option`
interface: We don't want to commit to having options printable but it's
nicer if they are.

### DIFF
--- a/app.go
+++ b/app.go
@@ -52,10 +52,6 @@ type Option interface {
 	apply(*App)
 }
 
-type optionFunc func(*App)
-
-func (f optionFunc) apply(app *App) { f(app) }
-
 // Provide registers any number of constructor functions, teaching the
 // application how to instantiate various types. The supplied constructor
 // function(s) may depend on other types available in the application, must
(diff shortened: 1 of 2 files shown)


app.go
```diff
@@ -52,10 +52,6 @@ type Option interface {
 	apply(*App)
```
glibsm: you might want to specify `fmt.Stringer` so an option can't get added without it

app.go
```diff
@@ -52,10 +52,6 @@ type Option interface {
...
 }
 
-type optionFunc func(*App)
-
-func (f optionFunc) apply(app *App) { f(app) }
-
```
glibsm: +1.

never liked this pattern

app.go
abhinav: I intentionally kept that out. Per the commit message,

> Note that we've intentionally kept fmt.Stringer out of the fx.Option
> interface: We don't want to commit to having options printable but it's
> nicer if they are.

glibsm: If downstream someone wants to print all the options that were used in the creation of the app, I don't really see a problem with that (I do see a problem of some of them being string-able and some not).

It seems like somewhat arbitrary line, why can't we commit to requiring them all to have a `String()`?

abhinav: I was envisioning arguments that could not reliably be turned into strings but I suppose everything can support a placeholder. I'll add it to the interface.

abhinav:
```suggestion
	fmt.Stringer

	apply(*App)
```

PR #687, opened 2019-11-15
fxreflect: Add CallStack

abhinav:
This adds an `fxreflect.CallStack` function that generalizes prior
`Caller()` logic.

Each commit should be reviewed individually.


twilly:
Seems good to me.

PR #688, opened 2019-11-16
Error reporting: Include Provide/Invoke stack traces

abhinav:
A number of errors we report during start up are fairly opaque because
they only reference the target function, not where the bad invocation
was made.

This is extra problematic if the target function was generated via
reflection. For example, @klnusbaum reported the following error which
was impossible to debug with just Fx.

    fx.Option should be passed to fx.New directly, not to fx.Provide:
    fx.Provide received fx.Provide(reflect.makeFuncStub())

To help debug issues like this, this PR changes Fx to capture a call
stack whenever `fx.Provide` or `fx.Invoke` are called. These stack
traces are later reported in the error messages for various errors
thrown during program startup.

With this change, the error above becomes similar to the following.

    fx.Option should be passed to fx.New directly, not to fx.Provide:
    fx.Provide received fx.Provide(reflect.makeFuncStub()) from:
    path/to/my/module.init
            /.../my/module.go:50
    path/to/your/app.main
            /.../your/app/main.go:42

Depends on #687
Ref T4487249

---

**Each commit should be reviewed separately.**

### DIFF
--- a/annotated_test.go
+++ b/annotated_test.go
@@ -24,6 +24,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"go.uber.org/fx"
 	"go.uber.org/fx/fxtest"
 )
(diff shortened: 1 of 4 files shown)


annotated_test.go
```diff
@@ -83,7 +84,20 @@ func TestAnnotatedWrongUsage(t *testing.T) {
...
+
+		err := app.Err()
+		require.Error(t, err)
+
+		// Example:
+		// fx.Annotated should be passed to fx.Provide directly, it should not be returned by the constructor: fx.Provide received go.uber.org/fx_test.TestAnnotatedWrongUsage.func2.1() from:
+		// go.uber.org/fx_test.TestAnnotatedWrongUsage.func2
+		//         /.../fx/annotated_test.go:76
+		// testing.tRunner
+		//         /.../go/1.13.3/libexec/src/testing/testing.go:909
```
SyrieBianco: Thanks for these comments

app.go
```diff
@@ -581,7 +581,9 @@ func (app *App) provide(p provide) {
 	app.logger.PrintProvide(constructor)
 
 	if _, ok := constructor.(Option); ok {
-		app.err = fmt.Errorf("fx.Option should be passed to fx.New directly, not to fx.Provide: fx.Provide received %v", constructor)
+		app.err = fmt.Errorf("fx.Option should be passed to fx.New directly, "+
```
zmt: nit: mixing `+` and format string interpolation

annotated_test.go
zmt: +1

app_test.go
```diff
@@ -520,10 +520,25 @@ func TestAppStart(t *testing.T) {
 	})
 
 	t.Run("InvokeNonFunction", func(t *testing.T) {
-		app := NewForTest(t, Invoke(struct{}{}))
+		spy := printerSpy{&bytes.Buffer{}}
```
zmt: nit (not for this diff stack): We might want to swap a `strings.Builder` into `printerSpy` at some point.

app.go
abhinav: Yeah, it's to keep the lines short. Go doesn't have lexical string concatenation so it has to be `+` to split a long line.

PR #690, opened 2019-12-14
Add fx.Optional for marking optional params

orishu:
Realized that fx can be improved so certain parameters provided, while in their absence, a default parameter value would be used.

Dependency: pull request in uber-go/dig: [255](https://github.com/uber-go/dig/pull/255)

Add a new `fx.Optional` type that can be embedded in structs to mark
that the dependency on them is optional.


abhinav:
Closing per discussion in https://github.com/uber-go/dig/pull/255#issuecomment-571348737

Issue #673, opened 2019-04-19
dig.As


rssathe:
How do I use dig.As with an fx app?

glibsm:
`dig.As` hasn't been released yet, you can see https://github.com/uber-go/dig/pull/235 for why it was halted slightly.

The plan for usage inside fx is through `fx.Annotated` (not implemented yet) that will function similarly to how `fx.Annotated` can do named and grouped params.

glibsm:
Going to leave this issue open to track the `fx.Annotated.As` work.

phamtai97:
Hi @glibsm
I am having the same problem as you, the issues have been resolved yet ?

sywhang:
`fx.As` has been implemented with #795.

AFMiziara:
I was testing this solution with the provided example in the docs, but I am getting error `type io.Writer is not an expression` . Am I missing something?

```go
fx.Provide(
  fx.Annotate(bytes.NewBuffer(...), fx.As(io.Writer))
)
```

sywhang:
@AFMiziara you're missing a `new`. Interface itself is not a valid expression in Go.

```go
fx.Provide(
  fx.Annotate(bytes.NewBuffer([]byte{}), fx.As(new(io.Writer)))
)
```

I see that in the docs we forgot to add the new there. Let me go fix that.

Issue #608, opened 2017-12-19
Can it resolve an interface function param?


DenisNeustroev:
Like this:
```go
type Repository struct {}
func (r *Repository) Fetch() {}

func NewRepository() *Repository {}

type IRepository interface {
    Fetch()
}

func NewService(r IRepository) *Service {}

app := fx.New(
    fx.Provide(
        NewRepository,
        NewService,
    ),
    fx.Invoke(func (s *Service) {
        s.Start()
    ),
)

app.Start()
```

 But I got the error: type IRepository is not in the container, did you mean to use *Repository?

abhinav:
Hello.

Resolving interfaces is supported but the interface type must be mentioned
explicitly by the constructor (at least at this time).

So this should work if you change NewRepository to,

```go
func NewRepository() IRepository {
    return &Repository{..}
}
```

That advice does go against the recommended pattern in Go of returning structs
and accepting interfaces in scenarios like this, so another option is to keep
NewRepository as-is, but introduce the following in your fx.Provide.

```go
fx.Provide(
    NewRepository,

    // Use *Repository for IRepository.
    func(r *Repository) IRepository { return r },

    NewService,
)
```

We chose not to do this automatically because Go interfaces are implicit.
Types may end up satisfying interfaces that the author did not mean to use
them for. That said, if we can figure out a clean way to expose auto-casting,
we fully intend to support it.

DenisNeustroev:
Thanks for the detailed answer. As you advised I wrote a converter function for casting pointers to interfaces. Works like a charm.

glibsm:
Great to hear that! Resolving based on previous comment.
Please file more issues if you encounter them, we're always looking to make Fx better.

Issue #604, opened 2017-12-01
Using `reflect.FuncOf` leads to a provide log without much context


prashantv:
If we have helper functions that return `fx.Options`, it's not possible to skip callers for logging, so we can end up with something like:
```
  fxtest.go:43: [Fx] PROVIDE  tchannelfx.Configuration <= reflect.makeFuncStub()
```

Ideally, we could customize the caller skip so we're showing something more instead of the helper functions.

prashantv:
While discussing this with @abhinav, one idea I had was to use an optional struct tag to tell `fx` what to log.

E.g., if we had a helper that provides a type `Channel`, it shouldn't just `fx.Provide(&Channel{..})`, but instead can provide an annotated version of the type which contains an extra tag,
```
struct {
  fx.Out

  Ch Channel `debuginfo:"chfx.Helper(..)"`
}
```

When `fx` is printing debug information for a type, it can check for this extra tag and use that in place of the function name.

prashantv:
This is coming up in #697 as well, so I think this is something we should prioritize.

Some more ideas:
re: https://github.com/uber-go/fx/pull/697

 - Caller skip (similar to zap), to indicate the number of caller frames to skip in the log.
 - t.Helper style or something similar where the helper marks itself as a helper, and we find the first call frame before that helper call (we might need some sort of fx.Helper(func(...) {}) to easily identify the helper
 - Just pass in a stack directly as field in `fx.Annotated` or similar to use instead of figuring it out ourselves. This is similar to the above, but we should try and capture the whole stack.

PR #697, opened 2020-03-22
Add fx.Supply()

roberth-k:
Adds `fx.Supply(values ...interface{})` to provide values for dependency injection that have been created externally (such as application configuration). `Supply` uses reflection to craft the appropriate function for `Provide`. As stated by the documentation:

Supply makes the given values available for dependency resolution as if they had been provided by a constructor. Specifically, given:

```go
type MyStruct struct{}
var myStruct = &MyStruct{}
```

The following two forms are equivalent:

```go
fx.Provide(func() *MyStruct { return myStruct })
fx.Supply(myStruct)
```

Supply operates by constructing a constructor of the first form based on the types of the supplied values. Supply accepts any number of arguments.

### DIFF
--- a/annotated.go
+++ b/annotated.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2019 Uber Technologies, Inc.
+// Copyright (c) 2020 Uber Technologies, Inc.
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
--- a/annotated.go
+++ b/annotated.go
@@ -52,6 +52,8 @@ import (
 //   })
 //
 // Annotated cannot be used with constructors which produce fx.Out objects.
+//
+// When used with fx.Supply, the target is a value rather than a constructor function.
 type Annotated struct {
 	// If specified, this will be used as the name for all non-error values returned
 	// by the constructor. For more information on named values, see the documentation
@@ -66,7 +68,7 @@ type Annotated struct {
 	// A group option may not be provided if a name option is provided.
 	Group string
 
-	// Target is the constructor being annotated with fx.Annotated.
+	// Target is the constructor or value being annotated with fx.Annotated.
 	Target interface{}
 }
 
--- a/app.go
+++ b/app.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2019 Uber Technologies, Inc.
+// Copyright (c) 2020 Uber Technologies, Inc.
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
--- a/app.go
+++ b/app.go
@@ -113,40 +113,6 @@ func (o provideOption) String() string {
 	return fmt.Sprintf("fx.Provide(%s)", strings.Join(items, ", "))
 }
 
-// Supply makes the given values available for dependency resolution as if
-// they had been provided by a constructor. Specifically, given:
-//
-//  type MyStruct struct{}
-//  var myStruct = &MyStruct{}
-//
-// The following two forms are equivalent:
-//
-//  fx.Provide(func() *MyStruct { return myStruct })
-//  fx.Supply(myStruct)
-//
-// Supply operates by constructing a constructor of the first form based on
-// the types of the supplied values. Supply accepts any number of arguments.
-func Supply(values ...interface{}) Option {
-	if len(values) == 0 {
-		return Options()
-	}
-
-	returnTypes := make([]reflect.Type, len(values))
-	returnValues := make([]reflect.Value, len(values))
-
-	for i, value := range values {
-		returnTypes[i] = reflect.TypeOf(value)
-		returnValues[i] = reflect.ValueOf(value)
-	}
-
-	ft := reflect.FuncOf([]reflect.Type{}, returnTypes, false)
-	fv := reflect.MakeFunc(ft, func([]reflect.Value) []reflect.Value {
-		return returnValues
-	})
-
-	return Provide(fv.Interface())
-}
-
 // Invoke registers functions that are executed eagerly on application start.
 // Arguments for these invocations are built using the constructors registered
 // by Provide. Passing multiple Invoke options appends the new invocations to
--- a/app.go
+++ b/app.go
@@ -351,6 +351,9 @@ type provide struct {
 
 	// Stack trace of where this provide was made.
 	Stack fxreflect.Stack
+
+	// IsSupply is true when the Target constructor was emitted by fx.Supply.
+	IsSupply bool
 }
 
 // invoke is a single invocation request to Fx.
@@ -574,11 +577,18 @@ func (app *App) dotGraph() (DotGraph, error) {
 }
 
 func (app *App) provide(p provide) {
-	constructor := p.Target
 	if app.err != nil {
 		return
 	}
-	app.logger.PrintProvide(constructor)
+
+	constructor := p.Target
+
+	switch {
+	case p.IsSupply:
+		app.logger.PrintSupply(constructor)
+	default:
+		app.logger.PrintProvide(constructor)
+	}
 
 	if _, ok := constructor.(Option); ok {
 		app.err = fmt.Errorf("fx.Option should be passed to fx.New directly, "+
--- a/app.go
+++ b/app.go
@@ -113,6 +113,40 @@ func (o provideOption) String() string {
 	return fmt.Sprintf("fx.Provide(%s)", strings.Join(items, ", "))
 }
 
+// Supply makes the given values available for dependency resolution as if
+// they had been provided by a constructor. Specifically, given:
+//
+//  type MyStruct struct{}
+//  var myStruct = &MyStruct{}
+//
+// The following two forms are equivalent:
+//
+//  fx.Provide(func() *MyStruct { return myStruct })
+//  fx.Supply(myStruct)
+//
+// Supply operates by constructing a constructor of the first form based on
+// the types of the supplied values. Supply accepts any number of arguments.
+func Supply(values ...interface{}) Option {
+	if len(values) == 0 {
+		return Options()
+	}
+
+	returnTypes := make([]reflect.Type, len(values))
+	returnValues := make([]reflect.Value, len(values))
+
+	for i, value := range values {
+		returnTypes[i] = reflect.TypeOf(value)
+		returnValues[i] = reflect.ValueOf(value)
+	}
+
+	ft := reflect.FuncOf([]reflect.Type{}, returnTypes, false)
+	fv := reflect.MakeFunc(ft, func([]reflect.Value) []reflect.Value {
+		return returnValues
+	})
+
+	return Provide(fv.Interface())
+}
+
 // Invoke registers functions that are executed eagerly on application start.
 // Arguments for these invocations are built using the constructors registered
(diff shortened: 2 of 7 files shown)


app.go
```diff
@@ -113,6 +113,40 @@ func (o provideOption) String() string {
 	return fmt.Sprintf("fx.Provide(%s)", strings.Join(items, ", "))
 }
 
+// Supply makes the given values available for dependency resolution as if
```
r-hang: The logic seems good to me, though i'm wondering if extending Provides would have been a better user experience. Would something like ValueProvide be more fitting than Supply?

yurishkuro: Why can't Provide() take both functions and/or values?

roberth-k: On the two points raised:

- There is no precedent so far about overloading the "provider" term. Since the connotation associated with a "provider" is that it's a function, a ___Provider that doesn't consume a function would be surprising.
- `Provide()` shouldn't be modified to implicitly provide values, as it would not be able to differentiate between (a) regular provider functions; and (b) function values that are to be provided. For example: given `type MyFunc func(context.Context) error`, one might choose to `fx.Provide(func() MyFunc { return ... })`, and thus use `MyFunc` from other providers or invokes.

(side note: that last part is a great alternative to traditional service-repository patterns)

abhinav: Historical context: `Provide` used to accept both, functions and
values. If the argument was not a function, it would stick it into the
container as-is using the reflected type.

An issue this caused was that the "most specific" type, as determined
by `reflect.TypeOf` would be used. This meant `fx.Provide(new(bytes.Buffer))`
would be used as a `*bytes.Buffer` instead of `io.Writer`. The obvious
fix for that was `fx.Provide(func() io.Writer { .. })`. We decided that
debugging issues from that design before you got to the `func()` based
solution would be frustrating, so we opted for `func()`-only with
explicit types everywhere, with the intention of adding value-based
provides if there was enough need for it.

Now that we've had a fair amount of production experience with Fx, I'm
supportive of this. A separate entrypoint makes sense, and fx.Supply
seems like a good name for it.

To @roberth-k's second point: We definitely have internal Fx modules
that produce naked `func`s into the container as well as newtype
wrapped funcs.

app.go
```diff
@@ -113,6 +113,40 @@ func (o provideOption) String() string {
...
+func Supply(values ...interface{}) Option {
+	if len(values) == 0 {
+		return Options()
+	}
+
+	returnTypes := make([]reflect.Type, len(values))
+	returnValues := make([]reflect.Value, len(values))
+
+	for i, value := range values {
+		returnTypes[i] = reflect.TypeOf(value)
```
abhinav: Some validation:

- If `reflect.ValueOf(value).IsZero()`, they passed in nil. We should panic (because bad usage).
- Since `fx.Provide` treats `error` returns differently, we should verify here that the type is not an `error` type. We can use [`fx.Error`][1] to short-circuit this and fail `fx.New`.
  
  [1]: https://pkg.go.dev/go.uber.org/fx?tab=doc#Error

app.go
yurishkuro: > would be used as a `*bytes.Buffer` instead of `io.Writer`

@abhinav slightly off topic, but why can't Fx do both? When something is expecting  an interface as input, it's easy to scan all objects in the DI context and check if they can be duck typed to that interface. The current behavior of most specific type always seemed counterintuitive to me (coming from DIs in Java, like Spring, which happily matches on interfaces).

app.go
abhinav: We'll also need to support [fx.Annotated][1] since that will be the only way
users will be able to use named values or value groups with fx.Supply.

[1]: https://pkg.go.dev/go.uber.org/fx?tab=doc#Annotated

```
fx.Supply(fx.Annotated{Target: v, Name: "foo})
```

To support it, we can do something like this:

```
for _, v := range values {
  var c interface{}
  if ann, ok := v.(Annotated); ok {
    ann.Target = constructorForValue(ann.Target)
    c = ann
  } else {
    c = constructorForValue(c)
  }
  ..
```

Where `constructorForValue` builds the `func() $Type` constructor.

This has one minor change. Instead of, `fx.Supply(1, "foo", true)` being
equivalent to,

    fx.Provide(func() (int, string, bool) {
      return 1, "foo", true
    })

It will now be equivalent to,

    fx.Provide(
      func() int    { return 1     },
      func() string { return "foo" },
      func() bool   { return true  },
    )

That way, `fx.Supply("foo", fx.Annotated{Target: "bar", Name: "x"})` becomes,

    fx.Provide(
      func() string { return "foo" },
      fx.Annotated{
        Target: func() string { return "foo" },
        Name: "x",
      },
    )

app.go
abhinav: @yurishkuro we chose to make this explicit specifically because Go
interfaces are implicit: types do not declare which interfaces they
implement. This would have the effect of matching types to interfaces
users did not intend (possibly non-deterministically). Given that Fx's
injection all happens at runtime, debugging issues arising from that
would be more difficult. There was some prior discussion on this in https://github.com/uber-go/fx/issues/608#issuecomment-352881873.

If we can come up with a safe method of opting into interface
conversion on a case-by-case basis, that would be more desirable. #673
gets close here—we can finish implementing that. There was a prior
attempt at implementing it but the behavior wasn't quite right so the
change was rolled back.

app.go
roberth-k: @abhinav I've introduced the proposed changes, but with the following differences:

- Panicking specifically on `interface{}(nil)`, not zero values in general, as zero values (such as empty structs) could be supplied for valid reasons.
- Panicking on error values rather than failing the app container, as supplied values are constructed (and thus validated) outside of a `fx.App`'s runtime. If a user had an error on hand when calling `fx.Supply`, they ought to have checked it earlier. This one's a semantic decision though, so I'm happy to change to `fx.Error` if that's more in line with what users would expect.

abhinav: > Panicking specifically on `interface{}(nil)`, not zero values in general, as zero values (such as empty structs) could be supplied for valid reasons.

Oops, I meant `Value.IsValid` (which would be true for `interface{}(nil)`).

RE: the second point: That's reasonable. There's no valid use case for `fx.Supply(err)` where `err` is known to be an `error`.

supply_test.go
```diff
@@ -0,0 +1,99 @@
...
+			t,
+			"untyped nil passed to fx.Supply",
+			func() { fx.Supply(A{}, nil) },
+			"a naked nil should panic")
+
+		require.NotPanicsf(
+			t,
+			func() {
+				fx.Supply(A{}, (*B)(nil))
+			},
```
abhinav: can match spacing of the cases above and below.
```suggestion
			func() { fx.Supply(A{}, (*B)(nil)) },
```

annotated.go
```diff
@@ -52,6 +52,8 @@ import (
 //   })
 //
 // Annotated cannot be used with constructors which produce fx.Out objects.
+//
+// When used with fx.Supply, the target is a value rather than a constructor function.
```
glibsm: Would it make sense to separate the two? I think we're still in the `is the function a value as well` dilemma that is being solved by separating `fx.Supply` from just extending the `fx.Provide`, right?

I think it could make sense to add another member like `Value interface{}`

supply.go
```diff
@@ -0,0 +1,87 @@
...
+//  fx.Supply(a, b, fx.Annotate{Target: c})
+//
+//  fx.Provide(
+//  	func() *TypeA { return a },
+//  	func() TypeB { return b },
+//  	fx.Annotate{Target: func() *TypeC { return c }},
+//  )
+//
+// Supply panics if a value (or annotation target) is an untyped nil or an error.
+//
```
glibsm: extra line of doc

supply_test.go
```diff
@@ -0,0 +1,99 @@
...
+		app := fxtest.New(t, fx.Supply(aIn, bIn), fx.Populate(&aOut, &bOut))
+		defer app.RequireStart().RequireStop()
+
+		require.NoError(t, app.Err())
+		require.Equal(t, aIn, aOut)
+		require.Equal(t, bIn, bOut)
+	})
+
+	t.Run("AnnotateIsSupplied", func(t *testing.T) {
+		firstIn, secondIn, thirdIn := &A{}, &A{}, &B{}
```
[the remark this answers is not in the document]
roberth-k: I wasn't sure why this would be necessary, but it makes sense given `require.Equal` checks for deep equality. I've kept the structs as-is, but used `require.Same` to be explicit.

annotated.go
roberth-k: The function-value ambiguity isn't an issue with `fx.Annotate`, as the interpretation of `Target` mirrors the option function it's used with: `fx.Provide` expects `Target` to be a function (proven by https://github.com/uber-go/fx/blob/master/app_test.go#L199), and `fx.Supply` will always wrap the target.

roberth-k:
@glibsm It's a fair point about the visibility. It's not completely zero fortunately, as one can still identify the type being provided. I can think of two options for improvement:

(a) Hiding `reflect.makeFuncStub()` with a dummy phrase like `fx.Supply()`.
(b) Getting the user's call site in `fx.Supply()`.

In either case, the debug messages originate from `fxlog`, and modifying that should be enough. The path of minimal refactoring would probably involve some marker function type that `fx.Supply` and `fxlog` understand -- I'll try out a few experiments in a bit. It'd definitely solve (a), but not necessarily (b). The downside of such an approach is the obvious spaghetti factor increase this would bring.

prashantv:
Agree on the visibility front, I think all the existing fx functionality tracks the stack correctly, but we've seen this same issue with helpers that exist outside of `fx`. I actually ran into exactly this issue when building the same functionality outside of fx, see #604 for the task related to improving the stack trace.

annotated.go
abhinav: FWIW we named the field `Target` instead of constructor specifically so that it could have values in the future. The original implementation named it something like `Func` or `Constructor`.

internal/fxlog/fxlog.go
```diff
@@ -58,6 +58,14 @@ func (l *Logger) PrintProvide(t interface{}) {
 	}
 }
 
+// PrintSupply logs a type supplied directly into the dig.Container
+// by the given constructor function.
+func (l *Logger) PrintSupply(constructor interface{}) {
```
abhinav, line 63: Not a problem in this PR but this exposed that we weren't logging names/groups for fx.Annotated constructors at all. Created #703 for us to fix that.

supply.go
```diff
@@ -0,0 +1,122 @@
+// Copyright (c) 2019 Uber Technologies, Inc.
```
abhinav, line 1:
```suggestion
// Copyright (c) 2020 Uber Technologies, Inc.
```

supply_test.go
```diff
@@ -0,0 +1,104 @@
+// Copyright (c) 2019 Uber Technologies, Inc.
```
abhinav, line 1:
```suggestion
// Copyright (c) 2020 Uber Technologies, Inc.
```

roberth-k:
@glibsm Supplies are now logged as such:

```
[Fx] SUPPLY	*fx_test.A
[Fx] SUPPLY	*fx_test.B
```

This could be made to log `fx.Supply`'s caller using the stack trace of the `provide`; I'm punting this decision, as one might find it confusing next to PROVIDE logs. https://github.com/uber-go/fx/issues/604 could be revisit candidate.

abhinav:
@roberth-k Thanks a lot! We'll release this shortly.

Issue #696, opened 2020-02-11
add ordered Invoke support?


runner-mei:
In my app,  I must first run db initsql.   run db initsql by Invoke.   but it isnot first Invoke.

abhinav:
Hello! We specifically planned for the ordering of invokes to be
unspecified. The reason for this is that if ordering matters, that
requirement should be specified in the form of dependencies: If A
depends on B to have started, then A should have a dependency on B.

Also, we do not recommend actually establishing the database connection
inside an Invoke, but in the OnStart hook. Inside an OnStart hook,
you're guaranteed that the dependencies of that hook have already
started.

PR #702, opened 2020-04-01
Add invoke sorting

glibsm:
Invoke ordering has been a huge problem in the last years
that Fx has been live. Executing invokes in the order they
are provided to the fx.New() call is simply not good enough
as it often requires developers to think through the import
order of their modules, and how they interact with one
another.

Luckily help is at hand with a new fx.Option -- fx.Sort! Now
developers can finally take control of their own destiny
when it comes to the order of execution of invokes with the
two very deterministic launch orders: alphabetical, and by
length of the fully qualified function name (comes out of
the box with package name support!!!).

I think this is going to take the possibilities through the
roof.

The framework is there for adding more sorting options in
the future. Some ideas that have been thrown out:
- Random
- Reverse
- Probability
- Time of day
- Leap year

The possibilities are literally endless.

### DIFF
--- a/app.go
+++ b/app.go
@@ -28,6 +28,7 @@ import (
 	"os"
 	"os/signal"
 	"reflect"
+	"sort"
 	"strings"
 	"sync"
 	"syscall"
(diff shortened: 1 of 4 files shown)


blampe:
This came up recently in https://github.com/uber-go/fx/issues/696.

This deserves better documentation for sure, but I don't think we want to add functionality to support ordering invokes.

yurishkuro:
What are those "huge problems" that this would solve? Ordering is already possible by declaring explicit (even if dummy) dependencies.

glibsm:
image

glibsm:
Since it's no longer 4/1, closing.

blampe:
You got me! Good one :)

PR #705, opened 2020-04-22
Fix a couple of docs problems

hsanjuan:
I am not signing your CLA because I don't know why you need my phone number for nor what you are going to do with that data.


abhinav:
Hey @hsanjuan, thanks for the change!

> I am not signing your CLA because I don't know why you need my phone number for nor what you are going to do with that data.

Unfortunately, we cannot accept the change without the CLA. However, this is a fair concern and your comment gave us the opportunity to reevaluate what information is absolutely necessary from people signing the CLA. We have made the phone number optional. We'll be able to accept this change once you sign the CLA.

Thanks!

hsanjuan:
Thanks @abhinav , glad to hear that. CLA is signed.

PR #706, opened 2020-05-29
Introduce fx.ValidateGraph which validates graph dependencies without invoking them

shirchen:
`fx.ValidateGraph` allows graph validation for correct dependencies and lack of cycles without running anything. This is useful if an Invoke has side-effects, does IO, etc.

Depends on https://github.com/uber-go/dig/pull/266

Ref T3706201


go.mod
```diff
@@ -10,3 +10,7 @@ require (
 	golang.org/x/lint v0.0.0-20190930215403-16217165b5de
 	golang.org/x/tools v0.0.0-20191114200427-caa0b0f7d508 // indirect
 )
+
+replace (
```
shirchen, line 14: TODO: remove after landing dig changes

app.go
```diff
@@ -399,6 +400,36 @@ func (ehl errorHandlerList) HandleError(err error) {
...
+	app.validate = o.validate
+}
+
+func (o validateOption) String() string {
+	return fmt.Sprintf("fx.validate(%v)", o.validate)
+}
+
+// ValidateApp validates that supplied graph would run and is not missing any dependencies. This
+// method does not invoke actual input functions.
+func ValidateApp(opts ...Option) error {
```
robbertvanginkel, line 424: I think originally this was proposed for the `fxtest` package. Is there a usecase for ValidateApp in the main package that we haven't discussed?

shirchen: Having `ValidateApp` in `fxtest` would require exposure of a `dry` getter/setter in Fx which isn't desired right now. We want a `dry` flag for us, not users (at least for this feature). So, in order to keep `dry` private, I moved the `ValidateApp` into main package. It also only returns an error anyway, so no need for any extra testing facilities. 

Advantage of exposing `Dry` to users would have been an easier migration for existing fxtests, where u would just need to add the extra option to existing tests and be able to remove all of the setup of side effects that you have. However, this is a minor advantage outweighed by exposure of `Dry` which we can avoid.

app.go
```diff
@@ -399,6 +400,36 @@ func (ehl errorHandlerList) HandleError(err error) {
...
+
+func (o validateOption) apply(app *App) {
+	app.validate = o.validate
+}
+
+func (o validateOption) String() string {
+	return fmt.Sprintf("fx.validate(%v)", o.validate)
+}
+
+// ValidateApp validates that supplied graph would run and is not missing any dependencies. This
```
SyrieBianco, line 422: If we're going to keep this in the main package, I would probably include in the comment that this is intended for testing.

app.go
robbertvanginkel: I was thinking about how to get it done using an internal package, but the only solution I came up with is refactoring the options with an extra level of indirection. The `Option` interface would change its `apply(*App)` to an `apply(*internal.AppParams)`, and after applying all options we'd convert an `internal.AppParam` to `App`.

Agree it seems a little much, having it on fx is probably fine.

app.go
```diff
@@ -399,6 +400,36 @@ func (ehl errorHandlerList) HandleError(err error) {
...
+// method does not invoke actual input functions.
```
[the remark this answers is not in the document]
abhinav: If we add that, instead of saying "X is for testing" I have a slight preference for "X may be used to test Y." The former can cause us to be lax with breaking the API because it's for testing.

app.go
```diff
@@ -399,6 +400,36 @@ func (ehl errorHandlerList) HandleError(err error) {
...
+func ValidateApp(opts ...Option) error {
+	opts = append(opts, validate(true))
+	app := New(opts...)
+	if err := app.Err(); err != nil {
+		return err
+	}
+	return nil
```
abhinav, line 430: nit: can simplify
```suggestion
	return app.Err()
```

app_test.go
```diff
@@ -681,6 +681,76 @@ func TestAppStop(t *testing.T) {
 	})
 }
 
+func TestValidateApp(t *testing.T) {
+	t.Run("do not run provides on graph validation", func(t *testing.T) {
+		type type1 struct{}
+		err := ValidateApp(
+			Provide(func() *type1 {
+				panic("always")
```
abhinav, line 689: Let's make this a t.Error.

```suggestion
				t.Error("provide must not be called")
				return nil
```

app_test.go
```diff
@@ -681,6 +681,76 @@ func TestAppStop(t *testing.T) {
...
+		require.NoError(t, err)
+	})
+	t.Run("do not run provides nor invokes on graph validation", func(t *testing.T) {
+		type type1 struct{}
+		err := ValidateApp(
+			Provide(func() *type1 {
+				panic("always")
+			}),
+			Invoke(func(*type1) {
+				panic("always")
```
abhinav, line 702: same
```suggestion
			Provide(func() *type1 {
				t.Error("provide must not be called")
				return nil
			}),
			Invoke(func(*type1) {
				t.Error("invoke must not be called")
```

PR #712, opened 2020-08-19
Setting minimum dependency on dig to 1.10 for released fx.ValidateGraph feature.

shirchen:
Forgot to increase minimum `dig` dependency as part of 1.13.0 `fx` release.


AlexAPB:
<3

PR #716, opened 2020-10-20
Add Priority field to Hook

dgladkov:
Proof of concept for priority idea. `PriorityLowest` would allow Hooks that start serving user IO (e.g. gRPC/HTTP servers) to run last as they require the whole application to be initialized and operational.

Few downsides of this approach:
1. No way to order the lowest priority hooks. Should not be a big problem since hooks are still expected to be independent of other hooks.
2. Requires fair use by Hooks author - if a hook is assigned a priority that does reflect it's use-case, we're going to have the same issues. For example, if a hook that initializes a database session is assigned the lowest priority same as RPC server, the same race condition will occur.

### DIFF
--- a/internal/lifecycle/lifecycle.go
+++ b/internal/lifecycle/lifecycle.go
@@ -33,13 +33,13 @@ import (
 type Priority int
 
 const (
-	// PriorityLowest hooks are executed last.
+	// PriorityLowest hooks are started last and stoppped first.
 	// For example, hooks that start serving network traffic might wish to execute
 	// only after all other application components are initialized.
 	PriorityLowest Priority = -1000
 	// PriorityNormal is the default priority level
 	PriorityNormal = 0
-	// PriorityHighest hooks are executed first.
+	// PriorityHighest hooks are started first and stopped last.
 	PriorityHighest = 1000
 )
 
--- a/internal/lifecycle/lifecycle.go
+++ b/internal/lifecycle/lifecycle.go
@@ -70,15 +70,17 @@ func New(logger *fxlog.Logger) *Lifecycle {
 // Append adds a Hook to the lifecycle.
 func (l *Lifecycle) Append(hook Hook) {
 	hook.caller = fxreflect.Caller()
+	// Can be replaced with priority double ended queue,
+	// but is it really necessary?
 	l.hooks = append(l.hooks, hook)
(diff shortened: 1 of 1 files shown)


internal/lifecycle/lifecycle.go
```diff
@@ -22,18 +22,34 @@ package lifecycle
...
+
 // A Hook is a pair of start and stop callbacks, either of which can be nil,
 // plus a string identifying the supplier of the hook.
 type Hook struct {
-	OnStart func(context.Context) error
-	OnStop  func(context.Context) error
-	caller  string
+	OnStart  func(context.Context) error
+	OnStop   func(context.Context) error
+	Priority Priority
```
jquirke, line 51: This line is technically a breaking change

jquirke: Yes it is a breaking change. If this structure is embedded anonymous with another structure with the field Priority
https://play.golang.org/p/TlpCWwQIjvP

abhinav: Hey! So, embedding of structs is not covered by the [Go 1 Compatibility Expectations](https://golang.org/doc/go1compat) so this wouldn't be a breaking change if we were to make it. Otherwise we would never be able to add new fields and methods to existing structs.

abhinav (changes requested):
Hey, thanks for the change!

I'd like to understand your problem better because I have a strong preference to avoid z-indexing as an API. The order of the hooks will be the order they are appended in, which will be the order in which items are initialized. So the `OnStart` for the dependencies of Foo will be before the `OnStart` for Foo.

Can you explain the problem you're running into in more detail and we can help you find a solution/workaround for it. If the solution requires changes to Fx, we're open to that but that'll likely require an internal discussion first. Lifecycle management is one of the more finnicky parts of Fx and we've had multiple attempts to "fix" it, but there are always some issues.

dgladkov:
Thanks for the review @abhinav. I thought a bit more and agree with you that this change is not necessary. If constructors and their Lifecycles are always executed in the topologic sort order, this would ensure that all dependencies are fully initialized.

In my example provided in the PR summary, the RPC handler lifecycle would be last to run anyway since it depends on the rest of the application objects, which in turn means IO naturally would not be served unless until app init, even without explicit lifecycle ordering.

The problem we faced is most likely is due to the structure of our app's dependency graph

PR #717, opened 2020-10-27
Add structured logger to internal Fx logging.

shirchen:
Introducing a structured logger, zap, to the internal logger. This is to address a common user complaint where logs are just a wall of text which is hard to parse through and find the error that's causing application to fail.

In this diff we implement current public API `Logger` via Zap, and adding private (for now) `withLogger` and `exitCode` which will be public once we iterate on the new logging interface. 

Following is a list of all new messages and fields.

msg | field | field
-- | -- | --
providing | type | constructor
supplying | type |  
invoke | function |  
received signal | signal |  
startup failed, rolling back | error |  
could not rollback cleanly | error |  
error encountered while applying options | error |  
fx.Invoke failed | function |  error |
failed to start | error |  
failed to stop cleanly | error |  
running |   |  

Ref T5424311 GO-77


app.go
```diff
@@ -261,22 +263,65 @@ func (t stopTimeoutOption) String() string {
...
 type loggerOption struct{ p Printer }
 
+type printerWrapper struct {
+	p Printer
+}
+
+// NewPrinter returns an implementation of zapcore.WriteSyncer
+// used to support Logger option which implements Printer
+// interface.
+func NewPrinter(p Printer) zapcore.WriteSyncer {
```
abhinav, line 304: It doesn't seem like this needs to be exported. It is an internal detail of how we convert a Printer to a WriteSyncer.

app.go
```diff
@@ -464,7 +529,12 @@ func New(opts ...Option) *App {
 	app.provide(provide{Target: app.dotGraph, Stack: frames})
 
 	if app.err != nil {
-		app.logger.Printf("Error after options were applied: %v", app.err)
+		fxlog.Info(
+			"error encountered while applying options",
+			fxlog.Field{
+				Key:   "error",
+				Value: app.err,
+			}).Write(app.log)
```
abhinav, line 537: You may optionally define a top-level `Err` function in the fxlog package that builds an `fxlog.Field` with the key "error"

```
func Err(err error) Field
```

Or better yet, instead of specifying `fxlog.Field{Key: ..., Value: ...}` everywhere, we can also switch to an `fxlog.F` helper function:

```
func F(string, interface{}) Field
```

It doesn't matter much because it's all internal—it's just a matter of how usable we find it inside Fx.

app.go
```diff
@@ -706,16 +798,26 @@ func (app *App) run(done <-chan os.Signal) {
...
+		fxlog.Error("failed to start", fxlog.Field{
+			Key:   "error",
+			Value: err,
+		}).Write(app.log)
+		// TODO: add option to override os.Exit behavior.
+		os.Exit(1)
 	}
 
-	app.logger.PrintSignal(<-done)
+	fxlog.Info(strings.ToUpper((<-done).String())).Write(app.log)
```
abhinav, line 809: I would change this to "received signal" and then make the signal a field of the log statement.

And just for readability, read from channel should be separate from stringifying it and logging it.

fxtest/printer.go
```diff
@@ -34,3 +37,23 @@ func NewTestPrinter(t TB) fx.Printer {
...
+func (t *testLogger) PrintSupply(interface{}) {
+
+}
+
+func (t *testLogger) PrintProvide(interface{}) {
+
+}
+
+func NewLogger(t TB) fxlog.Logger {
+	return &testLogger{t}
```
abhinav, line 58: I appreciate the intent but: if fxlog.Logger is private, then this shouldn't be public.

Also, my expectation for a test logger would be that it logs to the testing.TB, which is not the case here.

To make that true, we can export from fxlog, a means of building a `fxlog.Logger` from a `*zap.Logger`. And then we can simply pass a `zaptest.NewLogger(t)` to it.

internal/fxlog/fxlog.go
```diff
@@ -21,68 +21,121 @@
...
 
-// Printer is a formatting printer.
-type Printer interface {
-	Printf(string, ...interface{})
+// Level is the level of logging used by Logger.
+type Level int
+
+const (
+	InfoLevel = iota
+	ErrorLevel
```
abhinav, line 39: This should specify the type for the first case.

```suggestion
	InfoLevel Level = iota
	ErrorLevel
```

(ErrorLevel will take type level because Info has type level)

internal/fxlog/fxlog.go
```diff
@@ -21,68 +21,121 @@
...
+	Fields  []Field
+	Stack   string
 }
 
-// New returns a new Logger backed by the standard library's log package.
-func New() *Logger {
-	return &Logger{log.New(os.Stderr, "", log.LstdFlags)}
+func (e Entry) WithStack(stack string) Entry {
+	e.Stack = stack
+
```
abhinav, line 52: we should probably delete this line
```suggestion
```

internal/fxlog/fxlog.go
```diff
@@ -21,68 +21,121 @@
...
 }
 
-// PrintProvide logs a type provided into the dig.Container.
-func (l *Logger) PrintProvide(t interface{}) {
-	for _, rtype := range fxreflect.ReturnTypes(t) {
-		l.Printf("PROVIDE\t%s <= %s", rtype, fxreflect.FuncName(t))
+type Logger interface {
+	Log(entry Entry)
+	PrintProvide(interface{})
+	PrintSupply(interface{})
```
abhinav, line 68: We should try to keep the interface small. PrintProvide and PrintSupply can be implemented in terms of Logger.Log. In fact, the implementation will likely always be the same. So we can make them top-level functions that operate on a logger rather than methods of the logger:

```
func PrintProvide(Logger, interface{})
func PrintSupply(Logger, interface{})
```

internal/fxlog/fxlog.go
```diff
@@ -21,68 +21,121 @@
...
+	return fs
 }
 
-// PrintSupply logs a type supplied directly into the dig.Container
-// by the given constructor function.
-func (l *Logger) PrintSupply(constructor interface{}) {
-	for _, rtype := range fxreflect.ReturnTypes(constructor) {
-		l.Printf("SUPPLY\t%s", rtype)
+func (l *zapLogger) Log(entry Entry) {
+	switch entry.Level {
```
abhinav, line 87: We appear to be ignoring the stack trace?

internal/fxlog/fxlog.go
```diff
@@ -21,68 +21,121 @@
...
+	}
+}
+
+func Error(msg string, fields ...Field) Entry {
+	return Entry{
+		Level:   ErrorLevel,
+		Message: msg,
+		Fields:  fields,
+	}
+
```
abhinav, line 140: unnecessary empty line
```suggestion
```

internal/fxlog/fxlog.go
```diff
@@ -21,68 +21,121 @@
...
 
-// Fatalf logs an Fx line then fatals.
-func (l *Logger) Fatalf(format string, v ...interface{}) {
-	l.Printer.Printf(prepend(format), v...)
-	_exit()
+// DefaultLogger constructs a Logger out of io.Writer.
+func DefaultLogger(ws zapcore.WriteSyncer) Logger {
+	zcore := zapcore.NewCore(
+		zapcore.NewJSONEncoder(zap.NewProductionEncoderConfig()),
+		zapcore.Lock(ws),
```
abhinav, line 116: We shouldn't add a mutex around the write syncer.

This is for when we know that the underlying implementation isn't thread-safe,
which isn't something we can assume. We can let the caller decide if they want
to add a locking wrapper around their write syncer.

app.go
```diff
@@ -261,22 +263,65 @@ func (t stopTimeoutOption) String() string {
 	return fmt.Sprintf("fx.StopTimeout(%v)", time.Duration(t))
 }
 
+func WithLogger(l fxlog.Logger) Option {
```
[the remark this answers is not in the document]
shirchen: We have external tests in fx_test so unfortunately it looks like either we need to disable those tests or keep this public for those tests only.

.gitignore
```diff
@@ -12,3 +12,4 @@ coverage.txt
 /.bin
 /.cache
 /bin
+/.idea
```
twilly, line 15: This should't be a part of the this repo. Add IDE excludes to your [global gitignore](https://git-scm.com/docs/gitignore#_configuration).

app.go
```diff
@@ -261,22 +263,69 @@ func (t stopTimeoutOption) String() string {
...
+// Note, this will be deprecate with next release and you will need to implement
+// fxlog.Logger interface instead.
 type Printer interface {
 	Printf(string, ...interface{})
 }
 
 // Logger redirects the application's log output to the provided printer.
+//
+// Note, this will be deprecate with next release and you will need to use
+// WithLogger instead.
```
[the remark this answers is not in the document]
abhinav: This affects Fx's logging, not what the applications use. That should be fine, right?

app.go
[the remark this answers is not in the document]
abhinav: I don't want to expose this function while fxlog is private. There's a workaround to the test problem: keep this as `withLogger` here, then create a _test file in the same package that re-exports it as WithLogger, and use that from the external test.

```
// app.go
package fx

func withLogger(...) Option

// opt_test.go
package fx

func WithLogger(...) Option { return withLogger(...) }

// whatever_test.go
package fx_test

// can use fx.WithLogger now even though it's not part of the public API.
```

A variant of this is demonstrated in https://medium.com/@robiplus/golang-trick-export-for-test-aa16cbd7b8cd.

app.go
```diff
@@ -34,10 +34,12 @@ import (
 	"time"
 
 	"go.uber.org/dig"
+	"go.uber.org/multierr"
+	"go.uber.org/zap/zapcore"
```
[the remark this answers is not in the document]
abhinav: Part of the plan is to keep zap as an implementation detail of Fx's "internal" logging.
@shirchen perhaps the broad ideas of our design discussions should be part of the PR message.

app.go
```diff
@@ -261,22 +263,69 @@ func (t stopTimeoutOption) String() string {
...
+//
+// Note, this will be deprecate with next release and you will need to implement
+// fxlog.Logger interface instead.
 type Printer interface {
 	Printf(string, ...interface{})
 }
 
 // Logger redirects the application's log output to the provided printer.
+//
+// Note, this will be deprecate with next release and you will need to use
```
abhinav, line 293:
```suggestion
// Note, this will be deprecated with the next release and you will need to use
```

app.go
```diff
@@ -261,22 +263,69 @@ func (t stopTimeoutOption) String() string {
...
+// interface.
+func newPrinter(p Printer) zapcore.WriteSyncer {
+	return &printerWrapper{
+		p: p,
+	}
+}
+
+func (p *printerWrapper) Write(b []byte) (n int, err error) {
+	p.p.Printf(string(b))
+
```
abhinav, line 316: nit: unnecessary newline
```suggestion
```

app.go
```diff
@@ -261,22 +263,69 @@ func (t stopTimeoutOption) String() string {
...
+func (p *printerWrapper) Sync() error {
+	return nil
+}
+
 func (l loggerOption) apply(app *App) {
-	app.logger = &fxlog.Logger{Printer: l.p}
-	app.lifecycle = &lifecycleWrapper{lifecycle.New(app.logger)}
+	np := newPrinter(l.p)
+
+	app.log = fxlog.DefaultLogger(zapcore.Lock(np)) // assuming np isnt thread-safe.
```
abhinav, line 327: this may not be necessary. we don't do any concurrent logging in fx, and it may not be unreasonable to ask the caller to make fx.Printer thread safe. before this change, fx.Printer was expected to be a [log.Logger](https://golang.org/pkg/log/#Logger) which is already thread safe.

app.go
```diff
@@ -285,13 +334,15 @@ func (l loggerOption) String() string {
 
 // NopLogger disables the application's log output. Note that this makes some
 // failures difficult to debug, since no errors are printed to console.
-var NopLogger = Logger(nopLogger{})
+//
+// Note, when WithLogger is public we will make the change here as well.
+var NopLogger = WithLogger(nopLogger{})
```
abhinav, line 339:
```suggestion
var NopLogger = withLogger(nopLogger{})
```

app.go
```diff
@@ -618,9 +672,18 @@ func (app *App) provide(p provide) {
 
 	switch {
 	case p.IsSupply:
-		app.logger.PrintSupply(constructor)
+		// app.log.PrintSupply(constructor)
```
abhinav, line 675: did you mean to delete this

app.go
```diff
@@ -618,9 +672,18 @@ func (app *App) provide(p provide) {
...
+		for _, rtype := range fxreflect.ReturnTypes(constructor) {
+			fxlog.Info("supplying", fxlog.Field{Key: "constructor", Value: rtype}).Write(app.log)
+		}
 	default:
-		app.logger.PrintProvide(constructor)
+		// app.log.PrintProvide(constructor)
```
abhinav, line 680: same

app.go
```diff
@@ -618,9 +672,18 @@ func (app *App) provide(p provide) {
...
+		for _, rtype := range fxreflect.ReturnTypes(constructor) {
+			fxlog.Info("providing",
+				fxlog.Field{Key: "return value", Value: rtype},
```
abhinav, line 683: field names should be alphanumeric without spaces. these are typically json object keys that are ideally addressable in the form `obj.foo`. so, `returnValue`, not `return value`.

app.go
```diff
@@ -681,7 +744,7 @@ func (app *App) executeInvokes() error {
 	for _, i := range app.invokes {
 		fn := i.Target
 		fname := fxreflect.FuncName(fn)
-		app.logger.Printf("INVOKE\t\t%s", fname)
+		fxlog.Info("invoke", fxlog.F("funcname", fname)).Write(app.log)
```
abhinav, line 747: we should try to go for consistent key naming patterns. if we're using "return value" (or the camelCase variant), then let's use "function" or "function name"  (or its camelCase variant). otherwise, we'd want to use retval which is less helpful.

app.go
```diff
@@ -693,7 +756,11 @@ func (app *App) executeInvokes() error {
 		}
 
 		if err != nil {
-			app.logger.Printf("fx.Invoke(%v) called from:\n%+vFailed: %v", fname, i.Stack, err)
+			fxlog.Error(
+				"fx.Invoke failed",
+				fxlog.F("called from", fname),
```
abhinav, line 761: calledFrom

app.go
```diff
@@ -706,16 +773,20 @@ func (app *App) run(done <-chan os.Signal) {
 	defer cancel()
 
 	if err := app.Start(startCtx); err != nil {
-		app.logger.Fatalf("ERROR\t\tFailed to start: %v", err)
+		fxlog.Error("failed to start", fxlog.F("error", err)).Write(app.log)
```
abhinav, line 776: optional:

```
package fxlog

func Err(err error) Field
```

This can be short for `fxlog.F("error", err)`

internal/fxlog/fxlog.go
```diff
@@ -21,68 +21,108 @@
...
+	return &zapLogger{
+		logger: log,
+	}
 }
 
-// Fatalf logs an Fx line then fatals.
-func (l *Logger) Fatalf(format string, v ...interface{}) {
-	l.Printer.Printf(prepend(format), v...)
-	_exit()
+func Info(msg string, fields ...Field) Entry {
```
abhinav, line 114: we should document this and Error

app.go
shirchen: A message would then look like `fx.Invoke failed calledFrom` which is a bit awkward, no?

abhinav: No, because the structured logs will be written as JSON like:

```
{"msg": "fx.Invoke failed", "calledFrom": "..."}
```

app.go
```diff
@@ -34,10 +34,12 @@ import (
 	"time"
 
 	"go.uber.org/dig"
+	"go.uber.org/multierr"
+	"go.uber.org/zap/zapcore"
+
```
abhinav, line 39: we don't need to separate the import groups here.
```suggestion
```

app_internal_test.go
```diff
@@ -46,6 +46,29 @@ func TestAppRun(t *testing.T) {
 	wg.Wait()
 }
 
+// TestAppRunStart tries to trigger a startup error when running an app and not exit(1).
+func TestAppRunStart(t *testing.T) {
+	// Setting exit code to be successful.
+	app := New(Invoke(struct{}{}), ExitCode(0))
```
abhinav, line 52: If deleting this option causes the test to fail, that means that the test binary is being exited. We definitely do not want to exit the test binary.

app.go
```diff
@@ -261,22 +263,83 @@ func (t stopTimeoutOption) String() string {
...
+type exitCodeOption int
+
+func (e exitCodeOption) apply(app *App) {
+	app.exitCode = int(e)
+}
+
+func (e exitCodeOption) String() string {
+	return fmt.Sprintf("fx.exitCode(%v)", int(e))
+}
```
abhinav, line 280: we don't need control of the exit code for the failure path, and we definitely should not have tests that run os.Exit under any circumstances because that exits the test binary. I would rather lose the coverage on the lines with `os.Exit` than have tests silently succeed with `os.Exit(0)`.

app.go
```diff
@@ -261,22 +262,67 @@ func (t stopTimeoutOption) String() string {
...
+type printerWrapper struct {
+	p Printer
+}
+
+// newPrinter returns an implementation of zapcore.WriteSyncer
+// used to support Logger option which implements Printer
+// interface.
+func newPrinter(p Printer) zapcore.WriteSyncer {
+	return &printerWrapper{
+		p: p,
```
abhinav, line 309: I would move this and its implementation either to its own file (`printer.go`?) or into fxlog (`fxlog.WriteSyncerFrom(Printer)`?).

app.go
```diff
@@ -261,22 +262,67 @@ func (t stopTimeoutOption) String() string {
...
+func (p *printerWrapper) Sync() error {
+	return nil
+}
+
 func (l loggerOption) apply(app *App) {
-	app.logger = &fxlog.Logger{Printer: l.p}
-	app.lifecycle = &lifecycleWrapper{lifecycle.New(app.logger)}
+	np := newPrinter(l.p)
+	app.log = fxlog.DefaultLogger(np) // assuming np is thread-safe.
+	app.lifecycle = &lifecycleWrapper{lifecycle.New(app.log)}
```
abhinav, line 325: Instead of setting up the lifecycle here as well as on line 482, why not set the lifecycle after applying options.

```
app := &App{
  // ...
}

for _, opt := options { opt.apply(app) }

app.lifecycle = ...
```

app.go
```diff
@@ -618,9 +667,16 @@ func (app *App) provide(p provide) {
 
 	switch {
 	case p.IsSupply:
-		app.logger.PrintSupply(constructor)
+		for _, rtype := range fxreflect.ReturnTypes(constructor) {
+			fxlog.Info("supplying", fxlog.Field{Key: "constructor", Value: rtype}).Write(app.log)
```
abhinav, line 671: the key should be type or returnType. `fx.Supply(v)` is used to provide direct values.

```suggestion
			fxlog.Info("supplying", fxlog.Field{Key: "type", Value: rtype}).Write(app.log)
```

app.go
```diff
@@ -618,9 +667,16 @@ func (app *App) provide(p provide) {
...
+		}
 	default:
-		app.logger.PrintProvide(constructor)
+		for _, rtype := range fxreflect.ReturnTypes(constructor) {
+			fxlog.Info("providing",
+				fxlog.Field{Key: "returnValue", Value: rtype},
```
abhinav, line 676: this is a type, not a value. returnType or type.

```suggestion
				fxlog.Field{Key: "type", Value: rtype},
```

app.go
```diff
@@ -693,7 +749,11 @@ func (app *App) executeInvokes() error {
 		}
 
 		if err != nil {
-			app.logger.Printf("fx.Invoke(%v) called from:\n%+vFailed: %v", fname, i.Stack, err)
+			fxlog.Error(
+				"fx.Invoke failed",
+				fxlog.F("calledFrom", fname),
```
abhinav, line 754: I know we already renamed this but it may be better to align it with the field name we already use for functions.

```suggestion
				fxlog.F("function", fname),
```

app_test.go
```diff
@@ -29,13 +29,15 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/multierr"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+
```
abhinav, line 37: please delete this line 

```suggestion
```

app_test.go
```diff
@@ -64,9 +66,14 @@ func TestNewApp(t *testing.T) {
 		// (e.g., logging) from changing halfway through our provides.
 
 		spy := new(fxlog.Spy)
-		app := fxtest.New(t, Provide(func() struct{} { return struct{}{} }), Logger(spy))
+		app := fxtest.New(t, Provide(func() struct{} { return struct{}{} }), WithLogger(spy))
 		defer app.RequireStart().RequireStop()
-		assert.Contains(t, spy.String(), "PROVIDE\tstruct {}")
+		assert.Contains(t, spy.String(), "providing")
+		assert.Contains(t, spy.Fields(), zap.Field{
+			Key:    "returnValue",
```
abhinav, line 73:
```suggestion
			Key:    "type",
```

app_test.go
```diff
@@ -420,9 +427,15 @@ func TestOptions(t *testing.T) {
...
+			WithLogger(spy),
 		)
-		assert.Contains(t, spy.String(), "must provide constructor function")
+		assert.Contains(t, spy.String(), "error encountered while applying options")
+		fields := spy.Fields()
+		assert.Len(t, fields, 1)
+		s := fields[0]
+		assert.Equal(t, "error", s.Key)
+		assert.Equal(t, zapcore.ErrorType, s.Type)
+		assert.Contains(t, fmt.Sprintf("%v", s.Interface), "must provide constructor function")
```
abhinav, line 438: can do fmt.Sprint instead of `fmt.Sprintf` with a `%v`

```suggestion
		assert.Contains(t, fmt.Sprint(s.Interface), "must provide constructor function")
```

app_test.go
```diff
@@ -547,10 +560,13 @@ func TestAppStart(t *testing.T) {
 		//         /.../go/1.13.3/libexec/src/testing/testing.go:909
 		// Failed: can't invoke non-function {} (type struct {})
 		output := spy.String()
-		assert.Contains(t, output, "fx.Invoke({}) called from:")
+		fields := spy.Fields()
+		assert.Len(t, fields, 9)
+
+		assert.Contains(t, output, "fx.Invoke failed calledFrom:")
```
abhinav, line 566:
```suggestion
		assert.Contains(t, output, "fx.Invoke failed function:")
```

fxtest/fxtest_test.go
```diff
@@ -47,6 +47,10 @@ func (t *tb) FailNow() {
 	t.failures++
 }
 
+func (t *tb) Error(format string) {
+
+}
```
abhinav, line 52: are you sure we need this?

this isn't the signature of t.Error

fxtest/printer.go
```diff
@@ -20,7 +20,9 @@
 
 package fxtest
 
-import "go.uber.org/fx"
+import (
+	"go.uber.org/fx"
+)
```
abhinav, line 25: can revert this change.
```suggestion
import "go.uber.org/fx"
```

internal/fxlog/fxlog.go
```diff
@@ -21,68 +21,117 @@
...
-	Printer
+func (e Entry) Write(logger Logger) {
+	logger.Log(e)
 }
 
-// Printf logs a formatted Fx line.
-func (l *Logger) Printf(format string, v ...interface{}) {
-	l.Printer.Printf(prepend(format), v...)
+// Err is a helper for error Fields.
+func Err(value interface{}) Field {
```
abhinav, line 57: shouldn't this always be an error

```suggestion
func Err(value error) Field {
```

internal/fxlog/fxlog.go
```diff
@@ -21,68 +21,117 @@
...
+	logger *zap.Logger
+}
+
+func encodeFields(fields []Field, stack string) []zap.Field {
+	var fs []zap.Field
+	for _, field := range fields {
+		fs = append(fs, zap.Any(field.Key, field.Value))
+	}
+	if stack != "" {
+		fs = append(fs, zap.Stack(stack))
```
abhinav, line 92: this doesn't work like we want it to work.

https://pkg.go.dev/go.uber.org/zap#Stack accepts the name of the JSON key that should contain the stack trace, but it captures its own stack trace.

For now, I would do this:

```suggestion
	if stack != "" {
		fs = append(fs, zap.String("stack", stack))
```

internal/fxlog/fxlog_test.go
```diff
@@ -22,52 +22,64 @@ package fxlog
...
 	"bytes"
-	"errors"
 	"os"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"go.uber.org/dig"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+
```
abhinav, line 32:
```suggestion
```

internal/fxlog/fxlog_test.go
```diff
@@ -22,52 +22,64 @@ package fxlog
...
+		assert.Equal(t, "foo 42\n", sink.String())
 	})
 
 	t.Run("printProvide", func(t *testing.T) {
 		sink.Reset()
-		logger.PrintProvide(bytes.NewBuffer)
-		assert.Equal(t, "[Fx] PROVIDE\t*bytes.Buffer <= bytes.NewBuffer()\n", sink.String())
+		for _, rtype := range fxreflect.ReturnTypes(bytes.NewBuffer) {
+			Info("providing",
+				Field{Key: "returnValue", Value: rtype},
```
abhinav, line 55:
```suggestion
				Field{Key: "type", Value: rtype},
```

internal/fxlog/fxlog_test.go
```diff
@@ -22,52 +22,64 @@ package fxlog
...
+				Field{Key: "constructor", Value: fxreflect.FuncName(bytes.NewBuffer)},
+			).Write(sink)
+		}
+		assert.Contains(t, sink.String(), "providing")
+		assert.Contains(t, sink.Fields(), zap.Field{
+			Key:    "returnValue",
```
abhinav, line 61:
```suggestion
			Key:    "type",
```

internal/fxlog/fxlog_test.go
```diff
@@ -22,52 +22,64 @@ package fxlog
...
+			String: "bytes.NewBuffer()",
+		})
 	})
 
 	t.Run("PrintSupply", func(t *testing.T) {
 		sink.Reset()
-		logger.PrintSupply(func() *bytes.Buffer { return bytes.NewBuffer(nil) })
-		assert.Equal(t, "[Fx] SUPPLY\t*bytes.Buffer\n", sink.String())
+		for _, rtype := range fxreflect.ReturnTypes(func() *bytes.Buffer { return bytes.NewBuffer(nil) }) {
+			Info("supplying", Field{Key: "constructor", Value: rtype}).Write(sink)
```
abhinav, line 75:
```suggestion
			Info("type", Field{Key: "constructor", Value: rtype}).Write(sink)
```

internal/fxlog/fxlog_test.go
```diff
@@ -82,19 +94,45 @@ func TestPrint(t *testing.T) {
...
 		}
-		logger.PrintProvide(func() Ret { return Ret{} })
-
-		s := sink.String()
-		assert.Contains(t, s, "[Fx] PROVIDE\t*fxlog.A <=")
-		assert.Contains(t, s, "[Fx] PROVIDE\tfxlog.B <=")
-		assert.Contains(t, s, "[Fx] PROVIDE\tfxlog.C:foo <=")
+		for _, rtype := range fxreflect.ReturnTypes(func() Ret { return Ret{} }) {
+			Info("providing",
+				Field{Key: "returnValue", Value: rtype},
```
abhinav, line 99:
```suggestion
				Field{Key: "type", Value: rtype},
```

internal/fxlog/fxlog_test.go
```diff
@@ -82,19 +94,45 @@ func TestPrint(t *testing.T) {
...
+				Field{Key: "constructor", Value: fxreflect.FuncName(func() Ret { return Ret{} })},
+			).Write(sink)
+		}
+		assert.Contains(t, sink.String(), "providing")
+		assert.Contains(t, sink.Fields(), zap.Field{
+			Key:    "returnValue",
```
abhinav, line 105:
```suggestion
			Key:    "type",
```

internal/fxlog/fxlog_test.go
```diff
@@ -82,19 +94,45 @@ func TestPrint(t *testing.T) {
...
+			Type:   zapcore.StringType,
+			String: "*fxlog.A",
+		})
+		assert.Contains(t, sink.Fields(), zap.Field{
+			Key:    "returnValue",
```
abhinav, line 110:
```suggestion
			Key:    "type",
```

internal/fxlog/fxlog_test.go
```diff
@@ -82,19 +94,45 @@ func TestPrint(t *testing.T) {
...
+			Type:   zapcore.StringType,
+			String: "fxlog.B",
+		})
+		assert.Contains(t, sink.Fields(), zap.Field{
+			Key:    "returnValue",
```
abhinav, line 115:
```suggestion
			Key:    "type",
```

internal/fxlog/fxlog_test.go
```diff
@@ -82,19 +94,45 @@ func TestPrint(t *testing.T) {
...
+			String: "fxlog.C:foo",
+		})
 	})
 
 	t.Run("printHandlesDotGitCorrectly", func(t *testing.T) {
 		sink.Reset()
-		logger.PrintProvide(sample.New)
+		for _, rtype := range fxreflect.ReturnTypes(sample.New) {
+			Info("providing",
+				Field{Key: "returnValue", Value: rtype},
```
abhinav, line 125:
```suggestion
				Field{Key: "type", Value: rtype},
```

internal/fxlog/fxlog_test.go
```diff
@@ -109,61 +147,90 @@ func TestPrint(t *testing.T) {
...
 		}
-		logger.PrintProvide(func() Ret { return Ret{} })
-
-		s := sink.String()
-		assert.Contains(t, s, "[Fx] PROVIDE\t*fxlog.A:primary <=")
-		assert.Contains(t, s, "[Fx] PROVIDE\t*fxlog.A:secondary <=")
-		assert.Contains(t, s, "[Fx] PROVIDE\t*fxlog.B:foo <=")
+		for _, rtype := range fxreflect.ReturnTypes(func() Ret { return Ret{} }) {
+			Info("providing",
+				Field{Key: "returnValue", Value: rtype},
```
abhinav, line 152:
```suggestion
				Field{Key: "type", Value: rtype},
```

internal/fxlog/fxlog_test.go
```diff
@@ -109,61 +147,90 @@ func TestPrint(t *testing.T) {
...
+				Field{Key: "constructor", Value: fxreflect.FuncName(func() Ret { return Ret{} })},
+			).Write(sink)
+		}
+		assert.Contains(t, sink.String(), "providing")
+		assert.Contains(t, sink.Fields(), zap.Field{
+			Key:    "returnValue",
```
abhinav, line 158:
```suggestion
			Key:    "type",
```

internal/fxlog/fxlog_test.go
```diff
@@ -109,61 +147,90 @@ func TestPrint(t *testing.T) {
...
+			Type:   zapcore.StringType,
+			String: "*fxlog.A:primary",
+		})
+		assert.Contains(t, sink.Fields(), zap.Field{
+			Key:    "returnValue",
```
abhinav, line 163:
```suggestion
			Key:    "type",
```

internal/fxlog/fxlog_test.go
```diff
@@ -109,61 +147,90 @@ func TestPrint(t *testing.T) {
...
+			Type:   zapcore.StringType,
+			String: "*fxlog.A:secondary",
+		})
+		assert.Contains(t, sink.Fields(), zap.Field{
+			Key:    "returnValue",
```
abhinav, line 168:
```suggestion
			Key:    "type",
```

internal/fxlog/fxlog_test.go
```diff
@@ -109,61 +147,90 @@ func TestPrint(t *testing.T) {
...
 	})
 
 	t.Run("printProvideInvalid", func(t *testing.T) {
 		sink.Reset()
 		// No logging on invalid provides, since we're already logging an error
 		// elsewhere.
-		logger.PrintProvide(bytes.NewBuffer(nil))
+		for _, rtype := range fxreflect.ReturnTypes(bytes.NewBuffer(nil)) {
+			Info("providing",
+				Field{Key: "returnValue", Value: rtype},
```
abhinav, line 180:
```suggestion
				Field{Key: "type", Value: rtype},
```

internal/fxlog/fxlog_test.go
```diff
@@ -109,61 +147,90 @@ func TestPrint(t *testing.T) {
...
 	t.Run("printStripsVendorPath", func(t *testing.T) {
 		sink.Reset()
 		// assert is vendored within fx and is a good test case
-		logger.PrintProvide(assert.New)
-		assert.Contains(
-			t, sink.String(),
-			"*assert.Assertions <= github.com/stretchr/testify/assert.New()")
+		for _, rtype := range fxreflect.ReturnTypes(assert.New) {
+			Info("providing",
+				Field{Key: "returnValue", Value: rtype},
```
abhinav, line 192:
```suggestion
				Field{Key: "type", Value: rtype},
```

internal/fxlog/fxlog_test.go
```diff
@@ -109,61 +147,90 @@ func TestPrint(t *testing.T) {
...
+				Field{Key: "constructor", Value: fxreflect.FuncName(assert.New)},
+			).Write(sink)
+		}
+		assert.Contains(t, sink.Fields(), zap.Field{
+			Key:    "constructor",
+			Type:   zapcore.StringType,
+			String: "github.com/stretchr/testify/assert.New()",
+		})
+		assert.Contains(t, sink.Fields(), zap.Field{
+			Key:    "returnValue",
```
abhinav, line 202:
```suggestion
			Key:    "type",
```

internal/fxlog/fxlog_test.go
```diff
@@ -109,61 +147,90 @@ func TestPrint(t *testing.T) {
...
 	t.Run("printFooVendorPath", func(t *testing.T) {
 		sink.Reset()
 		// assert is vendored within fx and is a good test case
-		logger.PrintProvide(foovendor.New)
-		assert.Contains(
-			t, sink.String(),
-			"string <= go.uber.org/fx/internal/fxlog/foovendor.New()")
+		for _, rtype := range fxreflect.ReturnTypes(foovendor.New) {
+			Info("providing",
+				Field{Key: "returnValue", Value: rtype},
```
abhinav, line 213:
```suggestion
				Field{Key: "type", Value: rtype},
```

internal/fxlog/fxlog_test.go
```diff
@@ -109,61 +147,90 @@ func TestPrint(t *testing.T) {
...
+				Field{Key: "constructor", Value: fxreflect.FuncName(foovendor.New)},
+			).Write(sink)
+		}
+		assert.Contains(t, sink.String(), "providing")
+		assert.Contains(t, sink.Fields(), zap.Field{
+			Key:    "returnValue",
```
abhinav, line 219:
```suggestion
			Key:    "type",
```

internal/fxlog/spy.go
```diff
@@ -22,41 +22,61 @@ package fxlog
...
-	if len(s.msgs) == 0 {
+	if len(s.entries) == 0 {
 		return ""
 	}
+	var msg string
+	for _, m := range s.entries {
+		msg += m.Message
+		for _, f := range m.Fields {
+			// extra space before f.Key to separate out message.
+			msg += fmt.Sprintf(" %s: %v", f.Key, f.Value)
```
abhinav, line 59: we could tab delimit the items here

```suggestion
			msg += fmt.Sprintf("\t%s: %v", f.Key, f.Value)
```

internal/fxlog/spy.go
```diff
@@ -22,41 +22,61 @@ package fxlog
...
 	return msgs
 }
 
 // String returns all logged messages as a single newline-delimited string.
 func (s *Spy) String() string {
-	if len(s.msgs) == 0 {
+	if len(s.entries) == 0 {
 		return ""
 	}
+	var msg string
```
abhinav, line 54: not that it matters much here, but this could be optimized with the user of a strings.Builder

```
var msg strings.Builder
...
    fmt.Fprintf(msg, "...", ...)
...
return msg
```

internal/fxlog/spy.go
```diff
@@ -22,41 +22,61 @@ package fxlog
...
+	for _, m := range s.entries {
+		msg += m.Message
+		for _, f := range m.Fields {
+			// extra space before f.Key to separate out message.
+			msg += fmt.Sprintf(" %s: %v", f.Key, f.Value)
+		}
+		msg += m.Stack
```
abhinav, line 61: I would quote the stack trace so that it continues to be one statement per line

    fmt.Fprintf(msg, "%q", m.Stack)

internal/fxlog/spy.go
```diff
@@ -22,41 +22,61 @@ package fxlog
...
+
+	return msg
+}
+
+// Fields returns all Fields members of entries.
+func (s *Spy) Fields() []zap.Field {
+	var fields []zap.Field
+	for _, e := range s.entries {
+		fs := encodeFields(e.Fields, "")
+		fields = append(fields, fs...)
```
abhinav, line 73: optional

```suggestion
		fields = append(fields, encodeFields(e.Fields, "")...)
```

internal/lifecycle/lifecycle.go
```diff
@@ -38,17 +39,19 @@ type Hook struct {
...
 	numStarted int
 }
 
 // New constructs a new Lifecycle.
-func New(logger *fxlog.Logger) *Lifecycle {
+func New(logger fxlog.Logger) *Lifecycle {
 	if logger == nil {
-		logger = fxlog.New()
+		logger = fxlog.DefaultLogger(os.Stderr)
 	}
```
abhinav, line 51: i know that the old code did this, but i would drop the default logge so that we don't accidentally write to stderr.

internal/lifecycle/lifecycle.go
```diff
@@ -38,17 +39,19 @@ type Hook struct {
...
+
 	return &Lifecycle{logger: logger}
+
```
abhinav, line 54:
```suggestion
```

internal/fxlog/spy.go
```diff
@@ -23,40 +23,63 @@ package fxlog
...
+	for _, m := range s.entries {
+		_, _ = fmt.Fprintf(&msg, m.Message)
+		for _, f := range m.Fields {
+			// extra space before f.Key to separate out message.
+			_, _ = fmt.Fprintf(&msg, "\t%s: %v", f.Key, f.Value)
+		}
+		if m.Stack != "" {
+			_, _ = fmt.Fprintf(&msg, "%q", m.Stack)
+		}
+		_, _ = fmt.Fprintf(&msg, "\n")
```
abhinav, line 66: since we're ignoring the errors (writing to strings.builder cannot fail), we don't need to do the `_, _ =`.
```suggestion
		fmt.Fprintf(&msg, m.Message)
		for _, f := range m.Fields {
			// extra space before f.Key to separate out message.
			fmt.Fprintf(&msg, "\t%s: %v", f.Key, f.Value)
		}
		if m.Stack != "" {
			fmt.Fprintf(&msg, "\t%q", m.Stack)
		}
		msg.WriteString("\n")
```

internal/fxlog/spy.go
```diff
@@ -23,40 +23,63 @@ package fxlog
...
 }
 
 // String returns all logged messages as a single newline-delimited string.
 func (s *Spy) String() string {
-	if len(s.msgs) == 0 {
+	if len(s.entries) == 0 {
 		return ""
 	}
+	var msg strings.Builder
+
```
abhinav, line 56:
```suggestion
```

abhinav (approved):
minor changes but lgtm otherwise

PR #721, opened 2020-12-09
fxtest.Lifecycle: Write to testing.T

abhinav:
fxtest.Lifecycle writes to stderr right now. This makes for noisy test
output.

Seeing as a testing.TB is already available, we should log to it
instead.

Refs GO-281

### DIFF
--- a/fxtest/app_test.go
+++ b/fxtest/app_test.go
@@ -21,44 +21,15 @@
 package fxtest
 
 import (
-	"bytes"
 	"context"
 	"errors"
-	"fmt"
 	"testing"
 
-	"go.uber.org/fx"
-
 	"github.com/stretchr/testify/assert"
-	"go.uber.org/goleak"
+	"go.uber.org/fx"
 )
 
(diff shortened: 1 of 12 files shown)


fxtest/app_test.go
```diff
@@ -0,0 +1,89 @@
...
+		assert.Contains(t, spy.logs.String(), "running", "Expected to write logs to TB.")
+	})
+
+	t.Run("NewFailure", func(t *testing.T) {
+		spy := newTB()
+
+		New(
+			spy,
+
+			// Missing string dependency in container.
```
shirchen, line 48: Fix formatting?

shirchen (approved):

fxtest/app_test.go
abhinav: What would you like changed?

(This is correctly formatted as far as gofmt is concerned.  `gofmt -l fxtest/ app_test.go` is empty.)

Issue #682, opened 2019-10-18
Use /x/sys instead of syscall


abhinav:
As per the documentation of the syscall package,

> Deprecated: this package is locked down. Callers should use the corresponding package in the golang.org/x/sys repository instead.

PR #727, opened 2021-02-22
Use x/sys instead of syscall (fixes uber-go/fx#682)

blizzy78:
This removes the use of package syscall in favor of x/sys/{unix,windows}.

Please note that I've removed the explicit check for SIGTERM in shutdown_test.go, and replaced it with a non-nil check. In theory you could explicitly check for the signal as had been done before, but since Shutdown's documentation only mentions "a signal" being sent, I don't see it necessary to check for an implementation detail.

### DIFF
--- a/app.go
+++ b/app.go
@@ -30,7 +30,6 @@ import (
 	"reflect"
 	"strings"
 	"sync"
-	"syscall"
 	"time"
 
 	"go.uber.org/dig"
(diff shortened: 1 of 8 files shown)


abhinav (commented):
This looks great! Thanks for the fix.

Two requests:

- Would you mind adding license headers to the new files with the year 2021?
- Can we prefix the sigINT and sigTERM constants with `_`? (https://github.com/uber-go/guide/blob/master/style.md#prefix-unexported-globals-with-_)

Thanks!

blizzy78:
Done and done. I had added the license headers locally as suggested by `make lint`, then forgot to push 🙈

blizzy78:
I believe [this Travis CI run](https://travis-ci.com/github/uber-go/fx/jobs/485183856) should have failed earlier due to missing license headers? It looks like it did not.

abhinav:
> I believe [this Travis CI run](https://travis-ci.com/github/uber-go/fx/jobs/485183856) should have failed earlier due to missing license headers? It looks like it did not.

Weirdly, the oldstable run failed instead. That appears to be a mistake; fixing.

abhinav (approved):
Thanks!

PR #733, opened 2021-05-04
Add WithLogger option to supply a fxevent.Logger

shirchen:
This is a follow up to #717. This change adds a `WithLogger` option that
accepts a constructor that returns an `fxevent.Logger`. Unlike
`fx.Printer`, this constructor can have dependencies on other types
provided to the Fx graph. If supplied, we'll use this logger to log
events rather than our default behavior.

To make this work, we do the following:

- If `WithLogger` was not supplied, we use the old behavior of logging
  to stderr or the user-supplied `fx.Printer`.
- If `WithLogger` was supplied, we buffer log messages in-memory until
  we can build the user-supplied logger. If we successfully build the
  logger, we flush the buffered messages to it, and log to it from that
  point onwards. If we were unable to build the logger, we flush the
  buffered messages to stderr, or the user-supplied `fx.Printer` and
  fail application startup.

Further, to match our old behavior more accurately, this change includes
a `ConsoleLogger` implementation of `fxevent.Logger` that behaves almost
exactly like our logger before #717. This will make it easy to roll this
feature out without affecting users negatively, and also give us the
option of selecting between readable and non-readable log outputs based
on runtime environment (production vs development, for example).

Ref GO-77


app.go
```diff
@@ -449,22 +456,46 @@ func ValidateApp(opts ...Option) error {
 	return app.Err()
 }
 
+// logBuffer will buffer all messages until a logger has been
+// initialized.
+type logBuffer struct {
+	entries []fxlog.Entry
+	logger  fxlog.Logger
```
r-hang, line 463: As I read it, the value of logBuffer is to store log entries to write before the constructed structured logger is initialized so that those entries can be logged with the structure logger as opposed to another default logger for a more consistent experience.

Is this logger field within logBuffer required? Having a buffer with a logger allows someone to log with both the 'app.log' when its setup as well as the logBuffer once' it's connected, where ideally there should only be a single choice?

The alternative approach would be to replace Connect with a `func (l *logBuffer) Sync(logger fxlog.Logger)`
that only flushes and does not set a logger. Once the structured logger is available, the buffer shouldn't be used anymore? Is there a strong argument against going the other way and even closing off the logBuffer from further writes once its Synced?

app.go
```diff
@@ -449,22 +456,46 @@ func ValidateApp(opts ...Option) error {
...
+	} else {
+		l.logger.Log(entry)
+	}
+}
+
+// Connect flushes out all buffered entries to a logger and resets them.
+func (l *logBuffer) Connect(logger fxlog.Logger) {
+	l.logger = logger
+	for _, e := range l.entries {
+		logger.Log(e)
```
r-hang, line 479: Use l.logger.Log(e) since it's already set.

app.go
```diff
@@ -475,6 +506,39 @@ func New(opts ...Option) *App {
...
+			app.err = err
+
+			return app
+		}
+		var userLogger fxlog.Logger
+		if err := app.container.Invoke(func(got fxlog.Logger) { userLogger = got }); err != nil {
+			fxlog.Error("error constructing logger, did not get back fxlog.Logger").Write(app.logBuffer)
+			app.err = err
+
+			return app
```
r-hang, line 525: If app returns here, will anything written to app.logBuffer ever get flushed out of the buffer?

app.go
```diff
@@ -449,22 +456,46 @@ func ValidateApp(opts ...Option) error {
 	return app.Err()
 }
 
+// logBuffer will buffer all messages until a logger has been
+// initialized.
+type logBuffer struct {
```
abhinav, line 461: readability nit: this file has gotten pretty large. consider moving logBuffer to a sibling file (log.go?) so that we can also independently unit test it.

app.go
```diff
@@ -449,22 +456,46 @@ func ValidateApp(opts ...Option) error {
...
 // New creates and initializes an App, immediately executing any functions
 // registered via Invoke options. See the documentation of the App struct for
 // details on the application's initialization, startup, and shutdown logic.
 func New(opts ...Option) *App {
-	logger := fxlog.DefaultLogger(os.Stderr)
+	logBufferer := &logBuffer{}
 
 	app := &App{
-		log:          logger,
+		logBuffer:    logBufferer,
```
abhinav, line 491: nit: just logBuffer as the name should suffice. Although in this case, we don't actually need the variable.

```suggestion
	app := &App{
		logBuffer:    new(logBuffer),
```

app.go
```diff
@@ -475,6 +506,39 @@ func New(opts ...Option) *App {
 		app.provide(p)
 	}
 
+	// If WithLogger and Printer are both provided, WithLogger takes precedence.
+	// If user supplied an fxlog constructor, we need to have dig's container
+	// initialize it and grab all of arguments from the graph and then return.
+	if app.fxlogConstructor != nil {
```
abhinav, line 512: I'm not sure but I think this check should be after `opt.apply` but before `app.provide`? By this point in time, we've already provided a bunch of things to the container and logged about them. We also probably need an `else` clause to use the default logger when this is unspecified?

So what we need here is probably something like:

```
for _, opt := range opts {
  opt.apply(app)
}

var connectLogger interface{}
if app.fxLogConstructor != nil {
    // There's a user provided constructor. Log to buffer until we connect.
    app.log = app.logBuffer
    connectLogger = func(log fxlog.Logger) {
        app.logBuffer.Connect(log)
    }
} else {
    app.log = // put the default logger here
}

// ... app.provide(p)

if connectLogger != nil {
    app.container.Invoke(connectLogger)
}
```

The net result of this will be:

- Existing logging code remains unchanged and always logs to app.log
- app.log is a buffer if the user provided a logger. Otherwise it's not.
- If the user provided a logger, then the first thing we do after all provides is connect the logger

---

On top of this, we can actually get away with not having logBuffer as a field on app:

```
if app.fxLogConstructor != nil {
  var buffer logBuffer
  app.log = &buffer
  connectLogger = buffer.Connect // func(fxlog.Logger)
}
```

I slightly prefer this so that we don't accidentally use the logBuffer directly ever.

---

Finally, and this is optional, but the the Invoke can instead be this:

```
func(log fxlog.Logger) {
  buffer.FlushTo(log)
  app.log = log  // replace buffer with real logger
}
```

This will remove the buffer completely from the equation once things are ready and remove the if-else from the buffer. The buffer would just record things in memory until the replacement is available at which point we flush to replacement and replace.

app.go
```diff
@@ -475,6 +506,39 @@ func New(opts ...Option) *App {
...
+		app.log = userLogger
+		fxlog.Info("setting custom fxlog.Logger",
+			fxlog.F("constructor", fxreflect.FuncName(app.fxlogConstructor))).Write(app.logBuffer)
+	}
+	// Initialize a logger if user hasn't provided one via WithLogger.
+	if app.log == nil {
+		app.log = fxlog.DefaultLogger(os.Stderr)
+	}
+	// Flush out all previously buffered logs and connect newly initialized logger.
+	app.logBuffer.Connect(app.log)
```
abhinav, line 537: We probably want the logBuffer.Connect as suggested above so that it happens after all the other Provides.

That way, the log constructor can actually have a dependency on the fx.Lifecycle. (Which makes sense in case the logger has some kind of Shutdown functionality.)

app.go
```diff
@@ -638,18 +699,16 @@ func (app *App) provide(p provide) {
 	}
 
 	constructor := p.Target
-
 	switch {
 	case p.IsSupply:
 		for _, rtype := range fxreflect.ReturnTypes(constructor) {
-			fxlog.Info("supplying", fxlog.Field{Key: "type", Value: rtype}).Write(app.log)
+			entry := fxlog.Info("supplying", fxlog.Field{Key: "type", Value: rtype})
+			entry.Write(app.logBuffer)
```
abhinav, line 706: This should be reverted so that we're always writing to app.log, and app.log just points to something else.

app.go
abhinav: This is valid. If the Invoke of the fxlog.Logger fails, that means we failed to construct the user provided logger.
In that case, we should as you're doing here, but then flush to the default logger and exit with an error.

app.go
shirchen: Is this for readability? I'm curious to learn about which way is better and why.

app.go
shirchen: Yup, good call. Tests didn't catch this since I constructed a spy logger.

app.go
```diff
@@ -260,20 +260,22 @@ func (t stopTimeoutOption) String() string {
 	return fmt.Sprintf("fx.StopTimeout(%v)", time.Duration(t))
 }
 
-// withLogger will stay private until we export fxlog.Logger once
-// we finalize the API.
-func withLogger(l fxlog.Logger) Option {
-	return withLoggerOption{l}
+// WithLogger redirects the logging output to the instance of fx.Logger
+// provided by the constructor.
+func WithLogger(constructor interface{}) Option {
```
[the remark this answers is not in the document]
shirchen: I see, is there harm in leaving this public so we can keep on using this in tests and elsewhere and isolate exporting fxlog.Logger?

app.go
shirchen: _I'm not sure but I think this check should be after opt.apply but before app.provide_

So WithLogger is going to be part of Provide, so I think we need this after app.provide as `fxlogConstructor` is populated by WithLogger opt.apply but it'll need `zap.Logger` initialized by something like `zapfx`.

Good call on trying to simplify logBuffer, let me see how I can make it work.

app.go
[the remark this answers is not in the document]
abhinav: The reason I'd want to unexport is because it's currently unusable. As a user of Fx, there's no valid way for me to call this function right now.

That said, if you think we won't release this change until we also finish exporting fxlog.Logger, I'm okay with keeping it exported in this PR.

app.go
abhinav: If we do this after the `app.provides`, then we've already called `container.Provide` and logged a bunch before we've even considered whether there was a user-provided logger, right?

It's an option, but it's a "more important" option than provides. Like, we run Provides first, and Invokes later. We should examine WithLogger *before* we start logging things the same way we consider the current fx.Printer before anything else.

To be clear: we won't construct the logger until after app.provides, but we need to know whether we're buffering or using the default logger before we do app.provides. Please read the code samples above again.

app.go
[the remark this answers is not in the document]
shirchen: Yea, I think there isn't a point in releasing until this API is usable, so exporting fxlog.Logger should be a blocker for next release.

app.go
```diff
@@ -453,28 +464,92 @@ func ValidateApp(opts ...Option) error {
 // registered via Invoke options. See the documentation of the App struct for
 // details on the application's initialization, startup, and shutdown logic.
 func New(opts ...Option) *App {
-	logger := fxlog.DefaultLogger(os.Stderr)
-
 	app := &App{
-		log:          logger,
+		log:          new(logBuffer),
```
abhinav, line 468: We don't need to initialize this field here because it's always overwritten (L481-493).
```suggestion
```

app.go
```diff
@@ -453,28 +464,92 @@ func ValidateApp(opts ...Option) error {
...
+		// then use default one.
+		app.log = fxlog.DefaultLogger(os.Stderr)
+	}
 
 	for _, p := range app.provides {
 		app.provide(p)
 	}
 
+	if app.err != nil {
+		return app
```
abhinav, line 500: If a custom logger was provided, and this fails (because there was an error with one of the provides, like `fx.Provide(NewFoo, NewFoo)`), we won't print anything and fail silently.

We need to flush the buffer if there was a buffer, so maybe something like this:

```suggestion
	if app.err != nil {
		connectLogger(fxlog.DefaultLogger(os.Stderr))
		return app
```

app.go
```diff
@@ -453,28 +464,92 @@ func ValidateApp(opts ...Option) error {
...
+	// initialize it and grab all of arguments from the graph and then return.
+	if app.fxlogConstructor != nil {
+		// Here we add the custom logger into the container.
+		if err := app.container.Provide(app.fxlogConstructor); err != nil {
+			app.err = multierr.Combine(app.err,
+				fmt.Errorf("could not construct customer logger via fx.WithLogger: %w", err))
+			if connectLogger != nil {
+				err = connectLogger(fxlog.DefaultLogger(os.Stderr))
+				app.err = multierr.Combine(app.err, err)
+			} else {
```
abhinav, line 514: This will never be true unless there's a massive bug up above, in which case we should just panic. I would drop this else block and remove the `if` on connectLogger. Part of the contract of this code here is that if a custom constructor was provided, connectLogger was set.

app.go
```diff
@@ -453,28 +464,92 @@ func ValidateApp(opts ...Option) error {
...
+	}
+
+	// If WithLogger and Printer are both provided, WithLogger takes precedence.
+	// If user supplied an fxlog constructor, we need to have dig's container
+	// initialize it and grab all of arguments from the graph and then return.
+	if app.fxlogConstructor != nil {
+		// Here we add the custom logger into the container.
+		if err := app.container.Provide(app.fxlogConstructor); err != nil {
+			app.err = multierr.Combine(app.err,
+				fmt.Errorf("could not construct customer logger via fx.WithLogger: %w", err))
```
abhinav, line 510: nit: When there are just two errors, multierr.Append is more efficient.
```suggestion
			app.err = multierr.Append(app.err,
				fmt.Errorf("could not construct customer logger via fx.WithLogger: %w", err))
```

app.go
```diff
@@ -453,28 +464,92 @@ func ValidateApp(opts ...Option) error {
...
+			if connectLogger != nil {
+				err = connectLogger(fxlog.DefaultLogger(os.Stderr))
+				app.err = multierr.Combine(app.err, err)
```
abhinav, line 513:
```suggestion
				app.err = multierr.Append(app.err, err)
```

app.go
```diff
@@ -453,28 +464,92 @@ func ValidateApp(opts ...Option) error {
...
 
+	if app.err != nil {
+		return app
+	}
+
+	// If WithLogger and Printer are both provided, WithLogger takes precedence.
+	// If user supplied an fxlog constructor, we need to have dig's container
+	// initialize it and grab all of arguments from the graph and then return.
+	if app.fxlogConstructor != nil {
+		// Here we add the custom logger into the container.
```
abhinav, line 507: This if block is massive. Can we pull it into a method responsible for building the logger and nothing else?

```
if app.fxlogConstructor != nil {
  log, err := app.buildCustomLogger()  // better name TBD
  if err != nil {
    // failed to build custom logger. Flush to stderr.
    connectLogger(fxlog.DefaultLogger(os.Stderr))
    app.err = err
    return app
  }
  connectLogger(log)
}
```

app.go
```diff
@@ -453,28 +464,92 @@ func ValidateApp(opts ...Option) error {
...
+		}
+		err := app.container.Invoke(
+			func(log fxlog.Logger) error {
+				app.log = log
+				app.err = multierr.Combine(app.err, connectLogger(app.log))
+				if app.err != nil {
+					return app.err
+				}
+
+				return nil
```
abhinav, line 531: We shouldn't manipulate app.err inside the Invoke. If we got to the Invoke, we know app.err is blank, so we can just return the error.

```
err := Invoke(func(log fxlog.Logger) error {
  app.log = log
  return connectLogger(app.log)
})

if err != nil {
  app.err = err
  // ...
}
```

If there are any errors in invoking (because the logger could not be constructed), or because `connectLogger` failed, that'll be surfaced in `err`.

log.go
```diff
@@ -20,9 +20,38 @@
...
+	} else {
+		l.logger.Log(entry)
+	}
+}
+
+// Connect flushes out all buffered entries to a logger and resets them.
+func (l *logBuffer) Connect(logger fxlog.Logger) error {
+	if logger == nil {
+		return errors.New("missing a logger to connect to")
+	}
```
abhinav, line 49: We should remove this check. This check is the sole reason that Connect has to return an error, and that's making all the logic above significantly more complicated. If we make `Connect` non-error-able, the log-construction logic gets simplified signfiicantly.

We don't really need to handle the "no logger" case here. This isn't an API external users can call. We have control over it. We can assume that we'l always call it correctly, or make necessary checks in the callers.

log.go
```diff
@@ -20,9 +20,38 @@
...
+	l.logger = logger
+	for _, e := range l.entries {
+		logger.Log(e)
+	}
+	l.entries = []fxlog.Entry{}
```
abhinav, line 54: nil is a valid slice and costs no allocs

```suggestion
	l.entries = nil
```

app.go
shirchen: I have/had this check due to a nil deference linter, which points out that its possible to have connectLogger be nil here. I thought that we wanted to stay on the good side of it, but happy to remove this if you dont think this adds value.

app.go
shirchen: ugh, we already have a test for this case (line 513 in `app_test.go` but since it's using a spy, we didn't have a test about flushing out the actual message. i think testing this will be easier once #747 lands since i added a way to setup zap and see what exactly gets printed out. i'll address this comment after rebase.

app.go
abhinav: If we're trying to appease the linter here, I'm okay with a manual `panic` here because that's an irrecoverable "there;s a massive bug in Fx" sort of scenario.

app.go
```diff
@@ -308,7 +312,9 @@ func (l loggerOption) String() string {
 // failures difficult to debug, since no errors are printed to console.
 //
 // Note, when withLogger is public we will make the change here as well.
```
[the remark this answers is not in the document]
abhinav: +1

app.go
```diff
@@ -445,31 +450,101 @@ func ValidateApp(opts ...Option) error {
 	return app.Err()
 }
 
+func (app *App) constructCustomLogger(connectLogger func(logger fxevent.Logger)) error {
+	if err := app.container.Provide(app.fxlogConstructor); err != nil {
+		app.err = multierr.Append(app.err,
+			fmt.Errorf("could not construct custom logger via fx.WithLogger: %w", err))
+		connectLogger(fxlog.DefaultLogger(os.Stderr))
+
```
abhinav, line 458: nit:

 ```suggestion
```

File: app.go [hunk 21]
```diff
@@ -445,31 +450,101 @@ func ValidateApp(opts ...Option) error {
...
+		return app.err
+	}
+	err := app.container.Invoke(
+		func(log fxevent.Logger) {
+			app.log = log
+
```
abhinav, line 464: nit:
```suggestion
```

app.go
```diff
@@ -445,31 +450,101 @@ func ValidateApp(opts ...Option) error {
...
 
+func (app *App) constructCustomLogger(connectLogger func(logger fxevent.Logger)) error {
+	if err := app.container.Provide(app.fxlogConstructor); err != nil {
+		app.err = multierr.Append(app.err,
+			fmt.Errorf("could not construct custom logger via fx.WithLogger: %w", err))
+		connectLogger(fxlog.DefaultLogger(os.Stderr))
+
+		return app.err
+	}
+	err := app.container.Invoke(
```
abhinav, line 461: nit:
```suggestion

	err := app.container.Invoke(
```

app.go
```diff
@@ -445,31 +450,101 @@ func ValidateApp(opts ...Option) error {
...
+			app.log = log
+
+			connectLogger(app.log)
+		})
+
+	if err != nil {
+		return err
+	}
+
+	return nil
```
abhinav, line 472: nit: can shorten
```suggestion
	return err
```

app.go
```diff
@@ -445,31 +450,101 @@ func ValidateApp(opts ...Option) error {
...
 		dig.DryRun(app.validate),
 	)
+	var connectLogger func(logger fxevent.Logger)
+	if app.fxlogConstructor != nil {
+		// Since user supplied a custom logger, use a buffered logger to hold
+		// all messages until user supplied logger is instantiated. Then we flush
+		// those messages after fully constructing the custom logger.
+		var buffer logBuffer
+		app.log = &buffer
+		connectLogger = buffer.Connect // func(fxlog.Logger) error
```
abhinav, line 499: outdated comment
```suggestion
		connectLogger = buffer.Connect // func(fxlog.Logger)
```

app.go
```diff
@@ -445,31 +450,101 @@ func ValidateApp(opts ...Option) error {
...
+		app.log.LogEvent(&fxevent.ProvideError{Err: app.err})
+		// Here we do not return to give custom logger a chance to initialize itself.
+		// We will flush later.
+	}
+
+	// If WithLogger and Printer are both provided, WithLogger takes precedence.
+	// If user supplied an fxlog constructor, we need to have dig's container
+	// initialize it and grab all of arguments from the graph and then return.
+	if app.fxlogConstructor != nil {
+		// Here we add the custom logger into the container and flush out the buffer on success.
```
abhinav, line 519: nit: the other comments are helpful but this is a bit extraneous
```suggestion
```

app.go
```diff
@@ -445,31 +450,101 @@ func ValidateApp(opts ...Option) error {
...
+		err := app.constructCustomLogger(connectLogger)
+		if err != nil {
```
abhinav, line 521: nit: can inline
```suggestion
		if err := app.constructCustomLogger(connectLogger); err != nil {
```

app_test.go
```diff
@@ -258,6 +262,75 @@ func TestNewApp(t *testing.T) {
...
+func setupFxLogger(l *zap.Logger) fxevent.Logger {
+	return &eventLogger{l: l}
+}
+
+func TestSetupLogger(t *testing.T) {
+	t.Parallel()
+	t.Run("initializing custom logger", func(t *testing.T) {
+		t.Parallel()
+		app := fxtest.New(t,
+			Provide(zap.NewExample),
```
abhinav, line 282: This will write to stdout. Maybe we should use a zaptest.NewLogger here so it only writes to the testing.T.

```suggestion
			Supply(zaptest.NewLogger(t)),
```

app_test.go
```diff
@@ -258,6 +262,75 @@ func TestNewApp(t *testing.T) {
...
+		)
+		defer app.RequireStart().RequireStop()
+		require.NoError(t, app.Err())
+		require.NoError(t, app.Start(context.Background()))
+	})
+
+	t.Run("initializing zap production logger", func(t *testing.T) {
+		t.Parallel()
+		app := fxtest.New(t,
+			Provide(zap.NewProduction),
```
abhinav, line 293: This and the development logger are the same tests as the one above, with the difference being whether they'll write to stdout or stderr. In the interest of keeping stdout/stderr as silent as possible, I would suggest omitting these two since the one above will already cover the "instantiate a logger with no dependencies" case.

app_test.go
```diff
@@ -258,6 +262,75 @@ func TestNewApp(t *testing.T) {
...
+		app := New(
+			WithLogger(&bytes.Buffer{}),
+		)
+		err := app.Err()
+		require.Error(t, err)
+		assert.Contains(t,
+			err.Error(),
+			"could not construct custom logger via fx.WithLogger: must provide constructor function, got  (type *bytes.Buffer)",
+		)
+	})
```
abhinav, line 331: Can we add a couple more test cases? In particular, we should test a logger with a dependency on other provided objects. This matches what we'll be using these APIs for. We should have these cases:

- Success
- Dependency failed to build because of an error
- Logger failed to build because of an error

We can test the output of the success case pretty easily: have the logger write to a buffer.
However, to test the output of the failure cases that write to stderr, we'll have to temporarily hijack os.Stderr by replacing it temporarily with an `os.File` for a temporary file (`ioutil.TempFile(t.TempDir(), "stderr")`?) and asserting its output.

fxevent/event.go
```diff
@@ -41,6 +41,8 @@ func (*StopError) event()          {}
 func (*Rollback) event()           {}
 func (*RollbackError) event()      {}
 func (*Running) event()            {}
+func (*LoggerError) event()        {}
+func (*CustomLogger) event()       {}
```
abhinav, line 45: Forgot to add these to the `event()` test?

fxevent/event.go
```diff
@@ -99,3 +101,11 @@ type RollbackError struct{ Err error }
 
 // Running is emitted whenever an application is started successfully.
 type Running struct{}
+
+// LoggerError is emitted whenever a custom logger fails to construct.
+type LoggerError struct{ Err error }
```
abhinav, line 106: CustomLoggerError to match CustomLogger?

fxevent/zap.go
```diff
@@ -71,5 +71,9 @@ func (l *ZapLogger) LogEvent(event Event) {
 		l.Logger.Error("startup failed, rolling back", zap.Error(e.StartErr))
 	case *Running:
 		l.Logger.Info("running")
+	case *LoggerError:
+		l.Logger.Error("error constructing logger, did not get back fxevent.Logger from WithLogger", zap.Error(e.Err))
+	case *CustomLogger:
+		l.Logger.Info("setting custom fxevent.Logger", zap.String("function", fxreflect.FuncName(e.Function)))
```
abhinav, line 77: nit:
```suggestion
		l.Logger.Info("installing custom fxevent.Logger", zap.String("function", fxreflect.FuncName(e.Function)))
```

fxevent/zap.go
```diff
@@ -71,5 +71,9 @@ func (l *ZapLogger) LogEvent(event Event) {
 		l.Logger.Error("startup failed, rolling back", zap.Error(e.StartErr))
 	case *Running:
 		l.Logger.Info("running")
+	case *LoggerError:
+		l.Logger.Error("error constructing logger, did not get back fxevent.Logger from WithLogger", zap.Error(e.Err))
```
abhinav, line 75: nit: why we were unable to build the logger should be in the error message; the extra string can be skipped
```suggestion
		l.Logger.Error("error constructing logger", zap.Error(e.Err))
```

abhinav (commented):
Minus need for test cases for logger dependencies LGTM!

app.go
shirchen: sg, though adding test coverage will be impossible so i changed my mind: its easier to ignore the lint warning.

app.go
shirchen: Here, I try and follow https://github.com/ssgreg/nlreturn. Do we disagree that this is a valid lint warning?

abhinav: That linter seems pretty arbitrary, TBH. I feel like whether a newline should be added in the function body or not is a matter of "paragraphs" of code with a shared purpose, and not something you can always lint for.

This return here is connected to the two statements above it, in that all three together are how we're choosing to handle that error. Like, how we handle that error is `(record, connect, return)`, not `(record, connect)` and an independent `return`. So to me the `return` is part of the paragraph containing the other two statements.

shirchen: ok, sounds good. i'll disable it from list that i use.

abhinav (approved):
LGTM; please review the additions I've made as well.

CC @r-hang

sywhang (approved):
LGTM!

PR #735, opened 2021-05-10
Switch from Travis to GH Actions

shirchen:
Travis is getting deprecated for our OSS, so switching to GitHub Actions.


.github/workflows/go.yml
```diff
@@ -0,0 +1,52 @@
...
+      run: make lint
+
+    - name: Test
+      run: make cover
+
+    - name: Upload coverage to codecov.io
+      uses: codecov/codecov-action@v1
+
+    - name: Benchmark
+      run: make bench
```
abhinav, line 52: There's no `bench` target on Fx. We can delete this.
```suggestion
```

abhinav (approved):

PR #740, opened 2021-06-04
In/Out: Use type aliases

### LINKED ISSUE #730: In/Out: Use type aliases
Fx declares In and Out as

```
type In struct{ dig.In }
type Out struct{ dig.Out }
```

This choice is a holdover from when Go didn't support type aliases.
That's no longer true so we can switch the implementation:

```
type In = dig.In

type Out = dig.Out
```

Refs https://github.com/uber-go/dig/pull/274

abhinav:
When `fx.In` and `fx.Out` were originally implemented on top of `dig.In`
and `dig.Out`, type aliases didn't exist so they were implemented by
embedding the structs.

Now that we're well past Go 1.9 (when type aliases were added), we
should be more than safe to switch over to type aliases.

Note that even though we're changing the definition of the type, this is
not a breaking change because the `dig.In` and `dig.Out` types were not
usable directly before either as they embedded an internal sentinel
type.

Resolves #730
Refs GO-374

### DIFF
--- a/inout.go
+++ b/inout.go
@@ -162,7 +162,7 @@ import "go.uber.org/dig"
 //
 // Note that values in a value group are unordered. Fx makes no guarantees
 // about the order in which these values will be produced.
-type In struct{ dig.In }
+type In = dig.In
 
 // Out is the inverse of In: it can be embedded in result structs to take
 // advantage of advanced features.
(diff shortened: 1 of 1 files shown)


manjari25 (approved):

abhinav:
Verified against a bunch of internal systems that this will cause no issues.

However, this does require a manual change to some auto-generated code
turning it from `fx.In{In: dig.In{}}` to `fx.In`. I was able to do that
with the following [gopatch](https://github.com/uber-go/gopatch) patch.

```diff
@@
@@
 import "go.uber.org/fx"
-import "go.uber.org/dig"

 fx.In{
-  In: dig.In{},
 }
```

Going to move forward with this change.

PR #746, opened 2021-06-14
Report running hook and executed hooks' runtimes as part of startup timeout error

sywhang:
When Fx.Start() timeouts during startup, it's difficult for the user to debug why it failed, because the only error message being printed is "context deadline exceeded". 

This enhances the error returned from timeout by adding a report of the currently executing hook's caller (the function that called `Append`) as well as a report of all the hooks that were successfully executed - specifically their runtimes, functions that ran as hooks, and their callers.

Refs GO-628.


sywhang:
Marking this as WIP till I add some more tests : )

fxtest/lifecycle.go
```diff
@@ -116,7 +133,23 @@ func (l *Lifecycle) RequireStart() *Lifecycle {
...
+		select {
+		case err := <-c:
+			return err
+		case <-callerChan:
+			continue
+		case <-recordChan:
+			continue
+		}
+	}
+}
```
shirchen, line 152: eh, this is identical (other than `lc` component) to above and an extra `// Ignore caller/hookrecord channel`. i dunno if its worth refactoring a bit for easier testing?

app.go
```diff
@@ -756,36 +766,68 @@ func (app *App) run(done <-chan os.Signal) {
...
-		return ctx.Err()
-	case err := <-c:
-		return err
+func withTimeout(param *withTimeoutParams) error {
+	c := make(chan error, 1)
+	// Channel to keep track of the currently executing hooks.
+	callerChan := make(chan string, 1)
+	// Channel to keep track of the each hook's execution time.
+	recordChan := make(chan lifecycle.HookRecord, 1)
+	ctx := param.Ctx
```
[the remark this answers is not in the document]
sywhang: that wasn't on purpose - thanks for catching that.

app.go
```diff
@@ -563,6 +563,11 @@ func (app *App) Err() error {
 	return app.err
 }
 
+var (
+	onStartHook = "OnStart"
+	onStopHook  = "OnStop"
+)
```
abhinav, line 569: nit: prefix with _

https://github.com/uber-go/guide/blob/master/style.md#prefix-unexported-globals-with-_

abhinav (changes requested):
I have some reservations about the proposed API changes (even though they're largely internal).

It's a bit difficult to follow the flow of information with the way channels are being used right now. Consider using narrower read-only (`<-chan string`) or write-only (`chan<- string`) types to clarify intent (and add safety) when using channels. But more generally, avoid using channels as part of APIs if possible. It's very easy to mess them up with regards to buffering, closing, etc.

Instead of having the lifecycle publish reports as part of Start/Stop, we can make the fact that this information is recorded an implementation detail of the internal lifecycle.Lifecycle implementation. It can record everything internally, and then when we call lifecycle.Start, and it fails or times out, we can query our implementation of lifecycle for the execution records.

```
switch {
case err := <-ctx.Done():
  if err == context.DeadlineExceeded { 
    hookRecords := lc.Records()
    // ...
  }
case err := <-c:
  return err
}
```

Separately, we could even log the runtimes for the various hooks with the logger we already have:

```
fxlog.Debug("starting", fxlog.Field{...}).Write(l.logger)
if err := hook.OnStart(ctx); err != nil {
  return err
} 
fxlog.Info("started", fxlog.Field{...}, fxlog.Field{Key: "duration", ...}).Write(l.logger)
```

sywhang:
@abhinav @shirchen FYI I made the changes to the implementation to not use channels to report the times and get it directly from Lifecycle instead -- Since #747 is still underway, I'll hold off on making additional changes until that's done and rebase this with the changes from that.

abhinav (commented):
Oh, now that #747 is in, this can be rebased.

app.go
```diff
@@ -757,20 +774,46 @@ func (app *App) start(ctx context.Context) error {
...
+		r := param.lifecycle.hookRecords()
+		caller := param.lifecycle.runningHookCaller()
+		if len(r) > 0 {
+			err = multierr.Append(err, fmt.Errorf(`
+timed out while executing hook %s (Caller: %s). Hooks successfully ran so far: %s`,
+				param.hook, caller, r))
+		} else {
+			err = multierr.Append(err, fmt.Errorf(`
+timed out while executing hook %s (Caller: %s)`, param.hook, caller))
+		}
```
abhinav, line 815: optionally, instead of appending, we can wrap:

```
type hookRunError struct {
  Records []HookRecord
  Err error
}

// (this is needed for errors.Is and errors.As APIs to function.)
// https://golang.org/pkg/errors/
func (e hookRunError) Unwrap() error  { return e.Err }

func (hookRunError) Error() string {
  // something like the string you're generating right now, maybe multi-line
}
```

This can handle both `len == 0` and `len > 0` cases in the Error string which can be independently tested.

---

Since ctx.Err will normally be one of cancelled or timeout, we can probably wrap the ctx.Err unconditionally.

app.go
```diff
@@ -757,20 +774,46 @@ func (app *App) start(ctx context.Context) error {
...
+		// runtimes of hooks successfully run till end.
+		r := param.lifecycle.hookRecords()
+		caller := param.lifecycle.runningHookCaller()
+		if len(r) > 0 {
+			err = multierr.Append(err, fmt.Errorf(`
+timed out while executing hook %s (Caller: %s). Hooks successfully ran so far: %s`,
+				param.hook, caller, r))
+		} else {
+			err = multierr.Append(err, fmt.Errorf(`
+timed out while executing hook %s (Caller: %s)`, param.hook, caller))
```
abhinav, line 814: nit about the message: how about something like,

    "%v hook added by %v failed: %w\n%s", param.hook, caller, err, r

So we get something like:

    OnStart hook added by whatever.func() failed: deadline exceeded
    [hook run report here]

(`%w` above is equivalent to creating a wrapper; if you create a wrapper error type instead, that `%w` can be a `%v` in the `fmt.Sprintf` call.)

Maybe we can also include the name of the function that failed to run? I don't know the exact framing of the message for that, though.

internal/lifecycle/lifecycle.go
```diff
@@ -74,18 +96,71 @@ func (l *Lifecycle) Start(ctx context.Context) error {
...
+
+// HookRecords is a Stringer wrapper of HookRecord slice.
+type HookRecords []HookRecord
+
+// Used for logging startup errors.
+func (r HookRecords) String() string {
+	var b strings.Builder
+	sort.Slice(r, func(i, j int) bool { return r[i].Runtime > r[j].Runtime })
+	for _, r := range r {
+		b.WriteString(fmt.Sprintf("%s took %d ms to run. (Caller: %s)\n", fxreflect.FuncName(r.Func), r.Runtime.Milliseconds(), r.Caller))
```
abhinav, line 163: I'm considering the error presentation here. We have two pieces of information to report for each entry: the name of the function in the hook, and name of the function that added it.

Check out how we present information in fxreflect.Stack: https://github.com/uber-go/fx/blob/master/internal/fxreflect/stack.go#L74. We use that in error messages when a Provide or Invoke fails: the full stack trace to the function that called fx.Provide or fx.Invoke: https://github.com/uber-go/fx/blob/95337c559ed988a843a305ef45dae1b0d77ddc2f/app_test.go#L231.

I'm still only toying with presentation right now, but what do you think about multi-line formatting like that? Like the following for each entry?

```
foo/bar.Baz() took 3ms from:
    baz/qux.Whatever() (path/to/baz/qux/foo/whatever.go:42)
```

---

Separately: We don't need to truncate Runtime to milliseconds. time.Duration's default `String()` should be good enough, I think. (Use `%v` to get the default formatting for a type.)

abhinav (commented):
I'll give the implementation another pass later.
Right now, I've mostly been looking at the presentation of information.
Left some comments on that end.

internal/lifecycle/lifecycle.go
sywhang: Thanks. I wasn't aware of the +v formatter. I tried using it and it looks much nicer so I just pushed a commit to make it look like this:

```
OnStart hook added by go.uber.org/fx_test.TestAppStart.func2.3 failed: context deadline exceeded
go.uber.org/fx_test.TestAppStart.func2.2.1() took 10.819917ms from:
    go.uber.org/fx_test.TestAppStart.func2.2 (/Users/sungyoonwhang/go/src/github.com/uber-go/fx/app_test.go:516)
go.uber.org/fx_test.TestAppStart.func2.1.1() took 291ns from:
    go.uber.org/fx_test.TestAppStart.func2.1 (/Users/sungyoonwhang/go/src/github.com/uber-go/fx/app_test.go:506)
```

internal/lifecycle/lifecycle.go
```diff
@@ -74,18 +103,92 @@ func (l *Lifecycle) Start(ctx context.Context) error {
 // hooks run in reverse order.
 func (l *Lifecycle) Stop(ctx context.Context) error {
 	var errs []error
+	l.records = make(HookRecords, 0, l.numStarted)
```
abhinav, line 106: I slightly dislike that the start and stop records are on the same field, so the Stop records will overwrite the Start records, but I don't foresee any issues with this in the near-term because if Start was successful, we won't need the records anymore -- at least not in the current iteration of this change.

We may want to do this slightly differently once we're emitting this information to the event logger.

internal/lifecycle/lifecycle.go
```diff
@@ -74,18 +103,92 @@ func (l *Lifecycle) Start(ctx context.Context) error {
...
+	Runtime     time.Duration               // how long the hook ran
+}
+
+// HookRecords is a Stringer wrapper of HookRecord slice.
+type HookRecords []HookRecord
+
+// Used for logging startup errors.
+func (rs HookRecords) String() string {
+	var b strings.Builder
+	sort.Slice(rs, func(i, j int) bool { return rs[i].Runtime > rs[j].Runtime })
```
abhinav, line 168: `sort.Slice` is useful to avoid declaring a custom data type for the slice. Seeing as we already have a HookRecords data type, we might want to just implement `sort.Interface` on it and use `sort.Sort(rs)` here instead.

abhinav: Separately: I wouldn't expect `String()` to mutate the slice in any way. That should be a read-only operation.
Should we have  the caller sort the slice in advance based on their preferred criteria instead?

internal/lifecycle/lifecycle.go
```diff
@@ -74,18 +103,92 @@ func (l *Lifecycle) Start(ctx context.Context) error {
...
+	for _, r := range rs {
+		b.WriteString(fmt.Sprintf("%s took %v from %s",
+			fxreflect.FuncName(r.Func),
+			r.Runtime,
+			r.CallerFrame))
```
abhinav, line 173: strings.Builder implements io.Writer. We can shorten this to:

```suggestion
		fmt.Fprintf(&b, "%s took %v from %s",
			fxreflect.FuncName(r.Func),
			r.Runtime,
			r.CallerFrame)
```

app_test.go
```diff
@@ -496,7 +497,95 @@ func TestAppStart(t *testing.T) {
...
+		}
+		app := fxtest.New(
+			t,
+			Provide(newA),
+			Invoke(func(*A) {}),
+		)
+
+		ctx, cancel := context.WithCancel(context.Background())
+		go func() {
+			time.Sleep(100 * time.Millisecond)
```
abhinav, line 582: This is probably fine, but if you want to avoid using the sleep to wait to cancel the request, you can make OnStart `close` a random channel to signal that it's running, and in this goroutine, wait for that channel instead of sleep:

```
running := make(chan struct{})
(... OnStart: func(context.Context) error {
  close(running)
  select {}
})

ctx, cancel := ...
go func() {
  <-running
  cancel()
}()
```

internal/lifecycle/lifecycle.go
```diff
@@ -74,18 +101,105 @@ func (l *Lifecycle) Start(ctx context.Context) error {
...
+
+// Format implements fmt.Formatter to handle "%+v".
+func (rs HookRecords) Format(w fmt.State, c rune) {
+	if !w.Flag('+') {
+		// Without %+v, fall back to String().
+		io.WriteString(w, rs.String())
+		return
+	}
+
+	sort.Slice(rs, func(i, j int) bool { return rs[i].Runtime > rs[j].Runtime })
```
abhinav, line 197: Oh, same as String: Format should be read-only.

internal/lifecycle/lifecycle.go
```diff
@@ -74,18 +101,105 @@ func (l *Lifecycle) Start(ctx context.Context) error {
...
+	for _, r := range rs {
+		fmt.Fprintf(w, "\n%s took %v from:\n\t%+v",
+			fxreflect.FuncName(r.Func),
+			r.Runtime,
+			r.CallerFrame)
+	}
+	fmt.Printf("\n")
```
abhinav, line 204: Did you mean Fprintf.
```suggestion
	fmt.Fprintf(w, "\n")
```

internal/lifecycle/lifecycle_test.go
```diff
@@ -220,6 +225,40 @@ func TestLifecycleStop(t *testing.T) {
 	})
 }
 
+func TestHookRecordsFormat(t *testing.T) {
+	t.Run("SortRecords", func(t *testing.T) {
+		r := make(HookRecords, 2)
+		t1, _ := time.ParseDuration("10ms")
+		t2, _ := time.ParseDuration("20ms")
```
abhinav, line 232: we should handle these errors.
```suggestion
		t1, err := time.ParseDuration("10ms")
		require.NoError(t, err)
		t2, err := time.ParseDuration("20ms")
		require.NoError(t, err)
```

internal/lifecycle/lifecycle_test.go
```diff
@@ -220,6 +225,40 @@ func TestLifecycleStop(t *testing.T) {
...
+			Line:     1,
+		}
+		r[0] = HookRecord{
+			CallerFrame: f,
+			Func: func(context.Context) error {
+				return nil
+			},
+			Runtime: t1,
+		}
+		r[1] = HookRecord{
```
abhinav, line 245: when building a slice of things, prefer to use the slice literal form:

```
r := []HookRecord{
  {
    CallerFrame: f,
    Func: ...
    Runtime: t1,
  },
  {
    // ...
  },
}
```

internal/lifecycle/lifecycle_test.go
```diff
@@ -220,6 +225,40 @@ func TestLifecycleStop(t *testing.T) {
...
+			CallerFrame: f,
+			Func: func(context.Context) error {
+				return nil
+			},
+			Runtime: t2,
+		}
+		sort.Sort(r)
+		for _, s := range []string{fmt.Sprintf("%v", r), fmt.Sprintf("%+v", r), fmt.Sprintf("%s", r)} {
```
abhinav, line 253: optional: seeing as the thing that's different in each case is the format, this could be `[]string{"%v", "%+v", and "%s"}` and the loop body can `fmt.Sprintf` on it.

abhinav (approved):
LGTM minus minor issues

PR #747, opened 2021-06-22
Expose public structured logging interface: fxevent

shirchen:
As part of #733 we need to finalize and expose a logging interface for
users to implement. After some internal discussion, we decided that
instead of the `Field{string, interface{}}` approach that can only be
used for logging, we'll emit structured events.

    package fxevent

    type Logger interface{ LogEvent(Event) }
    type Event interface{ /* closed interface */ }

Each event that Fx was previously logging will now be emitted as a
structured event type that implements the Event interface.

    package fxevent

    type Provide struct{ Constructor interface{} }

This gives implementations a lot more control over what is logged and
how, and leaves us room for implementing metrics and other
observability on Fx events in the future.

We also completely replace existing internal logging implementation to
use this new interface.

Note that #733 will need to be rebased on top of this.

Ref GO-666


internal/fxlog/fxlog.go
```diff
@@ -21,86 +21,127 @@
...
+	Err      error
+	Stack    fxreflect.Stack
 }
 
+// StartFailureError is emitted right before exiting after failing to start.
+type StartFailureError struct{ Err error }
+
+// StopSignalEvent is emitted whenever application receives a signal after
+// starting the application.
+type StopSignalEvent struct{ Signal string }
```
abhinav, line 83: We should name the Signal "SignalName" or include the `os.Signal` here.
In the code where we supply this, we currently do:

    sig := (<-ch).String()

Instead, if  we do the following, we'll have the os.Signal that fired.

    sig := <-ch

internal/fxlog/fxlog.go
```diff
@@ -21,86 +21,127 @@
...
-	return fs
-}
-
-func (l *zapLogger) Log(entry Entry) {
-	switch entry.Level {
-	case InfoLevel:
-		l.logger.Info(entry.Message, encodeFields(entry.Fields, entry.Stack)...)
-	case ErrorLevel:
-		l.logger.Error(entry.Message, encodeFields(entry.Fields, entry.Stack)...)
+func (l *zapLogger) LogEvent(event Event) {
```
abhinav, line 104: This is great!

internal/fxlog/fxlog.go
```diff
@@ -21,86 +21,127 @@
...
+		for _, rtype := range fxreflect.ReturnTypes(e.Constructor) {
+			l.logger.Info("supplying",
+				zap.String("constructor", fxreflect.FuncName(e.Constructor)),
+				zap.String("type", rtype),
+			)
+		}
+	case ProvideEvent:
+		for _, rtype := range fxreflect.ReturnTypes(e.Constructor) {
+			l.logger.Info("providing",
+				zap.String("constructor", fxreflect.FuncName(e.Constructor)),
```
abhinav, line 122: Not a blocker for this diff but FYI: once https://github.com/uber-go/dig/pull/280 lands, we might be able to get rid of fxreflect.FuncName completely.

internal/fxlog/fxlog.go
```diff
@@ -21,86 +21,127 @@
...
+	case StartFailureError:
+		l.logger.Error("failed to start", zap.Error(event.(StartFailureError).Err))
+	case StopSignalEvent:
+		l.logger.Info("received signal", zap.String("signal", strings.ToUpper(e.Signal)))
+	case StopErrorEvent:
+		l.logger.Error("failed to stop cleanly", zap.Error(event.(StopErrorEvent).Err))
+	case StartRollbackError:
+		l.logger.Error("could not rollback cleanly", zap.Error(event.(StartRollbackError).Err))
+	case StartErrorEvent:
+		l.logger.Error("startup failed, rolling back", zap.Error(event.(StartErrorEvent).Err))
```
abhinav, line 142: Since we're doing `e := event.(type)`, we can use `e` here to reference the variable with the matched type instead of casting `event` again.
```suggestion
		l.logger.Error("failed to start", zap.Error(e.Err))
	case StopSignalEvent:
		l.logger.Info("received signal", zap.String("signal", strings.ToUpper(e.Signal)))
	case StopErrorEvent:
		l.logger.Error("failed to stop cleanly", zap.Error(e.Err))
	case StartRollbackError:
		l.logger.Error("could not rollback cleanly", zap.Error(e.Err))
	case StartErrorEvent:
		l.logger.Error("startup failed, rolling back", zap.Error(e.Err))
```

abhinav (commented):
Wow, that was quick! This is looking great. I've mostly got minor comments/side-notes. 

CC @sywhang; this ties in a little with your work too. We might end up with events for #746.

sywhang:
> We might end up with events for #746.

Yeah, I like this change so I'll hold off on #746 until this one is merged and rebase that branch using these.

internal/fxlog/spy.go
```diff
@@ -21,64 +21,41 @@
...
-		}
-		msg.WriteString("\n")
-	}
+// Events returns all captured events.
+func (s *Spy) Events() []Event {
+	events := make([]Event, len(s.events))
+	copy(events, s.events)
 
-	return msg.String()
+	return s.events
```
sywhang, line 45: Looks like this is returning the original slice and not the copied one:
```suggestion
	return events
```

app.go
```diff
@@ -765,15 +755,15 @@ func (app *App) start(ctx context.Context) error {
 	// Attempt to start cleanly.
 	if err := app.lifecycle.Start(ctx); err != nil {
 		// Start failed, rolling back.
-		fxlog.Info("startup failed, rolling back", fxlog.Err(err)).Write(app.log)
+		app.log.LogEvent(&fxlog.StartErrorEvent{Err: err})
```
abhinav, line 758: Oh, I see that we have a StartErrorEvent here. We could name this RollbackEvent, since that's what we're doing next. (We're usually emitting events right we do a thing, unless it can only be done after -- like receiving an error.)

If we call this RollbackEvent, the Err parameter can be StartErr: the error that caused the rollback to happen.

app.go
```diff
@@ -765,15 +755,15 @@ func (app *App) start(ctx context.Context) error {
...
 		if stopErr := app.lifecycle.Stop(ctx); stopErr != nil {
-			fxlog.Info("could not rollback cleanly", fxlog.Err(stopErr)).Write(app.log)
+			app.log.LogEvent(&fxlog.StartRollbackError{Err: stopErr})
```
abhinav, line 760: This can be RollbackErrorEvent if we name the error above RollbackEvent.

internal/fxlog/fxlog.go
```diff
@@ -21,86 +21,145 @@
...
+func (*InvokeFailedEvent) event()     {}
+func (*StartFailureError) event()     {}
+func (*StopSignalEvent) event()       {}
+func (*StopErrorEvent) event()        {}
+func (*StartErrorEvent) event()       {}
+func (*StartRollbackError) event()    {}
+func (*RunningEvent) event()          {}
+
+// LifecycleOnStartEvent is emitted for whenever an OnStart hook is executed
+type LifecycleOnStartEvent struct {
```
abhinav, line 59: suggestion: LifecycleHookStartEvent (since the thing you put into the lifecycle is called a Hook)

internal/fxlog/fxlog.go
```diff
@@ -21,86 +21,145 @@
...
+	CallerName string
 }
 
-// WithStack mutates an Entry to add a Stack field.
-func (e Entry) WithStack(stack string) Entry {
-	e.Stack = stack
-	return e
+// LifecycleOnStopEvent is emitted for whenever an OnStart hook is executed
+type LifecycleOnStopEvent struct {
```
abhinav, line 64: suggestion: LifecycleHookStopEvent

internal/fxlog/fxlog.go
```diff
@@ -21,86 +21,145 @@
...
+	CallerName string
 }
 
-func (e Entry) Write(logger Logger) {
-	logger.Log(e)
+// ApplyOptionsError is emitted whenever there is an error applying options.
+type ApplyOptionsError struct {
```
abhinav, line 69: Already suggested an alternative name for this.

internal/fxlog/fxlog.go
```diff
@@ -21,86 +21,145 @@
...
+// InvokeEvent is emitted whenever a function is invoked.
+type InvokeEvent struct {
+	Function interface{}
 }
 
-// Logger defines interface used for logging.
-type Logger interface {
-	Log(entry Entry)
+// InvokeFailedEvent is emitted when fx.Invoke has failed.
+type InvokeFailedEvent struct {
```
abhinav, line 90: suggestion: InvokeErrorEvent

internal/fxlog/fxlog.go
```diff
@@ -21,86 +21,127 @@
...
+	Err error
 }
 
-// Err is a helper for error Fields.
-func Err(value error) Field {
-	return F("error", value)
+// SupplyEvent is emitted whenever a Provide was called with a constructor provided
+// by fx.Supply.
+type SupplyEvent struct {
+	Constructor interface{}
```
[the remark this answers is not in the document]
abhinav: Oh, hold up. fx.Supply doesn't take a function. It takes a value.

```
x := T1{}
fx.Supply(x)
```

We internally turn it into a constructor so that we can call Provide. So it's equivalent to,

```
x := T1{}
fx.Provide(func() T1 { return x })
```

---

For these, we only log the types of values (as seen in the deleted code above). So for now, I would make this a TypeName field.

```suggestion
	TypeName string
```

internal/fxlog/fxlog.go
```diff
@@ -21,86 +21,145 @@
...
-	Log(entry Entry)
+// InvokeFailedEvent is emitted when fx.Invoke has failed.
+type InvokeFailedEvent struct {
+	Function   interface{}
+	Err        error
+	Stacktrace string
 }
 
+// StartFailureError is emitted right before exiting after failing to start.
+type StartFailureError struct{ Err error }
```
abhinav, line 97: StartErrorEvent

internal/fxlog/fxlog.go
```diff
@@ -21,86 +21,145 @@
...
+
+// StopSignalEvent is emitted whenever application receives a signal after
+// starting the application.
+type StopSignalEvent struct{ Signal os.Signal }
+
+// StopErrorEvent is emitted whenever we fail to stop cleanly.
+type StopErrorEvent struct{ Err error }
+
+// StartErrorEvent is emitted whenever a service fails to start.
+type StartErrorEvent struct{ Err error }
```
abhinav, line 107: RollbackEvent with a StartErr field

internal/fxlog/fxlog.go
```diff
@@ -21,86 +21,145 @@
...
+
+// StartRollbackError is emitted whenever we fail to rollback cleanly after
+// a start error.
+type StartRollbackError struct{ Err error }
```
abhinav, line 111: RollbackErrorEvent

internal/fxlog/fxlog.go
```diff
@@ -21,86 +21,145 @@
...
+		l.logger.Info("starting", zap.String("caller", e.CallerName))
+	case *LifecycleOnStopEvent:
+		l.logger.Info("stopping", zap.String("caller", e.CallerName))
+	case *ApplyOptionsError:
+		l.logger.Error("error encountered while applying options", zap.Error(e.Err))
+	case *SupplyEvent:
+		for _, rtype := range fxreflect.ReturnTypes(e.Constructor) {
+			l.logger.Info("supplying",
+				zap.String("constructor", fxreflect.FuncName(e.Constructor)),
+				zap.String("type", rtype),
```
abhinav, line 134: this doesn't match what we were doing for SupplyEvent earlier. we don't log the constructor here because the constructor is made-up. the user only provided a value, so we only log its type.

```suggestion
				zap.String("type", e.TypeName),
```

abhinav (commented):
The core logic looks fine minus the supply logging mismatch.
I have naming suggestions for some of the events, though.

fxevent/event.go
```diff
@@ -0,0 +1,101 @@
...
+func (*Invoke) event()             {}
+func (*InvokeError) event()        {}
+func (*StartError) event()         {}
+func (*StopSignal) event()         {}
+func (*StopError) event()          {}
+func (*Rollback) event()           {}
+func (*RollbackError) event()      {}
+func (*Running) event()            {}
+
+// LifecycleHookStart is emitted for whenever an OnStart hook is executed
```
sywhang, line 45:
```suggestion
// LifecycleHookStart is emitted whenever an OnStart hook is executed
```

fxevent/event.go
```diff
@@ -0,0 +1,101 @@
...
+type LifecycleHookStart struct {
+	CallerName string
+}
+
+// LifecycleHookStop is emitted for whenever an OnStart hook is executed
```
sywhang, line 50:
```suggestion
// LifecycleHookStop is emitted whenever an OnStart hook is executed
```

sywhang (approved):
LGTM modulo couple of nits

PR #750, opened 2021-06-24
fxreflect: Delete ReturnTypes in favor of dig.Info

abhinav:
With uber-go/dig#280 done, we can use the ProvideInfo produced by Dig to
determine what types will be produced by a constructor and log them, so we
no longer need to reimplement the `dig.Out` inspection logic in Fx.

To surface this information with the new event API introduced in #747, this
adds a new field to fxevent.Provide:

    OutputTypeNames []string

This, in lieu of exposing `dig.Info` wholesale ensures that we have some
wiggle room here.

Note that this introduces one behavioral difference on when the Provide
event is logged: previously, the Provide event would be logged whether
it succeeded or failed. I've created GO-677 internally to discuss
consistency here.

With this change, it will be logged only if it succeeds -- after it
succeeds. This makes sense because otherwise we can't extract information
about the function from Dig.

This temporarily pins Fx to the unreleased `master` branch of Dig. We'll
have to tag a releaes of Dig before we can release this feature.

Refs GO-676

### DIFF
--- a/internal/fxreflect/fxreflect.go
+++ b/internal/fxreflect/fxreflect.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2019 Uber Technologies, Inc.
+// Copyright (c) 2019-2021 Uber Technologies, Inc.
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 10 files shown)


sywhang (approved):
LGTM

internal/fxreflect/fxreflect.go
```diff
@@ -27,77 +27,11 @@ import (
...
-func (k *key) String() string {
-	if k.name != "" {
-		return fmt.Sprintf("%v:%s", k.t, k.name)
-	}
-	return k.t.String()
-}
-
-func traverseOuts(k key, f func(s string)) {
-	// skip errors
-	if isErr(k.t) {
```
sywhang, line 72: I think you can remove `isErr` since it's only being used here.

abhinav: Oh yeah, good catch. This should have been caught by staticcheck, but it looks like we don't have it set up for Fx. I'll follow up with that in a separate PR.

PR #751, opened 2021-06-28
Enable staticcheck

abhinav:
Enable staticcheck on Fx and fix issues found by it.

Since this introduces a bunch of new dependencies, move the tools dependencies
to an unpublished submodule of Fx similar to uber-go/zap#914.


sywhang (approved):
:shipit:

shirchen (approved):

PR #752, opened 2021-07-01
tools: Add internal linter, lint fxevent.Loggers

abhinav:
With #733, we have two fxevent.Logger implementations in Fx, and we plan
to change the event types a fair bit in the near future. Rather than
manually verify that the two linters are in-sync and handle all event
types, it's preferable to verify this automatically.

Add a custom linter based on go/analysis meant for use only inside Fx.
This linter verifies that any fxevent.Logger implementations it finds
handle all fxevent.Event types, or none of them.

For testing with analysistest, this creates a fake fxevent package
similar to the real fxevent package, and lints code that uses it.


abhinav:
FOSSA CLI appears to be broken. Please ignore that failure.
image

tools/analysis/passes/allfxevents/analysis.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+
+	// Each function declaration gets its own copy of the typeSet to track
+	// events in.
+	v.loggerType = t
+	v.funcEvents = v.Fxevent.Events.Clone()
+	return true
+}
+
+func (v *visitor) funcDeclExit(n *ast.FuncDecl) bool {
```
sywhang, line 152: `v.funcEvents.Len()` is repeatedly used. Can we pull it out to a variable and reuse it?
```suggestion
func (v *visitor) funcDeclExit(n *ast.FuncDecl) bool {
    nEvents := v.funcEvents.Len()
```

tools/analysis/passes/allfxevents/analysis.go
```diff
@@ -0,0 +1,278 @@
...
+// Iterate iterates through the type set in an unspecified order.
+func (ts *typeSet) Iterate(f func(types.Type)) {
+	ts.m.Iterate(func(t types.Type, _ interface{}) {
+		f(t)
+	})
+}
+
+func (ts *typeSet) Clone() *typeSet {
+	var out typeSet
+	ts.Iterate(out.Put)
```
sywhang, line 270: nice! 😍

sywhang (approved):
LGTM

PR #754, opened 2021-07-01
Fix race in timeout error report

sywhang:
In the timeout tests where we intentionally make hooks timeout and check the error messages, I noticed that it sometimes fails with errors like this:

```
        writer.go:40: [Fx] ERROR		Start failed, rolling back: context deadline exceeded
        app_test.go:698: 
            	Error Trace:	app_test.go:698
            	Error:      	"context deadline exceeded" does not contain "OnStart hook added by go.uber.org/fx_test.TestAppStart.func2.3 failed: context deadline exceeded"
            	Test:       	TestAppStart/TimeoutWithFinishedHooks
        app_test.go:701: 
            	Error Trace:	app_test.go:701
            	Error:      	Expect "context deadline exceeded" to match "app_test.go:\d+"
            	Test:       	TestAppStart/TimeoutWithFinishedHooks
        app_test.go:706: 
            	Error Trace:	app_test.go:706
            	Error:      	"-1" is not greater than "-1"
            	Test:       	TestAppStart/TimeoutWithFinishedHooks
    writer.go:40: [Fx] ERROR		Start failed, rolling back: context canceled
```

There were two races:

1. It's possible for the goroutine executing the OnStart hook itself to see the timeout error first, then return it, which will hit the `err := <- c` case in `withTimeout`, which just returns the error as it is. 

Fixed this race by checking for the timeout error in the `err := <- c` case and format the error the same way as when the parent goroutine sees the timeout error first. 

2. When the timeout error is returned from the onStart to the main goroutine, the main goroutine pokes at the lifecycle's records field while the onStart-executing goroutine is trying to unroll using Stop(), which overwrites that records field. 

Fixed this race by splitting the records field into start records and stop records which are reported separately.

### DIFF
--- a/app.go
+++ b/app.go
@@ -912,23 +912,16 @@ func withTimeout(ctx context.Context, param *withTimeoutParams) error {
 	c := make(chan error, 1)
 	go func() { c <- param.callback(ctx) }()
 
+	var err error
+
 	select {
 	case <-ctx.Done():
-		err := ctx.Err()
-		if err != context.DeadlineExceeded {
-			return err
-		}
-		return formatTimeoutError(err, param)
-	case err := <-c:
-		if err != context.DeadlineExceeded {
-			return err
-		}
-		return formatTimeoutError(err, param)
+		err = ctx.Err()
+	case err = <-c:
 	}
-}
-
-// formatTimeoutError is used by withTimeout to wrap timeout error with hook report
-func formatTimeoutError(err error, param *withTimeoutParams) error {
+	if err != context.DeadlineExceeded {
+		return err
+	}
 	// On timeout, report running hook's caller and recorded
 	// runtimes of hooks successfully run till end.
 	var r lifecycle.HookRecords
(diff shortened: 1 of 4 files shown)


app.go
```diff
@@ -918,28 +918,41 @@ func withTimeout(ctx context.Context, param *withTimeoutParams) error {
...
-		// TODO: Once this is integrated into fxevent, we can
-		// leave error unchanged and send this to fxevent.Logger, whose
-		// implementation can then determine how the error is presented.
-		if len(r) > 0 {
-			return fmt.Errorf("%v hook added by %v failed: %w\n%+v",
-				param.hook,
-				caller,
-				err,
-				r)
+		return formatTimeoutError(err, param)
```
abhinav, line 921: We can dedupe this a bit:

```
var err error
select {
case <-ctx.Done():
  err = ctx.Err()
case err = <-c:
}

if err != context.DeadlineExceeded {
  return err
}

return formatTimeoutError(err, param)
```

(at which point this can go back to inlining formatTimeoutError into the withTimeout body.)

abhinav (commented):
minor deduping but lgtm otherwise

abhinav (approved):
thanks!

PR #756, opened 2021-07-09
Rename event types and messages with consistent tenses.

shirchen:
Use active voice for events before we execute the events and passive after they are run. Here we also combine similar events to add an optional Err field if said event errored. 

Note, here for backwards compatibility we do not yet change event verbs for previous default console logger. New events will also be added in a followup diff as necessary.

Ref GO-677


fxevent/event.go
```diff
@@ -56,40 +52,34 @@ type LifecycleHookStop struct {
...
-// Invoke is emitted whenever a function is invoked.
-type Invoke struct {
-	Function interface{}
+	// Err is emitted if there was an error applying options.
+	Err error
 }
 
-// InvokeError is emitted when fx.Invoke has failed.
-type InvokeError struct {
+// Invoked is emitted whenever a function is invoked and whether it errored.
```
sywhang, line 73:
```suggestion
// Invoked is emitted whenever a function has been invoked and whether it errored.
```

fxevent/event.go
```diff
@@ -56,40 +52,34 @@ type LifecycleHookStop struct {
...
+type Invoked struct {
 	Function   interface{}
 	Err        error
 	Stacktrace string
 }
 
-// StartError is emitted right before exiting after failing to start.
-type StartError struct{ Err error }
+// Started is emitted whenever an application is started successfully or
+// it errors.
```
sywhang, line 81:
```suggestion
// Started is emitted whenever an application has been started and whether
// it errored.
```

fxevent/event.go
```diff
@@ -98,20 +88,15 @@ type StopSignal struct{ Signal os.Signal }
...
 type RollbackError struct{ Err error }
 
-// Running is emitted whenever an application is started successfully.
-type Running struct{}
-
-// CustomLoggerError is emitted whenever a custom logger fails to construct.
-type CustomLoggerError struct{ Err error }
-
-// CustomLogger is emitted whenever a custom logger is set.
+// CustomLogger is emitted whenever a custom logger is set or produces an error.
```
sywhang, line 98: Can we change this to a verb form as well? Maybe `CustomLoggerSet`?

app.go
```diff
@@ -833,7 +832,7 @@ func (app *App) executeInvokes() error {
 
 	for _, i := range app.invokes {
 		fn := i.Target
-		app.log.LogEvent(&fxevent.Invoke{Function: fn})
+		app.log.LogEvent(&fxevent.Invoked{Function: fn})
```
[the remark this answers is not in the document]
shirchen: So yes and no. This one is indeed right before, but the error is after so we can't have it both ways for a single Event name. I guess on second thought, I should rename this back to `Invoke` and then use correct verbs within each logger implementation. I think in this diff I'll keep it before as if this action fails user will get the error back.

fxevent/event.go
shirchen: I'll change this to is being invoked...

fxevent/event.go
shirchen: I think it's an antipattern to add `Set` or `Get` in function names so I don't like that in structs either. In any case, I think this is redundant anyway since this is a generic event.

fxevent/event.go
```diff
@@ -32,18 +32,14 @@ type Event interface {
...
-func (*Supply) event()             {}
-func (*Provide) event()            {}
-func (*Invoke) event()             {}
-func (*InvokeError) event()        {}
-func (*StartError) event()         {}
+func (*Supplied) event()           {}
+func (*Provided) event()           {}
+func (*Invoked) event()            {}
 func (*StopSignal) event()         {}
 func (*StopError) event()          {}
```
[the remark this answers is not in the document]
shirchen: Good point. I'll combine.

app.go
```diff
@@ -887,7 +886,7 @@ func (app *App) start(ctx context.Context) error {
 	// Attempt to start cleanly.
 	if err := app.lifecycle.Start(ctx); err != nil {
 		// Start failed, rolling back.
-		app.log.LogEvent(&fxevent.Rollback{StartErr: err})
+		app.log.LogEvent(&fxevent.RollingBack{StartErr: err})
 		if stopErr := app.lifecycle.Stop(ctx); stopErr != nil {
 			app.log.LogEvent(&fxevent.RollbackError{Err: stopErr})
```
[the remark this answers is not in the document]
shirchen: Ugh, yea, I struggled with this one as it would have two error fields which are easy to confuse if combined. Let me update this by either combining and making error fields super distinct or leaving them separate.

fxevent/event.go
```diff
@@ -32,18 +32,14 @@ type Event interface {
 // Passing events by type to make Event hashable in the future.
 func (*LifecycleHookStart) event() {}
 func (*LifecycleHookStop) event()  {}
```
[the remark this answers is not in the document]
shirchen: Hmm.. yea, maybe I should leave it for now as these could be used for both before action and after if we ever emit an error.

sywhang (approved):
Mostly LGTM! One thing I'm a little bit sad about is `CustomLogger` because it doesn't really fit well with the rest of the events' naming scheme (verb vs noun) but I can't really come up with a better name for it at the moment. Maybe @abhinav might have ideas there?

app.go
```diff
@@ -770,7 +770,7 @@ func (app *App) provide(p provide) {
 
 		switch {
 		case p.IsSupply:
-			app.log.LogEvent(&fxevent.Supply{TypeName: p.SupplyType.String()})
+			app.log.LogEvent(&fxevent.Supplied{TypeName: p.SupplyType.String()})
```
abhinav, line 773: Both, Provide and Supply events are emitted after the operation. (This is a `defer` block.)

app.go
```diff
@@ -562,7 +562,7 @@ func New(opts ...Option) *App {
 	app.provide(provide{Target: app.dotGraph, Stack: frames})
 
 	if app.err != nil {
-		app.log.LogEvent(&fxevent.ProvideError{Err: app.err})
+		app.log.LogEvent(&fxevent.Provide{Err: app.err})
```
abhinav, line 565: we should probably do this in the `app.provide`, including both, the constructor that we failed to provide and the error we got back in one event. (So we just need to update that `defer`.)

abhinav (commented):
I'll look more closely later today, but

RE: CustomLogger: Yeah, not a fan of the name. Just an idea: what if we just name it `Logger`, `LoggerInitialized`, `LoggerConstructed` or similar and invoke it regardless of whether the logger is custom or not?  So if you had a custom logger, you get `LoggerInitialized{Logger: log, Err: err, Constructor: fn}`, and if you didn't provide a custom logger, you get `LoggerInitialized{Logger: log}`.

fxevent/event.go
mh-park: A small nit on the documentation wording: as someone that doesn't have a lot of context, it's not clear that the the "produced error" mentioned here is related to setting the custom logger. maybe something like: 
```
// CustomLogger is emitted whenever a custom logger is set or an error is produced in the process.
```

fxevent/event.go
```diff
@@ -56,14 +50,9 @@ type LifecycleHookStop struct {
...
-// ProvideError is emitted whenever there is an error applying options.
-type ProvideError struct {
-	Err error
-}
-
-// Supply is emitted whenever a Provide was called with a constructor provided
+// Supplied is emitted whenever a Provide was called with a constructor provided
 // by fx.Supply.
-type Supply struct {
+type Supplied struct {
```
mh-park, line 55: Is there a reason why some of the types here are past tense and some of these are present tense?

sywhang: You might want to refer to GO-677 for discussion on that. Tl;dr: present tense is for events that can be fired both before/after somethign happens, and past tense is for events that are fired after something happens.

mh-park: Ah I see, thanks for the clarification :)

mh-park: Maybe we should add that to the package documentation here; if a future contributor came along and added a new event type here, I don't think they would know that there are naming guidelines.

app.go
abhinav: Offline discussion suggested that this PR is limited to renames only, and not changing which events get emitted when.

fxevent/event.go
abhinav: I suppose we can stick with CustomLogger for now, and change it up in the future PR where we're adding/changing events.

abhinav (approved):
Reducing scope of this PR to just renames, LGTM.

This is not how we would want to release, though, since we need to merge some of the emissions and move them around. Like Provide would become Provided and only emitted from one place, where we have access to both, the constructor that failed to be provided, and the error for it. We might add a separate Providing event if we also want to log the "before" for it, but I don't think that would be necessary. For Invokes, on the other hand, we'd probably want to do both, Invoking and Invoked.

PR #757, opened 2021-07-10
Add fx.Annotate

sywhang:
This adds fx.Annotate, along with fx.Annotation, fx.ParamTags, and fx.ResultTags.

fx.Annotate takes an arbitrary function and annotates it with supplied fx.Annotations.

fx.ParamTags takes in a list of strings in the form of struct tags that can be used as tags to each parameter to the supplied function, and returns an fx.Annotation that can be used with fx.Annotate. 

fx.ResultTags also takes in a list of strings in the form of struct tags that can be used as tags to each result of the supplied function, and returns an fx.Annotation that can be used with fx.Annotate.

Using this, one can annotate both parameter and result to `fx.Provide`d or `Invoke`d functions.

Current implementation makes heavy use of reflection to construct a struct type at runtime, both for the parameters and results depending on which tags are specified. fx.Annotate uses reflection to generate a function wrapper that invokes the provided function with an a struct that contains all the arguments with annotated parameters, and wraps the result inside a struct with the specified result tags.

We may decide in the future to replace the implementation similar to how fx.Annotated is implemented, but for now this should still work.

Refs GO-549.

### DIFF
--- a/annotated.go
+++ b/annotated.go
@@ -101,6 +101,17 @@ type paramTags struct {
 	tags []string
 }
 
+// Given func(T1, T2, T3, ..., TN), this generates a type roughly
+// equivalent to,
+//
+//   struct {
+//     fx.In
+//
+//     Field1 T1 `$tags[0]`
+//     Field2 T2 `$tags[1]`
+//     ...
+//     FieldN TN `$tags[N-1]`
+//   }
 func (p paramTags) getAnnotatedType(fType reflect.Type) []reflect.Type {
 	annotatedParams := []reflect.StructField{{
 		Name:      "In",
@@ -135,6 +146,17 @@ type resultTags struct {
 	tags []string
 }
 
+// Given func(T1, T2, T3, ..., TN), this generates a type roughly
+// equivalent to,
+//
+//   struct {
+//     fx.Out
+//
+//     Field1 T1 `$tags[0]`
+//     Field2 T2 `$tags[1]`
+//     ...
+//     FieldN TN `$tags[N-1]`
+//   }
 func (resultTags) getAnnotatedType(fType reflect.Type) []reflect.Type {
 	annotatedResult := []reflect.StructField{{
 		Name:      "Out",
@@ -199,14 +221,14 @@ func ResultTags(tags ...string) Annotation {
 //
 // For example,
 //
-// fx.Provide(
-//   fx.Annotate(
-//     NewGateWay,
-//     fx.ParamTags(`name:"ro" optional:"true"`),
-//     fx.ParamTags(`name:"rw"),
-//     fx.ResultTags(`name:"foo"`)
-//   )
-// )
+//  fx.Provide(
+//    fx.Annotate(
+//      NewGateWay,
+//      fx.ParamTags(`name:"ro" optional:"true"`),
+//      fx.ParamTags(`name:"rw"),
+//      fx.ResultTags(`name:"foo"`)
+//    )
+//  )
 //
 // is considered an invalid usage and will not apply any of the
 // Annotations to NewGateway.
--- a/annotated.go
+++ b/annotated.go
@@ -157,18 +157,29 @@ type resultTags struct {
 //     ...
 //     FieldN TN `$tags[N-1]`
 //   }
-func (resultTags) getAnnotatedType(fType reflect.Type) []reflect.Type {
+func (r resultTags) getAnnotatedType(fType reflect.Type) []reflect.Type {
 	annotatedResult := []reflect.StructField{{
 		Name:      "Out",
 		Type:      reflect.TypeOf(Out{}),
 		Anonymous: true,
 	}}
 
+	errInterface := reflect.TypeOf((*error)(nil)).Elem()
+	numAnnotated := 0
+
 	for i := 0; i < fType.NumOut(); i++ {
+		// guard against error results
+		if fType.Out(i).Implements(errInterface) {
+			continue
+		}
 		structField := reflect.StructField{
-			Name: fmt.Sprintf("Field%d", i),
+			Name: fmt.Sprintf("Field%d", numAnnotated),
 			Type: fType.Out(i),
 		}
+		if numAnnotated < len(r.tags) {
+			structField.Tag = reflect.StructTag(r.tags[numAnnotated])
+		}
+		numAnnotated += 1
 		annotatedResult = append(annotatedResult, structField)
 	}
 
@@ -290,9 +301,16 @@ func Annotate(f interface{}, anns ...Annotation) interface{} {
 		fResults = fVal.Call(fParams)
 		if annotatedOut {
 			// wrap the result in an annotated struct
+			errInterface := reflect.TypeOf((*error)(nil)).Elem()
+			numAnnotated := 0
 			results := reflect.New(outs[0]).Elem()
 			for i := 0; i < numOut; i++ {
-				results.FieldByName(fmt.Sprintf("Field%d", i)).Set(fResults[i])
+				if fResults[i].Type().Implements(errInterface) {
+					continue
+				}
+				results.FieldByName(fmt.Sprintf("Field%d",
+					numAnnotated)).Set(fResults[numAnnotated])
+				numAnnotated += 1
 			}
 			return []reflect.Value{results}
 		}
--- a/annotated.go
+++ b/annotated.go
@@ -94,15 +94,21 @@ func (a Annotated) String() string {
 type annotations struct {
 	ParamTags  []string
 	ResultTags []string
+
+	Ins  []reflect.Type
+	Outs []reflect.Type
 }
 
 // Annotation can be passed to Annotate(f interface{}, anns ...Annotation)
 // for annotating the parameter and result types of a function.
 type Annotation interface {
 	apply(*annotations) error
-	getAnnotatedType(reflect.Type) []reflect.Type
 }
 
+// annotationError is a wrapper for an error that was encountered while
+// applying annotation to a function. It contains the specific error
+// that it encountered as well as the target interface that was attempted
+// to be annotated.
 type annotationError struct {
 	target interface{}
 	err    error
@@ -112,11 +118,13 @@ func (e *annotationError) Error() string {
 	return e.err.Error()
 }
 
-type paramTags struct {
+type paramTagsAnnotation struct {
 	tags []string
 }
 
-func (pt paramTags) apply(ann *annotations) error {
+var _ paramTagsAnnotation = paramTagsAnnotation{}
+
+func (pt paramTagsAnnotation) apply(ann *annotations) error {
 	if len(ann.ParamTags) > 0 {
 		return fmt.Errorf("cannot apply more than one line of ParamTags")
 	}
@@ -135,7 +143,7 @@ func (pt paramTags) apply(ann *annotations) error {
 //     ...
 //     FieldN TN `$tags[N-1]`
 //   }
-func (pt paramTags) getAnnotatedType(fType reflect.Type) []reflect.Type {
+func (pt paramTagsAnnotation) getAnnotatedType(fType reflect.Type) []reflect.Type {
 	annotatedParams := []reflect.StructField{{
 		Name:      "In",
 		Type:      reflect.TypeOf(In{}),
@@ -153,29 +161,31 @@ func (pt paramTags) getAnnotatedType(fType reflect.Type) []reflect.Type {
 		annotatedParams = append(annotatedParams, structField)
 	}
 
-	paramTypes := []reflect.Type{reflect.StructOf(annotatedParams)}
-	return paramTypes
+	return []reflect.Type{reflect.StructOf(annotatedParams)}
 }
 
 // ParamTags is an Annotation that annotates the parameter(s) of a function.
 // When multiple tags are specified, each tag is mapped to the corresponding
 // positional parameter.
 func ParamTags(tags ...string) Annotation {
-	p := paramTags{tags}
-	return p
+	return paramTagsAnnotation{tags}
 }
 
-type resultTags struct {
+type resultTagsAnnotation struct {
 	tags []string
 }
 
-func (rt resultTags) apply(ann *annotations) error {
+var _ resultTagsAnnotation = resultTagsAnnotation{}
+
+func (rt resultTagsAnnotation) apply(ann *annotations) error {
 	if len(ann.ResultTags) > 0 {
 		return fmt.Errorf("cannot apply more than one line of ResultTags")
 	}
 	ann.ResultTags = rt.tags
 	return nil
 }
+
+var _typeOfError reflect.Type = reflect.TypeOf((*error)(nil)).Elem()
 
 // Given func(T1, T2, T3, ..., TN), this generates a type roughly
 // equivalent to,
@@ -188,19 +198,18 @@ func (rt resultTags) apply(ann *annotations) error {
 //     ...
 //     FieldN TN `$tags[N-1]`
 //   }
-func (rt resultTags) getAnnotatedType(fType reflect.Type) []reflect.Type {
+func (rt resultTagsAnnotation) getAnnotatedType(fType reflect.Type) []reflect.Type {
 	annotatedResult := []reflect.StructField{{
 		Name:      "Out",
 		Type:      reflect.TypeOf(Out{}),
 		Anonymous: true,
 	}}
 
-	errInterface := reflect.TypeOf((*error)(nil)).Elem()
-	numAnnotated := 0
+	var numAnnotated int
 
 	for i := 0; i < fType.NumOut(); i++ {
 		// guard against error results
-		if fType.Out(i).Implements(errInterface) {
+		if fType.Out(i) == _typeOfError {
 			continue
 		}
 		structField := reflect.StructField{
@@ -214,16 +223,14 @@ func (rt resultTags) getAnnotatedType(fType reflect.Type) []reflect.Type {
 		annotatedResult = append(annotatedResult, structField)
 	}
 
-	resultTypes := []reflect.Type{reflect.StructOf(annotatedResult)}
-	return resultTypes
+	return []reflect.Type{reflect.StructOf(annotatedResult)}
 }
 
 // ResultTags is an Annotation that annotates the result(s) of a function.
 // When multiple tags are specified, each tag is mapped to the corresponding
 // positional result.
 func ResultTags(tags ...string) Annotation {
-	r := resultTags{tags}
-	return r
+	return resultTagsAnnotation{tags}
 }
 
 // Annotate lets you annotate a function's paramter and returns with tags
@@ -283,12 +290,10 @@ func Annotate(f interface{}, anns ...Annotation) interface{} {
 	numIn := fType.NumIn()
 	numOut := fType.NumOut()
 
-	var annotations annotations
-	var ins []reflect.Type
-	var outs []reflect.Type
-
-	annotatedIn := false
-	annotatedOut := false
+	var (
+		annotations               annotations
+		annotatedIn, annotatedOut bool
+	)
 
 	for _, ann := range anns {
 		if e := ann.apply(&annotations); e != nil {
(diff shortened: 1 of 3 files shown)


annotated.go
```diff
@@ -89,3 +90,228 @@ func (a Annotated) String() string {
...
+// For example,
+//
+// fx.Provide(
+//   fx.Annotate(
+//     NewGateWay,
+//     fx.ParamTags(`name:"ro" optional:"true"`),
+//     fx.ParamTags(`name:"rw"),
+//     fx.ResultTags(`name:"foo"`)
+//   )
+// )
```
abhinav, line 209: Code samples should be indented at least one more space for godoc to render them as code blocks.
```suggestion
//  fx.Provide(
//    fx.Annotate(
//      NewGateWay,
//      fx.ParamTags(`name:"ro" optional:"true"`),
//      fx.ParamTags(`name:"rw"),
//      fx.ResultTags(`name:"foo"`)
//    )
//  )
```

annotated.go
```diff
@@ -89,3 +90,228 @@ func (a Annotated) String() string {
...
+//
+//    GW *Gateway `name:"foo"`
+//   }
+//
+//   fx.Provide(func(p params) result {
+//     return result{GW: NewGateway(p.RO, p.RW)}
+//   })
+//
+// If there are more Annotations are specified than than one per
+// Annotation type, none of the Annotations will be applied.
```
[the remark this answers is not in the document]
abhinav: But this seems like a strange silent failure. Is this because we don't have any means of communicating failure?

annotated.go
```diff
@@ -89,3 +90,228 @@ func (a Annotated) String() string {
...
+	if !annotatedOut {
+		outs = make([]reflect.Type, numOut)
+		for i := 0; i < fType.NumOut(); i++ {
+			outs[i] = fType.Out(i)
+		}
+	}
+
+	var newF func([]reflect.Value) []reflect.Value
+
+	if annotatedIn && annotatedOut {
```
abhinav, line 258: I suspect we can clean this up a bit. If we annotated the input, then we need a `func([]reflect.Value) []reflect.Value` that turns the fake fx.In struct into a list of parameters. If we annotated the output, we need one that turns a list of results into an fx.Out struct. The default implementation is "return as-is" for when we didn't annotate. Using that, we'll have only one code block generating `newF`.

annotated.go
```diff
@@ -89,3 +90,228 @@ func (a Annotated) String() string {
...
+	}
+
+	annotatedFuncType := reflect.FuncOf(ins, outs, false)
+	annotatedFunc := reflect.MakeFunc(annotatedFuncType, newF)
+	return annotatedFunc.Interface()
+}
+
+func verifyAnnotation(numIn int, numOut int, anns ...Annotation) bool {
+	sawParamAnn := false
+	sawResultAnn := false
```
abhinav, line 300: This seems like interpreting the annotations twice. There's an optional, slightly different way to go about this:

Instead of having paramTags and resultTags generate `[]reflect.Type`, have them update an annotations struct (like regular functional options), failing if the struct is already filled:

```
type annotations struct {
  ParamTags []string
  ResultTags []string
}

type Annotation interface{ apply(*annotations) }

func (pt paramTags) apply(ann *annotations) error {
  if len(pt.ParamTags) > 0 { return fmt.Errorf(...) }
  // ...
}
```

Then in `func Annotate`, as we apply these options, we'll get back an error with some amount of information.

We can generate the `[]reflect.Type`s inside `func Annotate` as needed (or inside a helper) elsewhere, probably with some shared code.

annotated.go
```diff
@@ -89,3 +90,228 @@ func (a Annotated) String() string {
...
+	} else if annotatedIn {
+		newF = func(args []reflect.Value) []reflect.Value {
+			fParams := make([]reflect.Value, numIn)
+			params := args[0]
+			for i := 0; i < numIn; i++ {
+				fParams[i] = params.Field(i + 1)
+			}
+			return fVal.Call(fParams)
+		}
+	} else { // annotatedOut
```
abhinav, line 281: I think annotatedOut might want to guard against `error` results. There can be an extraneous error argument at the end which should not be in the `fx.Out`.

annotated.go
```diff
@@ -89,3 +90,268 @@ func (a Annotated) String() string {
...
+}
+
+// Annotation can be passed to Annotate(f interface{}, anns ...Annotation)
+// for annotating the parameter and result types of a function.
+type Annotation interface {
+	apply(*annotations) error
+	getAnnotatedType(reflect.Type) []reflect.Type
+}
+
+type annotationError struct {
```
abhinav, line 106: Since this is used in a special way, consider adding some comments here to explain it.

annotated.go
```diff
@@ -89,3 +90,268 @@ func (a Annotated) String() string {
...
+	target interface{}
+	err    error
+}
+
+func (e *annotationError) Error() string {
+	return e.err.Error()
+}
+
+type paramTags struct {
```
abhinav, line 115: Consider naming this paramTagsAnnotation to make the interface compliance clearer.
Also, optionally, add an interface compliance check.

https://github.com/uber-go/guide/blob/master/style.md#verify-interface-compliance

annotated.go
```diff
@@ -89,3 +90,268 @@ func (a Annotated) String() string {
...
+	paramTypes := []reflect.Type{reflect.StructOf(annotatedParams)}
+	return paramTypes
+}
+
+// ParamTags is an Annotation that annotates the parameter(s) of a function.
+// When multiple tags are specified, each tag is mapped to the corresponding
+// positional parameter.
+func ParamTags(tags ...string) Annotation {
+	p := paramTags{tags}
+	return p
```
abhinav, line 165: nit: can simplify
```suggestion
	return paramTags{tags}
```

annotated.go
```diff
@@ -89,3 +90,268 @@ func (a Annotated) String() string {
...
+//     FieldN TN `$tags[N-1]`
+//   }
+func (rt resultTags) getAnnotatedType(fType reflect.Type) []reflect.Type {
+	annotatedResult := []reflect.StructField{{
+		Name:      "Out",
+		Type:      reflect.TypeOf(Out{}),
+		Anonymous: true,
+	}}
+
+	errInterface := reflect.TypeOf((*error)(nil)).Elem()
```
abhinav, line 198: This can be done once at package-level:

```
var _typeOfError = reflect.TypeOf((*error)(nil)).Elem()

func (rt resultTags) getAnnotatedType(..)
```

annotated.go
```diff
@@ -89,3 +90,268 @@ func (a Annotated) String() string {
...
+	numAnnotated := 0
```
abhinav, line 199: optional nitty nit: prefer using `var` form to instantiate zero value objects.

```
var numAnnotated int
```

slightly related: https://github.com/uber-go/guide/blob/master/style.md#use-var-for-zero-value-structs

annotated.go
```diff
@@ -89,3 +90,268 @@ func (a Annotated) String() string {
...
+
+	for i := 0; i < fType.NumOut(); i++ {
+		// guard against error results
+		if fType.Out(i).Implements(errInterface) {
+			continue
+		}
```
abhinav, line 205: Instead of Implements, we should maybe look at `==` here. The reason for this is that it is okay for me to do,

```
type MyThing struct{ ... }

func (*MyThing) Error() string

func NewMyThing(...) *MyThing
```

We should ignore it only if it is returned as `error`.

(I suspect we have a test for that somewhere for a different case in dig.)

annotated.go
```diff
@@ -89,3 +90,268 @@ func (a Annotated) String() string {
...
+		structField := reflect.StructField{
+			Name: fmt.Sprintf("Field%d", numAnnotated),
+			Type: fType.Out(i),
+		}
+		if numAnnotated < len(rt.tags) {
+			structField.Tag = reflect.StructTag(rt.tags[numAnnotated])
+		}
+		numAnnotated++
```
abhinav, line 213: Why is numAnnotated different from `i` here? If this is to handle the `error` case, I guess we should decide what we want the user code to look like. I assumed,

```
fx.Annotate(func(...) (Thing, error), fx.ResultTags(`name:"foo"`))
```

If there's a result after error, then,

```
fx.Annotate(func(...) (Thing, error, AnotherThing), fx.ResultTags(`name:"foo"`, "", `name:"bar"`))
```

In which case, `i` should be used, not `numAnnotated`. I think.

Of course, this might all be moot because I _think_ (but have not verified) that Fx will reject constructors that return errors as anything but their last result.

annotated.go
```diff
@@ -89,3 +90,268 @@ func (a Annotated) String() string {
...
+	fType := fVal.Type()
+	numIn := fType.NumIn()
+	numOut := fType.NumOut()
+
+	var annotations annotations
+	var ins []reflect.Type
+	var outs []reflect.Type
+
+	annotatedIn := false
+	annotatedOut := false
```
abhinav, line 291: optional stylistic nit: `var` group please.

```
var (
  annotations annoatations
  ins, outs []reflect.Type
  annotatedIn, annotatedOut bool
)
```

annotated.go
```diff
@@ -89,3 +90,268 @@ func (a Annotated) String() string {
...
+
+	for _, ann := range anns {
+		if e := ann.apply(&annotations); e != nil {
+			return annotationError{
+				target: f,
+				err:    e,
+			}
+		}
+		if pTags, ok := ann.(paramTags); ok {
```
abhinav, line 300: style nit: type switch might be more readable here.

```
switch ann := ann.(type) {
case paramTags:
  ann.getAnnotatedType(..)

case resultTags:
  ann.getAnnotatedType(..)

default:
  panic(...)
}
```

(I added a panic in this example because if we see an annotation that is not one of the two we support, there's a catastrophic bugs somewhere. There's a [blurb that Dig uses for panics](https://github.com/uber-go/dig/blob/b79edae8e13848e8ca4c84f54f8be680d5ba7605/result.go#L167-L172) like that that inform the user that this is a bug and that they should file a bug report. We can use something like that here.)

annotated.go
```diff
@@ -89,3 +90,268 @@ func (a Annotated) String() string {
...
+			ins = pTags.getAnnotatedType(fType)
+			annotatedIn = true
+		}
+		if rTags, ok := ann.(resultTags); ok {
+			outs = rTags.getAnnotatedType(fType)
+			annotatedOut = true
+		}
```
abhinav, line 307: I may not fully grok this yet, but I think we have some redundancy in the Annotation interface here.
We have two methods:

- apply: fills an annotations struct with information similar to [how functional options usually work](https://github.com/uber-go/guide/blob/master/style.md#functional-options)
- getAnnotatedType: contains behavior based on that single option

I feel like it should be one or the other? We should either have just getAnnotatedType, and that holds the behavior, or we have apply fill the annotations struct above, and then we build the annotated type out of the information recorded in the annotations struct. As it stands, the annotations struct doesn't serve much purpose otherwise.

annotated.go
```diff
@@ -89,3 +90,268 @@ func (a Annotated) String() string {
...
+	if !annotatedOut {
+		outs = make([]reflect.Type, numOut)
+		for i := 0; i < fType.NumOut(); i++ {
+			outs[i] = fType.Out(i)
+		}
+	}
+
+	newF := func(args []reflect.Value) []reflect.Value {
+		var fParams []reflect.Value
+		var fResults []reflect.Value
```
abhinav, line 325:
```suggestion
		var fParams, fResults []reflect.Value
```

annotated.go
sywhang: Oh, I suppose I could make the resulttag be an empty string for an error. Yeah, this is the handle the exact case you have there, but I was thinking the user would then need to do:
```
fx.Annotate(func(...) (Thing, error, AnotherThing), fx.ResultTags(`name:"foo"`, `name:"bar"`))
```
instead, which is why I have `numAnnotated` instead of `i`. 

But I suppose the user could just pass in an empty string here. Hm.. I'm not sure what would be more intuitive thing for the user to do.

sywhang: it's been updated to expect an empty string for the error result :) Thanks for the feedback!

annotated.go
```diff
@@ -89,3 +90,263 @@ func (a Annotated) String() string {
 	}
 	return fmt.Sprintf("fx.Annotated{%v}", strings.Join(fields, ", "))
 }
+
+// annotations is used for building out the info needed to generate struct
+// with tags using reflection.
+type annotations struct {
```
abhinav, line 96: minor: This is now an internal representation of the annotated func. Consider renaming this to `annotatedFunc` or similar.

annotated.go
```diff
@@ -89,3 +90,263 @@ func (a Annotated) String() string {
...
+	numOut := fType.NumOut()
+	annotations := newAnnotations(fType)
+
+	for _, ann := range anns {
+		if e := ann.apply(&annotations); e != nil {
+			return annotationError{
+				target: f,
+				err:    e,
+			}
+		}
```
abhinav, line 317: nit: prefer using `err` for error variables unless there's another error in context that needs to remain unchanged.
```suggestion
		if err := ann.apply(&annotations); err != nil {
			return annotationError{target: f, err: err}
		}
```

annotated.go
```diff
@@ -89,3 +90,263 @@ func (a Annotated) String() string {
...
+
+	ins := annotations.Ins
+	outs := annotations.Outs
+	newF := func(args []reflect.Value) []reflect.Value {
+		var fParams, fResults []reflect.Value
+		if annotations.annotatedIn {
+			fParams = make([]reflect.Value, numIn)
+			params := args[0]
+			for i := 0; i < numIn; i++ {
+				fParams[i] = params.Field(i + 1)
```
abhinav, line 328: it might be worth adding a comment about the `+1`

```suggestion
				fParams[i] = params.Field(i + 1) // + 1 to skip fx.In
```

annotated.go
```diff
@@ -89,3 +90,263 @@ func (a Annotated) String() string {
...
+		fResults = fVal.Call(fParams)
+		if annotations.annotatedOut {
+			// wrap the result in an annotated struct
+			results := reflect.New(outs[0]).Elem()
+			for i := 0; i < numOut; i++ {
+				if fResults[i].Type() == _typeOfError {
+					continue
+				}
+				results.FieldByName(fmt.Sprintf("Field%d",
+					i)).Set(fResults[i])
```
abhinav, line 342: This seems a bit iffy in that we're using a made-up field name to resolve this. Can we do fields by offset, possibly (with [Value.Field](https://pkg.go.dev/reflect#Value.Field))?

Since we skip errors, one way to do that would be to track a list of indexes `fields []int` of indexes in the generated `fx.Out` struct corresponding to the result value. That is, for each `fResults[i]`, `fields[i]` is the index of the field in `results` that we want to set.

    results.Field(fields[i]).Set(fResults[i])

annotated.go
```diff
@@ -89,3 +90,263 @@ func (a Annotated) String() string {
...
+		Name:      "Out",
+		Type:      reflect.TypeOf(Out{}),
+		Anonymous: true,
+	}}
+
+	for i := 0; i < fType.NumOut(); i++ {
+		// guard against error results
+		if fType.Out(i) == _typeOfError {
+			continue
+		}
```
abhinav, line 231: Oh, this seems slightly broken. We'll ignore errors completely:

```
fx.Annotate(func() (Foo, error), fx.ResultTags(`name:"foo"`))
```

will become

```
type whatever struct{
  fx.Out

  Field1 Foo `name:"foo"`
}

func() whatever
```

It will ignore the error.

What we want is,

```
func() (whatever, error)
```

annotated_test.go
```diff
@@ -176,3 +176,159 @@ func TestAnnotatedString(t *testing.T) {
...
+
+	t.Run("Provide with optional", func(t *testing.T) {
+		app := fxtest.New(t,
+			fx.Provide(
+				fx.Annotate(newB, fx.ParamTags(`name:"a" optional:"true"`)),
+			),
+			fx.Invoke(newC),
+		)
+		defer app.RequireStart().RequireStop()
+		require.NoError(t, app.Err())
```
abhinav, line 200: If this is a failure case, we should assert something about the error message.
It's common to say something like the following in a test:

```
assert.Contains(t, err.Error(), "subset of the message")
```

abhinav (commented):
Thanks for the changes! This looks better but I didn't realize before that we were silently ignoring errors.

annotated.go
```diff
@@ -89,3 +91,296 @@ func (a Annotated) String() string {
...
+}
+
+// Annotation can be passed to Annotate(f interface{}, anns ...Annotation)
+// for annotating the parameter and result types of a function.
+type Annotation interface {
+	apply(*annotations) error
+}
+
+var _typeOfError reflect.Type = reflect.TypeOf((*error)(nil)).Elem()
+var _nilError = reflect.Zero(_typeOfError)
```
abhinav, line 133: nit:

```suggestion
var (
	_typeOfError = reflect.TypeOf((*error)(nil)).Elem()
	_nilError    = reflect.Zero(_typeOfError)
)
```

annotated.go
```diff
@@ -89,3 +91,296 @@ func (a Annotated) String() string {
...
+//
+//   struct {
+//     fx.Out
+//
+//     Field1 T1 `$tags[0]`
+//     Field2 T2 `$tags[1]`
+//     ...
+//     FieldN TN `$tags[N-1]`
+//   }
+func (rt resultTagsAnnotation) apply(ann *annotations) error {
```
abhinav, line 216: might be worth documenting what happens with error returns here

annotated.go
```diff
@@ -89,3 +91,296 @@ func (a Annotated) String() string {
...
+		annotatedResult = append(annotatedResult, structField)
+	}
+	if returnsError {
+		ann.Outs = []reflect.Type{
+			reflect.StructOf(annotatedResult),
+			_typeOfError,
+		}
+	} else {
+		ann.Outs = []reflect.Type{reflect.StructOf(annotatedResult)}
+	}
```
abhinav, line 251: optional simplification:

```suggestion
	ann.Outs = []reflect.Type{reflect.StructOf(annotatedResult)}
	if returnsError {
		ann.Outs = append(ann.Outs, _typeOfError)
	}
```

annotated.go
```diff
@@ -89,3 +91,296 @@ func (a Annotated) String() string {
...
+//     fx.Out
+//
+//     Field1 T1 `$tags[0]`
+//     Field2 T2 `$tags[1]`
+//     ...
+//     FieldN TN `$tags[N-1]`
+//   }
+func (rt resultTagsAnnotation) apply(ann *annotations) error {
+	if ann.annotatedOut {
+		return fmt.Errorf("cannot apply more than one line of ResultTags")
```
abhinav, line 218: nit: If there are no formatting arguments, we can use errors.New instead of fmt.Errorf.
```suggestion
		return errors.New("cannot apply more than one line of ResultTags")
```

annotated.go
```diff
@@ -89,3 +91,296 @@ func (a Annotated) String() string {
...
+//
+//    GW *Gateway `name:"foo"`
+//   }
+//
+//   fx.Provide(func(p params) result {
+//     return result{GW: NewGateway(p.RO, p.RW)}
+//   })
+//
+// If a single annotation appears multiple times, all annotations
+// will be ignored.
```
abhinav, line 295: No longer accurate. We now detect and fail on this with an error, so we can just say "Using the same annotation multiple times is invalid. For example, the following will fail with an error."

annotated.go
```diff
@@ -89,3 +91,296 @@ func (a Annotated) String() string {
...
+//
+// For example,
+//
+//  fx.Provide(
+//    fx.Annotate(
+//      NewGateWay,
+//      fx.ParamTags(`name:"ro" optional:"true"`),
+//      fx.ParamTags(`name:"rw"),
```
abhinav, line 303: Worth highlighting that this is bad:
```suggestion
//      fx.ParamTags(`name:"rw"),  // ERROR: ParamTags was already used above
```

annotated.go
```diff
@@ -89,3 +91,296 @@ func (a Annotated) String() string {
...
+			fParams = make([]reflect.Value, numIn)
+			params := args[0]
+			for i := 0; i < numIn; i++ {
+				fParams[i] = params.Field(i + 1)
+			}
+		} else {
+			fParams = args
+		}
+		fResults = fVal.Call(fParams)
+		if annotations.annotatedOut {
```
abhinav, line 344: minor styling nit: let's reduce nesting here by returning early if annotatedOut is false:

```
if !annotations.annotatedOut { return fResults }

var errValue ...
```

https://github.com/uber-go/guide/blob/master/style.md#reduce-nesting

annotated.go
```diff
@@ -89,3 +91,296 @@ func (a Annotated) String() string {
...
+				errValue = reflect.New(outs[1]).Elem()
+			}
+
+			var errResults error
+			for i := 0; i < numOut; i++ {
+				if fResults[i].Type() == _typeOfError {
+					var errResult error
+					if !fResults[i].IsNil() {
+						errResult = fResults[i].Interface().(error)
+						errResults = multierr.Append(errResults, errResult)
```
abhinav, line 362: nit: This `var` declaration appears premature since it's not needed until inside the if. 

Further, you can simplify this a little by using the `foo, ok := bar.(baz)` style cast, and ignoring the `ok` value. To explain that further, there are two ways of casting:

```
x := foo.(Bar) // panics if foo is not a Bar, or if it's nil

x, ok := foo.(Bar)  // returns false and sets x to zero value if foo is not a Bar or if it's nil
```

So you can do the following when you don't care about the `ok` and the zero-value check will suffice.

```
x, _ := foo.(Bar)
```

Which shortens the above to:

```
if err, _ := fResults[i].Interface().(error); err != nil {
    errResults = multierr.Append(errResults, err)
}
```

annotated.go
abhinav: This is still valid. I think if we can keep track on `annotations` of the `results` field index mapping and use `Field(int i)`, this would feel slightly more secure. That said, we can do it in a follow up because this should still work even if it feels a bit unsafe

abhinav (approved):
LGTM minus nits

annotated.go
sywhang: thanks, I will address this in a follow-up PR.

PR #759, opened 2021-07-12
Add events for lifecycle hook execution

sywhang:
This adds events for lifecycle hook execution. `LifecycleHookStart` and `LifecycleHookStop` was being used to fire events separately for `OnStart` and `OnStop` events previously. Since the recorded fields for both events aren't different, this change merges the two events into a single set of events, and adds another event so that an event is fired before and another is fired after a hook is executed.

Specifically, `LifecycleHookExecuting` is fired right before either an `OnStart` or `OnStop` hook is executed. After the hook has finished executing, `LifecycleHookExecuted` is fired with either an error or the runtime. Both `LifecycleHookExecuting` and `LifecycleHookExecuted` events contain the name of the hook being executed, the name of the caller that appended the lifecycle, and the method that's being executed (either `OnStart` or `OnStop`).

Refs: GO-707.

### DIFF
--- a/fxevent/console.go
+++ b/fxevent/console.go
@@ -49,9 +49,9 @@ func (l *ConsoleLogger) LogEvent(event Event) {
 		l.logf("HOOK %s\t\t%s executing (caller: %s)", e.Method, e.FunctionName, e.CallerName)
 	case *LifecycleHookExecuted:
 		if e.Err != nil {
-			l.logf("HOOK %s\t\t%s called by %s failed: %v", e.Method, e.FunctionName, e.CallerName, e.Err)
+			l.logf("HOOK %s\t\t%s called by %s failed in %s: %v", e.Method, e.FunctionName, e.CallerName, e.Runtime, e.Err)
 		} else {
-			l.logf("HOOK %s\t\t%s called by %s ran successfully. Runtime: %s", e.Method, e.FunctionName, e.CallerName, e.Runtime)
+			l.logf("HOOK %s\t\t%s called by %s ran successfully in %s", e.Method, e.FunctionName, e.CallerName, e.Runtime)
 		}
 	case *ProvideError:
 		l.logf("Error after options were applied: %v", e.Err)
--- a/fxevent/console.go
+++ b/fxevent/console.go
@@ -45,10 +45,14 @@ func (l *ConsoleLogger) logf(msg string, args ...interface{}) {
 // LogEvent logs the given event to the provided Zap logger.
 func (l *ConsoleLogger) LogEvent(event Event) {
 	switch e := event.(type) {
-	case *LifecycleHookStart:
-		l.logf("START\t\t%s", e.CallerName)
-	case *LifecycleHookStop:
-		l.logf("STOP\t\t%s", e.CallerName)
+	case *LifecycleHookExecuting:
+		l.logf("HOOK %s\t\t%s executing (caller: %s)", e.Method, e.FunctionName, e.CallerName)
+	case *LifecycleHookExecuted:
+		if e.Err != nil {
+			l.logf("HOOK %s\t\t%s called by %s failed: %v", e.Method, e.FunctionName, e.CallerName, e.Err)
+		} else {
+			l.logf("HOOK %s\t\t%s called by %s ran successfully. Runtime: %s", e.Method, e.FunctionName, e.CallerName, e.Runtime)
+		}
 	case *ProvideError:
 		l.logf("Error after options were applied: %v", e.Err)
 	case *Supply:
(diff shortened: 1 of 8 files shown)


internal/lifecycle/lifecycle.go
```diff
@@ -121,14 +149,29 @@ func (l *Lifecycle) Stop(ctx context.Context) error {
...
 			Func:        hook.OnStop,
 			Runtime:     time.Since(begin),
 		})
+
+		l.logger.LogEvent(&fxevent.LifecycleHookExecuted{
+			FunctionName: funcName,
+			CallerName:   hook.callerFrame.Function,
+			Method:       _hookStop,
+			Runtime:      runtime,
+		})
```
shirchen, line 173: any reason why the fields are in different order when you call this on 110?

internal/lifecycle/lifecycle.go
```diff
@@ -121,14 +149,29 @@ func (l *Lifecycle) Stop(ctx context.Context) error {
...
 		}
 		l.mu.Lock()
+		runtime := time.Since(begin)
 		l.stopRecords = append(l.stopRecords, HookRecord{
 			CallerFrame: hook.callerFrame,
 			Func:        hook.OnStop,
 			Runtime:     time.Since(begin),
 		})
+
+		l.logger.LogEvent(&fxevent.LifecycleHookExecuted{
```
shirchen, line 168: isn't it a bit confusing if you get two similar events, one on 152 and almost exact same one here w/out the Err but with timing into which you can get on 152 anyway? difference here from `OnStart` is that there we return the error right away but here we accumulate.

shirchen (approved):

fxevent/console.go
```diff
@@ -45,10 +45,14 @@ func (l *ConsoleLogger) logf(msg string, args ...interface{}) {
...
 	switch e := event.(type) {
-	case *LifecycleHookStart:
-		l.logf("START\t\t%s", e.CallerName)
-	case *LifecycleHookStop:
-		l.logf("STOP\t\t%s", e.CallerName)
+	case *LifecycleHookExecuting:
+		l.logf("HOOK %s\t\t%s executing (caller: %s)", e.Method, e.FunctionName, e.CallerName)
+	case *LifecycleHookExecuted:
+		if e.Err != nil {
+			l.logf("HOOK %s\t\t%s called by %s failed: %v", e.Method, e.FunctionName, e.CallerName, e.Err)
```
abhinav, line 52: should we include how long we took to fail here?

```suggestion
			l.logf("HOOK %s\t\t%s called by %s failed in %v: %v", e.Method, e.FunctionName, e.CallerName, e.Runtime, e.Err)
```

fxevent/console.go
```diff
@@ -45,10 +45,14 @@ func (l *ConsoleLogger) logf(msg string, args ...interface{}) {
...
+		} else {
+			l.logf("HOOK %s\t\t%s called by %s ran successfully. Runtime: %s", e.Method, e.FunctionName, e.CallerName, e.Runtime)
```
abhinav, line 54: optional nit: can do this in one sentence
```suggestion
			l.logf("HOOK %s\t\t%s called by %s ran successfully in %s", e.Method, e.FunctionName, e.CallerName, e.Runtime)
```

fxevent/event.go
```diff
@@ -30,30 +31,39 @@ type Event interface {
...
+func (*CustomLogger) event()           {}
+
+// LifecycleHookExecuting is emitted before an OnStart hook is about to be executed.
+type LifecycleHookExecuting struct {
+	// FunctionName is the name of the hook being executed.
+	FunctionName string
+	// CallerName is the name of the caller that appended the hook.
 	CallerName string
+	// Method is the lifecycle hook method getting called.
+	Method string
```
abhinav, line 57: Would it be overkill to introduce a LifecycleHookType enum here and use that with a Type field?

```
type LifecycleHookType int

const (
  LifecycleHookOnStart LifecycleHookType = iota + 1
  LifecycleHookOnStop
)

// We' can hand-write it, or use stringer or enumer to generate a String method
// for the enum. The trimprefix flag will be needed to make it be only "OnStop" or "OnStart".

type LifecycleHookExecuting struct {
  // ...
  Type LifecycleHookType
}
```

abhinav (commented):
Nice. Some nits, and an optional "upgrade".

fxevent/event.go
sywhang: I'll actually do this as a follow-up PR because it looks like we could use this enum in lifecycle implementation as well. Thanks for the suggestion!

Issue #758, opened 2021-07-10
data race detected when testing in master branch


sywhang:
Sometimes, `make test` in master branch fails with the following error: 

```
==================
WARNING: DATA RACE
Read at 0x00c0005024c3 by goroutine 17:
  testing.(*common).logDepth()
      /usr/local/go/src/testing/testing.go:754 +0x74
  testing.(*common).log()
      /usr/local/go/src/testing/testing.go:747 +0x6c
  testing.(*common).Logf()
      /usr/local/go/src/testing/testing.go:793 +0x1c
  testing.(*T).Logf()
      <autogenerated>:1 +0x5c
  go.uber.org/fx/internal/testutil.WriteSyncer.Write()
      /Users/sungyoonwhang/go/src/github.com/uber-go/fx/internal/testutil/writer.go:40 +0xf4
  go.uber.org/fx/internal/testutil.(*WriteSyncer).Write()
      <autogenerated>:1 +0x24
  fmt.Fprintf()
      /usr/local/go/src/fmt/print.go:205 +0x9c
  go.uber.org/fx/fxevent.(*ConsoleLogger).logf()
      /Users/sungyoonwhang/go/src/github.com/uber-go/fx/fxevent/console.go:42 +0xf54
  go.uber.org/fx/fxevent.(*ConsoleLogger).LogEvent()
      /Users/sungyoonwhang/go/src/github.com/uber-go/fx/fxevent/console.go:74 +0xe84
  go.uber.org/fx.(*App).start()
      /Users/sungyoonwhang/go/src/github.com/uber-go/fx/app.go:890 +0x11c
  go.uber.org/fx.(*App).start-fm()
      /Users/sungyoonwhang/go/src/github.com/uber-go/fx/app.go:881 +0x48
  go.uber.org/fx.withTimeout.func1()
      /Users/sungyoonwhang/go/src/github.com/uber-go/fx/app.go:913 +0x50

Previous write at 0x00c0005024c3 by goroutine 13:
  testing.tRunner.func1()
      /usr/local/go/src/testing/testing.go:1181 +0x36c
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1198 +0x190

Goroutine 17 (running) created at:
  go.uber.org/fx.withTimeout()
      /Users/sungyoonwhang/go/src/github.com/uber-go/fx/app.go:913 +0x74
  go.uber.org/fx.(*App).Start()
[...]

PR #760, opened 2021-07-13
Fix races in test

sywhang:
The goroutine that executes lifecycle hooks may leak by design. With the recent changes in fxevent Logger and lifecycle execution hooks logging events, it's possible for that goroutine to then race with the main thread because they both touch testing.T. 

This fixes the issue by changing the tests that may cause a race pass `fxlog.Spy` with `WithLogger` - this makes the lifecycle goroutine to write to `fxlog.Spy` instead of testing.T's log buffers which races with the main goroutine. 

Refs GO-714

### DIFF
--- a/app.go
+++ b/app.go
@@ -910,15 +910,7 @@ type withTimeoutParams struct {
 
 func withTimeout(ctx context.Context, param *withTimeoutParams) error {
 	c := make(chan error, 1)
-	var wg sync.WaitGroup
-	// Wait for lifecycle goroutine to exit before returning.
-	defer wg.Wait()
-
-	wg.Add(1)
-	go func() {
-		c <- param.callback(ctx)
-		wg.Done()
-	}()
+	go func() { c <- param.callback(ctx) }()
 
 	var err error
 
--- a/app.go
+++ b/app.go
@@ -911,6 +911,7 @@ type withTimeoutParams struct {
 func withTimeout(ctx context.Context, param *withTimeoutParams) error {
 	c := make(chan error, 1)
 	var wg sync.WaitGroup
+	// Wait for lifecycle goroutine to exit before returning.
 	defer wg.Wait()
 
 	wg.Add(1)
@@ -927,7 +928,6 @@ func withTimeout(ctx context.Context, param *withTimeoutParams) error {
 	case err = <-c:
 	}
 	if err != context.DeadlineExceeded {
-		// Wait for lifecycle goroutine to exit before returning.
 		return err
 	}
 	// On timeout, report running hook's caller and recorded
--- a/app.go
+++ b/app.go
@@ -913,8 +913,8 @@ func withTimeout(ctx context.Context, param *withTimeoutParams) error {
 	var wg sync.WaitGroup
 	defer wg.Wait()
 
+	wg.Add(1)
 	go func() {
-		wg.Add(1)
 		c <- param.callback(ctx)
 		wg.Done()
 	}()
--- a/app.go
+++ b/app.go
@@ -910,10 +910,13 @@ type withTimeoutParams struct {
 
 func withTimeout(ctx context.Context, param *withTimeoutParams) error {
 	c := make(chan error, 1)
-	done := make(chan struct{}, 1)
+	var wg sync.WaitGroup
+	defer wg.Wait()
+
 	go func() {
+		wg.Add(1)
 		c <- param.callback(ctx)
-		close(done)
+		wg.Done()
 	}()
 
 	var err error
@@ -925,7 +928,6 @@ func withTimeout(ctx context.Context, param *withTimeoutParams) error {
 	}
 	if err != context.DeadlineExceeded {
 		// Wait for lifecycle goroutine to exit before returning.
-		<-done
 		return err
 	}
 	// On timeout, report running hook's caller and recorded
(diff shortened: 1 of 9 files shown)


internal/lifecycle/lifecycle.go
```diff
@@ -72,26 +72,31 @@ func (l *Lifecycle) Append(hook Hook) {
...
-			l.mu.Unlock()
-
-			begin := time.Now()
-			if err := hook.OnStart(ctx); err != nil {
-				return err
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+			if hook.OnStart != nil {
```
tfrench-uber, line 79: You can reduce indentation:
```
if hook.OnStart == nil {
  continue
}

File: internal/lifecycle/lifecycle.go [hunk 2]
```diff
@@ -72,26 +72,31 @@ func (l *Lifecycle) Append(hook Hook) {
...
-
-			l.mu.Lock()
-			l.runningHook = hook
-			l.mu.Unlock()
-
-			begin := time.Now()
-			if err := hook.OnStart(ctx); err != nil {
-				return err
+		select {
+		case <-ctx.Done():
```
tfrench-uber, line 76: Since you have a return in your `case <-ctx.Done():`, you could git rid of `default:` and pull the rest out of the select

sywhang: So in that case, `select` will block till `ctx.Done()`, which is not what we want. We need these under `default:` so that when `ctx.Done()` hasn't yet happened, we keep executing the hooks. The select on `ctx.Done()` here is specifically for the case when `ctx` is cancelled/timed out while we're in the middle of executing the hooks, we can bail before we finish executing all the hooks instead of waiting for all of them.

tfrench-uber: Makes sense, thanks for clarifying!

app.go
```diff
@@ -910,7 +910,15 @@ type withTimeoutParams struct {
 
 func withTimeout(ctx context.Context, param *withTimeoutParams) error {
 	c := make(chan error, 1)
-	go func() { c <- param.callback(ctx) }()
+	var wg sync.WaitGroup
+	// Wait for lifecycle goroutine to exit before returning.
+	defer wg.Wait()
```
mh-park, line 915: A bit confused; shouldn't the `wg.Wait()` be at line 924 (after the `var err error` decl)?

If the ordering of the code execution is:

context is cancelled while code being executed in `internal/lifecycle/lifecycle.go:79-90`, then `app.go:937` executes before `internal/lifecycle/lifecycle.go:90-96`, we'd be missing records, right?

Shouldn't we wait for the execution of callback entirely before we start executing past `app.go:933`?

mh-park (commented):
Not sure if this is possible, but could we somehow add a test case for this? It's a flaky one so I suspect it won't be easy to do, but maybe at the least, executing the race-detecting test multiple times?

app.go
sywhang: So yes - in the case you described, it's possible that the record for the hook that was executing when the context timed out is missing. But we don't need the record of that hook because records are specifically for hooks that successfully till the end before a timeout happened. If timeout happened in the middle of executing some hook, then we still report that hook in L953, so I think we should be OK here.

sywhang:
> Not sure if this is possible, but could we somehow add a test case for this? It's a flaky one so I suspect it won't be easy to do, but maybe at the least, executing the race-detecting test multiple times?

Yeah this issue was caught by an existing test (see #758) - I tested the fix by running the test 10000 times overnight and they all passed. I'm not sure how I can add that into an automated test but... reproing behaviors like this is a quite hard w/o modification of the actual product code.

abhinav (commented):
I'm not sure this is desirable. The reason withTimeout works like this is that it shouldn't block if the callback doesn't respect the context. For example, if callback is `func(context.Context) { select{} }}`, it will block forever. That shouldn't block withTimeout; it should stop after the timeout passes.

abhinav:
I've only taken a cursory look at this, but from the trace in #758, it looks like the issue is at:

- https://cs.opensource.google/go/go/+/refs/tags/go1.16.6:src/testing/testing.go;l=1180;drc=refs%2Ftags%2Fgo1.16.6
- https://cs.opensource.google/go/go/+/refs/tags/go1.16.6:src/testing/testing.go;l=1180;drc=refs%2Ftags%2Fgo1.16.6

Which indicates that we're `t.Log`ging to the `testing.T` after the test finished, which makes sense; the timed-out goroutine has something to say, but the writer we gave it is no longer valid.

So I think maybe the spirit of this change is correct, in that we want to wait for the background goroutine to finish, but we should probably do that in the test case rather than the withTimeout implementation, because withTimeout should be okay with leaking the goroutine by design.

sywhang:
@abhinav Hmm, that's true... Haven't thought about ill hooks. 

Then again, not syncing it makes this inherently race with not only whatever is in the hooks, but other things we touch inside that loop which include things like the log (that's where the race happens in the test failure).

Let me think again about what might be the best option for us.

sywhang:
> So I think maybe the spirit of this change is correct, in that we want to wait for the background goroutine to finish, but we should probably do that in the test case rather than the withTimeout implementation, because withTimeout should be okay with leaking the goroutine by design.

OK. I'll do that instead then.

abhinav (approved):

PR #761, opened 2021-07-21
Introduce Invoking and rename other event types

shirchen:
Follow up to #756 where we introduce a new `Invoking` event to be emitted before an `fx.Invoke`, rename `Invoke` to `Invoked`.

Additionally, `Provide` event will have access to constructor field in case it's emitted as a result of an error.

Ref GO-743

### DIFF
--- a/fxevent/console.go
+++ b/fxevent/console.go
@@ -62,13 +62,11 @@ func (l *ConsoleLogger) LogEvent(event Event) {
 		if e.Err != nil {
 			l.logf("Error after options were applied: %v", e.Err)
 		}
-	case *Invoke:
-		if e.Err != nil {
-			l.logf("fx.Invoke(%v) called from:\n%+vFailed: %v",
-				fxreflect.FuncName(e.Function), e.Stacktrace, e.Err)
-		} else {
-			l.logf("INVOKE\t\t%s", fxreflect.FuncName(e.Function))
-		}
+	case *Invoking:
+		l.logf("INVOKE\t\t%s", fxreflect.FuncName(e.Function))
+	case *Invoked:
+		l.logf("fx.Invoke(%v) called from:\n%+vFailed: %v",
+			fxreflect.FuncName(e.Function), e.Stacktrace, e.Err)
 	case *Stop:
 		if e.Err != nil {
 			l.logf("ERROR\t\tFailed to stop cleanly: %v", e.Err)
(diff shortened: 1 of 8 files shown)


fxevent/event.go
```diff
@@ -78,8 +79,13 @@ type Provide struct {
...
 }
 
-// Invoke is emitted whenever a function is being invoked and/or it errored.
-type Invoke struct {
+// Invoking is emitted whenever a function is being invoked.
+type Invoking struct {
+	Function interface{}
+}
+
+// Invoked is emitted whenever a function errored.
```
sywhang, line 87: Might sound a little ambiguous as it is. How about:
```suggestion
// Invoked is emitted whenever a function being invoked errored.
```

fxevent/console.go
```diff
@@ -87,11 +85,11 @@ func (l *ConsoleLogger) LogEvent(event Event) {
...
 			l.logf("RUNNING")
 		}
-	case *CustomLogger:
+	case *LoggerInitialized:
 		if e.Err != nil {
-			l.logf("ERROR\t\tFailed to construct custom logger: %v", e.Err)
+			l.logf("ERROR\t\tFailed to initialize custom logger: %v", e.Err)
 		} else {
-			l.logf("LOGGER\tSetting up custom logger from %v", fxreflect.FuncName(e.Function))
+			l.logf("LOGGER\tInitializing custom logger from %v", fxreflect.FuncName(e.Constructor))
```
sywhang, line 92:
```suggestion
			l.logf("LOGGER\tInitialized custom logger from %v", fxreflect.FuncName(e.Constructor))
```

fxevent/console_test.go
```diff
@@ -135,15 +135,15 @@ func TestConsoleLogger(t *testing.T) {
...
 		{
-			name: "CustomLogger",
-			give: &CustomLogger{
-				Function: func() Logger { panic("should not run") },
+			name: "LoggerInitialized",
+			give: &LoggerInitialized{
+				Constructor: func() Logger { panic("should not run") },
 			},
-			want: "[Fx] LOGGER	Setting up custom logger from go.uber.org/fx/fxevent.TestConsoleLogger.func1()\n",
+			want: "[Fx] LOGGER	Initializing custom logger from go.uber.org/fx/fxevent.TestConsoleLogger.func1()\n",
```
sywhang, line 146:
```suggestion
			want: "[Fx] LOGGER	Initialized custom logger from go.uber.org/fx/fxevent.TestConsoleLogger.func1()\n",
```

sywhang (approved):
few suggestions but LGTM otherwise!

fxevent/event.go
```diff
@@ -103,8 +109,8 @@ type Rollback struct {
 	Err      error
 }
 
-// CustomLogger is emitted whenever a custom logger is set or produces an error.
-type CustomLogger struct {
-	Function interface{}
-	Err      error
+// LoggerInitialized is emitted whenever a custom logger is set or produces an error.
```
sywhang, line 112: The latter bit of this comment sounds a little confusing since it sounds like it can be emitted when the custom logger produces an error (i.e. after initialization as well). Can we reword it such that it's emitted when an error is produced only during initialization of the custom logger?

sywhang (approved):
Is this the last bit of logger changes before we release the new version of fx? If so, it might be helpful to add package-level documentation on how the tenses of fxevent events relates to the timing of the events that are fired. Otherwise I think this is good to go 👍

sywhang (approved):
👍

fxevent/event.go
shirchen: Synced offline. Added constructor to case where we emit LoggerInitialized due to error. Today, it's emitted in either case as documented.

PR #765, opened 2021-08-04
fxevent: Rename, split, and simplify

abhinav:
This PR includes several individually reviewable commits. The changes
are summarized below, but first, I want to explain my strategy here:

I went over every event defined in fxevent, and validated the tense for
its name for all its emissions outside the fxevent package.
Specifically,

- The tense of verbs inside names should accurately reflect when those
  events occur. Names for events using present progressive tense
  ("starting") should emit before the event occurs, and those using past
  tense ("started") should emit after the event. Prior PRs already took
  care of most of this, but they missed a couple spots. This PR further
  splits the following events:
    - Stop into Stopping and Stopped
    - Rollback into RollingBack and RolledBack
- Events should not include raw references into objects owned by the
  containers. We were previously supplying raw constructors, functions,
  etc. as `interface{}` objects to event loggers. For safety, these have
  all been switched to string variants that only include the name of the
  function--which is all those loggers were extracting out of those
  `interface{}` objects in the first place.
- Past tense events should emit for both, success and failure where
  applicable.
    - The following events were previously emitted only for failure
      cases: RolledBack, Stopped, Invoked.
    - The following events were previously emitted only for success
      cases: Supplied.
- I didn't like that the `LifecycleHook*` events had a `Mode string`
  field. I considered making that an enum, but decided that it was
  cleaner to split them into `OnStart*` and `OnStop*` events, since
  they can have different semantics.
- An event type should be instantiated and emitted exactly once in
  non-test code. This is a maintainability requirement: it ensures that
  if we add new fields to events, we don't forget to update all emission
  sites, because there's exactly one. This required minor refactoring in
  a few places.


shirchen (commented):
thanks!

shirchen (approved):
thanks!

app.go
```diff
@@ -648,7 +650,7 @@ func VisualizeError(err error) (string, error) {
 // Start, Done, and Stop methods. Applications with more specialized needs
 // can use those methods directly instead of relying on Run.
 func (app *App) Run() {
-	app.run(app.Done())
+	os.Exit(app.run(app.Done()))
```
sywhang, line 653: Looks like when the exit code is 0, we're changing the behavior? i.e. we didn't `os.Exit` before with exit code 0 ever, but it would now. If someone relied on call to this returning to the caller, wouldn't that break them?

I suppose the docs do say apps with more specialized needs should use `Start`, `Done` and `Stop` methods but now that we're changing that behavior, would this be considered a breaking change?

sywhang (approved):
LGTM but had a question on compatibility on the behavior of `Run`.

app.go
```diff
@@ -835,53 +838,63 @@ func (app *App) executeInvokes() error {
...
+		return 1
 	}
+	return 0
 }
 
-func (app *App) start(ctx context.Context) error {
+func (app *App) start(ctx context.Context) (err error) {
+	defer func() {
+		app.log.LogEvent(&fxevent.Started{Err: err})
+	}()
```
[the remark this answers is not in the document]
abhinav: No, because when this `defer` runs, `err` will always refer to the error being returned.

Example: https://play.golang.org/p/YjlkSzp_9CU

app.go
abhinav: Ah, good catch. Typically, people who use app.Run are ceding control of the application, but I'd like to not change the behavior here in case someone is relying on this returning at some point. Fixing.

PR #774, opened 2021-08-16
Change when we log `fxevent.Invoking` and `fxevent.Invoked` events

shirchen:
For `fxevent` console implementation: no longer log non-error case for `fxevent.Invoke` event, while for zap implementation, start logging `fx.Invoking` case without stack.

### DIFF
--- a/fxevent/zap.go
+++ b/fxevent/zap.go
@@ -88,16 +88,13 @@ func (l *ZapLogger) LogEvent(event Event) {
 				zap.Error(e.Err))
 		}
 	case *Invoking:
-		// Do nothing. Will log on Invoked.
-
+		// Do not log stack as it will make logs hard to read.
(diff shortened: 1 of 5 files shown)


sywhang (approved):
👍

fxevent/zap.go
```diff
@@ -88,18 +88,15 @@ func (l *ZapLogger) LogEvent(event Event) {
 				zap.Error(e.Err))
 		}
 	case *Invoking:
-		// Do nothing. Will log on Invoked.
-
+		// Do not log stack as it will make logs hard to read.
+		l.Logger.Info("invoked",
+			zap.String("function", e.FunctionName))
```
abhinav, line 93: invoking

abhinav (approved):
👍 

this and the previous change are bugfixes; patch release works for me

PR #776, opened 2021-08-16
CHANGELOG: Reorganize reference links

abhinav:
We were previously placing reference links for changes in a version at
the bottom of the file.

Besides bloating up the bottom of the file, this makes it easy to forget
to add these references on new releases.

Move the reference links to the section for each release so that it's
easy to pattern match on this next release.


sywhang (approved):
LGTM 👍

shirchen (approved):

PR #783, opened 2021-09-04
Remove dependency on parent folder name from error tests

alexykot:
I happened to check out this repo locally into a folder named differently to git's default (`FX`, not `fx`), and unit tests were failing for me.

Turns out - some of the error-checking negative tests contain hardcoded references to the name of the parent folder, and it's hardcoded to `fx/`. Checking error text output makes sense in this context, but referencing the parent folder name does not make sense to me, it's not a part of the codebase, and has no effect on the code validity.

This PR removes those references and removes checks of the parent folder name entirely. 

The other possible option I reasoned about was to read the name of the parent folder and use that in the test assertions, but that seems like an overkill to me and unnecessary magic, it doesn't really add any value to asserting code behaviour. So I discarded that option.


abhinav (approved):
Thanks for the fix! The reliance on the parent folder name is likely a leftover from the Before Go Modules Times when the repository was required to be at `$GOPATH/src/go.uber.org/fx`. With Go modules, that's no longer a valid assumption.

Issue #762, opened 2021-07-24
Allow to unit test and detect an attempt to `fx.Populate` non-pointer types


alexykot:
We're using `fx` for wireup of both long-living servers and short-living CLI commands, and using `fx.Populate` in the wireup of CLI commands to extract and use CLI dependencies. 

We're also using `fx.ValidateApp` to unit test the completeness of the wireup, and it works pretty well in most situations. It does not work for one specific problem though - when the targets provided to `fx.Populate` are unsuitable for populating (e.g. non-pointer objects).

Consider this piece of wireup:
```
func cliDepOptions(app *microservice.App, dep ...interface{}) fx.Option {
	return fx.Options(
		coreOptions(app),
		fx.Populate(dep),
	)
}
```
This produces a set of core options that are needed for any CLI command, and it populates the supplied list of dependencies. There may be different dependencies for different CLI commands, so we use variadic `...interface{}` for it. 

This wireup is subtly wrong, and this cannot be detected by compiler or the unit tests using `fx.ValidateApp`. `fx.Populate(dep)` should really be `fx.Populate(dep...)`. Both will compile, but the first version effectively supplies a single `[]interface{}` object, and that one cannot be populated.

This situation cannot be unit tested because of the interplay between how the `fx.Populate` is designed and how `dig.DryRun` option works. `fx.Populate()` returns an `fx.Option`, and in case of a failure this option is an invoker that takes no arguments and simply returns the detected error, it looks like this - `Invoke(func() error { return err })`. 

However, in a unit test, with `dig.DryRun` option enabled, no invokers are triggered so the `fx.New` does not receive the error produced by `fx.Populate` and this mistake goes unnoticed.

As a result - humans writing the CLI wireup can make a small mistake that will go unnoticed by both the compiler and the unit tests, and will only be caught later in the integration testing or after deployment. 

I'm not sure what would be the best approach here, whether it's an `fx` or `dig` issue, but would be good to fix this and allow these situations to be unit tested.

alexykot:
Seems like the solution for this would be not that complicated. 

In case of an error the `Populate()` call should return an `errorOption` instead of `invokeOption`. That will allow the app to return an error even if `dig` container is in `dryRun` mode and all dependencies are met.

I would be happy to make a PR for this if this approach seems reasonable.

abhinav:
Resolved in #784. Thanks for the fix, @alexykot!

PR #784, opened 2021-09-04
Make Populate return errorOption allowing ValidateApp to recognise wireup problems

alexykot:
This PR solves issue #762 that I've reported earlier. 

This slightly changes the internal behaviour of `Populate`, instead of returning a dedicated custom `invokeErr` option it returns a standard `errorOption`. This allows the `ValidateApp()` helper to recognise wire-up issues that arise from incorrectly configured `Populate` options. This wasn't possible earlier because `ValidateApp` sets `DryRun=true` for the `dig` container, and that makes it ignore and not execute any invocations, so the `invokeErr` was never called and the related errors could not be discovered in the unit tests relying on `ValidateApp`.

Using `errorOption` sets the `App.err` explicitly on applying option (rather than on calling invocations), so it allows to surface errors produced by `Populate` in dry-run mode without calling the invocations.

This also makes the same change to the deprecated `Extract` function, since I'm not sure when/if it will be removed.

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -65,7 +65,7 @@ func NewSpied(opts ...Option) (*App, *fxlog.Spy) {
 	return New(opts...), spy
 }
 
-func NewValidateAppErr(tb testing.TB, opts ...Option) error {
+func validateTestApp(tb testing.TB, opts ...Option) error {
 	testOpts := []Option{
 		// Provide both: Logger and WithLogger so that if the test
 		// WithLogger fails, we don't pollute stderr.
--- a/app_test.go
+++ b/app_test.go
@@ -65,6 +65,18 @@ func NewSpied(opts ...Option) (*App, *fxlog.Spy) {
 	return New(opts...), spy
 }
 
+func NewValidateAppErr(tb testing.TB, opts ...Option) error {
+	testOpts := []Option{
+		// Provide both: Logger and WithLogger so that if the test
+		// WithLogger fails, we don't pollute stderr.
+		Logger(fxtest.NewTestPrinter(tb)),
+		WithLogger(func() fxevent.Logger { return fxtest.NewTestLogger(tb) }),
+	}
+	opts = append(testOpts, opts...)
+
+	return ValidateApp(opts...)
+}
+
 func TestNewApp(t *testing.T) {
 	t.Run("ProvidesLifecycleAndShutdowner", func(t *testing.T) {
 		var (
--- a/extract_test.go
+++ b/extract_test.go
@@ -59,6 +59,27 @@ func TestExtract(t *testing.T) {
 		}
 	})
 
+	t.Run("ValidateApp", func(t *testing.T) {
+		tests := []interface{}{
+			3,
+			func() {},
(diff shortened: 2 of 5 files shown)


app_test.go
```diff
@@ -64,6 +64,18 @@ func NewSpied(opts ...Option) (*App, *fxlog.Spy) {
 	return New(opts...), spy
 }
 
+func NewValidateAppErr(tb testing.TB, opts ...Option) error {
```
abhinav, line 67: minor: would you please rename this to validateApp or validateTestApp, since the net effect of this is to validate the provided application (comprised of the options) with ValidateApp?

(If this returned the App, then the `New*` would work better, but the current behavior is preferable for the use case.)

abhinav (approved):
This looks great. Thanks for the fix!

CC @shirchen @sywhang

sywhang (approved):
LGTM - thank you for the contribution!

populate_test.go
```diff
@@ -244,3 +245,74 @@ func TestPopulateErrors(t *testing.T) {
...
+			wantErr: "target 2 is nil",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.msg, func(t *testing.T) {
+			testOpts := []Option{
+				NopLogger,
+				Provide(func() *t1 { return &t1{} }),
+
```
shirchen, line 309: nit: extra line?

extract_test.go
```diff
@@ -59,6 +59,27 @@ func TestExtract(t *testing.T) {
 		}
 	})
 
+	t.Run("ValidateApp", func(t *testing.T) {
+		tests := []interface{}{
```
shirchen, line 63: add t.Parallel if these can run in parallel.

populate_test.go
```diff
@@ -244,3 +245,74 @@ func TestPopulateErrors(t *testing.T) {
...
+				tt.opt,
```
shirchen, line 310: any reason for not taking a list of arguments to pass into `Populate` and pass in an `Option` instead? to me, it'll be clearer what's tested if you do 
```
Provide(....),
Populate(...)
```

shirchen (commented):
lgtm, with some nits. thanks for your contribution and thorough tests.

populate_test.go
alexykot: No specific reason, just copypasted from another test. Yeah, that will be clearer, fixed.

extract_test.go
alexykot: Doesn't look like it can immediately, I get data races with `TestWithLogger` when trying to add that. 
Didn't look deeper into how this could be resolved. Let me know if you think this is a blocker.

abhinav: Dug into this. Looks like there's a legitimate race. TestWithLogger is marked as Parallel, but one of its tests modifies `os.Stderr`, which is what causes the data race. Fixing.

shirchen: nope, we can do this separately!

shirchen (approved):

PR #785, opened 2021-09-08
Parallelize all tests

abhinav:
This adds `t.Parallel()` for all tests and subtests.

This change was made automatically with the following gopatch patch.

```diff
@@
var Test identifier
@@
 func Test(t *testing.T) {
+  t.Parallel()
   ...
 }

@@
var tt identifier
var tests expression
@@
 for _, tt := range tests {
+  tt := tt
   t.Run(..., func(t *testing.T) {
+     t.Parallel()
      ...
   })
 }

@@
@@
   t.Run(..., func(t *testing.T) {
+     t.Parallel()
      ...
   })

@@
@@
 t.Parallel()
-t.Parallel()
```

To keep style consistent, I added a newline after each `t.Parallel()` where
absent with the following commands.

```
$ git diff --name-only | xargs perl -p -i -e 's/(t.Parallel\(\))/$1\n/'
$ gofmt -s -w $(git diff --name-only)
```

Finally, I reverted or adjusted the two cases where we cannot run tests in
parallel:

- app_test/WithLogger: One of the subtests alters `os.Stderr` temporarily, so
  it cannot run in parallel with any other test. However, because the outer
  test is marked as parallel, it ends up running in parallel with other tests
  after all.
- fxlog/spy_test: The test's subtests specifically rely on sharing the object
  and cannot run in parallel with each other.

### DIFF
--- a/annotated_test.go
+++ b/annotated_test.go
@@ -31,6 +31,8 @@ import (
 )
 
 func TestAnnotated(t *testing.T) {
+	t.Parallel()
+
 	type a struct {
 		name string
 	}
@@ -43,6 +45,8 @@ func TestAnnotated(t *testing.T) {
 		return &a{name: "foo"}
 	}
 	t.Run("Provide", func(t *testing.T) {
+		t.Parallel()
+
 		var in in
 		app := fxtest.New(t,
 			fx.Provide(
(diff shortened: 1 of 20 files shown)


sywhang (approved):
LGTM - gopatch FTW! Do you want to add the patch you used to gopatch Wiki and maybe link this PR?

abhinav:
> LGTM - gopatch FTW! Do you want to add the patch you used to gopatch Wiki and maybe link this PR?

Yes https://github.com/uber-go/gopatch/wiki/In-the-Wild :)

PR #792, opened 2021-09-17
app: Add internal option to override os.Exit

abhinav:
Add an option exposed only to tests in the go.uber.org/fx package that
allows overriding how `os.Exit` behaves.

The way the new option works is,

- `App` gets a new `osExit func(int)` field which defaults to `os.Exit`
- app_internal_test.go uses `package fx` and therefore, it can access
  internal fields of the `App`. We define a new exported option in this
  package that sets `osExit`. Because this option is defined in a test
  file, though, it is not made part of the public API of the package.
- app_test.go uses `package fx_test` so it's an external test--it cannot
  access unexported fields, but it can access the exported option
  defined in app_internal_test.go.

This will allow testing the `app.Run` function directly.

### DIFF
--- a/app.go
+++ b/app.go
@@ -381,6 +381,8 @@ type App struct {
 	// Used to signal shutdowns.
 	donesMu sync.RWMutex
 	dones   []chan os.Signal
+
+	osExit func(code int) // os.Exit override; used for testing only
 }
 
 // provide is a single constructor provided to Fx.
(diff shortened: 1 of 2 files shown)


shirchen (approved):

sywhang (approved):
LGTM other than maybe adding a comment for the osExit field

PR #795, opened 2021-10-12
Implement fx.As Annotation

### LINKED ISSUE #800: Annotated structs cannot be used to resolve interfaces
Currently, FX offers some limitations when resolving interfaces to implementations. This brings complications when following the concept of accepting interfaces (and defining them on the client's side) and returning values.
Given two similar interfaces to be bound to the same struct, what I usually do to overcome this is to make a wrapper constructor to resolve structs to the interface.  Such as:
```go
package foo

type interface Performer{
    Perform()
}

type Foo{
    performer Performer
}

func NewFoo(p Performer) *Foo {
    return &Foo{performer: p}
}
```
```go
package bar

type interface Performer{
    Perform()
}

type Bar{
    performer Performer
}

func NewBar(p Performer) *Bar {
    return &Bar{performer: p}
}
```
```go
package performer

type struct Performer{}

func (Performer) Perform() {}

func NewPerformer() *Performer {
    return &Performer{}
}
```
And eventually in a base package:
```go
fx.Provide(
    performer.NewPerformer()
    func(p *performer.Performer) foo.Performer { return p }
    func(p *performer.Performer) bar.Performer { return p }
)
```
By doing this, I can effectively bind several interfaces to the same (or multiple) implementations. It's not that bad but has the drawback of having to redefine these constructors for every interface.
I've been using FX for a while doing this. But lately, I was reworking some of my providers and decided to try `fx.In`, `fx.Out`, and `fx.Annotaed`. 
It would be perfect to bind these interfaces based on their names. NestJS does something [similar](https://docs.nestjs.com/fundamentals/custom-providers#non-class-based-provider-tokens).
I'm not sure if the lack of support for this is a limitation or is by design, especially because of the fx.In docs, it says "Note that both the name AND type of the fields on the parameter struct must match the corresponding result struct.".

Anyway, would you be open to a PR to change this to allow named values to be injected based solely on the name and interface, and not on the actual type?

sywhang:
This implements `fx.As` Annotation, which lets you annotate a constructor's result types to be provided to the container as a different interface type. 

For example,
```go
fx.Provide(
  fx.Annotate(bytes.NewBuffer(...), fx.As(io.Writer))
)
```
Specifies that the return type of `bytes.NewBuffer` (`bytes.Buffer`) should be provided as `io.Writer` type. 

We leverage the existing reflection code for generating an fx.Out struct type whose field type is the annotated type. This lets us merge the code with `ResultTags` Annotation for cleaner implementation.

Resolves #800, #673

### DIFF
--- a/annotated.go
+++ b/annotated.go
@@ -487,7 +487,7 @@ func Annotate(f interface{}, anns ...Annotation) interface{} {
 		}
 		if annotations.returnsError {
 			if errResults != nil {
-				errValue = reflect.ValueOf(errResults).Elem()
+				errValue = reflect.ValueOf(errResults)
 				return []reflect.Value{results, errValue}
 			}
 			// error is nil. Return nil error Value.
--- a/annotated.go
+++ b/annotated.go
@@ -293,37 +293,6 @@ func (rt resultTagsAnnotation) apply(ann *annotations) error {
 	}
 	ann.annotatedOut = true
 	ann.outTags = rt.tags
-	/*
-		fType := ann.fType
-
-		annotatedResult := []reflect.StructField{{
-			Name:      "Out",
-			Type:      reflect.TypeOf(Out{}),
-			Anonymous: true,
-		}}
-
-		offsets := make([]int, fType.NumOut())
-		for i := 0; i < fType.NumOut(); i++ {
-			if fType.Out(i) == _typeOfError {
-				ann.returnsError = true
-				continue
-			}
-			structField := reflect.StructField{
-				Name: fmt.Sprintf("Field%d", i),
-				Type: fType.Out(i),
-			}
-			if i < len(rt.tags) {
-				structField.Tag = reflect.StructTag(rt.tags[i])
-			}
-			offsets[i] = len(annotatedResult)
-			annotatedResult = append(annotatedResult, structField)
-		}
-		ann.Outs = []reflect.Type{reflect.StructOf(annotatedResult)}
-		ann.resultOffsets = offsets
-		if ann.returnsError {
-			ann.Outs = append(ann.Outs, _typeOfError)
-		}
-	*/
 	return nil
 }
 
@@ -349,63 +318,7 @@ func (at asAnnotation) apply(ann *annotations) error {
 		return errors.New("cannot apply more than one line of As")
 	}
 	ann.annotatedAs = true
-	// Check if the annotations are of same interface.
 	ann.asTargets = at.targets
-
-	/*
-		fType := ann.fType
-
-		// Generate stub function for mapping in/out types.
-		asInputs := []reflect.StructField{{
-			Name:      "AsIn",
-			Type:      reflect.TypeOf(In{}),
-			Anonymous: true,
-		}}
-
-		asOutputs := []reflect.StructField{{
-			Name:      "AsOut",
-			Type:      reflect.TypeOf(Out{}),
-			Anonymous: true,
-		}}
-
-		numOut := fType.NumOut()
-		offsets := make([]int, numOut)
-		for i := 0; i < numOut; i++ {
-			asType := reflect.TypeOf(ann.asTargets[i]).Elem()
-
-			if fType.Out(i) == _typeOfError {
-				ann.returnsError = true
-				continue
-			}
-
-			if !fType.Out(i).Implements(asType) {
-				return fmt.Errorf("invalid fx.As: %v does not implement %v",
-					ann.fType,
-					asType)
-			}
-			inStructField := reflect.StructField{
-				Name: fmt.Sprintf("InField%d", i),
-				Type: fType.Out(i),
-			}
-			outStructField := reflect.StructField{
-				Name: fmt.Sprintf("OutField%d", i),
-				Type: asType,
-			}
-			if i < len(ann.outTags) {
-				outStructField.Tag = reflect.StructTag(ann.outTags[i])
-			}
-			asInputs = append(asInputs, inStructField)
-			outIdx := len(asOutputs)
-			offsets[i] = outIdx
-			asOutputs = append(asOutputs, outStructField)
-		}
-		ann.asIns = []reflect.Type{reflect.StructOf(asInputs)}
-		ann.asOuts = []reflect.Type{reflect.StructOf(asOutputs)}
-		ann.resultOffsets = offsets
-		if ann.returnsError {
-			ann.asOuts = append(ann.asOuts, _typeOfError)
-		}
-	*/
 	return nil
 }
 
--- a/annotated.go
+++ b/annotated.go
@@ -309,6 +309,57 @@ type asAnnotation struct {
 
 var _ asAnnotation = asAnnotation{}
 
+// As is an Annotation that annotates the result(s) of a function to be
+// provided as another interface.
+//
+// For example,
+//   fx.Provide(
+//     fx.Annotate(bytes.NewBuffer(...), fx.As(io.Writer))
+//   )
+//
+// specifies that the return type of bytes.NewBuffer (bytes.Buffer) should be
+// provided as io.Writer type.
+//
+// In other words, the code above is equivalent to:
+//
+//  type result struct {
+//    fx.Out
+//
+//    w io.Writer
+//  }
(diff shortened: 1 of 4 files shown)


annotated_test.go
```diff
@@ -65,6 +67,196 @@ func TestAnnotated(t *testing.T) {
...
+				}, fx.As(new(io.Writer))),
+			),
+			invoke: func(s fmt.Stringer, w io.Writer) {
+				assert.Equal(t, "stringer", s.String())
+				_, err := w.Write([]byte{1})
+				require.NoError(t, err)
+			},
+		},
+		{
+			desc: "provide the same provider as multiple types",
```
Multiply, line 166: Would it be possible to design this so it's possible to provide it as multiple interfaces in a si gle provide call, or does that go against the overall design?

We tend to write smaller interfaces, but sometimes larger implementations, which could implement 3-5 interfaces. We would probably end up making a wrapper for it, instead of manually providing the same constructor 5 times.

sywhang: @Multiply thanks for the feedback, and to answer your question, yes. 

A possible workaround for doing what you described is something like the following where `Buffer` implements `io.Writer` and `io.Reader`:
```go
func newBuffer() (*Buffer, *Buffer) {
  b := newBuffer()
  return b, b
}

fx.Annotate(newBuffer, fx.As(new(io.Writer), new(io.Reader)))
```

But that wrapper func is annoying and I agree it'd be nice to not have to do this; in fact, @abhinav and I discussed this earlier today and while it's out of scope for this PR, we're currently thinking it'd look something like:
```go
fx.Provide(
  fx.Annotate(newBuffer, fx.As(new(io.Writer)), fx.As(new(io.Reader)))
)
```

Currently if you try to annotate a constructor with multiple `As`, it'll error out. But we can change that to be allowed once we have the implementation detail sorted out in a near-term future release.

Multiply: That's also the design I had in mind when first looking at it, but didn't seem it was supported in this PR, as you just mentioned. 👍 

Feel free to tag me in the follow up PR, I'll be happy to test the functionality in our codebase, as this is one of the main feature requests from our team! 🚀

sywhang: will do :)

annotated.go
```diff
@@ -126,6 +130,66 @@ func newAnnotations(fType reflect.Type) annotations {
...
+}
+
+// This generates the annotated Out struct when function is annotated
+// with fx.As and/or fx.ResultTags options.
+func (ann *annotations) genAnnotatedOutStruct() error {
+	if !ann.annotatedOut && !ann.annotatedAs {
+		return nil
+	}
+	fType := ann.fType
+	offsets := make([]int, fType.NumOut())
```
abhinav, line 147: A comment here would be helpful:

```suggestion
	// offsets[i] is the index of this field in annotatedResult.
	// This will always be >0 when valid.
	offsets := make([]int, fType.NumOut())
```

annotated.go
```diff
@@ -126,6 +130,66 @@ func newAnnotations(fType reflect.Type) annotations {
...
+// helper for getting type of a fx.Out struct field
+func (ann *annotations) structFieldType(i int) (reflect.Type, error) {
+	if ann.annotatedAs {
+		asType := reflect.TypeOf(ann.asTargets[i]).Elem()
+		if !ann.fType.Out(i).Implements(asType) {
+			return nil, fmt.Errorf("invalid fx.As: %v does not implement %v", ann.fType, asType)
+		}
+		return asType, nil
+	}
+	return ann.fType.Out(i), nil
```
abhinav, line 182: nit: we should invert the if condition here to reduce nesting

```
if !ann.annotatedAs { return ann.fType.Out(i), nil }

asType := ...
```

https://github.com/uber-go/guide/blob/master/style.md#reduce-nesting

annotated.go
```diff
@@ -126,6 +130,66 @@ func newAnnotations(fType reflect.Type) annotations {
...
+}
+
+// helper for generating an fx.Out struct field
+func genAnnotatedOutStructField(i int, t reflect.Type) reflect.StructField {
```
abhinav, line 186: nit: I would inline this into genAnnotatedOutStruct since it's just a struct initialization.

annotated.go
```diff
@@ -126,6 +130,66 @@ func newAnnotations(fType reflect.Type) annotations {
...
+	if ann.returnsError {
+		ann.Outs = append(ann.Outs, _typeOfError)
+	}
+	return nil
+}
+
+// helper for getting type of a fx.Out struct field
+func (ann *annotations) structFieldType(i int) (reflect.Type, error) {
+	if ann.annotatedAs {
+		asType := reflect.TypeOf(ann.asTargets[i]).Elem()
```
abhinav, line 176: Should we turn the asTargets into the destination type in asAnnotation.apply?
That would make it clear that ann.asTargets is the actual target type.

annotated.go
```diff
@@ -335,6 +439,13 @@ func Annotate(f interface{}, anns ...Annotation) interface{} {
 		}
 	}
 
+	if e := annotations.genAnnotatedOutStruct(); e != nil {
+		return annotationError{
+			target: f,
+			err:    e,
```
abhinav, line 445: error variables are typically named err unless there's another err in scope that also has to be valid.
```suggestion
	if err := annotations.genAnnotatedOutStruct(); err != nil {
		return annotationError{
			target: f,
			err:    err,
```

annotated_test.go
```diff
@@ -65,6 +67,196 @@ func TestAnnotated(t *testing.T) {
 	})
 }
 
+type asStringer struct {
+	name string
+}
+
+func (as *asStringer) String() string {
+	return as.name
+}
```
abhinav, line 76: for one of the test cases, we should also test the case where a value type implements the interface instead of the pointer type.

```
type anotherStringer struct{ name string }

return (a anotherStringer) String() { return a.name }
```

annotated_test.go
```diff
@@ -65,6 +67,196 @@ func TestAnnotated(t *testing.T) {
...
+			},
+		},
+		{
+			desc: "provide as with result annotation",
+			provide: fx.Provide(
+				fx.Annotate(func() *asStringer {
+					return &asStringer{name: "stringer"}
+				},
+					fx.ResultTags(`name:"goodStringer"`),
+					fx.As(new(fmt.Stringer))),
```
abhinav, line 127: did we run gofmt here? indentation seems off.

abhinav (approved):
Looks reasonable.

I'm not a big fan of how we apply the annotations right now, but we can fix that in a follow up diff since that affects the other annotations too.

I'm not set on this, and may not have thought it fully, but the general shape I'm thinking that would work here is:

- Annotations that affect parameters come together and build a function or type with the following rough signature.

    ```go
    func(argTypes []reflect.Type) (
        newTypes []reflect.Type,
        convertArgs func([]reflect.Value) []reflect.Value),
    )
    // (we can use interfaces/objects with methods instead of closures)
    ```

    That is, the annotation takes the current types of the parameters of the annotated function (`[ftype.In(0), ftype.In(1), ...]`), and returns the new list of types of the parameter, and a function that converts actual arguments with that new type back to the old type. So for example, `ParamTags` will return a single `struct{ fx.In }` type in `newTypes`, and a `convertArgs` that maps the fields of that struct back to the original parameter list, which we can call the original function with.

- Annotations that affect results will take a similar form, and will similarly be applied to convert the output of the original function to the new `struct{ fx.Out }` type.

    ```go
    func(resultTypes []reflect.Type) (
        newTypes []reflect.Type,
        convertResults func([]reflect.Value) []reflect.Value
    )
    ```

This may lead to a cleaner application of annotations here.

annotated_test.go
sywhang: Idk why but gofmt is the thing that rewrote it this way.

annotated_test.go
sywhang: @Multiply just FYI, https://github.com/uber-go/fx/pull/817 implements the change we talked about here. Thought I should let you know :)

Multiply: Awesome! I'll try it out right away 🚀

PR #805, opened 2021-11-09
Support app shutdown before app.Done() is called

manjari25:
In the event of shutdown of an app before app.Done() is called, we currently run into a data race. This leads to a situation where an app.Done() is blocked forever. 

In order to fix this race condition, we record the shutdown signal and modify app.Done() to offer it on a channel, independent of the done channels in the app.

Refs: GO-982, #679

### DIFF
--- a/app.go
+++ b/app.go
@@ -381,6 +381,7 @@ type App struct {
 	// Used to signal shutdowns.
 	donesMu sync.RWMutex
 	dones   []chan os.Signal
+	shutdownSig os.Signal
 
 	osExit func(code int) // os.Exit override; used for testing only
 }
@@ -787,7 +788,15 @@ func (app *App) Stop(ctx context.Context) (err error) {
 // using the Shutdown functionality (see the Shutdowner documentation for details).
 func (app *App) Done() <-chan os.Signal {
 	c := make(chan os.Signal, 1)
-	signal.Notify(c, _sigINT, _sigTERM)
+
+	// If shutdown signal has been received already
+	// send it else wait for user to send a termination
+	// signal.
+	if app.shutdownSig != nil {
+		c <- app.shutdownSig
+	} else {
+		signal.Notify(c, _sigINT, _sigTERM)
+	}
 
 	app.donesMu.Lock()
 	app.dones = append(app.dones, c)
--- a/app.go
+++ b/app.go
@@ -379,7 +379,7 @@ type App struct {
 	errorHooks []ErrorHandler
 	validate   bool
 	// Used to signal shutdowns.
-	donesMu     sync.Mutex
+	donesMu     sync.Mutex  // guards dones and shutdownSig
 	dones       []chan os.Signal
 	shutdownSig os.Signal
 
@@ -798,8 +798,8 @@ func (app *App) Done() <-chan os.Signal {
 		c <- app.shutdownSig
 		return c
 	}
-	signal.Notify(c, _sigINT, _sigTERM)
 
+	signal.Notify(c, _sigINT, _sigTERM)
 	app.dones = append(app.dones, c)
 	return c
 }
--- a/app.go
+++ b/app.go
@@ -379,8 +379,8 @@ type App struct {
 	errorHooks []ErrorHandler
 	validate   bool
 	// Used to signal shutdowns.
-	donesMu sync.RWMutex
-	dones   []chan os.Signal
+	donesMu     sync.Mutex
+	dones       []chan os.Signal
 	shutdownSig os.Signal
 
 	osExit func(code int) // os.Exit override; used for testing only
@@ -792,13 +792,12 @@ func (app *App) Done() <-chan os.Signal {
 	// If shutdown signal has been received already
 	// send it else wait for user to send a termination
 	// signal.
+	app.donesMu.Lock()
 	if app.shutdownSig != nil {
 		c <- app.shutdownSig
 	} else {
 		signal.Notify(c, _sigINT, _sigTERM)
 	}
-
-	app.donesMu.Lock()
 	app.dones = append(app.dones, c)
 	app.donesMu.Unlock()
 	return c
--- a/app.go
+++ b/app.go
@@ -789,17 +789,18 @@ func (app *App) Stop(ctx context.Context) (err error) {
 func (app *App) Done() <-chan os.Signal {
 	c := make(chan os.Signal, 1)
 
-	// If shutdown signal has been received already
-	// send it else wait for user to send a termination
-	// signal.
 	app.donesMu.Lock()
+	defer app.donesMu.Unlock()
+	// If shutdown signal has been received already
+	// send it and return. If not, wait for user to send a termination
+	// signal.
 	if app.shutdownSig != nil {
 		c <- app.shutdownSig
-	} else {
-		signal.Notify(c, _sigINT, _sigTERM)
+		return c
 	}
+	signal.Notify(c, _sigINT, _sigTERM)
+
 	app.dones = append(app.dones, c)
-	app.donesMu.Unlock()
 	return c
 }
 
--- a/app.go
+++ b/app.go
@@ -379,7 +379,7 @@ type App struct {
 	errorHooks []ErrorHandler
 	validate   bool
 	// Used to signal shutdowns.
-	donesMu     sync.Mutex  // guards dones and shutdownSig
+	donesMu     sync.Mutex // guards dones and shutdownSig
 	dones       []chan os.Signal
 	shutdownSig os.Signal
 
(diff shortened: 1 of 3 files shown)


app.go
```diff
@@ -787,7 +788,15 @@ func (app *App) Stop(ctx context.Context) (err error) {
...
-	signal.Notify(c, _sigINT, _sigTERM)
+
+	// If shutdown signal has been received already
+	// send it else wait for user to send a termination
+	// signal.
+	if app.shutdownSig != nil {
+		c <- app.shutdownSig
+	} else {
+		signal.Notify(c, _sigINT, _sigTERM)
+	}
```
abhinav, line 799: This isn't thread safe. Someone might call Shutdown (which calls broadcastSignal) in another goroutine at the same time this is happening. This needs to be done after we have a lock.

shutdown.go
```diff
@@ -60,6 +60,9 @@ func (app *App) broadcastSignal(signal os.Signal) error {
 	app.donesMu.RLock()
 	defer app.donesMu.RUnlock()
 
+	// Record a shutdown signal
+	app.shutdownSig = signal
```
abhinav, line 64: This isn't thread-safe. We're modifying state with a read-lock. We need a write lock if we're modifying state, or we can use an atomic.

shutdown_test.go
```diff
@@ -65,4 +67,21 @@ func TestShutdown(t *testing.T) {
...
+			t,
+			fx.Populate(&s),
+		)
+
+		require.NoError(t, app.Start(context.Background()), "error starting app")
+		assert.NoError(t, s.Shutdown(), "error in app shutdown")
+		done1, done2 := app.Done(), app.Done()
+		defer app.Stop(context.Background())
+		assert.NotNil(t, <-done1, "done channel 1 did not receive signal")
+		assert.NotNil(t, <-done2, "done channel 2 did not receive signal")
```
abhinav, line 85: We should add a comment here that these will deadlock if app.Done misbehaves

abhinav (commented):
Let's also add a test for the data race:
have a test spawn off N (say, 50) goroutines, all blocked on a channel.
When signaled, have them call Done().
Separately, have another goroutine also blocked on the same channel,
and have that call shutdown when signaled.

Perhaps something like:

```
const N = 50

var s Shutdowner
app := ...

ready := make(chan struct{}) // closed when goroutines should run
var wg sync.WaitGroup // (use this to keep track of and wait for all goroutines)
wg.Add(N)
for i := 0; i < N; i++ {
  go func() { ... }()  // calls app.Done and inspects the result
}

wg.Add(1)
go func() { ... }() // calls Shutdown

close(ready) // goroutines race to call app.Done and shutdown
wg.Wait()
```

I suggest writing this test before fixing the data races I mentioned in the code review. `go test -race` should explode; if it doesn't, the test isn't quite right and we should figure out why.

shutdown_test.go
```diff
@@ -65,4 +69,59 @@ func TestShutdown(t *testing.T) {
...
+		const N = 50
+		ready := make(chan struct{})
+		var wg sync.WaitGroup // tracks and waits for all goroutines
+
+		// call app.Done()
+		wg.Add(N)
+		for i := 0; i < N; i++ {
+			i := i
+			go func() {
+				defer wg.Done()
```
sywhang, line 111: i think you meant to make these goroutines block on `ready`?

shutdown_test.go
```diff
@@ -65,4 +69,59 @@ func TestShutdown(t *testing.T) {
...
+				done := app.Done()
+				assert.NotNil(t, <-done, fmt.Sprintf("done channel %v did not receive signal", i))
+			}()
+		}
+
+		// call Shutdown()
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
```
sywhang, line 120: same as above

shutdown_test.go
```diff
@@ -65,4 +69,59 @@ func TestShutdown(t *testing.T) {
...
+		assert.NoError(t, s.Shutdown(), "error in app shutdown")
+		done1, done2 := app.Done(), app.Done()
+		defer app.Stop(context.Background())
+		// Receiving on done1 and done2 will deadlock in the event that app.Done()
+		// doesn't work as expected.
+		assert.NotNil(t, <-done1, "done channel 1 did not receive signal")
+		assert.NotNil(t, <-done2, "done channel 2 did not receive signal")
+	})
+
+	t.Run("test data race: shutdown app before calling Done()", func(t *testing.T) {
```
abhinav, line 92: also this test is independent enough that i would put it in its own Test function

app.go
```diff
@@ -787,9 +788,16 @@ func (app *App) Stop(ctx context.Context) (err error) {
...
 	c := make(chan os.Signal, 1)
-	signal.Notify(c, _sigINT, _sigTERM)
 
+	// If shutdown signal has been received already
+	// send it else wait for user to send a termination
+	// signal.
 	app.donesMu.Lock()
+	if app.shutdownSig != nil {
+		c <- app.shutdownSig
+	} else {
```
abhinav, line 798: In this case, we'll want to return the channel early.
We don't need to hold onto the channel.
(It doesn't cause any problems,
but we'll never hit that codepath again after shutdown.)

I'd suggest using a `defer unlock` on the mutex above
so that we can return early here.

```
mu.Lock()
defer mu.Unlock()

if sig != nil {
  c <- sig
  return c
}

...
```

shutdown.go
```diff
@@ -57,8 +57,11 @@ func (app *App) shutdowner() Shutdowner {
 }
 
 func (app *App) broadcastSignal(signal os.Signal) error {
-	app.donesMu.RLock()
-	defer app.donesMu.RUnlock()
+	app.donesMu.Lock()
+	defer app.donesMu.Unlock()
+
+	// Record a shutdown signal
```
abhinav, line 63: nit: Don't need this comment.
```suggestion
```

shutdown_test.go
```diff
@@ -65,4 +69,61 @@ func TestShutdown(t *testing.T) {
...
+		t,
+		fx.Populate(&s),
+	)
+	require.NoError(t, app.Start(context.Background()), "error starting app")
+
+	const N = 50
+	ready := make(chan struct{}) // used to orchestrate goroutines for Done() and ShutdownOption()
+	var wg sync.WaitGroup        // tracks and waits for all goroutines
+
+	// call app.Done()
```
abhinav, line 107: Expand this comment please.
I was confused by what this meant.
Clarify that you're spawning N goroutines that call app.Done.

shutdown_test.go
```diff
@@ -65,4 +69,61 @@ func TestShutdown(t *testing.T) {
...
+	wg.Add(N)
+	for i := 0; i < N; i++ {
+		i := i
+		go func() {
+			defer wg.Done()
+			<-ready
+			done := app.Done()
+			assert.NotNil(t, <-done, fmt.Sprintf("done channel %v did not receive signal", i))
```
abhinav, line 115: testify supports printf-style strings out of the box. don't need to cal fmt.Sprintf.
```suggestion
			assert.NotNil(t, <-done, "done channel %v did not receive signal", i)
```

shutdown_test.go
```diff
@@ -65,4 +69,61 @@ func TestShutdown(t *testing.T) {
...
+		}()
+	}
+
+	// call Shutdown()
+	wg.Add(1)
+	go func() {
+		<-ready
+		defer wg.Done()
+		s.Shutdown()
```
abhinav, line 124: assert.NoError here please

abhinav (commented):
much better/almost ready to land.
thanks!

sywhang (approved):
👍

app.go
```diff
@@ -379,8 +379,9 @@ type App struct {
 	errorHooks []ErrorHandler
 	validate   bool
 	// Used to signal shutdowns.
-	donesMu sync.RWMutex
-	dones   []chan os.Signal
+	donesMu     sync.Mutex
```
abhinav, line 382: I would add a comment here that this guards shutdownSig too.
```suggestion
	donesMu     sync.Mutex  // guards dones and shutdownSig
```

app.go
```diff
@@ -787,11 +788,19 @@ func (app *App) Stop(ctx context.Context) (err error) {
...
+	defer app.donesMu.Unlock()
+	// If shutdown signal has been received already
+	// send it and return. If not, wait for user to send a termination
+	// signal.
+	if app.shutdownSig != nil {
+		c <- app.shutdownSig
+		return c
+	}
+	signal.Notify(c, _sigINT, _sigTERM)
+
```
abhinav, line 802: nit: signal.Notify is part of the second "paragraph" in that the two blocks are "already shutdown" and "not yet shutdown" so the statements for those should be grouped together.
```suggestion

	signal.Notify(c, _sigINT, _sigTERM)
```

abhinav (approved):

Issue #763, opened 2021-07-28
Shutdowner: provide option to communicate an exit code or even a value for fx.Run to return


oakad:
`Shutdowner` is a useful mechanism for willfully exiting applications, but it can be even better. To this end, the application channel can be upgraded to carry a value of type

```
type SignalWithValue struct {
    Signal os.Signal
    ExitValue interface{}
}
```
The exit value can then be recovered from the `App.Done` channel value or from `App.Run` return value.

This is mostly useful to set the application exit code to some value, but in a slightly more generic way than simply returning int.

For a bit of context: Dotnet Core allows applications to set the eventual application exit code without actually exiting. This enables a pattern of the following sort:

```
System.Environment.ExitCode = 0;
// do things
// bad things happen
System.Environment.ExitCode = 255;
// Perform an orderly shutdown
(IHostApplicationLifetime).StopApplication();
```

Thus, Dotnet Core can avoid passing any values to `StopApplication`, but we don't have such a luxury in Go.

abhinav:
This is a reasonable feature request,
although I'd like for us to do it slightly differently.

We intentionally left room in the Shutdowner API for functional options.
We can add a new functional option to specify the shutdown exit code:

```
func ShutdownCode(exitCode int) ShutdownOption
```

If unspecified, we'll continue to default to zero.
With this option, users will be able to control the app.Run exit code.

To control the exit code for applications that do not use app.Run,
we'll likely have to offer a replacement for app.Done that returns a different kind of channel,
with a struct type similar to yours.

```
// name TBD
type Shutdown struct {
  Signal os.Signal
  ExitCode int
}
```

Usage:

```
app.Start(...)

ch := app.Something() // name TBD
result := <-ch

app.Stop(...)

os.Exit(result.ExitCode)
```

Internal issue ref: GO-981

oakad:
Should work.

Though, I still think an interface or a particular subtype of `error` (something with `GetErrorCode` in it) there may add a bit of genericity to the feature.

jmmills:
@sywhang This issue should be closed

PR #808, opened 2021-11-10
Support app shutdown with exit codes

manjari25:
We provide a way to shutdown an app via the Shutdowner interface but, we do not provide a way to specify/change the exit code. This commit adds support for shutting down apps with an exit code. Exit code can be controlled by providing a new Option to configure the Shutdowner. This can be done via `shutdowner.Shutdown(WithExitCode(...))`

* For apps that use `app.Run()`, the exit code is read directly and returned.
* For apps that are orchestrated with `app.Start()` and `app.Stop()`, a new function `app.DoneWithCode()` is presented that works similar to `app.Done()`. The following code illustrats its usage -
```
app := ...
app.Start(...)
done := app.DoneWithCode()
res := <-done
app.Stop(...)
res.ExitCode // captures the exit code
```

Refs: GO-981, #763

### DIFF
--- a/app.go
+++ b/app.go
@@ -666,11 +666,7 @@ func (app *App) Run() {
 	// Historically, we do not os.Exit(0) even though most applications
 	// cede control to Fx with they call app.Run. To avoid a breaking
 	// change, never os.Exit for success.
-	code := app.run(app.Done())
-	if app.exitCode != 0 {
-		code = app.exitCode
-	}
-	if code != 0 {
+	if code := app.run(app.Done()); code != 0 {
 		app.exit(code)
 	}
 }
@@ -679,8 +675,13 @@ func (app *App) run(done <-chan os.Signal) (exitCode int) {
 	startCtx, cancel := context.WithTimeout(context.Background(), app.StartTimeout())
 	defer cancel()
 
+	exitCode = 1
+	if app.exitCode != 0 {
+		exitCode = app.exitCode
+	}
+
 	if err := app.Start(startCtx); err != nil {
-		return 1
+		return exitCode
 	}
 
 	sig := <-done
@@ -690,10 +691,10 @@ func (app *App) run(done <-chan os.Signal) (exitCode int) {
 	defer cancel()
 
 	if err := app.Stop(stopCtx); err != nil {
-		return 1
+		return exitCode
 	}
 
-	return 0
+	return app.exitCode
 }
 
 // Err returns any error encountered during New's initialization. See the
--- a/app.go
+++ b/app.go
@@ -666,7 +666,11 @@ func (app *App) Run() {
 	// Historically, we do not os.Exit(0) even though most applications
 	// cede control to Fx with they call app.Run. To avoid a breaking
 	// change, never os.Exit for success.
-	if code := app.run(app.Done()); code != 0 {
+	code := app.run(app.Done())
+	if app.exitCode != 0 {
+		code = app.exitCode
+	}
+	if code != 0 {
 		app.exit(code)
 	}
 }
--- a/app.go
+++ b/app.go
@@ -817,7 +817,7 @@ type Shutdown struct {
 
 // DoneWithCode returns a channel of the Shutdown struct to block on after starting
 // the application. This behaves exactly as Done() does with the addition of
-// capturing an application exit code, if one exists.
+// capturing the application's exit code, if one is set.
 func (app *App) DoneWithCode() <-chan Shutdown {
 	c := make(chan Shutdown, 1)
 	// farm out a goroutine to wait for
--- a/app.go
+++ b/app.go
@@ -809,6 +809,25 @@ func (app *App) Done() <-chan os.Signal {
 	return c
 }
 
+// Shutdown captures a signal and exit code.
+type Shutdown struct {
+	Signal   os.Signal
+	ExitCode int
+}
+
+// DoneWithCode returns a channel of the Shutdown struct to block on after starting
+// the application. This behaves exactly as Done() does with the addition of
+// capturing an application exit code, if one exists.
+func (app *App) DoneWithCode() <-chan Shutdown {
+	c := make(chan Shutdown, 1)
+	// farm out a goroutine to wait for
+	// app.Done() channel to be populated
+	go func() {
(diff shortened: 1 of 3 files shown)


shutdown.go
```diff
@@ -39,6 +39,18 @@ type ShutdownOption interface {
 	apply(*shutdowner)
 }
 
+type exitCodeOption int
+
+func (e exitCodeOption) apply(s *shutdowner) {
```
abhinav, line 44: let's make the option implement `String()` so that it can be printed when debugging.

abhinav (commented):
RE: testing:
check out the hidden WithExit option (only available in tests).
https://github.com/uber-go/fx/blob/14f0965fc20acdf272e662cefd1776faa8aaff0e/app_internal_test.go#L77

usage:
https://github.com/uber-go/fx/blob/master/app_test.go#L842

Using that, you can capture the exit code that an app.Run would have exited with.

app.go
```diff
@@ -665,7 +666,11 @@ func (app *App) Run() {
 	// Historically, we do not os.Exit(0) even though most applications
 	// cede control to Fx with they call app.Run. To avoid a breaking
 	// change, never os.Exit for success.
-	if code := app.run(app.Done()); code != 0 {
+	code := app.run(app.Done())
+	if app.exitCode != 0 {
+		code = app.exitCode
```
sywhang, line 671: this is kinda ugly. Rather than doing this, can you change `exitCode`'s default value to be 1 and change `run` to return `app.exitCode` on error cases?

sywhang (commented):
Can you modify the PR title/description so that it includes the public API's name (DoneWithCode) in it? It'll make it easier to locate it from commit history later :)

app.go
manjari25: Hmm, will that not rewrite the exitCode set upon app shutdown?

manjari25:
> Can you modify the PR title/description so that it includes the public API's name (DoneWithCode) in it? It'll make it easier to locate it from commit history later :)

Done

shutdown.go
```diff
@@ -34,11 +34,27 @@ type Shutdowner interface {
...
+	s.app.exitCode = int(e)
+}
+
+func (e exitCodeOption) String() string {
+	return fmt.Sprintf("s.Shutdown(%v)", int(e))
```
abhinav, line 49: This should probably be fx.WithExitCode.

shutdown.go
```diff
@@ -34,11 +34,27 @@ type Shutdowner interface {
...
+}
+
+// WithExitCode allows the user to configure the exitCode upon application
+// shutdown.
```
abhinav, line 53: How about,

> WithExitCode specifies the exit code for an application shutdown. Applications using App.Run will exit the program with this exit code. Applications using App.DoneWithCode will return this exit code in the Shutdown struct.
>
> Defaults to 0.

app.go
```diff
@@ -804,6 +810,25 @@ func (app *App) Done() <-chan os.Signal {
...
+// Shutdown captures a signal and exit code.
+type Shutdown struct {
+	Signal   os.Signal
+	ExitCode int
+}
+
+// DoneWithCode returns a channel of the Shutdown struct to block on after starting
+// the application. This behaves exactly as Done() does with the addition of
+// capturing the application's exit code, if one is set.
+func (app *App) DoneWithCode() <-chan Shutdown {
```
abhinav, line 822: Can we find an alternative name for this? DoneWithCode isn't enough because we're returning a whole Shutdown struct that can conceivably contain more information in the future.

I can't think of anything obvious yet. Maybe "Wait" or similar? We can discuss this.

app.go
```diff
@@ -804,6 +810,25 @@ func (app *App) Done() <-chan os.Signal {
 	return c
 }
 
+// Shutdown captures a signal and exit code.
```
abhinav, line 813: > Shutdown is a request to shut down the application.

app.go
```diff
@@ -804,6 +810,25 @@ func (app *App) Done() <-chan os.Signal {
...
+type Shutdown struct {
+	Signal   os.Signal
+	ExitCode int
```
abhinav, line 816: Please document these fields.

> Signal is the signal that caused the shutdown.
> For manual shutdowns with Shutdowner, it is TODO.

(Figure out the value.)

> ExitCode specifies the code the program should exit with.
> For programs terminated with SIGINT, this is zero.
> For programs terminated with Shutdowner, callers may specify the exit code using the WithExitCode function.

app.go
```diff
@@ -674,8 +675,13 @@ func (app *App) run(done <-chan os.Signal) (exitCode int) {
 	startCtx, cancel := context.WithTimeout(context.Background(), app.StartTimeout())
 	defer cancel()
 
+	exitCode = 1
+	if app.exitCode != 0 {
+		exitCode = app.exitCode
```
abhinav, line 680: This won't give us what we want for normal use cases.
Per our documentation, Shutdowner should be invoked from a lifecycle hook.
That's run when we call App.Start.
By that point, we'll have already captured the exit code.

We need to use the new DoneWithCode API below after the Start, and use the exit code from that.

shutdown_test.go
```diff
@@ -87,6 +87,59 @@ func TestShutdown(t *testing.T) {
...
+		expectedExitCode := 2
+		shutdown := func(s fx.Shutdowner) error {
+			return s.Shutdown(fx.WithExitCode(expectedExitCode))
+		}
+
+		app := fxtest.New(
+			t,
+			fx.Populate(&s),
+			fx.WithExit(exit),
+			fx.Invoke(shutdown),
```
abhinav, line 113: This is incorrect usage per our documentation of Shutdowner:

```
// In practice this means Shutdowner.Shutdown should not be called from an
// fx.Invoke, but from a fx.Lifecycle.OnStart hook.
```

Previously, this just wouldn't work.
That has likely changed since #805 so we can update the documentation,
but all usages of Shutdowner up until this point are inside a lifecycle hook (on or after App.Start),
so we need to test for that too.

abhinav (changes requested):

sywhang:
This has been completed with https://github.com/uber-go/fx/pull/989.

PR #809, opened 2021-11-14
fx.Annotate: Support variadic functions

xqbumu:
Add support to fx.Annotate for variadic functions.
The last parameter of variadic functions is now treated as a slice.
With this users can feed value groups into variadic functions.

Ref https://github.com/uber-go/dig/issues/120

### DIFF
--- a/annotated.go
+++ b/annotated.go
@@ -287,7 +287,8 @@ func (ann annotated) String() string {
 // Build builds and returns a constructor based on fx.In/fx.Out params and
 // results wrapping the original constructor passed to fx.Annotate.
 func (ann *annotated) Build() (interface{}, error) {
-	if reflect.TypeOf(ann.Target).Kind() != reflect.Func {
+	ft := reflect.TypeOf(ann.Target)
+	if ft.Kind() != reflect.Func {
 		return nil, fmt.Errorf("must provide constructor function, got %v (%T)", ann.Target, ann.Target)
 	}
 
@@ -301,7 +302,12 @@ func (ann *annotated) Build() (interface{}, error) {
 	origFn := reflect.ValueOf(ann.Target)
 	newFn := reflect.MakeFunc(newFnType, func(args []reflect.Value) []reflect.Value {
 		args = remapParams(args)
-		results := origFn.Call(args)
+		var results []reflect.Value
+		if ft.IsVariadic() {
+			results = origFn.CallSlice(args)
+		} else {
+			results = origFn.Call(args)
+		}
 		results = remapResults(results)
 		return results
 	})
--- a/annotated.go
+++ b/annotated.go
@@ -552,7 +552,7 @@ func (ann *annotated) results() (
 //  }, fx.ParamTags(``, `group:"server"`))
 //
 // If we provide the above to the application,
-// any constructor in the Fx application can feed it HTTP handlers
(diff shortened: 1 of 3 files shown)


sywhang (approved):
Thanks for this change. Looks good to me

xqbumu:
> 
> 
> Thanks for this change. Looks good to me

There is something wrong that 'value groups cannot be optional' when use optional tags in ParamTags.
And that message is from dig where we can delve deeper.

Refer: https://github.com/uber-go/dig/issues/120

abhinav:
Hey, on a second look, this doesn't match the behavior of dig.
Dig specifically just ignores variadic arguments.
That is, if a function accepts variadic arguments, dig will fill everything except that.
I originally thought that's what this change was doing, but the test indicates otherwise.

If we want to support filling variadic arguments with value groups using this,
we can do that, but we should make that choice explicitly.
CC @sywhang @shirchen 

RE: value groups cannot be optional:
They cannot be optional, in that you cannot tag them optional.
They are already optional, in that if no values are provided to that group, you get an empty slice.

xqbumu:
> 
> 
> Hey, on a second look, this doesn't match the behavior of dig. Dig specifically just ignores variadic arguments. That is, if a function accepts variadic arguments, dig will fill everything except that. I originally thought that's what this change was doing, but the test indicates otherwise.
> 
> If we want to support filling variadic arguments with value groups using this, we can do that, but we should make that choice explicitly. CC @sywhang @shirchen
> 
> RE: value groups cannot be optional: They cannot be optional, in that you cannot tag them optional. They are already optional, in that if no values are provided to that group, you get an empty slice.

Oh, I got it, very thanks for you explain.

annotated_test.go
```diff
@@ -494,6 +498,46 @@ func TestAnnotate(t *testing.T) {
...
+
+		type T1 struct{ s string }
+
+		app := fxtest.New(t,
+			fx.Supply(
+				fx.Annotate(T1{"foo"}, fx.ResultTags(`group:"t"`)),
+				fx.Annotate(T1{"bar"}, fx.ResultTags(`group:"t"`)),
+				fx.Annotate(T1{"baz"}, fx.ResultTags(`group:"t"`)),
+			),
+			fx.Invoke(fx.Annotate(func(got []T1) {
```
sywhang, line 532: Should this be `got ...T1` instead?

```suggestion
			fx.Invoke(fx.Annotate(func(got ...T1) {
```

annotated.go
```diff
@@ -522,6 +528,42 @@ func (ann *annotated) results() (
...
+//
+// You can use variadic parameters with Fx's value groups.
+// For example,
+//
+//  fx.Annotate(func(mux *http.ServeMux, handlers ...http.Handler) {
+//    // ...
+//  }, fx.ParamTags(``, `group:"server"`))
+//
+// If we provide the above to the application,
+// any constructor in the Fx application can feed it HTTP handlers
```
r-hang, line 555: nit: feed it's HTTP handlers

annotated_test.go
abhinav: derp. yes.

annotated.go
abhinav: ack

sywhang (approved):
LGTM

abhinav:
Thanks! @xqbumu

PR #813, opened 2021-11-18
Use mock clock, unflake tests

### LINKED ISSUE #799: OnStop_timeout test occasionally fails
TestAppRunTimeout/OnStop timeout test occasionally fails on Windows leg because the Stopped event isn't emitted for some reason. Probably a flaky test and we should fix this. 
```
FIND: Parameter format not correct
go test -race -coverprofile=cover.out -coverpkg=./... ./...
--- FAIL: TestAppRunTimeout (0.00s)
    --- FAIL: TestAppRunTimeout/OnStop_timeout (0.00s)
        app_test.go:828: 
            	Error Trace:	app_test.go:828
            	Error:      	"[]" should have 1 item(s), but has 0
            	Test:       	TestAppRunTimeout/OnStop_timeout
            	Messages:   	expected a "Stopped" event
FAIL
```

Ref: GO-1001

abhinav:
This PR resolves flakiness of some tests in Fx
by abstracting our use of time operations behind a Clock interface
and using a mock clock for the relevant tests.

This adds an internal/fxclock package that defines the Clock interface.
This interface is a compliant subset of ["benbjohnson/clock"][1].Clock.
This intentionally does not use benbjohnson/clock directly outside of tests
to keep it a test-only dependency.

[1]: https://github.com/benbjohnson/clock

The new *internal-only* `WithClock` option
accepts and plumbs an implementation of `fxclock.Clock`.
Note that this option is declared in app_internal_test.go.
It's not part of the public API,
but it's accessible to external tests in the top-level Fx directory.

Finally, this switches to using the Clock for time operations where relevant,
and updates tests that rely on sleep/time-based blocking to use a mock clock.
Now,

- `time.Now` is only present in fxclock.System.
- `time.Since` is only present in fxclock.System.
- `time.Sleep` is only present in fxclock.System.
- `context.WithTimeout` is present in fxclock.System,
  `fxtest.App.Require{Start,Stop}`, and `example_test.go`.
  The latter two because fxclock.Clock is not part of our public API.

Resolves #799
Ref: GO-1001

### DIFF
--- a/app_internal_test.go
+++ b/app_internal_test.go
@@ -29,6 +29,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/fx/fxevent"
+	"go.uber.org/fx/internal/fxclock"
 	"go.uber.org/fx/internal/fxlog"
 	"go.uber.org/fx/internal/fxreflect"
 )
@@ -86,4 +87,21 @@ func (o withExitOption) String() string {
 
 func (o withExitOption) apply(app *App) {
 	app.osExit = o
+}
+
+// WithClock specifies how Fx accesses time operations.
+//
+// This is an internal option available only to tests defined in this package.
+func WithClock(clock fxclock.Clock) Option {
+	return withClockOption{clock}
+}
+
+type withClockOption struct{ clock fxclock.Clock }
+
+func (o withClockOption) apply(app *App) {
+	app.clock = o.clock
+}
+
+func (o withClockOption) String() string {
+	return fmt.Sprintf("WithClock(%v)", o.clock)
 }
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module go.uber.org/fx
 go 1.13
 
 require (
+	github.com/benbjohnson/clock v1.3.0
 	github.com/stretchr/testify v1.7.0
 	go.uber.org/dig v1.12.0
 	go.uber.org/goleak v1.1.11
--- a/app.go
+++ b/app.go
@@ -35,6 +35,7 @@ import (
 
 	"go.uber.org/dig"
 	"go.uber.org/fx/fxevent"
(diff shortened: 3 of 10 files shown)


abhinav:
Fixing data race in https://github.com/benbjohnson/clock/pull/42.
Temporarily pinning to the fork while the fix is in review to validate the fix further.

go.mod
```diff
@@ -3,10 +3,13 @@ module go.uber.org/fx
...
+	github.com/benbjohnson/clock v1.3.0
 	github.com/stretchr/testify v1.7.0
 	go.uber.org/dig v1.12.0
 	go.uber.org/goleak v1.1.11
 	go.uber.org/multierr v1.5.0
 	go.uber.org/zap v1.16.0
 	golang.org/x/sys v0.0.0-20210903071746-97244b99971b
 )
+
+replace github.com/benbjohnson/clock => github.com/abhinav/clock v1.3.1-0.20211118204451-38de6f8a55b9
```
abhinav, line 15: We probably will not land this `replace` directive.
I'd like to wait a few days for the change to be merged upstream.

sywhang (approved):
the code changes looks good to me, and yeah, waiting on merging this until the upstream change gets merged is probably better than pinning the fork :)

app_internal_test.go
```diff
@@ -87,3 +88,20 @@ func (o withExitOption) String() string {
 func (o withExitOption) apply(app *App) {
 	app.osExit = o
 }
+
+// WithClock specifies how Fx accesses time operaitons.
```
r-hang, line 92: nit: operations.

r-hang (approved):

abhinav:
This PR is now WIP while I'm digging further into this.

abhinav:
@sywhang @r-hang 
This is good to go now.
We're also no longer pinned to the fork of clock—I was able to resolve that by not sharing the clock between subtests.

Issue #811, opened 2021-11-17
Ability to annotate multiple interfaces for single provider


apawelec:
Lets say I have two command/query interfaces: `AccountCommandRepository` and `AccountQueryRepository` and one infrastructure component `databaseAccountRepository` that implements them both. 

My current configurations looks like this:
```
fx.Provide(func(...) (command.AccountCommandRepository, query.AccountQueryRepository) {
    r := newDatabaseAccountRepository(...)
    return r, r
}),
```

If I'm not wrong, there is currently no easy way to mimic this with `fx.Annotate`/`fx.As`, the only way I've figured is:

```
fx.Provide(
    fx.Annotate(
        func(...) (*databaseAccountRepository, *databaseAccountRepository) {
            r := newDatabaseAccountRepository(...)
            return r, r
        },
        fx.As(new(command.AccountCommandRepository), new(query.AccountQueryRepository)),
    ),
),
```
but it's even more verbose (of course I can modify `newDatabaseAccountRepository` constructor to return same value twice, to simplify this, but I prefer don't do this).

It would be nice if this can be replaced by simple:

```
fx.Provide(
    fx.Annotate(
        newDatabaseAccountRepository,
        fx.As(new(command.AccountCommandRepository)),
        fx.As(new(query.AccountQueryRepository)),
    ),
),
```
but in current version it's not possible and results with error `cannot apply more than one line of As`.

Could you consider to change this behaviour?

abhinav:
Hello! Yes, supporting that was next on the list for `fx.As`.
The latter version, specifically:

```go
fx.Provide(
    fx.Annotate(
        newDatabaseAccountRepository,
        fx.As(new(command.AccountCommandRepository)),
        fx.As(new(query.AccountQueryRepository)),
    ),
),
```

We'll welcome a PR adding support for that,
but if you don't have time to contribute this,
we plan to work on this in the coming weeks.

Internal issue ref: GO-1023

sywhang:
This has been implemented in #817. Closing this issue as done.

PR #817, opened 2021-11-20
Allow multiple As Annotations to be provided

sywhang:
Currently, Fx errors out when two `As` annotations are provided to a constructor. This is inconvenient as there may be times where a user wants to annotate a single type as many interfaces that it implements.

This change allows such annotations to be provided. 

In particular, the following annotation is now possible:
```go
fx.Annotate(func() *bytes.Buffer {...},
  fx.As(new(io.Writer)),
  fx.As(new(io.Reader)),
}
```

`bytes.Buffer`  implements both `io.Writer` and `io.Reader`, hence this provides the same instance of `bytes.Buffer` as both `io.Writer` and `io.Reader` interface. Note that this provides the same *instances* as arguments that takes in these, and should NOT be used as if the constructor got invoked twice to provide two different instances of type `bytes.Buffer`, one as `io.Writer` and the other as `io.Writer`.

To allow this change, few modifications were made to internal representation of As tags. Most of the changes are changing the dimensions of the slices from 1D to 2D. The following changes were made:
- `annotated.As` is now an array of array of `reflect.Type`, where As[i][j] is the i-th type that j-th result needs to be annotated as. 
- offsets and results are now 2D slices as well, similar to `annotated.As`.
- The out type of annotated function generated by reflection is now multiple `fx.Out` structs, each of which correspond to one line of `fx.As` annotation, followed by an error type if one exists in the output type of the annotated constructor.

Refs: #811
GO-1023

### DIFF
--- a/annotated.go
+++ b/annotated.go
@@ -244,9 +244,6 @@ func As(interfaces ...interface{}) Annotation {
 }
 
 func (at asAnnotation) apply(ann *annotated) error {
-	if len(ann.As) > 0 {
-		return errors.New("cannot apply more than one line of As")
-	}
 	types := make([]reflect.Type, len(at.targets))
 	for i, typ := range at.targets {
 		t := reflect.TypeOf(typ)
@@ -257,7 +254,7 @@ func (at asAnnotation) apply(ann *annotated) error {
 		types[i] = t
 	}
 
-	ann.As = types
+	ann.As = append(ann.As, types)
 	return nil
 }
 
@@ -265,7 +262,7 @@ type annotated struct {
 	Target     interface{}
 	ParamTags  []string
 	ResultTags []string
-	As         []reflect.Type
+	As         [][]reflect.Type
 }
 
 func (ann annotated) String() string {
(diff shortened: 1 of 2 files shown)


annotated.go
```diff
@@ -421,61 +428,66 @@ func (ann *annotated) results() (
...
+			out := reflect.New(resTypes[structNum]).Elem()
+			for i, r := range results {
+				if i == len(results)-1 && hasError {
+					// If hasError and this is the last item,
+					// we are guaranteed that this is an error
+					// object.
+					if err, _ := r.Interface().(error); err != nil {
+						outErr = err
+					}
+					continue
```
abhinav, line 474: I think this nested loop could be inverted.
Right now, we iterate through all results for each output struct,
and try to check the error for each output struct.

If we invert the nested loop so that `range` results is outside and `range resTypes` is inside,
we'll check error only once,
and for each result, we'll assign different output fields as needed.
We'll have to populate remappedResults with `reflect.New(..)` in advance, though.
So something like:

```go
var outputs []reflect.Value  // final return values
for _, structType := range resTypes {
  outputs = append(outputs, reflect.New(structType).Elem())
}

for i, r := range results {
  if /* error case */ {
    // ...
  }

  // Fill each struct that cares about this field
  // with the value.
  for j, structType := range resTypes {
    if idx := offsets[j][i]; idx > 0 {
      outputs[j][i].Set(r)
    }
  }
}
```

annotated.go
```diff
@@ -461,21 +478,22 @@ func (ann *annotated) results() (
 				}
 				continue
 			}
-
-			out.Field(offsets[i]).Set(r)
+			for j := range resTypes {
+				if fieldIdx := offsets[j][i]; fieldIdx > 0 {
+					outResults[j].Field(fieldIdx).Set(r)
```
abhinav, line 483: consider adding a comment here about why `fieldIdx` has to be non-zero to be valid

annotated.go
```diff
@@ -393,19 +390,29 @@ func (ann *annotated) results() (
...
+	if len(ann.As) > 0 {
+		numStructs = len(ann.As)
 	}
 
-	// offsets[i] is index of result i in the generated fx.Out
+	// offsets[i][j] is index of result j in the ith generated fx.Out
 	// struct.
-	offsets := make([]int, ft.NumOut())
+	offsets := make([][]int, numStructs)
+
```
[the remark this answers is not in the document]
abhinav: This slight re-order of declarations would still be nice.

abhinav (approved):
LGTM w nits

PR #824, opened 2021-12-09
Validate arguments to fx.Annotate for fx.In and fx.Out

sywhang:
Functions that either take in fx.In as input parameters or outputs
an fx.Out struct should not be used as parameters to fx.Annotate.

This validates the input function to fx.Annotate to check for such
case and bail out early.

Ref: GO-1087

### DIFF
--- a/annotated.go
+++ b/annotated.go
@@ -326,8 +326,10 @@ func (ann *annotated) typeCheckOrigFn() error {
 		if ot.Kind() != reflect.Struct {
 			continue
 		}
-		if ot.NumField() > 0 && ot.Field(0).Type == reflect.TypeOf(Out{}) {
-			return fmt.Errorf("fx.Out structs cannot be annotated")
+		for i := 0; i < ot.NumField(); i++ {
+			if ot.Field(i).Type == reflect.TypeOf(Out{}) {
+				return errors.New("fx.Out structs cannot be annotated")
+			}
 		}
 	}
 
@@ -336,8 +338,10 @@ func (ann *annotated) typeCheckOrigFn() error {
 		if it.Kind() != reflect.Struct {
 			continue
 		}
-		if it.NumField() > 0 && it.Field(0).Type == reflect.TypeOf(In{}) {
-			return fmt.Errorf("fx.In structs cannot be annotated")
+		for i := 0; i < it.NumField(); i++ {
+			if it.Field(i).Type == reflect.TypeOf(In{}) {
+				return errors.New("fx.In structs cannot be annotated")
+			}
 		}
 	}
 	return nil
--- a/annotated.go
+++ b/annotated.go
@@ -289,6 +289,10 @@ func (ann *annotated) Build() (interface{}, error) {
 		return nil, fmt.Errorf("must provide constructor function, got %v (%T)", ann.Target, ann.Target)
 	}
 
+	if err := ann.typeCheckOrigFn(); err != nil {
+		return nil, fmt.Errorf("error while applying annotation to function %T: %w", ann.Target, err)
+	}
+
 	paramTypes, remapParams := ann.parameters()
 	resultTypes, remapResults, err := ann.results()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


annotated.go
```diff
@@ -312,6 +316,33 @@ func (ann *annotated) Build() (interface{}, error) {
...
+func (ann *annotated) typeCheckOrigFn() error {
+	ft := reflect.TypeOf(ann.Target)
+	for i := 0; i < ft.NumOut(); i++ {
+		ot := ft.Out(i)
+		if ot.Kind() != reflect.Struct {
+			continue
+		}
+		if ot.NumField() > 0 && ot.Field(0).Type == reflect.TypeOf(Out{}) {
+			return fmt.Errorf("fx.Out structs cannot be annotated")
+		}
```
manjari25, line 331: Could we collapse this into a single if condition?

```lang=go
if ot.Kind() == reflect.Struct && ot.NumField() > 0 && ... {
...
}
```
Although maybe it is more readable right now. I'll leave it up to you.

annotated.go
```diff
@@ -312,6 +316,33 @@ func (ann *annotated) Build() (interface{}, error) {
...
+// fx.In struct as a parameter.
+func (ann *annotated) typeCheckOrigFn() error {
+	ft := reflect.TypeOf(ann.Target)
+	for i := 0; i < ft.NumOut(); i++ {
+		ot := ft.Out(i)
+		if ot.Kind() != reflect.Struct {
+			continue
+		}
+		if ot.NumField() > 0 && ot.Field(0).Type == reflect.TypeOf(Out{}) {
+			return fmt.Errorf("fx.Out structs cannot be annotated")
```
manjari25, line 330: Consider defining a const for this error. It could be reused in the test during assertion.

manjari25 (approved):

annotated.go
moisesvega:
```suggestion
			return errors.New("fx.Out structs cannot be annotated")
```

annotated.go
```diff
@@ -312,6 +316,33 @@ func (ann *annotated) Build() (interface{}, error) {
...
+		}
+	}
+
+	for i := 0; i < ft.NumIn(); i++ {
+		it := ft.In(i)
+		if it.Kind() != reflect.Struct {
+			continue
+		}
+		if it.NumField() > 0 && it.Field(0).Type == reflect.TypeOf(In{}) {
+			return fmt.Errorf("fx.In structs cannot be annotated")
```
moisesvega, line 340:
```suggestion
			return errors.New("fx.In structs cannot be annotated")
```

annotated.go
sywhang: yeah that's a bit messy.. i separated it out on purpose for readability. I'll leave it as is.

annotated.go
```diff
@@ -312,6 +316,33 @@ func (ann *annotated) Build() (interface{}, error) {
...
+// returning an fx.Out struct or an taking in a
+// fx.In struct as a parameter.
+func (ann *annotated) typeCheckOrigFn() error {
+	ft := reflect.TypeOf(ann.Target)
+	for i := 0; i < ft.NumOut(); i++ {
+		ot := ft.Out(i)
+		if ot.Kind() != reflect.Struct {
+			continue
+		}
+		if ot.NumField() > 0 && ot.Field(0).Type == reflect.TypeOf(Out{}) {
```
moisesvega, line 329: Are we assuming the first fill will be fx.Out ?

moisesvega, line 329: Are we assuming the first Field is always `Out{}` ?

manjari25: I swear I remember reading somewhere that In/Out embeddings should be the first field in a struct, but I am unable to find any supporting docs now. Maybe @sywhang can find some. 🤷‍♀️

abhinav: There's no such restriction. The `fx.Out` can be any field.

I haven't reviewed the PR yet, but I think Sung is validating the generated function, in which the `fx.Out` will indeed be the first field.

sywhang: oops. I'm validating the input function, so I think this should be fixed. lemme follow up with a fix

moisesvega (approved):

PR #830, opened 2022-02-01
Add fx.Module

sywhang:
This adds `fx.Module` Option which is a first-class object for supporting scoped operations on dependencies. A Module can consist of zero or more `fx.Option`s.

By default, `Provide`s to a `Module` is provided to the entire `App`, but there is a room for adding an option to scope that to a Module. `Module` can wrap `Option`s such as`Supply`/`Extract`, `Provide`, and `Invoke` but there are some `Option`s that don't make sense to put under `Module`. For example, `StartTimeout`, `StopTimeout`, `WithLogger` explicitly errors out when supplied to a `Module`. 

Implementation-wise, a `Module` corresponds to `dig.Scope` which was added in https://github.com/uber-go/dig/pull/305. Extra bookkeeping is done by the `module` struct which contains the provides and invokes to a `Scope`.


module_test.go
```diff
@@ -0,0 +1,341 @@
...
+			fx.Invoke(fx.Annotate(func(l *Logger) {
+				assert.Equal(t, "good logger", l.Name)
+			}, fx.ParamTags(`name:"goodLogger"`))),
+		)
+		defer app.RequireStart().RequireStop()
+	})
+
+	t.Run("invoke in module with annotate", func(t *testing.T) {
+		t.Parallel()
+		ranInvoke := false
```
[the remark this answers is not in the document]
sywhang: I'd rather explicitly declare it as false than relying on default initialization of the `bool` for readability purposes : )

module_test.go
```diff
@@ -0,0 +1,333 @@
...
+	})
+
+	t.Run("invoke from nested module", func(t *testing.T) {
+		t.Parallel()
+
+		app := fxtest.New(t,
+			fx.Module("child",
+				fx.Module("grandchild",
+					fx.Invoke(func(l *Logger) {
+						assert.Equal(t, "redis", l.Name)
```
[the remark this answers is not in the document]
sywhang: good catch - fixed the test to `Provide` the `Logger` above.

abhinav:
I've reverted some of the code reorganization to make the core logic changes easier to review.
The code reorg is available in https://github.com/sywhang/fx/pull/1.

module.go
```diff
@@ -0,0 +1,103 @@
...
+import (
+	"fmt"
+
+	"go.uber.org/dig"
+)
+
+// Module is a named group of zero or more fx.Options. A Module is a
+// dependency graph with limited scope, and can be used for scoping
+// graph modifications (not implemented yet).
+func Module(name string, opts ...Option) Option {
```
abhinav, line 32: minor: I would drop the second sentence since it's not yet true.
```suggestion
// Module is a named group of zero or more fx.Options.
func Module(name string, opts ...Option) Option {
```

module.go
```diff
@@ -0,0 +1,103 @@
...
+	app      *App
+}
+
+// builds the Scopes using the App's Container. Note that this happens
+// after applyModules' are called because the App's Container needs to
+// be built for any Scopes to be initialized, and applys' should be called
+// before the Container can get initialized.
+func (m *module) build(app *App, root *dig.Container) {
+	if m.parent == nil {
+		m.scope = root.Scope(m.name)
```
abhinav, line 84: maybe:

```suggestion
		m.scope = root.Scope(m.name)
		// TODO: Once fx.Decorate is in-place,
		// use the root container instead of subscope.
```

module.go
```diff
@@ -0,0 +1,103 @@
...
+	} else {
+		parentScope := m.parent.scope
+		m.scope = parentScope.Scope(m.name)
+	}
+
+	for i := 0; i < len(m.modules); i++ {
+		m.modules[i].build(app, root)
+	}
```
abhinav, line 92: nit: why the triple form over range?

```suggestion
	for _, mod := range m.modules {
		mod.build(app, root)
	}
```

abhinav (approved):
Pretty straightforward after removing the reorg.

PR #831, opened 2022-02-08
fx.Annotate: make variadic params optional by default

josephinedotlee:
If annotate is passed a variadic function, the dependency
listed as the variadic parameter should be optional
unless otherwise specified with the `optional:false`
parameter tag. This commit addresses that bug.

Resolves GO-1169


annotated_test.go
```diff
@@ -599,6 +603,50 @@ func TestAnnotate(t *testing.T) {
...
+				fx.Annotate(newB(newA())),
+			),
+			fx.Invoke(fx.Annotate(newSliceAWithB)),
+		)
+
+		defer app.RequireStart().RequireStop()
+		require.NoError(t, app.Err())
+	})
+
+	t.Run("Invoke variadic with a missing dependency", func(t *testing.T) {
```
sywhang, line 636:
```suggestion
	t.Run("Invoke non-optional variadic function with a missing dependency", func(t *testing.T) {
```

annotated_test.go
```diff
@@ -599,6 +603,50 @@ func TestAnnotate(t *testing.T) {
...
+
+		assert.Len(t, got.sa, 0)
+	})
+
+	t.Run("Invoke variadic function with multiple params", func(t *testing.T) {
+		t.Parallel()
+
+		app := fxtest.New(t,
+			fx.Supply(
+				fx.Annotate(newB(newA())),
```
sywhang, line 627: Looks like this has no annotation applied - why is this `Annotate`d?

annotated.go
```diff
@@ -381,9 +381,15 @@ func (ann *annotated) parameters() (
...
 
+		var paramTag string
 		if i < len(ann.ParamTags) {
-			field.Tag = reflect.StructTag(ann.ParamTags[i])
+			paramTag = ann.ParamTags[i]
 		}
+		if i == ft.NumIn()-1 && ft.IsVariadic() && !strings.Contains(paramTag, "optional:") &&
+			!strings.Contains(paramTag, "group:") {
+			paramTag = paramTag + " optional:\"true\""
+		}
```
abhinav, line 391: I feel like the string matching here will be fragile. Normally, since the format of struct tags is structured, we would talk about parsing it and checking, but that might be moot because:

Based on the issue description, we want to add `optional:"true"` only if the variadic parameter has no other tags. So the `[]Foo` for `whatever(...Foo)` defaults to optional only if the user didn't explicitly specify anything for it.
So the check we actually need is: param tags for this argument is empty, or is too short to hit this argument.

annotated_test.go
abhinav: +1. I think fx.Annotate will no-op if there are no annotations.

abhinav (commented):
Let's add a test for the case where some parameters of the variadic function have annotations, but not all do.

For example, given,

```
func NewFoo(A, B, ...Option)
```

We should test at least:

    fx.ParamTags(`name:"foo"`)  // only A has an annotation
    fx.ParamTags(`optional:"true"`, `name:"bar"`)  // A and B have annotations
    fx.ParamTags(`name:"foo"`, `optional:"true"`, `name:"whatever"`)  // everything has an annotation

annotated_test.go
```diff
@@ -624,9 +624,15 @@ func TestAnnotate(t *testing.T) {
 
 		app := fxtest.New(t,
 			fx.Supply(
+<<<<<<< HEAD
```
sywhang: looks like git did something funky here?

annotated.go
abhinav: Ping. This comment is still valid.

annotated.go
```diff
@@ -383,6 +383,8 @@ func (ann *annotated) parameters() (
 
 		if i < len(ann.ParamTags) {
 			field.Tag = reflect.StructTag(ann.ParamTags[i])
+		} else if i == ft.NumIn()-1 && ft.IsVariadic() {
+			field.Tag = reflect.StructTag("optional:\"true\"")
```
sywhang, line 387:
```suggestion
			field.Tag = reflect.StructTag(`optional:"true"`)
```

annotated.go
```diff
@@ -383,6 +383,8 @@ func (ann *annotated) parameters() (
 
 		if i < len(ann.ParamTags) {
 			field.Tag = reflect.StructTag(ann.ParamTags[i])
+		} else if i == ft.NumIn()-1 && ft.IsVariadic() {
+			field.Tag = reflect.StructTag(`optional:"true"`)
```
abhinav, line 387: worth explaining

```suggestion
			// If a variadic argument is unannotated, mark it optional,
			// so that just wrapping a function in fx.Annotate does not
			// suddenly introduce a required []arg dependency.
			field.Tag = reflect.StructTag(`optional:"true"`)
```

abhinav:
oops i broke it

abhinav (approved):

sywhang (approved):
👍

Issue #825, opened 2021-12-15
[Feature Request] OverrideSupply and OverrideProvide 


Wondertan:
## What
Currently, the FX errors if you Provide/Supply for an unnamed/ungrouped type multiple times. However, there are use cases where overriding an existing Provide or Supply becomes useful, like in testing(#649). Also, this becomes useful for big applications with a liberal API allowing their users to override the default setup of modules. That is, the API devs prepare the default set of modules, while a user can swap out a module of their choice through overriding.

## Why
Essentially, the feature would allow APIs like:
```go
some.New(Modules, WithModule(custom.Module))
```

Currently, users have to decouple Modules, understand what needs to drop out, and pass their custom one. With this feature, users would add desired things on top.

### API
```go
// OverrideSupply overwrites values passed to Supply and constructed with Provide.
// It mimics Supply with one difference - it does not error if there are multiple instances of a type
// are supplied/provided. Instead, OverrideSupply forces the Supply.
// Only one override can happen for a type.
func OverrideSupply(vals ...interface{}) Option

// OverrideProvide mimics Provide with one difference - it does not error if there are multiple instances of a type
// are provided or supplied. Instead, OverrideProvide forces the provide.
// Only one override can happen for a type.
func OverrideProvide(vals ...interface{}) Option
```

### Implementation ideas
* A naive idea would be to collect all the overrides and exclude Provides and Supplies with conflicting types. However, the problem arises when provided ctor returns multiple types, and overriding happens for only one of them, as the whole ctor is excluded. The issue might be ok for some cases and easily detected, thus errored. The good thing is that this can be implemented beneath FX only, without digging into `dig`. Also, this should be fine enough for testing needs in #649. 

* A more thorough approach without the issue of the approach above would go dipper into `dig` and somehow be part of graph construction. Unfortunately, I am not that familiar with `dig` implementation to understand the complexity and feasibility, so kindly ask you to help. Ideally, the implementation should still execute a conflicting ctor but exclude an overridden type from the dependency graph.

Happy to submit a patch once there is a decision on how to proceed.

abhinav:
Hey there! Apologies for the delay in responding. This has indeed been a long-standing feature request in this or other forms. There's also some discussion #653. We've been unable to reach a satisfactory design for a while, especially because there's risk of random modules stomping on your injected entities.

For example, consider a random `redisfx` module:

```go
package redisfx

var Module = fx.Options(
    fx.Provide(New),
)

[...]

Issue #649, opened 2018-10-04
possible to use fx to inject mocks?


rubensayshi:
trying to write tests and provide a mock service instead of the real, but I get an "already provided" error, is the only way to do this by making sure the original service is not being provided when I want to provide a mocked version?

glibsm:
It's difficult to say without seeing your particular setup. There is currently no mechanism to override or replace existing objects in the graph (although that has been discussed in the past).

You could set your application in a way to accept a flag to omit certain components that will be added at a later time and add them in tests, or completely re-construct application from scratch in your test.

elliotchance:
Urgh, seriously? Under unit tests it makes perfect sense to supply values for mocked services or other static input. I get that it shouldn't be a feature of the `fx` package, but it should be a feature of the `fxtest` package. Perhaps something like?

```go
app := fxtest.New(t,
    // Provides a *Flag just like fx.Supply. However, will not trigger
    // an error when *Flags is provided later. Ideally, not call the
    // constructor at all.
    fxtest.Supply(&Flags{Service: "foo"}),

    // Normal app options.
    opts(),
)
```

Thinking out loud: If `*Flags` is provided by a constructor that also provides other types, it would be acceptable to just ignore the return value of `*Flags` with the already set value. If this causes issues under test, a person should break up their constructors so `*Flags` is provided independently (and hence won't be called at all).

sywhang:
Hey all, 

It's been a very long time since this request came in, but we finished working on improving fx in the past few months and you can now achieve this in v1.17.0 with the newly added `fx.Module` and `fx.Decorate` APIs.

Specifically, here is how you might go about injecting mocked types:

```
fx.New(
  fx.Provide(
    func() Service {...},
    // ...
  ),  

  fx.Module("test",
    fx.Decorate(func() *Service {
      // returns a mocked *Service
    }),
    fx.Invoke(func() { 
      // will see mocked *Service, only within this Module. 
    }),
  ),
)
```

rubensayshi:
hey @sywhang that's pretty sweet! I'll revisit fx in the next few weeks thanks!

PR #833, opened 2022-02-09
Add fx.Decorate

sywhang:
This adds `fx.Decorate`, which lets you specify decorators to an fx app. A decorator can take in one or more dependencies that have already been `Provide`d to the app, and produce one or more values that will be used as replacements in the object graph.

For example, suppose there is a simple app like this:
```go

fx.New(
  fx.Provide(func() *Logger {
    return &Logger{Name: "logger"}
  }),
  fx.Invoke(func(l *Logger) {
    fmt.Println(l.Name)
  }),
)
```

Running this app will print "logger" on the console.

Now let us suppose a decorator was provided:
```go
fx.New(
  fx.Provide(...), // Provide same function as above
  fx.Decorate(func(l *Logger) *Logger {
    return &Logger{Name: "decorated " + l.Name}
  }),
  fx.Invoke(...), // Invoke same function as above
)
```

The decorator here will take in the provided Logger and replace it with another logger whose `Name` is `decorated logger`. The `Invoke`d function is then executed with this replacement value, so running this app will print "decorated logger" on the console.

In terms of implementation, a decorator is represented by the target decorator function and the call stack it was provided from, similar to a provider. `module` contains a list of decorators that were specified within its scope.

The dig dependency had to be updated to the latest master branch of Dig to ensure the fix for https://github.com/uber-go/dig/issues/316 is in.

Following this PR, there are two additional pieces I will be adding:
1. An eventing system for fx.Decorate. 
2. fx.Replace, which takes in a value instead of a function to replace a value in the object graph. This is similar to what fx.Supply is to fx.Provide.

This PR along with the two PRs above should make the long-awaited feature of graph modifications in fx finally possible.

---

Refs #653, #649, #825, uber-go/dig#230, GO-1203, GO-736

### DIFF
--- a/decorate.go
+++ b/decorate.go
@@ -28,6 +28,27 @@ import (
 	"go.uber.org/fx/internal/fxreflect"
 )
 
+// Decorate specifies one or more decorator functions to an fx application.
+// Decorator functions let users augment objects in the graph. They can take in
+// zero or more dependencies that must be provided to the application with fx.Provide,
+// and produce one or more values that can be used by other invoked values.
+//
+// An example decorator is the following function which accepts a value, augments that value,
+// and returns the replacement value.
+//
+//  fx.Decorate(func(log *zap.Logger) *zap.Logger {
+//    return log.Named("myapp")
+//  })
+//
+// The following decorator accepts multiple dependencies from the graph, augments and returns
+// one of them.
+//
+//  fx.Decorate(func(log *zap.Logger, cfg *Config) *zap.Logger {
+//    return log.Named(cfg.Name)
+//  })
+//
+// All modifications in the object graph due to a decorator are scoped to the fx.Module it was
+// specified from.
 func Decorate(decorators ...interface{}) Option {
 	return decorateOption{
 		Targets: decorators,
--- a/decorate.go
+++ b/decorate.go
@@ -1,0 +1,88 @@
+// Copyright (c) 2022 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package fx
+
+import (
+	"fmt"
+	"strings"
+
+	"go.uber.org/dig"
+	"go.uber.org/fx/internal/fxreflect"
+)
+
+func Decorate(decorators ...interface{}) Option {
+	return decorateOption{
+		Targets: decorators,
+		Stack:   fxreflect.CallerStack(1, 0),
+	}
+}
+
+type decorateOption struct {
+	Targets []interface{}
+	Stack   fxreflect.Stack
+}
+
+func (o decorateOption) apply(mod *module) {
+	for _, target := range o.Targets {
+		mod.decorators = append(mod.decorators, decorator{
+			Target: target,
+			Stack:  o.Stack,
+		})
+	}
+}
+
+func (o decorateOption) String() string {
+	items := make([]string, len(o.Targets))
+	for i, f := range o.Targets {
+		items[i] = fxreflect.FuncName(f)
+	}
+	return fmt.Sprintf("fx.Decorate(%s)", strings.Join(items, ", "))
+}
+
+// provide is a single decorators used in Fx.
+type decorator struct {
+	// Constructor provided to Fx. This may be an fx.Annotated.
+	Target interface{}
+
+	// Stack trace of where this provide was made.
+	Stack fxreflect.Stack
+}
+
+func runDecorator(c container, d decorator, opts ...dig.DecorateOption) error {
+	decorator := d.Target
+
+	switch decorator := decorator.(type) {
+	case annotated:
+		dcor, err := decorator.Build()
+		if err != nil {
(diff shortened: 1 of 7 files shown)


decorate.go
```diff
@@ -0,0 +1,107 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"go.uber.org/dig"
+	"go.uber.org/fx/internal/fxreflect"
+)
+
+// Decorate specifies one or more decorator functions to an fx application.
```
abhinav, line 31:
```suggestion
// Decorate specifies one or more decorator functions to an Fx application.
```

decorate.go
```diff
@@ -0,0 +1,107 @@
...
+//
+// The following decorator accepts multiple dependencies from the graph, augments and returns
+// one of them.
+//
+//  fx.Decorate(func(log *zap.Logger, cfg *Config) *zap.Logger {
+//    return log.Named(cfg.Name)
+//  })
+//
+// All modifications in the object graph due to a decorator are scoped to the fx.Module it was
+// specified from.
```
abhinav, line 51: Worth adding:

> Decorations specified in the top-level fx.New call apply across the application.

(We can add more, like decoration chaining, value group decoration, etc. later)

decorate.go
```diff
@@ -0,0 +1,107 @@
...
+//    return log.Named("myapp")
+//  })
+//
+// The following decorator accepts multiple dependencies from the graph, augments and returns
+// one of them.
+//
+//  fx.Decorate(func(log *zap.Logger, cfg *Config) *zap.Logger {
+//    return log.Named(cfg.Name)
+//  })
+//
```
abhinav, line 49: Worth adding:

> Similar to fx.Provide, functions passed to fx.Decorate may optionally return an error as their last result. If a decorator returns a non-nil error, it will halt application startup.

decorate.go
```diff
@@ -0,0 +1,107 @@
...
+	items := make([]string, len(o.Targets))
+	for i, f := range o.Targets {
+		items[i] = fxreflect.FuncName(f)
+	}
+	return fmt.Sprintf("fx.Decorate(%s)", strings.Join(items, ", "))
+}
+
+// provide is a single decorators used in Fx.
+type decorator struct {
+	// Constructor provided to Fx. This may be an fx.Annotated.
```
abhinav, line 83: outdated: this may not be an fx.Annotated. This may be an fx.annotated, produced by fx.Annotate.

(Can probably document separately that decorators support fx.Annotate, but not fx.Annotated.)

decorate_test.go
```diff
@@ -0,0 +1,329 @@
...
+			fx.Decorate(func() *Logger {
+				return &Logger{Name: "testRedis"}
+			}),
+			fx.Invoke(func(l *Logger) {
+				assert.Equal(t, "testRedis", l.Name)
+			}),
+		)
+
+		app := fxtest.New(t,
+			testRedis,
```
abhinav, line 56: This is fine as a test, but in the real world, I expect we will not nest "redis" under "testredis".

decorate_test.go
```diff
@@ -0,0 +1,329 @@
...
+			fx.Invoke(func(l *Logger) {
+				assert.Equal(t, "redis", l.Name)
+			}),
+		)
+		defer app.RequireStart().RequireStop()
+	})
+
+	t.Run("decorate a dependency from root", func(t *testing.T) {
```
abhinav, line 64: description does not match test

decorate_test.go
```diff
@@ -0,0 +1,329 @@
...
+			redis,
+			fx.Provide(func() *Logger {
+				assert.Fail(t, "should not run this")
+				return &Logger{Name: "root"}
+			}),
+		)
+		defer app.RequireStart().RequireStop()
+	})
+
+	t.Run("use Decorate in root", func(t *testing.T) {
```
abhinav, line 83: nice

decorate_test.go
```diff
@@ -0,0 +1,329 @@
...
+
+		takeout := fx.Module("takeout",
+			cafe,
+			fx.Decorate(fx.Annotate(func(coffee []*Coffee) []*Coffee {
+				var newC []*Coffee
+				for _, c := range coffee {
+					newC = append(newC, &Coffee{
+						Name:  c.Name,
+						Price: c.Price + 1,
+					})
```
abhinav, line 130: 😆

module.go
```diff
@@ -174,3 +176,17 @@ func (m *module) executeInvoke(i invoke) (err error) {
...
+
+func (m *module) decorate() (err error) {
+	for _, decorator := range m.decorators {
+		if err := runDecorator(m.scope, decorator); err != nil {
+			return err
+		}
+	}
+	for _, m := range m.modules {
+		if err := m.decorate(); err != nil {
+			return err
```
abhinav, line 188: failure in a decorator inside a nested module should cover this, I think.

decorate.go
```diff
@@ -0,0 +1,107 @@
...
+func (o decorateOption) apply(mod *module) {
+	for _, target := range o.Targets {
+		mod.decorators = append(mod.decorators, decorator{
+			Target: target,
+			Stack:  o.Stack,
+		})
+	}
+}
+
+func (o decorateOption) String() string {
```
abhinav, line 73: Let's test this one too. there's a string test for all the other options, I believe.

abhinav (approved):
Surprisingly simple after all of the other changes. No major concerns.
Minus the couple missing untested corner cases, I think this is good to go.

Please get a review from someone else on the team as well.

---

As an aside: 🎉 🎉 🎉

decorate_test.go
```diff
@@ -0,0 +1,353 @@
...
+				assert.Fail(t, "this should never run")
+			}),
+		)
+
+		err := app.Err()
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "missing dependencies")
+	})
+
+	t.Run("use a decorator like a provider", func(t *testing.T) {
```
r-hang, line 335:
```suggestion
	t.Run("decorate cannot provide a non-existent type", func(t *testing.T) {
```

decorate_test.go
```diff
@@ -0,0 +1,353 @@
...
+				assert.Fail(t, "this should never run")
+			}, fx.ParamTags(tag))),
+		)
+
+		err := app.Err()
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "major sadness")
+	})
+
+	t.Run("decorator missing a dependency", func(t *testing.T) {
```
r-hang, line 310:
```suggestion
	t.Run("all decorator dependencies must be provided", func(t *testing.T) {
```

decorate_test.go
```diff
@@ -0,0 +1,353 @@
...
+				}),
+			),
+		)
+
+		err := app.Err()
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "minor sadness")
+	})
+
+	t.Run("decorate the same type twice from the same Module", func(t *testing.T) {
```
r-hang, line 265:
```suggestion
	t.Run("decorating a type more than once in the same Module errors", func(t *testing.T) {
```

r-hang (approved):

akshayjshah:
Heck yeah! Nice work :)

Ardagan:
Great work!
Is there any rough estimation when this will end up in released version of fx?

sywhang:
Thanks @Ardagan! Our current estimate is to ship this by end of the month. We have some more work to do to integrate this with the fxevent package, and add fx.Replace (though there's already a PR out for that - #837). We also need to verify this doesn't break any Go services here at Uber.

After all those have been done, we should be good to go!

sywhang:
@Ardagan it's been released now: https://github.com/uber-go/fx/releases/tag/v1.17.0

PR #837, opened 2022-02-11
Add Fx.Replace

sywhang:
This adds `fx.Replace`, which replaces a value already provided in the graph with another value. This is similar to how `fx.Supply` provides an instantiated value to the graph.

For example the following code that uses fx.Decorate to modify the `*bytes.Buffer` type...
```go
fx.New(
  fx.Provide(func() *bytes.Buffer { ... }),
  fx.Decorate(func() *bytes.Buffer {
    ...
  }),
  fx.Invoke(func(b *bytes.Buffer) {
    // I get a modified buffer!
  }),
)
```

is the same as this version that uses fx.Replace:
```go
b := func() *bytes.Buffer { ... }

fx.New(
  fx.Provide(func() *bytes.Buffer { ... }),
  fx.Replace(b),
  fx.Invoke(func(b *bytes.Buffer) {
    // I get a modified buffer!
  }),
)
```

Implementation-wise, it is very similar to how fx.Supply creates a function that produces the given value using reflection and `dig.Provide`s it - fx.Replace creates a function that produces the given value with no argument using reflection, and `dig.Decorate` with that function.

### DIFF
--- a/replace_test.go
+++ b/replace_test.go
@@ -104,7 +104,7 @@ func TestReplaceSuccess(t *testing.T) {
 
 		app := fxtest.New(t,
 			fx.Supply(
-				fx.Annotate([]string{"A", "B", "C"}, fx.ResultTags(`group:"t"`)),
+				fx.Annotate([]string{"A", "B", "C"}, fx.ResultTags(`group:"t,flatten"`)),
 			),
 			fx.Replace(fx.Annotate([]string{"a", "b", "c"}, fx.ResultTags(`group:"t"`))),
 			fx.Invoke(fx.Annotate(func(ss ...string) {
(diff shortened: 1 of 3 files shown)


replace_test.go
```diff
@@ -0,0 +1,185 @@
...
+		)
+		defer app.RequireStart().RequireStop()
+	})
+
+	t.Run("replace a value group with annotate", func(t *testing.T) {
+		t.Parallel()
+
+		app := fxtest.New(t,
+			fx.Supply(
+				fx.Annotate([]string{"A", "B", "C"}, fx.ResultTags(`group:"t"`)),
```
abhinav, line 107: Is this right? Won't this end up in a `[][]string` value group? Might need a flatten here.

replace_test.go
```diff
@@ -0,0 +1,185 @@
...
+			"untyped nil passed to fx.Replace",
+			func() { fx.Replace(A{}, nil) },
+			"a naked nil should panic",
+		)
+
+		require.PanicsWithValuef(
+			t,
+			"error value passed to fx.Replace",
+			func() { fx.Replace(A{}, errors.New("some error")) },
+			"a naked nil should panic",
```
r-hang, line 177: Is this the right error message for the error case?

sywhang:
@abg thanks for catching the issue with the test. Fixed that.
> Question: if I do this, what's the value in group:"t" at the top-level/what should it be?

I believe that should be `[]string{"d", "e", "f"}`? Unless that wasn't what you were thinking. Decoration at top-level should override anything provided by child modules.

PR #839, opened 2022-02-15
fx.Supply: small code cleanup

lookfirst:


abhinav:
Hello, @lookfirst!
I'll let CI run, but this won't work for two reasons:

- Go does not do automatic fallthrough in switch statements. You have to add `fallthrough` at the end of a `case` to go to the next case.
- Type switches are a bit unique because inside the `case` statement, the compiler knows that the type of `value` is A or B. `annotated` and `Annotated` are different types, so even though the code inside them looks the same, they're accessing different fields of different types; the fields just happened to be name the same.

lookfirst:
Oh interesting, I learned something new about golang. Apologies!

https://stackoverflow.com/questions/40821855/do-go-switch-cases-fallthrough-or-not

PR #844, opened 2022-02-18
Make provide's errors more expressive for fx.Annotate-ed functions

josephinedotlee:
Because of the use of reflection in fx.Annotate, the error
messages for provide tend to refer to the annotated function as
`"reflect".makeFuncStub` which is not helpful for debugging.

This change updates error messages that previously looked like:
```
Failed: cannot provide function "reflect".makeFuncStub (/usr/local/go/src/reflect/asm_arm64.s:12):
cannot provide *fx_test.Logger from [0].Field0: already provided by "reflect".makeFuncStub
(/usr/local/go/src/reflect/asm_arm64.s:12)
```

To looking like this:
```
Failed: cannot provide function "reflect".makeFuncStub (/usr/local/go/src/reflect/asm_amd64.s:12):
cannot provide *fx_test.Logger from [0].Field0: already provided by "go.uber.org/fx_test".TestAnnotate.func1
(/gocode/src/github.com/uber-go/fx/annotated_test.go:515)
```

WIP: Planning on making a future fix that gives a more
readable name to the first function which still is referred to
as "reflect".makeFuncStub after this change.


annotated.go
```diff
@@ -264,6 +264,7 @@ type annotated struct {
 	ParamTags  []string
 	ResultTags []string
 	As         [][]reflect.Type
+	FuncInfo   dig.ProvideOption
```
moisesvega, line 267: Can this be *
```suggestion
	FuncPtr   uintptr
```

Annotated should keep track of the minimum info needed to let Dig know where this function is does not need to keep track of the whole function.

annotated.go
```diff
@@ -302,6 +303,7 @@ func (ann *annotated) Build() (interface{}, error) {
 
 	newFnType := reflect.FuncOf(paramTypes, resultTypes, false)
 	origFn := reflect.ValueOf(ann.Target)
+	ann.FuncInfo = dig.LocationForPC(origFn.Pointer())
```
moisesvega, line 306:
```suggestion
	ann.FuncPtr = origFn.Pointer()
```

annotated_test.go
```diff
@@ -787,6 +787,22 @@ func TestAnnotate(t *testing.T) {
 		defer app.RequireStart().RequireStop()
 	})
 
+	t.Run("provide an already provided function error", func(t *testing.T) {
```
moisesvega, line 790:
```suggestion
	t.Run("provide an already provided function using Annotate", func(t *testing.T) {
```

provide.go
```diff
@@ -109,6 +109,7 @@ func runProvide(c container, p provide, opts ...dig.ProvideOption) error {
 			return fmt.Errorf("fx.Provide(%v) from:\n%+vFailed: %v", constructor, p.Stack, err)
 		}
 
+		opts = append(opts, constructor.FuncInfo)
```
moisesvega, line 112:
```suggestion
		opts = append(opts, dig.LocationForPC(constructor.FuncPtr))
```

sywhang (approved):
Please address the linter error before merging. Also it'd be nice if you can edit the commit message here to mention that the "other half" of the error message that wasn't fixed by this PR is due to an issue in Dig and will be addressed in follow-up work in Dig.

PR #846, opened 2022-02-22
Add more Fx.Decorate tests

sywhang:
This adds more cases of fx.Decorate tests, specifically for
transitive dependencies that are decorated. For example, when a
function depends on type A whose constructor depends on type B,
and B has a decorator associated with it, we expect type A's constructor
to use the decorated value of B in its constructor.


decorate_test.go
```diff
@@ -213,6 +213,54 @@ func TestDecorateSuccess(t *testing.T) {
...
+					Scope: cfg.Scope + " logger",
+				}}
+			}),
+			fx.Module("child",
+				fx.Decorate(func() *Config {
+					return &Config{Scope: "child"}
+				}),
+				fx.Invoke(func(l *Logger) {
+					assert.Equal(t, "child logger", l.Cfg.Scope)
+				}),
```
abhinav, line 259: 😮

I didn't think of this one. This is nice.

abhinav (approved):

dhackner (approved):
:-)

PR #857, opened 2022-03-09
doc: Clarify Supply/Replace behavior with interface values

abhinav:
Use of `fx.Replace` or `fx.Supply` with interface values can lead to
unexpected behavior. `fx.Supply((io.Reader)(r))` will use the type of
the implementation `r`, not `io.Reader`.

This clarifies the documentation for both APIs to make this explicit.

Refs GO-1253, GO-1254

---

**Previews**

fx.Replace: 
image

fx.Supply:
image

### DIFF
--- a/replace.go
+++ b/replace.go
@@ -28,14 +28,52 @@ import (
 	"go.uber.org/fx/internal/fxreflect"
 )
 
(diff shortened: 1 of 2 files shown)


sywhang (approved):
👍

Issue #855, opened 2022-03-03
Logging for Provided/Decorated/Invoked should include Module info


sywhang:
Now that `fx.Module` is a thing, and the exact `Module` within which `fx.Provide`/`fx.Decorate`/`fx.Invoke` is called from affects the result. 

Debugging with only which function got provided/decorated/invoked is difficult, so we should add the Module information to the logs.

PR #863, opened 2022-03-22
Add module info to fxevent

sywhang:
This adds the module info in the fxevent Provided/Decorated/Invoked
structs to make the logs more descriptive by including which module
each of these happened from.

This can be helpful especially for debugging decorates, since the
module from which decoration is applied changes the results.

Fix https://github.com/uber-go/fx/issues/855.
Refs GO-1285.


fxevent/console.go
```diff
@@ -63,24 +63,24 @@ func (l *ConsoleLogger) LogEvent(event Event) {
 		if e.Err != nil {
 			l.logf("ERROR\tFailed to supply %v: %v", e.TypeName, e.Err)
 		} else {
-			l.logf("SUPPLY\t%v", e.TypeName)
+			l.logf("SUPPLY\t%v from Module \"%v\"", e.TypeName, e.ModuleName)
```
abhinav, line 66: nit: use `%q` to let the system escape the module name.

also, let's just use "module" instead of "Module" since the concept of modules has already been introduced by this point. (there's some tech writing advice I read elsewhere that went along the lines of not capitalizing reusable concepts or components if they're not proper nouns.)

```suggestion
			l.logf("SUPPLY\t%v from module %q", e.TypeName, e.ModuleName)
```


same suggestion for the other log lines

fxevent/console_test.go
```diff
@@ -131,7 +133,7 @@ func TestConsoleLogger(t *testing.T) {
 		{
 			name: "Invoking",
 			give: &Invoking{FunctionName: "bytes.NewBuffer()"},
-			want: "[Fx] INVOKE		bytes.NewBuffer()\n",
+			want: "[Fx] INVOKE		bytes.NewBuffer() from Module \"\"\n",
```
abhinav, line 136: I think maybe we should omit the "from module ..." if module is empty.

(fx.New is empty, right? That'll help differentiate top-level provide/invoke/decorates from module-level.)

fxevent/zap_test.go
```diff
@@ -131,64 +131,73 @@ func TestZapLogger(t *testing.T) {
...
 			},
 		},
 		{
 			name:        "Provide with Error",
 			give:        &Provided{Err: someError},
 			wantMessage: "error encountered while applying options",
 			wantFields: map[string]interface{}{
-				"error": "some error",
+				"error":  "some error",
+				"module": "",
```
abhinav, line 168: same. maybe we should omit this field if it's empty.

on the zap end, you can do this easily like so:

```
func moduleField(name string) zap.Field {
  if len(name) == 0 { return zap.Skip() }
  return zap.String("module", name)
}
```

abhinav (commented):
some suggestions

abhinav (approved):

PR #869, opened 2022-04-01
Emit a separate log for fx.Replace

sywhang:
This adds a new type fxevent.Replaced, which is used for logging
values that are being decorated via fx.Replace. Currently, these
are being emitted as part of the fxevent.Decorated log, but that
can be misleading and confusing since the decorator name for such
values is not properly provided (they are generated at runtime using
reflection).

The new log entry for fx.Replace-d types contains the types being
replaced, as well as the module name (if there is one).

### DIFF
--- a/fxevent/console.go
+++ b/fxevent/console.go
@@ -88,7 +88,7 @@ func (l *ConsoleLogger) LogEvent(event Event) {
 			}
 		}
 		if e.Err != nil {
-			l.logf("ERROR\tFailed to replace: %v", e.Err)
+			l.logf("ERROR\tFailed to replace: %+v", e.Err)
 		}
 	case *Decorated:
 		for _, rtype := range e.OutputTypeNames {
--- a/fxevent/console.go
+++ b/fxevent/console.go
@@ -78,6 +78,18 @@ func (l *ConsoleLogger) LogEvent(event Event) {
 		if e.Err != nil {
 			l.logf("Error after options were applied: %v", e.Err)
 		}
+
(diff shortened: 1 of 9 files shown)


fxevent/console.go
```diff
@@ -78,6 +78,18 @@ func (l *ConsoleLogger) LogEvent(event Event) {
...
+	case *Replaced:
+		for _, rtype := range e.OutputTypeNames {
+			if e.ModuleName != "" {
+				l.logf("REPLACE\t%v from module %q", rtype, e.ModuleName)
+			} else {
+				l.logf("REPLACE\t%v", rtype)
+			}
+		}
+		if e.Err != nil {
+			l.logf("ERROR\tFailed to replace: %v", e.Err)
```
abhinav, line 91:
```suggestion
			l.logf("ERROR\tFailed to replace: %+v", e.Err)
```

(#870)

fxevent/event.go
```diff
@@ -133,6 +134,19 @@ type Provided struct {
 	Err error
 }
 
+// Replaced is emitted when a decorator is provided to Fx.
+type Replaced struct {
+	// OutputTypeNames is a list of names of types that are decorated by
+	// this replacement.
```
abhinav, line 140:
```suggestion
	// OutputTypeNames is a list of names of types that were replaced.
```

abhinav (approved):
"IsReplace" is a bit hacky but I don't see a good way around it right now since replace is just sugar for decorate.

PR #875, opened 2022-04-20
Do not run lifecycle hooks with expired context

sywhang:
Currently, the lifecycle hooks are still run even when the context
passed as parameter to Start/Stop calls has expired/DeadlineExceeded.

This changes the fx lifecycle hook execution loop to check for expired
context before executing each hook, and bailing out of the exec loop
upon encountering context expiration.

Refs GO-1026.

### DIFF
--- a/internal/lifecycle/lifecycle.go
+++ b/internal/lifecycle/lifecycle.go
@@ -22,7 +22,6 @@ package lifecycle
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"io"
 	"strings"
@@ -79,7 +78,7 @@ func (l *Lifecycle) Start(ctx context.Context) error {
 
 	for _, hook := range l.hooks {
 		// if ctx has cancelled, bail out of the loop.
-		if err := ctx.Err(); errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+		if err := ctx.Err(); err != nil {
 			return err
 		}
 
@@ -137,7 +136,7 @@ func (l *Lifecycle) Stop(ctx context.Context) error {
 	// Run backward from last successful OnStart.
 	var errs []error
 	for ; l.numStarted > 0; l.numStarted-- {
-		if err := ctx.Err(); errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+		if err := ctx.Err(); err != nil {
 			return err
 		}
 		hook := l.hooks[l.numStarted-1]
--- a/internal/lifecycle/lifecycle.go
+++ b/internal/lifecycle/lifecycle.go
@@ -22,6 +22,7 @@ package lifecycle
 
 import (
 	"context"
+	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


internal/lifecycle/lifecycle.go
```diff
@@ -77,6 +77,11 @@ func (l *Lifecycle) Start(ctx context.Context) error {
 	l.mu.Unlock()
 
 	for _, hook := range l.hooks {
+		// if ctx has cancelled, bail out of the loop.
+		if err := ctx.Err(); err == context.Canceled || err == context.DeadlineExceeded {
+			return err
+		}
```
abhinav, line 83: Should this be errors.Is?

abhinav, line 83:
```suggestion
		if err := ctx.Err(); errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
			return err
		}
```

Should we use errors.Is?

internal/lifecycle/lifecycle.go
```diff
@@ -131,6 +136,9 @@ func (l *Lifecycle) Stop(ctx context.Context) error {
 	// Run backward from last successful OnStart.
 	var errs []error
 	for ; l.numStarted > 0; l.numStarted-- {
+		if ctx.Err() == context.Canceled || ctx.Err() == context.DeadlineExceeded {
+			return ctx.Err()
+		}
```
abhinav, line 141:
```suggestion
		if err := ctx.Err(); errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
			return err
		}
```

abhinav (commented):
Is there a reason to do the allowlist of errors instead of just all of them?
The contract is:

```
// If Done is not yet closed, Err returns nil.
	// If Done is closed, Err returns a non-nil [error](https://pkg.go.dev/builtin#error) explaining why:
	// Canceled if the context was canceled
	// or DeadlineExceeded if the context's deadline passed.
	// After Err returns a non-nil error, successive calls to Err return the same error.
	Err() error
```

So if err is non nil, then the context has expired no matter what.

sywhang:
> Is there a reason to do the allowlist of errors instead of just all of them?

Oh, I missed this comment. That's fair. I'll just check for non-nil errors from the context.

internal/lifecycle/lifecycle_test.go
```diff
@@ -119,6 +119,30 @@ func TestLifecycleStart(t *testing.T) {
...
+			},
+		})
+		ctx, cancel := context.WithCancel(context.Background())
+		cancel()
+		err := l.Start(ctx)
+		require.Error(t, err)
+		// Note: Stop does not return an error here because no hooks
+		// have been started, so we don't end up any of the corresponding
+		// stop hooks.
+		require.NoError(t, l.Stop(ctx))
```
abhinav, line 144: Fair. App will still error as expected on Stop, which is good.

abhinav (approved):

PR #877, opened 2022-05-04
Ensure args match verbs in ConsoleLogger

jkanywhere:
Fix the format call for `Supplied` event.

Previously, `\n%!(EXTRA string=)` was appended to `SUPPLY` logs when
`ModuleName` was empty.

Add tests to ensure format string contains the same number of verbs as
there are arguments when logging an `Event` using `ConsoleLogger` for
all relevant event types,

Module info was added to fxevent in #863.

### DIFF
--- a/fxevent/console.go
+++ b/fxevent/console.go
@@ -65,7 +65,7 @@ func (l *ConsoleLogger) LogEvent(event Event) {
 		} else if e.ModuleName != "" {
 			l.logf("SUPPLY\t%v from module %q", e.TypeName, e.ModuleName)
 		} else {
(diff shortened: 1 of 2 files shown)


sywhang (approved):
Thanks!

jkanywhere:
Example of the previous malformed log
```
[Fx] SUPPLY	envfx.Context
%!(EXTRA string=)[Fx] SUPPLY	servicefx.Metadata
%!(EXTRA string=)[Fx] PROVIDE	tally.Scope <= mypackage.Test__Module_construction_errors.func2.1()
```

PR #893, opened 2022-06-23
bump go report to show A+ instead of C, lint README.md

piotrostr:
Screen Shot 2022-06-23 at 12 20 03 PM


sywhang (approved):
Thanks for the contribution!

PR #895, opened 2022-07-01
Adds OnStart/OnStop lifecycle Annotations

jasonmills:
This provides two new annotations, OnStart and OnStop. Each will append their respective hooks into the application Lifecycle when the provided option function is executed. 

Usage:
```go
type Interface interface {
    Start(context.Context) error
    Stop(context.Context) error
}

func New() (Interface, error)

var Module = fx.Provide(fx.Annotate(
    New,
    fx.OnStart(func(ctx context.Context, server Interface) error {
      return server.Start(ctx)
    }),
    fx.OnStop(func(ctx context.Context, server Interface) error {
       return server.Stop(ctx)
   }),
))   
```
Which is functionally equivalent to:

```go
var Module = fx.Provide(
  func(lifecycle fx.Lifecycle) (Interface, error) {
    server, err := New()
    lifecycle.Append(fx.Hook{
      OnStart: func(ctx context.Context) error { return server.Start(ctx) }, 
      OnStop: func(ctx context.Context) error { return server.Stop(ctx) },
    }
    return server, err
  },
)  
``` 

Note for reviewers:
* The underlying implementation is unified for each hook type and may be easily extended should additional hook fields be added to the Hook type.
* This slightly modifies the parameter implementation of the private annotation type, this was required in order to make the hook annotation types able to detect when a given annotated function provides a required hook (closure) parameter. Without this any hook that takes a dependency on a type that is constructed in the same annotation would trigger a circular dependency.

### DIFF
--- a/annotated.go
+++ b/annotated.go
@@ -306,7 +306,6 @@ func (la *lifecycleHookAnnotation) resolveMap(results []reflect.Type) (
 		}
 	}
 
-	fmt.Printf("Result map %+v\n", resultMap)
 	return
 }
 
@@ -319,7 +318,7 @@ func (la *lifecycleHookAnnotation) resolveLifecycleParamField(
 	if param.Kind() == reflect.Struct {
 		nf := param.NumField()
 		if n <= nf {
-			value = param.FieldByName(fmt.Sprintf("Field%d", n-1))
+			value = param.FieldByName(fmt.Sprintf("Field%d", n))
 		}
 	}
 
@@ -375,10 +374,7 @@ func (la *lifecycleHookAnnotation) parameters(results ...reflect.Type) (
 		}
 		params = append(params, field)
 
-		resolver := func(v reflect.Value, pos int) (value reflect.Value) {
-			value = la.resolveLifecycleParamField(v, i)
-			return
-		}
+		resolver := la.resolveLifecycleParamField
 
 		resolverIdx = append(resolverIdx, argSource{
 			pos:     i,
--- a/annotated.go
+++ b/annotated.go
@@ -223,7 +223,7 @@ func (la *lifecycleHookAnnotation) String() string {
 func (la *lifecycleHookAnnotation) apply(ann *annotated) error {
 	if la.Target == nil {
 		return fmt.Errorf(
-			"cannot use nil function for %v hook annotation",
+			"cannot use nil function for %q hook annotation",
 			la,
 		)
 	}
@@ -231,7 +231,7 @@ func (la *lifecycleHookAnnotation) apply(ann *annotated) error {
 	for _, h := range ann.Hooks {
 		if la.Type == h.Type {
 			return fmt.Errorf(
-				"cannot apply more than one %v hook annotation",
+				"cannot apply more than one %q hook annotation",
 				la,
 			)
 		}
@@ -240,7 +240,8 @@ func (la *lifecycleHookAnnotation) apply(ann *annotated) error {
 	ft := la.targetType()
 	if ft.Kind() != reflect.Func {
 		return fmt.Errorf(
-			"must provide function for hook, got %v (%T)",
+			"must provide function for %q hook, got %v (%T)",
+			la,
 			la.Target,
 			la.Target,
 		)
@@ -319,8 +320,7 @@ func (la *lifecycleHookAnnotation) resolveLifecycleParamField(
 	value reflect.Value,
 ) {
 	if param.Kind() == reflect.Struct {
-		nf := param.NumField()
-		if n <= nf {
+		if n <= param.NumField() {
 			value = param.FieldByName(fmt.Sprintf("Field%d", n))
 		}
 	}
@@ -471,7 +471,9 @@ func (la *lifecycleHookAnnotation) Build(results ...reflect.Type) (reflect.Value
 //    }),
 //  )
 //
-// Only one OnStart annotation may be applied to a given function at a time.
+// Only one OnStart annotation may be applied to a given function at a time,
+// however functions may be annotated with other types of lifecylce Hooks, such
+// as OnStop.
 func OnStart(onStart interface{}) Annotation {
 	return &lifecycleHookAnnotation{
 		Type:   _onStartHookType,
@@ -491,7 +493,9 @@ func OnStart(onStart interface{}) Annotation {
 //    }),
 //  )
 //
-// Only one OnStop annotation may be applied to a given function at a time.
+// Only one OnStop annotation may be applied to a given function at a time,
+// however functions may be annotated with other types of lifecylce Hooks, such
+// as OnStart.
 func OnStop(onStop interface{}) Annotation {
 	return &lifecycleHookAnnotation{
 		Type:   _onStopHookType,
--- a/annotated.go
+++ b/annotated.go
@@ -730,10 +730,10 @@ func (ann *annotated) parameters(results ...reflect.Type) (
 
 	// append required types for hooks to types field, but do not
 	// include them as params in constructor call
-	for h, t := range ann.Hooks {
+	for i, t := range ann.Hooks {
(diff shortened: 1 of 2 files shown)


annotated.go
```diff
@@ -195,6 +196,310 @@ func ResultTags(tags ...string) Annotation {
...
+// on the Lifecycle type.
+//
+//  fx.Annotate(
+//    func(...) Server { ... },
+//    fx.OnStart(func(ctx context.Context, server Server) error {
+//        return server.Listen(ctx)
+//    }),
+//  )
+//
+// Only one OnStart annotation may be applied to a given function at a time.
```
sywhang, line 475: Let's add a clarification on this - we can provide OnStart/OnStop annotations to the same function at the same time. Just not two OnStarts.

sywhang:
also it looks like this branch is out-of-date with the master branch - can you rebase on master?

annotated_test.go
```diff
@@ -983,4 +984,519 @@ func TestAnnotate(t *testing.T) {
...
+	t.Parallel()
+
+	validateApp := func(t *testing.T, opts ...fx.Option) error {
+		return fx.ValidateApp(
+			append(opts, fx.Logger(fxtest.NewTestPrinter(t)))...,
+		)
+	}
+
+	t.Run("depend on result interface of target", func(t *testing.T) {
+		//t.Skip()
```
moisesvega, line 1001:
```suggestion

```

annotated_test.go
```diff
@@ -983,4 +984,519 @@ func TestAnnotate(t *testing.T) {
...
+					},
+					fx.OnStart(func(_ context.Context, s stub) error {
+						started = true
+						if !assert.Equal(t, "expected", s.String()) {
+							return fmt.Errorf(
+								"expected %q got %q",
+								"expected",
+								s.String(),
+							)
+						}
```
moisesvega, line 1023: This will fail the test directly instead of waiting for `require.NoError`

```suggestion
						require.Equal(t, "expected", s.String()) 
```

annotated.go
```diff
@@ -195,6 +196,310 @@ func ResultTags(tags ...string) Annotation {
...
+}
+
+func (la *lifecycleHookAnnotation) apply(ann *annotated) error {
+	name := "UnknownHookAnnotation"
+	switch la.Type {
+	case _onStartHookType:
+		name = _onStartHook
+	case _onStopHookType:
+		name = _onStopHook
+	}
```
moisesvega, line 219: Looks like name is only used in errors here, can we save an alloc for successful cases by implementing a stringer interface for `_lifecycleHookAnnotationType` and moving this function there.

annotated.go
```diff
@@ -195,6 +196,310 @@ func ResultTags(tags ...string) Annotation {
...
+
+	if la.Target == nil {
+		return fmt.Errorf(
+			"cannot use nil function for %v hook annotation",
```
moisesvega, line 223:
```suggestion
			"cannot use nil function for %q hook annotation",
```

annotated.go
```diff
@@ -195,6 +196,310 @@ func ResultTags(tags ...string) Annotation {
...
+			name,
+		)
+	}
+
+	for _, h := range ann.Hooks {
+		if la.Type == h.Type {
+			return fmt.Errorf(
+				"cannot apply more than one %v hook annotation",
```
moisesvega, line 231:
```suggestion
				"cannot apply more than one %q hook annotation",
```

annotated.go
```diff
@@ -195,6 +196,318 @@ func ResultTags(tags ...string) Annotation {
...
+}
+
+var (
+	_typeOfLifecycle reflect.Type = reflect.TypeOf((*Lifecycle)(nil)).Elem()
+	_typeOfContext   reflect.Type = reflect.TypeOf((*context.Context)(nil)).Elem()
+)
+
+func (la *lifecycleHookAnnotation) targetType() (targetType reflect.Type) {
+	return reflect.TypeOf(la.Target)
+}
```
sywhang, line 288: do we really need this to be its own method when it's just a simple one liner function call?

annotated.go
```diff
@@ -195,6 +196,318 @@ func ResultTags(tags ...string) Annotation {
...
+			Type: t,
+		}
+		params = append(params, field)
+
+		resolver := la.resolveLifecycleParamField
+
+		resolverIdx = append(resolverIdx, argSource{
+			pos:     i,
+			resolve: resolver,
+		})
```
sywhang, line 385:
```suggestion
		resolverIdx = append(resolverIdx, argSource{
			pos:     i,
			resolve: la.resolveLifecycleParamField,
		})
```

sywhang (commented):
👏 Coming together nicely - left one more round of reviews, mainly around cleanliness and readability. 

Also try to edit the PR message here to include a brief description of the API and how one might go about using this, since a lot of people come across the PRs as part of the documentation (which we do lack but that's a separate story).

annotated.go
```diff
@@ -195,6 +196,318 @@ func ResultTags(tags ...string) Annotation {
...
+	}
+
+	in = reflect.StructOf(params)
+
+	argmap = func(
+		args []reflect.Value,
+	) (lc Lifecycle, remapped []reflect.Value) {
+		remapped = make([]reflect.Value, ft.NumIn())
```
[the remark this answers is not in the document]
jasonmills: This seems (as well as the other ft.NumIn offset suggestion above) to break things, I believe it's due to having to account for the always present Lifecycle parameter added to the parameters (in order to allow us to be able to append to it)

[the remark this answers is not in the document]
sywhang: ah, i see. sounds reasonable.

annotated_test.go
```diff
@@ -983,4 +984,460 @@ func TestAnnotate(t *testing.T) {
...
+
+func TestHookAnnotations(t *testing.T) {
+	t.Parallel()
+
+	t.Run("with hook on invoke", func(t *testing.T) {
+		t.Parallel()
+
+		var called bool
+		hook := fx.Annotate(
+			func() {},
```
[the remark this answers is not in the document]
jasonmills: Am empty constructor, I'll change it to something more meaningful.

annotated.go
```diff
@@ -195,6 +196,318 @@ func ResultTags(tags ...string) Annotation {
...
+	}
+
+	type valueResolver func(reflect.Value, int) reflect.Value
+	type argSource struct {
+		pos     int
+		result  bool
+		resolve valueResolver
+	}
+
+	resolverIdx := make([]argSource, 1)
```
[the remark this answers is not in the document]
jasonmills: Same as below, changing this messes with the lifecycle offset. Will discuss in message if this is still an issue after next review.

annotated.go
```diff
@@ -195,6 +196,318 @@ func ResultTags(tags ...string) Annotation {
...
+// on the Lifecycle type.
+//
+//  fx.Annotate(
+//    func(...) Server { ... },
+//    fx.OnStart(func(ctx context.Context, server Server) error {
+//        return server.Listen(ctx)
+//    }),
+//  )
+//
+// Only one OnStart annotation may be applied to a given function at a time,
```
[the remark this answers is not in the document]
jasonmills: Thanks, nesting annotations would definitely be nice.

annotated_test.go
```diff
@@ -983,4 +984,432 @@ func TestAnnotate(t *testing.T) {
...
+			),
+		},
+		{
+			name:        "with hook that errors",
+			errContains: "OnStart hook failed",
+			useNew:      true,
+			annotation: fx.Annotate(
+				func() (A, error) { return nil, nil },
+				fx.OnStart(func(context.Context) error {
+					return errors.New("hook failed")
```
sywhang, line 1321: This should be changed to "OnStart hook failed"

sywhang (approved):

Issue #892, opened 2022-06-10
Allow modules to be provided multiple times application


mikhasd:
I'm currently working on reusable modules for common infrastructure like database, MQ, ID provider, etc and these modules may also depends on other sub-modules, such as a config server.

Ideally, I'd like to declare my modules as follows:

```go
package config

var Module = fx.Module(
  "config",
  fx.Provide(
    NewConfigServer,
  ),
)

```
```go
package database

var Module = fx.Module(
  "database",
  config.Module,
  fx.Provide(
    NewOracleConn, // <- uses config server
  ),
)
```
```go
package oidc

var Module = fx.Module(
  "oidc",
  config.Module,
  fx.Provide(
    NewOidcProvider, // <- uses config server
  ),
)
```
```go
package main

var MainModule = fx.Module(
  "main",
  database.Module,
  oidc.Module,
)

func main(){
  fx.New(MainModule).Run()
}
```

Currently, if I declare my modules as previously mentioned, I'll get an "already provided" error, which is not ideal because the user of the database and oidc modules would not be aware of it's dependencies.

The framework, could keep track of the modules that have already populated the container and skip it if already processed.

sywhang:
Hey there, apologies for the delay in response. 

We do not allow the same type to be provided multiple times in the application. We're planning to work on a public documentation site for fx that will explain some of the core thoughts behind this, but just referring to the specific case you are bringing up here, it is easier if there is a top-level library that groups and provides the most commonly used dependencies (i.e. things like config, or logger, or metrics). 

> The framework, could keep track of the modules that have already populated the container and skip it if already processed.
The error is specifically coming from the same `fx.Provide` being provided multiple times. The reason we do not allow this currently is because `Provide`s don't run "in-order" and is run "lazily" as needed by `Invoke`s or other `Provide`s that depend on them. If we blindly 'skipped' a constructor that's provided, it's hard to find out which one actually ran. In the example above, the constructors are provided by the same `fx.Module`, so there is no side effect. But if there are multiple provides, that's a separate story.

That being said, if we can track things by a `Module` basis, that would make more sense. I'll have to think about the implications of this some more, but I'm open to this idea.

mikhasd:
I'm trying to achieve a functional behavior similar to spring boot's `@Configuration` (difficult not to compare) where I can import a  configuration class in many different points of my application and the container will only instantiate it once.

[...]

PR #896, opened 2022-07-07
Allow modules to be provided multiple times #892

mikhasd:
Naive implementation preventing modules from being processed multiple times by the container.

I find quite difficult to implement this feature in other points of the initialization process such as `func (m *module) build`  because `Option`s were (until my PR) value objects with no unique identity.

This PR probably won't be merged, but raise awareness that the feature can be implemented without breaking the library and its tests.


mikhasd:
Need to review with my employer before signing the CLA

PR #909, opened 2022-08-03
Add documentation and tests for soft group values

EstebanOlmedo:
This adds documentation regarding soft group values feature and also
adds test cases to check this feature works with modules, `Supply`,
`Decorate` and `Annotate`.

### DIFF
--- a/annotated_test.go
+++ b/annotated_test.go
@@ -444,6 +444,22 @@ func TestAnnotatedWrongUsage(t *testing.T) {
 		)
 		assert.Contains(t, app.Err().Error(), "embeds a dig.In", "expected error when result types were annotated")
 	})
+
+	t.Run("invalid group option", func(t *testing.T) {
+		t.Parallel()
+
+		app := NewForTest(t,
+			fx.Provide(
+				fx.Annotated{
+					Group: "foo,soft",
+					Target: func() string {
+						return "sad times"
+					},
+				},
+			),
+		)
+		assert.Contains(t, app.Err().Error(), "cannot use soft with result value groups", "expected error when invalid group option is provided")
+	})
 }
 
 func TestAnnotatedString(t *testing.T) {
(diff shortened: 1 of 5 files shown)


annotated_test.go
```diff
@@ -444,6 +444,22 @@ func TestAnnotatedWrongUsage(t *testing.T) {
...
+		t.Parallel()
+
+		app := NewForTest(t,
+			fx.Provide(
+				fx.Annotated{
+					Group: "foo,soft",
+					Target: func() string {
+						return "sad times"
+					},
+				},
```
sywhang, line 458: Can you use `fx.Annotate` instead? We're trying to deprecate `Annotated`

```suggestion
				fx.Annotate(func() string {
						return "sad times"
				}, fx.ResultTags(`group:"foo,soft"`))
```

decorate_test.go
```diff
@@ -185,6 +186,55 @@ func TestDecorateSuccess(t *testing.T) {
...
+		app := fxtest.New(t,
+			fx.Provide(
+				fx.Annotate(
+					func() (string, int) { return "cheeseburger", 15 },
+					fx.ResultTags(`group:"burger"`, `group:"potato"`),
+				),
+			),
+			fx.Provide(
+				fx.Annotate(
+					func() (string, int) { return "mushroomburger", 35 },
```
sywhang, line 199: non-review comment: mushroom burger...? 🤣

inout.go
```diff
@@ -163,6 +163,26 @@ import "go.uber.org/dig"
 // Note that values in a value group are unordered. Fx makes no guarantees
 // about the order in which these values will be produced.
 //
+// To declare a soft relationship between a group an its constructors, use the
+// `soft` option on the group tag (`group:"[groupname],soft"`). A soft group
+// will be populated only with values from already-executed constructors, this
+// means constructors of soft value groups will be called only if they provide
+// another requested value.
```
sywhang, line 170:
```suggestion
// will be populated only with values from already-executed constructors.
```

The last statement isn't quite true. If there's another function consuming the value then it will still populate it. 

It's better to describe soft value groups as specifying the parameter as an 'opportunistic' consumer - if the values are already provided in the DI container, it consumes them. Otherwise, it doesn't.

You also want to specify that this is only a valid option for input parameters in the docs.

inout.go
```diff
@@ -163,6 +163,26 @@ import "go.uber.org/dig"
...
+//
+//   type Params struct {
+//	  fx.In
+//
+//    Handlers []Handler `group:"server"`
+//    Logger   *zap.Logger
```
sywhang, line 176: indentation seems off here

inout.go
```diff
@@ -163,6 +163,26 @@ import "go.uber.org/dig"
...
+//   }
+//   func NewHandlerAndLogger() (Handler, *zap.Logger) { ... }
+//   func NewHandler() Handler { ... }
+//   func Foo(Params) { ... }
+//
+// When `NewHandlerAndLogger` and `NewHandler`are provided, and `Foo` invoked,
+// the only constructor called is `NewHandler`, because this also provides
+// `*zap.Logger` needed in the `Params` struct.
+//
```
sywhang, line 185: Apart from this example, can you also provide an example where:

1. If you have a constructor for a type, and the only consumer is the soft value group, it is not populated. 
2. In the exact same case above, add a consumer that also consumes that type. Then show that it is populated.

module_test.go
```diff
@@ -76,6 +76,42 @@ func TestModuleSuccess(t *testing.T) {
...
+		type Param struct {
+			fx.In
+
+			Foos []string `group:"foo,soft"`
+			Bar  int
+		}
+		type Result struct {
+			fx.Out
+
+			Foo string `group:"foo"`
```
sywhang, line 90: can you add an error case where `soft` is specified to an fx.Out struct? Also one where it's being speified using fx.Annotate?

i.e.
`fx.Annotate(someFunc, fx.ResultTags(`group:"something,soft"`))`
should error out. 
I believe that won't error out currently, because you haven't changed the implementation.

Take a look at annotated.go and look at the `apply` method for `resultTagsAnnotation`. You probably want to add something there 🙂

sywhang:
Also it looks like you have a merge conflict. You might want to rebase master :)

decorate_test.go
EstebanOlmedo: While I was writing this I couldn't remember the name of another burger hahaha

sywhang (approved):

PR #912, opened 2022-08-04
Implements ShutdownCode option and ShutdownSignal os.Signal wrapper

jasonmills:
Provides API for users of the `Shutdowner` interface to specify an exit code that should be used to exit an application.
Addresses issue #763

* Adds `ShutdownCode` shutdown option
* Adds `ShutdownSignal` type which includes fields for both os.Signal and exit code
* Adds `Wait()` method which operates like `Done()` but instead of returning a os.Signal channel, returns a `ShutdownSignal` channel
* Adds `ShutdownCode` option example test
* Includes broadcast signaling refactor by @abhinav 

```go
	app := fx.New(
		fx.Invoke(func(shutdowner fx.Shutdowner) {
			// Call the shutdowner Shutdown method with a shutdown code
			// option
			shutdowner.Shutdown(fx.ShutdownCode(1))
		}),
	)

	app.Run()

	wait := app.Wait()

	signal := <-wait

	os.Exit(signal.ExitCode)
```

### DIFF
--- a/app.go
+++ b/app.go
@@ -275,7 +275,9 @@ type App struct {
 	err       error
 	clock     fxclock.Clock
 	lifecycle *lifecycleWrapper
-	stopch    chan struct{} // closed when Stop is called
+
+	stopch     chan struct{} // closed when Stop is called
+	stopChLock sync.RWMutex  // mutex for init and closing of stopch
 
 	container *dig.Container
 	root      *module
@@ -400,7 +402,6 @@ func New(opts ...Option) *App {
 		clock:        fxclock.System,
 		startTimeout: DefaultTimeout,
 		stopTimeout:  DefaultTimeout,
-		stopch:       make(chan struct{}),
 	}
 
 	app.root = &module{
@@ -563,6 +564,7 @@ func (app *App) run(done <-chan ShutdownSignal) (exitCode int) {
 	defer cancel()
 
 	if err := app.Start(startCtx); err != nil {
+		app.closeStopChannel()
 		return 1
 	}
 
@@ -628,6 +630,8 @@ func (app *App) Start(ctx context.Context) (err error) {
 		return app.err
 	}
 
+	app.initStopChannel()
+
 	return withTimeout(ctx, &withTimeoutParams{
 		hook:      _onStartHook,
 		callback:  app.start,
@@ -653,6 +657,30 @@ func (app *App) start(ctx context.Context) error {
 	return nil
 }
 
+func (app *App) initStopChannel() {
+	app.stopChLock.Lock()
+	defer app.stopChLock.Unlock()
+	if app.stopch == nil {
+		app.stopch = make(chan struct{})
+	}
+}
+
+func (app *App) stopChannel() chan struct{} {
+	app.stopChLock.RLock()
+	defer app.stopChLock.RUnlock()
+	ch := app.stopch
+	return ch
+}
+
+func (app *App) closeStopChannel() {
+	app.stopChLock.Lock()
+	defer app.stopChLock.Unlock()
+	if app.stopch != nil {
+		close(app.stopch)
+		app.stopch = nil
+	}
+}
+
 // Stop gracefully stops the application. It executes any registered OnStop
 // hooks in reverse order, so that each constructor's stop hooks are called
 // before its dependencies' stop hooks.
@@ -666,7 +694,7 @@ func (app *App) Stop(ctx context.Context) (err error) {
 		// Protect the Stop hooks from being called multiple times.
 		app.runStop.Do(func() {
 			app.log().LogEvent(&fxevent.Stopped{Err: err})
-			close(app.stopch)
+			app.closeStopChannel()
 		})
 	}()
 
@@ -724,10 +752,14 @@ func (app *App) appendSignalReceiver(r signalReceiver) {
 		sigch := make(chan os.Signal, 1)
 		signal.Notify(sigch, os.Interrupt, _sigINT, _sigTERM)
 		go func() {
-			select {
-			case sig := <-sigch:
-				app.broadcastSignal(sig, 1)
-			case <-app.stopch:
+			// if the stop channel is nil; that means that the app was never started
+			// thus, do not broadcast any signals
+			if stopch := app.stopChannel(); stopch != nil {
+				select {
+				case sig := <-sigch:
+					app.broadcastSignal(sig, 1)
+				case <-stopch:
+				}
 			}
 		}()
 	})
--- a/app.go
+++ b/app.go
@@ -400,7 +400,9 @@ func New(opts ...Option) *App {
 		clock:        fxclock.System,
 		startTimeout: DefaultTimeout,
 		stopTimeout:  DefaultTimeout,
+		stopch:       make(chan struct{}),
 	}
+
 	app.root = &module{
 		app: app,
 		// We start with a logger that writes to stderr. One of the
@@ -415,11 +417,7 @@ func New(opts ...Option) *App {
 		// user gave us. For the last case, however, we need to fall
 		// back to what was provided to fx.Logger if fx.WithLogger
 		// fails.
-		log:          logger,
-		clock:        fxclock.System,
-		startTimeout: DefaultTimeout,
-		stopTimeout:  DefaultTimeout,
-		stopch:       make(chan struct{}),
+		log: logger,
 	}
 	app.modules = append(app.modules, app.root)
 
@@ -569,7 +567,7 @@ func (app *App) run(done <-chan ShutdownSignal) (exitCode int) {
 	}
 
 	sig := <-done
-	app.log.LogEvent(&fxevent.Stopping{Signal: sig.Signal})
+	app.log().LogEvent(&fxevent.Stopping{Signal: sig.Signal})
 
 	stopCtx, cancel := app.clock.WithTimeout(context.Background(), app.StopTimeout())
 	defer cancel()
@@ -621,7 +619,7 @@ var (
 func (app *App) Start(ctx context.Context) (err error) {
 	defer func() {
 		app.runStart.Do(func() {
-			app.log.LogEvent(&fxevent.Started{Err: err})
+			app.log().LogEvent(&fxevent.Started{Err: err})
 		})
 	}()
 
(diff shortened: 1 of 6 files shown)


sywhang (commented):
Great progress. Didn't fully go through the implementation, but I'm not so sure if this is what we want; it's a little awkward. So, what we have now is:

1. Call `Done()` to get a channel
2. Receive from that channel
3. Create a ShutdownSignal from that channel
4. Check its exit code

An easier thing to do might be exposing a method under `App` that returns a different channel than the one returned by `App.Done()`.

Say:
```
app := fx.App(
  // .. stuff
  fx.ShutdownCode(100),
)

exit := <-app.Exit() // name TBD
os.Exit(exit.ExitCode()) // exits with 100
```

so `exit` is a channel that receives `ShutdownerSignal` type.

With this, we get:
1. Call `Exit()` to get a channel
2. Receive from that channel 
3. Check its exit code.

which removes the awkward step of having to create a random type out of something returned from the Done channel.

shutdown_code_example_test.go
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) 2019 Uber Technologies, Inc.
```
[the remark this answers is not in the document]
jasonmills: copy & paste from another file, I'll update it.

shutdown_code_example_test.go
```diff
@@ -0,0 +1,48 @@
...
+		}),
+	)
+
+	app.Run()
+
+	// Extract the shutdown signal from the os.Signal returned by app.Done
+	signal := fx.ShutdownSignal(<-app.Done())
+
+	// Retrieve the exit code
+	fmt.Println(fmt.Sprintf("os.Exit(%v)", signal.ExitCode()))
```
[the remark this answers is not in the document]
jasonmills: Isn't the assertion on example tests contingent on a print line?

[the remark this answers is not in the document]
sywhang: sorry, yes. I meant to delete this review before posting it. My bad

shutdown_code_example_test.go
```diff
@@ -0,0 +1,48 @@
...
+func ExampleShutdownCode() {
+	app := fx.New(
+		fx.Invoke(func(shutdowner fx.Shutdowner) {
+			// Call the shutdowner Shutdown method with a shutdown code
+			// option
+			shutdowner.Shutdown(fx.ShutdownCode(1))
+		}),
+	)
+
+	app.Run()
```
jasonmills, line 38: Make sure Run method supports exiting with the correct error code.

app.go
```diff
@@ -296,9 +296,12 @@ type App struct {
 	errorHooks []ErrorHandler
 	validate   bool
 	// Used to signal shutdowns.
-	donesMu     sync.Mutex // guards dones and shutdownSig
-	dones       []chan os.Signal
-	shutdownSig os.Signal
+	donesMu        sync.Mutex // guards dones and shutdownSig
+	dones          []chan os.Signal
+	shutdownSig    os.Signal
+	waitsMu        sync.Mutex // guards waits and shutdownCode
```
abhinav, line 302:
```suggestion
	waitsMu        sync.Mutex // guards waits and shutdownSignal
```

app.go
```diff
@@ -749,6 +752,21 @@ func (app *App) Done() <-chan os.Signal {
...
 
+func (app *App) Wait() <-chan ShutdownSignal {
+	c := make(chan ShutdownSignal, 1)
+
+	app.waitsMu.Lock()
+	defer app.waitsMu.Unlock()
+
+	if app.shutdownSignal != nil {
+		c <- *app.shutdownSignal
+		return c
```
abhinav, line 763: Let's add a quick comment here indicating what this special case represents.

```suggestion
	if app.shutdownSignal != nil {
		// Already shut down.
		c <- *app.shutdownSignal
		return c
```

shutdown.go
```diff
@@ -74,10 +105,61 @@ func (app *App) broadcastSignal(signal os.Signal) error {
...
+			Code:     code,
+			Channels: len(app.waits),
+		}
+	}
+
+	return nil
+}
+
+// ErrOnUnsentSignal ... TBD
+type ErrOnUnsentSignal struct {
```
abhinav, line 151: Is there a good use case for users to want to match on this error?
If not, we can keep it unexported (errUnsentSignal).
(If we want to export it, we can call it ErrUnsentSignal.)

shutdown.go
```diff
@@ -49,14 +69,25 @@ type shutdowner struct {
...
 func (app *App) shutdowner() Shutdowner {
 	return &shutdowner{app: app}
 }
 
-func (app *App) broadcastSignal(signal os.Signal) error {
+func (app *App) broadcastSignal(signal os.Signal, code int) error {
+	return multierr.Combine(
+		app.broadcastDoneSignal(signal),
+		app.broadcastWaitSignal(signal, code),
+	)
```
abhinav, line 87: optional/ux improvement:
if both functions return errUnsentSignal,
we should maybe merge those numbers into a single error.
Otherwise we get an error like,

    The following errors occurred:
    - failed to send SIGINT signal to 5 out of 10 channels
    - failed to send SIGINT signal to 7 out of 9 channels

We can keep them separate if we annotate each error with `Wait()` and `Done()`:

    failed to send SIGINT signal to 7 out of 9 Done() channels
    failed to send SIGINT signal to 5 out of 10 Wait() channels

shutdown.go
```diff
@@ -74,10 +105,61 @@ func (app *App) broadcastSignal(signal os.Signal) error {
...
+type ErrOnUnsentSignal struct {
+	Signal   os.Signal
+	Unsent   int
+	Code     int
+	Channels int
+}
+
+func (err ErrOnUnsentSignal) Error() string {
+	return fmt.Sprintf(
+		"failed to send %v signal to %v out of %v channels",
```
abhinav, line 160: I think the error message can be more concise.
This error message was written when we hadn't internalized the error message guidance of omitting "failed to x" and "error doing x" from messages. (This is now part of our style guide.)

How about something like:

    send %v signal: %v/%v channels are blocked

app.go
```diff
@@ -296,9 +296,12 @@ type App struct {
...
 	// Used to signal shutdowns.
-	donesMu     sync.Mutex // guards dones and shutdownSig
-	dones       []chan os.Signal
-	shutdownSig os.Signal
+	donesMu        sync.Mutex // guards dones and shutdownSig
+	dones          []chan os.Signal
+	shutdownSig    os.Signal
+	waitsMu        sync.Mutex // guards waits and shutdownCode
+	waits          []chan ShutdownSignal
+	shutdownSignal *ShutdownSignal
```
abhinav, line 304: I think we can simplify state management here a bit by collapsing dones and waits into a single list/mutex and recording only shutdownSignal, not shutdownSig.

Basically, let's implement Done (the weaker API) in terms of Wait (the richer API).
The problem is `Wait()` returns a `chan ShutdownSignal` and `Done()` returns a `chan os.Signal`.
Mapping ShutdownSignal to Signal is possible, but using the current setup, requires a goroutine or similar to map. That's not a good idea.

Instead, we can unify the internal state here with an interface for the signal receiver (the channel that takes os.Signal or ShutdownSignal from App) and "discard" the extra information from ShutdownSignal in the os.Signal implementation.

Basically, imagine a private interface:

```go
// Returns an error if the channel wasn't receiving.
type signalRecevier interface{ ReceiveSignal(ShutdownSignal) error }
```

Then the two implementations are,

```go
type osSignalReceiver struct{ ch chan os.Signal<- }

func (r *osSignalReceiver) ReceiveSignal(sig ShutdownSignal) error {
  select {
  case r.ch <- sig.Signal:
    return nil
  default:
    return errChannelBlocked  // define this somewhere
  }
}
```

The shutdownSignalReceiver implementation will do the same but it'll stick the `sig` in as-is.

Then internal state management is just a `[]signalReceiver` instead of `[]chan os.Signal` and `[]chan ShutdownSignal`, and the os.Signal field can be dropped completely from App.

Both, App.Wait and App.Done will just build a signalReceiver and call a private method.
Something like this setup.

```go
func (app *App) Wait() <-chan ShutdownSignal {
  c := make(chan ShutdownSignal, 1)
  app.onShutdown(&shutdownSignalReceiver{ch: c})
  return c
}

func (app *App) Done() <-chan os.Signal {
  c := make(chan os.Signal, 1)
  app.onShutdown(&osSignalReceiver{ch: c})
  return c
}
```

shutdown.go
jasonmills: It was just a guess on my part, assuming people want to know the signal and exit code information to handle it, as opposed to some other error type. It's probably wiser to just make it private however, and if someone raises the use-case we can make it public then.

app.go
```diff
@@ -646,16 +689,23 @@ func (app *App) start(ctx context.Context) error {
...
 // fail.
 func (app *App) Stop(ctx context.Context) (err error) {
+
 	defer func() {
-		app.log().LogEvent(&fxevent.Stopped{Err: err})
+		// Protect the Stop hooks from being called multiple times.
+		app.runStop.Do(func() {
+			app.log().LogEvent(&fxevent.Stopped{Err: err})
+			app.closeStopChannel()
+		})
```
abhinav, line 698: Same. I'm not sure we need this guard? If there's a stop channel because of Start, then there should be a defer close.

app.go
```diff
@@ -286,10 +289,16 @@ type App struct {
...
 	// Used to signal shutdowns.
-	donesMu     sync.Mutex // guards dones and shutdownSig
-	dones       []chan os.Signal
-	shutdownSig os.Signal
+	shutdownMu   sync.Mutex
+	shutdownSig  *ShutdownSignal
+	sigReceivers []signalReceiver
+	signalOnce   sync.Once
+
+	// Used to make sure Start/Stop is called only once.
```
sywhang, line 299: do you need this to be included in this PR? I'm asking because this stuff was supposed to ship with the previous version (v1.17.2) and it broke some users: see https://github.com/uber-go/fx/issues/945 and https://github.com/uber-go/fx/pull/950 for context.

jasonmills:
I'm going to abandon this PR and separate it into some smaller chunks since it has some scope creep.

PR #913, opened 2022-08-05
Docs change for soft value groups

sywhang:
This modifies some wordings in the docs for soft value groups and
adds some clarifications.

### DIFF
--- a/inout.go
+++ b/inout.go
@@ -170,6 +170,11 @@ import "go.uber.org/dig"
 // if a constructor's output type is only consumed by a soft value group,
 // it will not be run.
 //
+// Note that Fx does not guarantee precise execution order of constructors
+// or invokers, which means that the change in code that affects execution
+// ordering of other constructors or functions will affect the values
+// populated in this group.
+//
 // To declare a soft relationship between a group and its constructors, use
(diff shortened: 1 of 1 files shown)


inout.go
```diff
@@ -163,22 +163,27 @@ import "go.uber.org/dig"
 // Note that values in a value group are unordered. Fx makes no guarantees
 // about the order in which these values will be produced.
 //
+// # Soft Value Groups
+//
+// A soft value group can be thought of as a best-attempt at populating the
+// group with values from constructors that have already run. In other words,
+// if a constructor's output type is only consumed by a soft value group,
+// it will not be run.
```
EstebanOlmedo, line 171: Do you think that it should clarify that fx doesn't guarantee to populate the soft value groups with all the values from constructors that are called after the creation of te `fx.App`, that it only guarantees to not call constructors that only provides values to the soft group?

EstebanOlmedo (approved):
I think it's fine for me, just left a comment about a clarification.

inout.go
sywhang: 👍 Sg. Let me update that.

PR #915, opened 2022-08-06
Populate add fx.Annotated support

kkHAIKE:
simply use `Populate(Annotated{Name: "n1", Target: &v1})` to Expect named depend. 😊


sywhang:
@kkHAIKE thanks for the contribution, but we're currently trying to deprecate `fx.Annotated` in favor of `fx.Annotate`. While we really appreciate you taking the effort to add this, we will not be merging this PR since we don't want to iterate on soon-to-be-deprecated API.

Issue #918, opened 2022-08-09
fx.Options and fx.Module behave differently


djmitche:
**Describe the bug**

It appears that `fx.Invoke`'s that are in an `fx.Module` do not run until after all `fx.Invoke`'s _outside_ the module.

This is not documented in `fx.Module`, and `fx.Invoke`'s docs say
> Invoke registers functions that are executed eagerly on application start. Arguments for these invocations are built using the constructors registered by Provide. Passing multiple Invoke options appends the new invocations to the application's existing list.
> 
> Unlike constructors, invocations are always executed, and they're always run in order. Invocations may have any number of returned values. If the final returned object is an error, it indicates whether the operation was successful. All other returned values are discarded.

emphasis on _in order_ :)

**To Reproduce**
https://go.dev/play/p/SP93TiMiji9
```go
package main

import (
	"context"
	"fmt"

	"go.uber.org/fx"
)

func main() {
	app := fx.New(
		fx.Invoke(func() { fmt.Printf("Outer invoke 1\n") }),
		fx.Module(
			"mod",
			fx.Invoke(func() { fmt.Printf("Module invoke\n") }),
		),
		fx.Invoke(func() { fmt.Printf("Outer invoke 2\n") }),
	)
	app.Start(context.Background())
	app.Stop(context.Background())
}
```
)

**Actual behavior**
```
Outer invoke 1
Outer invoke 2
Module invoke
```

**Expected behavior**
```
Outer invoke 1
Module invoke
Outer invoke 2
```

**Additional context**
Changing `fx.Module` to `fx.Options` (and removing the "mod") makes this behave as expected.

djmitche:
In the use-case I'm working on, I want to write
```go
app := fx.New(
  someModule,
  anotherModule,
  fx.Invoke(func(lc fx.Lifecycle) {
    lc.Append(fx.Hook{OnStart: ...})
  }),
)
```
and know that, because it's the last invocation, that fx.Hook is the last thing started in the lifecycle.  As it is, everything in the modules is starting _after_ my top-level hook.

sywhang:
This is a valid issue, thanks for reporting this.

Tracking internal issue: GO-1591.

sywhang:
BTW, just to clarify what the issue is - the issue isn't that fx.Options and fx.Module behaves differently. They do behave differently because they're meant for doing different things :) 

There are two issues that this issue is tracking:
1. Lack of documentation on the invoke order when fx.Module is involved. 
[...]

PR #925, opened 2022-08-19
Fix Invoke order when fx.Module is used

sywhang:
Currently, fx.Module's invokes are run after all the invokes provided
at the parent level are invoked.

This makes it difficult for module consumers to control the precise
invoke orders.

This changes the invoke order to run all the invokes in the child
module before running any of the invokes in the parent module, and
clarify that order in the documentation for Invoke.

Solves #918.
Refs GO-1591.

### DIFF
--- a/invoke.go
+++ b/invoke.go
@@ -44,10 +44,10 @@ import (
 //
 //	fx.New(
 //		fx.Invoke(func3),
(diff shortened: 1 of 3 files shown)


r-hang (approved):
lgtm!

PR #931, opened 2022-08-23
Ensure OnStart/OnStop hooks can only be called once

### LINKED ISSUE #930: onStop executed many times
### Discussed in https://github.com/uber-go/fx/discussions/929

<div type='discussions-op-text'>

<sup>Originally posted by **coolyrat** August 22, 2022</sup>
```go
package main

import (
	"context"
	"fmt"
	"log"
	"os"
	"os/signal"
	"syscall"
	"time"

	"go.uber.org/fx"
)

func main() {
	server := fx.New(
		fx.Invoke(func(lc fx.Lifecycle) {
			fmt.Println("first invoke...")
			lc.Append(fx.Hook{
					fmt.Println("#1 start")
					return nil
				},
				OnStop: func(ctx context.Context) error {
					fmt.Println("#1 stop")
					return nil
				},
			})
		}),

		fx.Invoke(func(lc fx.Lifecycle) {
			fmt.Println("last invoke...")
			lc.Append(fx.Hook{
					fmt.Println("#2 start")
					return nil
				},
				OnStop: func(ctx context.Context) error {
					fmt.Println("#2 stop")
					return nil
				},
			})
		}),
	)

	go func() {
		server.Run()
	}()

	kill := make(chan os.Signal, 1)
	signal.Notify(kill, syscall.SIGINT, syscall.SIGTERM, syscall.SIGKILL, syscall.SIGQUIT)
	<-kill
	fmt.Println("killed")

	ctx, _ := context.WithTimeout(context.Background(), 10*time.Second)
	if err := server.Stop(ctx); err != nil {
		log.Printf("error stopping gracefully")
	}
}

```

When I give the stop signal, the log
```
...
killed
#2 stop
#1 stop
#2 stop
```

And I am expected to be 
```
...
killed
#2 stop
#1 stop
```

Is there anything I'm doing wrong?</div>

sywhang:
It is possible for a user to erroneously call app.Run() then
follow up by calling app.Stop(). In such a case, it is possible
for the app.Stop() method to be called by two goroutines concurrently,
resulting in a race.

This adds a state in the App to keep track of whether Stop() has been
invoked so that such a race can be prevented.

Fix #930
Internal Ref: GO-1606

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -1282,6 +1282,47 @@ func TestAppStart(t *testing.T) {
 		err := app.Start(context.Background()).Error()
 		assert.Contains(t, err, "OnStart hook added by go.uber.org/fx_test.TestAppStart.func10.1 failed: goroutine exited without returning")
 	})
+
+	t.Run("Start/Stop should be called exactly once only.", func(t *testing.T) {
+		t.Parallel()
+		startCalled := 0
+		stopCalled := 0
+		app := fxtest.New(t,
+			Provide(Annotate(func() int { return 0 },
+				OnStart(func(context.Context) error {
+					startCalled += 1
+					return nil
+				}),
+				OnStop(func(context.Context) error {
+					stopCalled += 1
+					return nil
+				})),
+			),
+			Invoke(func(i int) {
+				assert.Equal(t, 0, i)
+			}),
+		)
+		var wg sync.WaitGroup
+		for i := 0; i < 10; i++ {
+			wg.Add(1)
+			go func() {
+				app.Start(context.Background())
+				wg.Done()
+			}()
+		}
+		wg.Wait()
+		assert.Equal(t, 1, startCalled)
+		for i := 0; i < 10; i++ {
+			wg.Add(1)
+			go func() {
+				app.Stop(context.Background())
+				wg.Done()
+			}()
+		}
+		wg.Wait()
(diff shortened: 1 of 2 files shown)


app_test.go
```diff
@@ -1281,6 +1282,47 @@ func TestAppStart(t *testing.T) {
...
+			Invoke(func(i int) {
+				assert.Equal(t, 0, i)
+			}),
+		)
+		var wg sync.WaitGroup
+		for i := 0; i < 10; i++ {
+			wg.Add(1)
+			go func() {
+				app.Start(context.Background())
+				wg.Done()
```
abhinav, line 1310: minor: We should defer the wg.Done so that a panic in Start() doesn't cause a deadlock.
```suggestion
				defer wg.Done()
				app.Start(context.Background())
```

app_test.go
```diff
@@ -1281,6 +1282,47 @@ func TestAppStart(t *testing.T) {
...
+			}()
+		}
+		wg.Wait()
+		assert.Equal(t, 1, startCalled)
+		for i := 0; i < 10; i++ {
+			wg.Add(1)
+			go func() {
+				app.Stop(context.Background())
+				wg.Done()
```
abhinav, line 1319:
```suggestion
				defer wg.Done()
				app.Stop(context.Background())
```

abhinav (approved):

PR #932, opened 2022-08-24
Fix incorrect documentation on soft value group

sywhang:


abhinav (commented):
The provides aren't annotated or using fx.Out so they won't go to the group.

sywhang:
yeah, realized the code sample was broken too after pushing this - updated it.

r-hang (approved):
lgtm!

abhinav (commented):
retroactive lgtm

Issue #878, opened 2022-05-06
Add fx.From Annotation to allow passing an implementation type as an interface parameter


rossb83:
The most common pattern I hear about to use in `go` is accept interfaces and return structs. The newest `fx` framework addition `fx.As(...)` makes following this paradigm much more intuitive and cleaner in `fx` modules. But it seems we're only halfway there. I would like to propose a new `fx` method `fx.From(...)` which serves as a companion for `fx.As(...)` to manipulate input types rather than the outputs. 

The current functionality works well in situations like this
```go
package a

type Bar interface {...} // Bar is implemented by Foo
type Foo struct {...}
func NewFoo() *Foo {...}

var Module = fx.Options(
    fx.Provide(fx.Annotate(
        NewFoo, fx.As(new(Bar))),
    ), 
)
```
```go
package b

type Baz struct {...}
func NewBaz(bar a.Bar) *Baz {...}

var Module = fx.Options(
    fx.Provide(NewBaz), 
    a.Module,
)
```
But other situations like below may fall short. For example if the interface were in `package b` generating the following diff to support it would cause a cyclic dependency error. This situation could come up often as it is common in `go` for the interface to be defined on the caller side.
```diff
package a

type Foo struct {...}
func NewFoo() *Foo {...}

var Module = fx.Options(
-    fx.Provide(NewFoo),
+    fx.Provide(fx.Annotate(
+        NewFoo, fx.As(new(b.Bar))),
+    ), 
)
```
```go
package b

type Bar interface {...} // Bar is implemented by Foo
type Baz struct {...}
func NewBaz(bar Bar) *Baz {...}

var Module = fx.Options(
    fx.Provide(NewBaz), 
    a.Module,
)
```
Yes, we could make use of new packages to hold the fx logic and/or interface to break the cycle, but it seems ideal for `fx` limitations not to enforce package management decisions. Similarly, `package a` might be part of some external codebase or very old legacy code that no one wants to touch and inserting an `fx.As(...)` in there years later may create unintended side-effects.

I'd like to propose `fx.From(...)` which works exactly like `fx.As(...)` except to aid in manipulating input parameter types rather than outputs. 

Taking the second example, it may work like this leaving `package a` entirely unaware of this dilemma.
```go
package b

type Bar interface {...} // Bar is implemented by Foo
type Baz struct {...}
func NewBaz(bar Bar) *Baz {...}

var Module = fx.Options(
    fx.Provide(fx.Annotate(
[...]

PR #938, opened 2022-09-21
adds fx.From annotation

rossb83:
Adds fx.From annotation to allow injecting struct ptr types for interface dependencies which they implement. This should address issue https://github.com/uber-go/fx/issues/878

### DIFF
--- a/annotated.go
+++ b/annotated.go
@@ -585,11 +585,39 @@ func (at asAnnotation) apply(ann *annotated) error {
 	return nil
 }
 
+type fromAnnotation struct {
+	targets []interface{}
+}
+
+var _ Annotation = fromAnnotation{}
+
+func From(interfaces ...interface{}) Annotation {
+	return fromAnnotation{interfaces}
+}
+
+func (fr fromAnnotation) apply(ann *annotated) error {
+	if len(ann.From) > 0 {
+		return fmt.Errorf("only single fx.From annotation allowed")
+	}
+	types := make([]reflect.Type, len(fr.targets))
+	for i, typ := range fr.targets {
+		t := reflect.TypeOf(typ)
+		if t.Kind() != reflect.Ptr || t.Elem().Kind() != reflect.Ptr || t.Elem().Elem().Kind() != reflect.Struct {
+			return fmt.Errorf("fx.From: argument must be a pointer a struct: got %v", t)
+		}
+		t = t.Elem()
+		types[i] = t
+	}
+	ann.From = append(ann.From, types)
+	return nil
+}
+
 type annotated struct {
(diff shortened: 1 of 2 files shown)


annotated.go
```diff
@@ -585,11 +585,41 @@ func (at asAnnotation) apply(ann *annotated) error {
...
+
+// From is an Annotation that annotates the parameters for a function (i.e. a
+// constructor) that accepts interfaces. Very much the opposite of fx.As(...).
+func From(interfaces ...interface{}) Annotation {
+	return fromAnnotation{interfaces}
+}
+
+func (fr fromAnnotation) apply(ann *annotated) error {
+	if len(ann.From) > 0 {
+		return fmt.Errorf("fx.From does not support multiple annotations")
```
sywhang, line 602:
```suggestion
		return fmt.Errorf("only one fx.From annotation can be specified")
```

annotated.go
```diff
@@ -585,11 +585,41 @@ func (at asAnnotation) apply(ann *annotated) error {
...
+	}
+	types := make([]reflect.Type, len(fr.targets))
+	for i, typ := range fr.targets {
+		t := reflect.TypeOf(typ)
+		if t.Kind() != reflect.Ptr || t.Elem().Kind() != reflect.Ptr && t.Elem().Kind() != reflect.Struct || t.Elem().Kind() == reflect.Ptr && t.Elem().Elem().Kind() != reflect.Struct {
```
sywhang, line 607: It's not immediately clear why we allow double pointer types to a struct.. What's the rationale behind allowing pointer to a struct type, and a pointer to a pointer to a struct type, but not pointer to a pointer to a struct type? 

I think it's more consistent if we _only_ allowed a pointer to a struct type.

sywhang (commented):
Thank you for putting this PR up - left some questions and comments.

annotated.go
rossb83: I agree with this so in this case the following test would fail (which is fine) due to `new` not being a ptr to a struct
`		{
			desc: "value type implementing interface",
			provide: fx.Provide(
				func() anotherStringer {
					return anotherStringer{
						"another stringer",
					}
				},
				fx.Annotate(
					func(myStringer myStringer) fmt.Stringer {
						return &fromStringer{
							name: myStringer.String(),
						}
					},
					fx.From(new(anotherStringer)),
				),
			),
			invoke: func(s fmt.Stringer) {
				assert.Equal(t, s.String(), "another stringer")
			},
		},`

kkHAIKE:
ping @rossb83 😊 LGTM

tchung1118:
This has been completed with #990.

PR #939, opened 2022-09-22
Add test for case: decorator required by a member of group

zenthangplus:
I add this test just to raise an issue:

Decorator does not execute when only required by a member of group.

Current, this test is will be failed. So does this behavior is expected?


sywhang:
@zenthangplus thanks for bringing this to our attention. I can confirm this is a bug. I filed #940 to track this issue. I'll close this PR.

sywhang:
This was fixed with #941

PR #941, opened 2022-09-22
Fix decorator not being applied to transient dependencies

### LINKED ISSUE #940: bug: Decorator is not applied properly
**Describe the bug**
Decorator is not applied properly when it is not required through a "direct" dependency.

**To Reproduce**
See https://github.com/uber-go/fx/pull/939.

**Expected behavior**
The test added in https://github.com/uber-go/fx/pull/939 

Internal Ref: GO-1668

sywhang:
There is a bug that only manifests when decorating objects in Fx where transient dependency types are not properly decorated. For example, if type A depends on type B, and type B depends on type C which has a decorator, it fails to recognize that.

This only occurs when all constructors and the decorators for all these types are provided at the "root" level fx.App.

This happens because fx injects a fake "root" Scope between the actual root Container (which is where constructors are provided to by default).

Fixed by making fx stop create this fake root Scope and use the dig.Container directly.

Fixes #940.

### DIFF
--- a/decorate_test.go
+++ b/decorate_test.go
@@ -313,6 +313,39 @@ func TestDecorateSuccess(t *testing.T) {
 		)
 		defer app.RequireStart().RequireStop()
 	})
+
+	t.Run("decoration must execute when required by a member of group", func(t *testing.T) {
+		type Drinks interface {
+		}
+		type Coffee struct {
+			Type  string
+			Name  string
+			Price int
+		}
+		type PriceService struct {
+			DefaultPrice int
+		}
+		app := fxtest.New(t,
+			fx.Provide(func() *PriceService {
+				return &PriceService{DefaultPrice: 3}
+			}),
+			fx.Decorate(func(service *PriceService) *PriceService {
+				service.DefaultPrice = 10
+				return service
+			}),
+			fx.Provide(fx.Annotate(func(service *PriceService) Drinks {
+				assert.Equal(t, 10, service.DefaultPrice)
+				return &Coffee{Type: "coffee", Name: "Americano", Price: service.DefaultPrice}
+			}, fx.ResultTags(`group:"drinks"`))),
+			fx.Provide(fx.Annotated{Group: "drinks", Target: func() Drinks {
+				return &Coffee{Type: "coffee", Name: "Cold Brew", Price: 4}
+			}}),
+			fx.Invoke(fx.Annotate(func(drinks []Drinks) {
+				assert.Len(t, drinks, 2)
+			}, fx.ParamTags(`group:"drinks"`))),
+		)
+		defer app.RequireStart().RequireStop()
+	})
 }
 
 func TestDecorateFailure(t *testing.T) {
--- a/module.go
+++ b/module.go
@@ -81,12 +81,23 @@ func (o moduleOption) apply(mod *module) {
 type module struct {
 	parent     *module
 	name       string
-	scope      *dig.Scope
+	scope      scope
 	provides   []provide
 	invokes    []invoke
 	decorators []decorator
 	modules    []*module
 	app        *App
+}
+
+// scope is a private wrapper interface for dig.Container and dig.Scope.
+// We can consider moving this into Fx using type constraints after Go 1.20
+// is released and 1.17 is deprecated.
+type scope interface {
+	Decorate(f interface{}, opts ...dig.DecorateOption) error
+	Invoke(f interface{}, opts ...dig.InvokeOption) error
+	Provide(f interface{}, opts ...dig.ProvideOption) error
+	Scope(name string, opts ...dig.ScopeOption) *dig.Scope
+	String() string
 }
 
 // builds the Scopes using the App's Container. Note that this happens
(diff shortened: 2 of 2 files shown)


module.go
```diff
@@ -81,23 +81,32 @@ func (o moduleOption) apply(mod *module) {
...
+
 // builds the Scopes using the App's Container. Note that this happens
 // after applyModules' are called because the App's Container needs to
 // be built for any Scopes to be initialized, and applys' should be called
 // before the Container can get initialized.
 func (m *module) build(app *App, root *dig.Container) {
 	if m.parent == nil {
-		m.scope = root.Scope(m.name)
-		// TODO: Once fx.Decorate is in-place,
-		// use the root container instead of subscope.
```
abhinav, line 100: heh

abhinav (approved):

zenthangplus:
Thank, could you publish a release?

sywhang:
@zenthangplus Sure, I can. But can you wait until Monday, please? I'll need to verify the new version with Uber's Go Monorepo and we don't do releases on Friday to avoid causing issues over the weekend for services.

zenthangplus:
@sywhang i've tested, fx.Decorate work well, but seems fx.Replace is not work as expected. As i known, Replace is equivalent with Decorate.

sywhang:
> @sywhang i've tested, fx.Decorate work well, but seems fx.Replace is not work as expected. As i known, Replace is equivalent with Decorate.

Can you be more specific? What's the code you wrote to test it?

zenthangplus:
@sywhang I've add two test cases, please take a look: https://github.com/uber-go/fx/pull/942

dannyposi:
@sywhang could you help to release, thank

dannyposi:
Oops, seems have some problem on master branch. When upgrading to latest code on master branch, my app is not run correctly, i don't know why, but i'm investigating.

I'm using the bellow function to handle request (for testing only). 
```
httpHandler := func(w http.ResponseWriter, r *http.Request) {
		ctx := context.Background()
		ctxWithResponseWriter := context.WithValue(ctx, "w", w)
		ctxWithRequest := context.WithValue(ctxWithResponseWriter, "r", r)
		_ = app.Start(ctxWithRequest)
	}

and append hook

lc.Append(
		fx.Hook{
				log.Info("start application") // Problem: only show once time
				r := ctx.Value("r").(*http.Request)
				w := ctx.Value("w").(*httptest.ResponseRecorder)
				httpServer.Handler.ServeHTTP(w, r)
				return nil
			},
		},
	)
```

But when run a test suite with two test cases, the last test case seems not run correctly. OnStart hook seems only run once time.

zenthangplus:
Oops, seems have some problem on master branch. When upgrading to latest code on master branch, my app is not run correctly, i don't know why, but i'm investigating.

I'm using the bellow function to handle request (for testing only). 
```
httpHandler := func(w http.ResponseWriter, r *http.Request) {
		ctx := context.Background()
		ctxWithResponseWriter := context.WithValue(ctx, "w", w)
		ctxWithRequest := context.WithValue(ctxWithResponseWriter, "r", r)
		_ = app.Start(ctxWithRequest)
	}

and append hook

lc.Append(
		fx.Hook{
				log.Info("start application") // Problem: only show once time
				r := ctx.Value("r").(*http.Request)
				w := ctx.Value("w").(*httptest.ResponseRecorder)
				httpServer.Handler.ServeHTTP(w, r)
				return nil
			},
		},
	)
```

But when run a test suite with two test cases, the last test case seems not run correctly. OnStart hook seems only run once time.

Then i've tried to revert to v1.18.1 and it works correctly. Seems latest code on master has problem.

abhinav:
Thanks for catching that @zenthangplus.
Would you mind creating a new issue for this? We nearly missed the comment.
If your investigation yields a reproduction of the bug, that would be really helpful for us to fix this.

zenthangplus:
@abhinav i've check, it related to this MR https://github.com/uber-go/fx/pull/931, it not been released yet

sywhang:
Hey @zenthangplus, could you please specify what you mean? OnStart/OnStop hooks are only supposed to run once. If you were relying on it being run twice, that's not the intended behavior.

zenthangplus:
@sywhang In my case, i've wrote test for integration api testing. every http call to my api will be handled by this function.

```
httpHandler := func(w http.ResponseWriter, r *http.Request) {
		ctx := context.Background()
		ctxWithResponseWriter := context.WithValue(ctx, "w", w)
		ctxWithRequest := context.WithValue(ctxWithResponseWriter, "r", r)
		_ = app.Start(ctxWithRequest)
	}
```

Before it works well, but after this MR https://github.com/uber-go/fx/pull/931 `app.Start` is only allow to run once. I can change my logic to handle `app.Stop` before handle another request but i think this is breaking change and it break my test logic.

decorate_test.go
```diff
@@ -313,6 +313,39 @@ func TestDecorateSuccess(t *testing.T) {
...
+		}
+		type Coffee struct {
+			Type  string
+			Name  string
+			Price int
+		}
+		type PriceService struct {
+			DefaultPrice int
+		}
+		app := fxtest.New(t,
```
jkanywhere, line 328: This test would pass even if none of the functions executed. What guarantees the assertions on 337 and 344 are called?

sywhang: Invokes are guaranteed to run (and the fact that it ran is guaranteed by `fx.RequireStart().RequireStop()` - if any Invokes fail to run, that fails the test). Invoke running successfully means assertion in line 344 are called.

And yes, line 337 in itself does not guarantee the execution, but the assertion in line 344 means both constructors `Provide`d here ran successfully.

PR #942, opened 2022-09-23
Add test for cases: replace by value and replace by constructor

zenthangplus:
Issue:
- Replace by value: work
- Replace by constructor: not work

### DIFF
--- a/replace_test.go
+++ b/replace_test.go
@@ -135,6 +135,72 @@ func TestReplaceSuccess(t *testing.T) {
 			),
 		)
 	})
+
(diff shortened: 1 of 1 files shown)


sywhang:
This is expected behavior. fx.Replace only works on values. 

Refer to the documentation: https://pkg.go.dev/go.uber.org/fx#Replace

```go
Replace provides instantiated values for graph modification as if they had been provided using a decorator with fx.Decorate. The most specific type of each value (as determined by reflection) is used.
```

sywhang:
Please also use our https://github.com/uber-go/fx/discussions section to see if it's a bug rather than opening PRs.

PR #947, opened 2022-09-28
Allow fx.Module-specific custom loggers

### LINKED ISSUE #894: Allow specifying fx.Module-specific custom loggers with WithLogger
Currently, passing `fx.WithLogger` to an `fx.Module` throws an error like this:

```
[Fx] ERROR              Failed to start: fx.WithLogger Option should be passed to top-level App, not to fx.Module
```

This isn't ideal because a library author that exports their app as an `fx.Module` cannot specify a custom logger whose scope applies to just their `Module`.

This tracks extending the current implementation of `WithLogger` option so that it can be passed directly to an `fx.Module`, such that when it is, it only affects the scope of that `Module`. 

Internal Ref: GO-1488

tchung1118:
Currently passing fx.WithLogger option to fx.Module does not work because fx currently does not support specifying module-level custom loggers.

Now fx.WithLogger can be passed to fx.Modules for specifying custom loggers for fx.Modules. Also added some test cases to test fx.Module-specific loggers.

Fixes #894

### DIFF
--- a/module.go
+++ b/module.go
@@ -26,6 +26,7 @@ import (
 	"go.uber.org/dig"
 	"go.uber.org/fx/fxevent"
 	"go.uber.org/fx/internal/fxreflect"
+	"go.uber.org/multierr"
 )
 
 // A container represents a set of constructors to provide
@@ -79,14 +80,16 @@ func (o moduleOption) apply(mod *module) {
 }
 
 type module struct {
-	parent     *module
-	name       string
-	scope      scope
-	provides   []provide
-	invokes    []invoke
-	decorators []decorator
-	modules    []*module
-	app        *App
+	parent         *module
+	name           string
+	scope          scope
+	provides       []provide
+	invokes        []invoke
+	decorators     []decorator
+	modules        []*module
+	app            *App
+	log            fxevent.Logger
+	logConstructor *provide
 }
 
 // scope is a private wrapper interface for dig.Container and dig.Scope.
@@ -110,6 +113,15 @@ func (m *module) build(app *App, root *dig.Container) {
 	} else {
 		parentScope := m.parent.scope
 		m.scope = parentScope.Scope(m.name)
+	}
+
+	m.log = app.log
+	if m.logConstructor != nil {
+		// Since user supplied a custom logger, use a buffered logger
+		// to hold all messages until user supplied logger is
+		// instantiated. Then we flush those messages after fully
+		// constructing the custom logger.
+		m.log = new(logBuffer)
 	}
 
 	for _, mod := range m.modules {
@@ -158,7 +170,50 @@ func (m *module) provide(p provide) {
 			Err:             m.app.err,
 		}
 	}
-	m.app.log.LogEvent(ev)
+	m.log.LogEvent(ev)
+}
+
+func (m *module) constructAllCustomLoggers() {
+	if m.logConstructor != nil {
+		buffer, ok := m.log.(*logBuffer)
+		if ok {
+			// default to app logger if custom logger constructor fails
+			if err := m.constructCustomLogger(buffer); err != nil {
+				m.app.err = multierr.Append(m.app.err, err)
+				m.log = m.app.log
+				buffer.Connect(m.log)
+			}
+		}
+	}
+
+	for _, mod := range m.modules {
+		mod.constructAllCustomLoggers()
+	}
+}
+
+// Mirroring the behavior of app.constructCustomLogger
+func (m *module) constructCustomLogger(buffer *logBuffer) (err error) {
+	p := m.logConstructor
+	fname := fxreflect.FuncName(p.Target)
+	defer func() {
+		m.log.LogEvent(&fxevent.LoggerInitialized{
+			Err:             err,
+			ConstructorName: fname,
+		})
+	}()
+
+	if err := m.scope.Provide(p.Target); err != nil {
+		return fmt.Errorf("fx.WithLogger(%v) from:\n%+vFailed: %v",
+			fname, p.Stack, err)
+	}
+
+	// TODO: Use dig.FillProvideInfo to inspect the provided constructor
+	// and fail the application if its signature didn't match.
+
+	return m.scope.Invoke(func(log fxevent.Logger) {
+		m.log = log
+		buffer.Connect(log)
+	})
 }
 
 func (m *module) executeInvokes() error {
@@ -179,12 +234,12 @@ func (m *module) executeInvokes() error {
 
 func (m *module) executeInvoke(i invoke) (err error) {
 	fnName := fxreflect.FuncName(i.Target)
-	m.app.log.LogEvent(&fxevent.Invoking{
+	m.log.LogEvent(&fxevent.Invoking{
 		FunctionName: fnName,
 		ModuleName:   m.name,
 	})
 	err = runInvoke(m.scope, i)
-	m.app.log.LogEvent(&fxevent.Invoked{
+	m.log.LogEvent(&fxevent.Invoked{
 		FunctionName: fnName,
 		ModuleName:   m.name,
 		Err:          err,
@@ -203,14 +258,14 @@ func (m *module) decorate() (err error) {
 		}
 
 		if decorator.IsReplace {
-			m.app.log.LogEvent(&fxevent.Replaced{
+			m.log.LogEvent(&fxevent.Replaced{
 				ModuleName:      m.name,
 				OutputTypeNames: outputNames,
 				Err:             err,
 			})
 		} else {
 
-			m.app.log.LogEvent(&fxevent.Decorated{
+			m.log.LogEvent(&fxevent.Decorated{
 				DecoratorName:   fxreflect.FuncName(decorator.Target),
 				ModuleName:      m.name,
 				OutputTypeNames: outputNames,
--- a/module.go
+++ b/module.go
@@ -110,12 +110,14 @@ type scope interface {
 func (m *module) build(app *App, root *dig.Container) {
 	if m.parent == nil {
 		m.scope = root
+		m.log = app.log
 	} else {
 		parentScope := m.parent.scope
 		m.scope = parentScope.Scope(m.name)
+		// use parent module's logger by default
+		m.log = m.parent.log
 	}
 
-	m.log = app.log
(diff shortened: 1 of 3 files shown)


module_test.go
```diff
@@ -241,6 +244,85 @@ func TestModuleSuccess(t *testing.T) {
 		)
 		require.NoError(t, app.Err())
 	})
+
+	t.Run("custom logger for module", func(t *testing.T) {
```
sywhang, line 248: Let's add a test case that Decorates the logger supplied inside a Module, and make sure that the decorator only gets applied to the Module's logger and not the root-level logger.

module_test.go
```diff
@@ -241,6 +244,85 @@ func TestModuleSuccess(t *testing.T) {
...
+			"Supplied", "Provided", "Provided", "Provided", "Provided",
+			"LoggerInitialized", "Invoking", "Invoked", "Invoking", "Invoked",
+		}, spy.EventTypes())
+
+		spy.Reset()
+		app.RequireStart().RequireStop()
+
+		require.NoError(t, app.Err())
+
+		assert.Equal(t, []string{"Started", "Stopped"}, spy.EventTypes())
```
sywhang, line 324: These tests may actually be better off written as a table test. 
See: https://github.com/uber-go/guide/blob/master/style.md#test-tables

module_test.go
```diff
@@ -420,4 +498,142 @@ func TestModuleFailures(t *testing.T) {
...
+			}),
+			fx.Invoke(func(s string) {
+				assert.Fail(t, "this should never run")
+			}),
+		)
+
+		var buff bytes.Buffer
+		app := fx.New(
+			redis,
+			fx.Logger(log.New(&buff, "", 0)),
```
sywhang, line 595: this is deprecated - can you change these with `WithLogger` and use spy like the other tests?

module.go
```diff
@@ -158,7 +170,50 @@ func (m *module) provide(p provide) {
 			Err:             m.app.err,
 		}
 	}
-	m.app.log.LogEvent(ev)
+	m.log.LogEvent(ev)
+}
+
+func (m *module) constructAllCustomLoggers() {
```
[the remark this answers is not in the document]
tchung1118: Added a comment

module.go
```diff
@@ -158,7 +170,50 @@ func (m *module) provide(p provide) {
...
+	if m.logConstructor != nil {
+		buffer, ok := m.log.(*logBuffer)
+		if ok {
```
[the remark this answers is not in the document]
tchung1118: Done 👍

module.go
```diff
@@ -158,7 +170,50 @@ func (m *module) provide(p provide) {
...
+		m.log.LogEvent(&fxevent.LoggerInitialized{
+			Err:             err,
+			ConstructorName: fname,
+		})
+	}()
+
+	if err := m.scope.Provide(p.Target); err != nil {
+		return fmt.Errorf("fx.WithLogger(%v) from:\n%+vFailed: %v",
+			fname, p.Stack, err)
+	}
```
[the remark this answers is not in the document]
tchung1118: Updated the error message to include module name

module_test.go
tchung1118: Added a test case using Replace. Also added another test case for nest Module using parent Module's logger.

module_test.go
```diff
@@ -420,4 +612,122 @@ func TestModuleFailures(t *testing.T) {
...
+				redis := fx.Module("redis",
+					append(
+						tt.giveModuleOpts,
+						fx.Provide(func() string {
+							return "redis"
+						}),
+						fx.Invoke(func(s string) {
+							assert.Fail(t, "this should never run")
+						}),
+					)...,
```
sywhang, line 705: collect the default options into `fx.Options` instead of appending and exploding it like this.

module_test.go
```diff
@@ -420,4 +612,122 @@ func TestModuleFailures(t *testing.T) {
...
+				giveAppOpts:     spyAsLogger,
+				wantErrContains: []string{"error building logger"},
+				wantEventTypes: []string{
+					"Supplied", "Provided", "Provided", "Provided",
+					"LoggerInitialized", "Provided", "LoggerInitialized",
+				},
+			},
+			{
+				desc: "logger dependency in module failed to build",
+				giveModuleOpts: []fx.Option{
```
sywhang, line 662:
```suggestion
				giveModuleOpts: fx.Options(
```

module_test.go
```diff
@@ -241,6 +243,187 @@ func TestModuleSuccess(t *testing.T) {
...
+		appSpy := NamedSpy{name: "app"}
+		childSpy := NamedSpy{name: "child"}
+
+		grandchild := fx.Module("grandchild",
+			fx.Provide(func() *Logger {
+				return &Logger{Name: "grandchild"}
+			}),
+			fx.Invoke(func(r *Logger) {
+				assert.Equal(t, "grandchild", r.Name)
+			}),
```
sywhang, line 380: I had to stare at this test for a while because it wasn't clear what this Logger thing was. Looks like this is just a random dependency that gets Provided to make sure that the "Provided" event gets logged by the loggers you supply, which are  `appSpy` and `childSupply`.

Can you change this so that it uses some other naive looking type, and not something that looks like an actual logger that gets passed to fx.WithLogger?

module_test.go
```diff
@@ -420,4 +599,118 @@ func TestModuleFailures(t *testing.T) {
...
+				assert.Fail(t, "this should never run")
+			}),
+		)
+
+		tests := []struct {
+			desc            string
+			giveModuleOpts  fx.Option
+			giveAppOpts     fx.Option
+			wantErrContains []string
+			wantEventTypes  []string
```
sywhang, line 629: nit: `:%s/wantEventTypes/wantEvents/g`

module_test.go
```diff
@@ -420,4 +599,118 @@ func TestModuleFailures(t *testing.T) {
...
+						assert.Fail(t, "this should never run")
+					}),
+				)
+				err := app.Err()
+				require.Error(t, err)
+				for _, contains := range tt.wantErrContains {
+					assert.Contains(t,
+						err.Error(),
+						contains,
+					)
```
sywhang, line 706:
```suggestion
					assert.Contains(t, err.Error(), contains)
```

module_test.go
```diff
@@ -241,6 +243,187 @@ func TestModuleSuccess(t *testing.T) {
...
+			"Supplied", "Provided", "Replaced", "LoggerInitialized",
+			//Invoke logged twice, once from child and another from grandchild
+			"Invoking", "Invoked", "Invoking", "Invoked",
+		}, childSpy.EventTypes())
+
+		// events from modules do not appear in app logger
+		assert.Equal(t, []string{
+			"Provided", "Provided", "Provided",
+			"LoggerInitialized", "Invoking", "Invoked",
+		}, appSpy.EventTypes())
```
sywhang, line 418: Add what you have as a comment as the last parameter of assert.Equal and it'll show up in the test failure message if it passes. i.e.
```suggestion
		// events from modules do not appear in app logger
		assert.Equal(t, []string{
			"Provided", "Provided", "Provided",
			"LoggerInitialized", "Invoking", "Invoked",
		}, appSpy.EventTypes(), "events from modules should not appear in app logger")
```

module.go
```diff
@@ -158,7 +172,49 @@ func (m *module) provide(p provide) {
...
+			}
+		}
+	}
+
+	for _, mod := range m.modules {
+		mod.constructAllCustomLoggers()
+	}
+}
+
+// Mirroring the behavior of app.constructCustomLogger
```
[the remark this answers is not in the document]
tchung1118: To do this, I would need to refactor the whole initialization process of `app.log` to be handled by module instead of fx.New. Is it reasonable to include this type of change in this PR's scope?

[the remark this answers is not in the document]
sywhang: You can file a follow-up task to refactor it if you believe this PR is getting too large.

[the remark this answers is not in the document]
tchung1118: GO-1690 created

module_test.go
```diff
@@ -241,6 +247,185 @@ func TestModuleSuccess(t *testing.T) {
...
+
+		appSpy.Reset()
+		app.RequireStart().RequireStop()
+
+		require.NoError(t, app.Err())
+
+		assert.Equal(t, []string{"Started", "Stopped"}, appSpy.EventTypes())
```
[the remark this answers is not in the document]
tchung1118: Added it right below this line

module_test.go
```diff
@@ -241,6 +247,191 @@ func TestModuleSuccess(t *testing.T) {
...
+
+		appSpy.Reset()
+		moduleSpy.Reset()
+
+		app.RequireStart().RequireStop()
+
+		require.NoError(t, app.Err())
+
+		assert.Equal(t, []string{"Started", "Stopped"}, appSpy.EventTypes())
+		assert.Equal(t, []string{}, moduleSpy.EventTypes())
```
sywhang, line 372:
```suggestion
		assert.Equal(t, 1, len(moduleSpy.EventTypes())
```

module_test.go
```diff
@@ -241,6 +247,191 @@ func TestModuleSuccess(t *testing.T) {
...
+
+		appSpy.Reset()
+		childSpy.Reset()
+
+		app.RequireStart().RequireStop()
+
+		require.NoError(t, app.Err())
+
+		assert.Equal(t, []string{"Started", "Stopped"}, appSpy.EventTypes())
+		assert.Equal(t, []string{}, childSpy.EventTypes())
```
sywhang, line 433:
```suggestion
		assert.Empty(t, childSpy.EventTypes())
```

module_test.go
```diff
@@ -241,6 +247,191 @@ func TestModuleSuccess(t *testing.T) {
...
+
+		appSpy.Reset()
+		moduleSpy.Reset()
+
+		app.RequireStart().RequireStop()
+
+		require.NoError(t, app.Err())
+
+		assert.Equal(t, []string{"Started", "Stopped"}, appSpy.EventTypes())
+		assert.Equal(t, 1, len(moduleSpy.EventTypes())
```
sywhang, line 372:
```suggestion
		assert.Empty(t, moduleSpy.EventTypes())
```

sywhang (approved):

PR #952, opened 2022-09-30
Add support for various Hook-convertible functions

mway:
Adds new `Hook`-based helper functions:

- `StartHook`
- `StopHook`
- `StartStopHook`

and enables those function to accept any function of the form `func([context.Context] [error]` (see the `HookFunc` constraint), for example:

- `func()`
- `func() error`
- `func(context.Context)`
- `func(context.Context) error`

So, rather than users needing to do things like:

```go
func start() error {
  // ...
}

func stop(context.Context) {
  // ...
}

lifecycle.Append(fx.Hook{
    return start()
  },
  OnStop: func(ctx context.Context) error {
    stop(ctx)
    return nil
  },
})
```

which is an increasingly common pattern, users can now do:

```go
lifecycle.Append(fx.StartStopHook(start, stop))
```

or

```go
lifecycle.Append(fx.StartHook(start))
lifecycle.Append(fx.StopHook(stop))
```

and the provided functions will be converted to `func(context.Context) error` under the hood.

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2020-2021 Uber Technologies, Inc.
+// Copyright (c) 2022 Uber Technologies, Inc.
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -205,7 +205,8 @@ func TestNewApp(t *testing.T) {
 				func(b struct {
 					In
 					Foos []int `group:"foo"`
-				}) {
+				},
+				) {
 					assert.Len(t, b.Foos, 1)
 				},
 			),
@@ -381,7 +382,6 @@ func TestNewApp(t *testing.T) {
 			[]string{"Provided", "Provided", "Provided", "Provided", "Decorated", "Decorated", "LoggerInitialized", "Started"},
 			spy.EventTypes())
 	})
-
 }
 
 func TestWithLoggerErrorUseDefault(t *testing.T) {
@@ -793,13 +793,13 @@ func TestAppRunTimeout(t *testing.T) {
 	t.Parallel()
 
 	// Fails with an error immediately.
-	failure := func(context.Context) error {
+	failure := func() error {
 		return errors.New("great sadness")
 	}
 
 	// Builds a hook that takes much longer than the application timeout.
-	takeVeryLong := func(clock *clock.Mock) func(ctx context.Context) error {
-		return func(ctx context.Context) error {
+	takeVeryLong := func(clock *clock.Mock) func() error {
+		return func() error {
 			// We'll exceed the start and stop timeouts,
 			// and then some.
 			for i := 0; i < 3; i++ {
@@ -825,7 +825,7 @@ func TestAppRunTimeout(t *testing.T) {
 			desc: "OnStart timeout",
 			buildHooks: func(clock *clock.Mock) []Hook {
 				return []Hook{
-					{OnStart: takeVeryLong(clock)},
+					StartHook(takeVeryLong(clock)),
 				}
 			},
 			wantEventType: &fxevent.Started{},
@@ -837,8 +837,8 @@ func TestAppRunTimeout(t *testing.T) {
 				return []Hook{
 					// The hooks are separate because
 					// OnStop will not be run if that hook failed.
-					{OnStop: takeVeryLong(clock)},
-					{OnStart: failure},
+					StartHook(takeVeryLong(clock)),
+					StopHook(failure),
 				}
 			},
 			wantEventType: &fxevent.Started{},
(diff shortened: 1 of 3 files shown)


lifecycle.go
```diff
@@ -26,6 +26,11 @@ import (
 	"go.uber.org/fx/internal/lifecycle"
 )
 
+// A HookFunc is a function that can be used as a [Hook].
+type HookFunc interface {
+	~func() | ~func() error | ~func(context.Context) | ~func(context.Context) error
+}
```
tchung1118, line 32: Since according to the comment in `internal/lifecycle/lifecycle.go`, `lifecycle.Callable` should be identical to `fx.HookFunc`, is there a particular reason why this can't be just a type alias to `lifecycle.Callable`? Will it break documentation?

app_test.go
```diff
@@ -1428,6 +1428,275 @@ func TestValidateApp(t *testing.T) {
...
+				))
+			}),
+		).RequireStart().RequireStop()
+
+		for name, count := range calls {
+			require.Equalf(t, 2, count, "bad call count: %s (%d)", name, count)
+		}
+	})
+
+	t.Run("start errors", func(t *testing.T) {
```
tchung1118, line 1534: I think this, along with "stop errors", can be combined into a [table test](https://github.com/uber-go/guide/blob/master/style.md#test-tables).

app_test.go
```diff
@@ -1428,6 +1428,275 @@ func TestValidateApp(t *testing.T) {
...
+		// expected error.
+		err = New(Invoke(func(lc Lifecycle) {
+			lc.Append(StartHook(func(ctx context.Context) error {
+				return wantErr
+			}))
+		})).Start(context.Background())
+		require.ErrorContains(t, err, wantErr.Error())
+	})
+
+	t.Run("start deadline", func(t *testing.T) {
```
tchung1118, line 1555: Similar to "start errors", maybe this can be combined with "stop deadline"?

lifecycle.go
```diff
@@ -40,14 +45,104 @@ type Lifecycle interface {
...
+	return Hook{
+		OnStop:     onstop,
+		onStopName: stopname,
+	}
+}
+
+// StartStopHook returns a new Hook with start as its [Hook.OnStart] function
+// and stop as its [Hook.OnStop] function, independently wrapping the signature
+// of each as needed.
+func StartStopHook[T1 HookFunc, T2 HookFunc](start T1, stop T2) Hook {
```
tchung1118, line 122: I'm not sure if this was considered, but what if we automatically wrap OnStart and OnStop hooks in fx.Hook passed to `lifecycle.Append` and just update the type definition for fx.Hook to take in the new types of function that we're allowing? I think that will eliminate the need for these helper functions, right? I'm just thinking if we ever add another type of hook to lifecycle, we would have to add a bunch of these helper functions again. Is there performance or backward compatibility issue that I'm not aware of with this approach?

tchung1118 (changes requested):
Looks good other than the comments I added.

lifecycle.go
mway: It can break docs, but more generally, I don't want to create a type alias for something that isn't itself exported. I don't think keeping them in sync here will require any effort.

lifecycle.go
mway: I had considered that first. Unfortunately, we can't update `fx.Hook` to do this without parameterizing it to use the `HookFunc` type constraint, which would (1) qualify as a breaking change and (2) make direct `fx.Hook` much less ergonomic:

```go
type Hook[T1 HookFunc, T2 HookFunc] struct {
  OnStart T1
  OnStop  T2
}
```

which, at a minimum, would require both (1) explicit parameterization for struct literals and (2) both type parameters to be specified even if only using one field.

app_test.go
mway: Big fan of table tests, however none of the other tests here are table tests, so I'd prefer to keep this consistent with other tests as part of this change.

Fx tests are also not a great fit for table tests as currently written. We could come back and refactor/restructure them later, if we decide that's worthwhile.

lifecycle.go
tchung1118: I see. Thank you for explaining this to me!

tchung1118 (approved):
Approved since all my comments were addressed.

sywhang (approved):
Nice. This looks good to me @mway! Thanks for doing this work.

app_test.go
```diff
@@ -205,7 +205,8 @@ func TestNewApp(t *testing.T) {
 				func(b struct {
 					In
 					Foos []int `group:"foo"`
-				}) {
+				},
+				) {
```
abhinav, line 209: nit: weird formatting change

abhinav (commented):
Retroactive LGTM

PR #953, opened 2022-09-30
Add support for context- and error-less annotated hook funcs

mway:
Similar to #952, this adds support for any `HookFunc`-compatible funcs as part of `fx.OnStart`/`fx.OnStop` annotations.

~Unrelated change: also changes bind addresses for tests to be localhost.~ Was done in #960

### DIFF
--- a/annotated.go
+++ b/annotated.go
@@ -250,23 +250,14 @@ func (la *lifecycleHookAnnotation) apply(ann *annotated) error {
 		)
 	}
 
-	if ft.NumIn() < 1 || ft.In(0) != _typeOfContext {
-		return fmt.Errorf(
-			"first argument of hook must be context.Context, got %v (%T)",
-			la.Target,
-			la.Target,
-		)
-	}
-
-	hasOut := ft.NumOut() == 1
-	returnsErr := hasOut && ft.Out(0) == _typeOfError
-
-	if !hasOut || !returnsErr {
-		return fmt.Errorf(
-			"hooks must return only an error type, got %v (%T)",
-			la.Target,
-			la.Target,
-		)
+	if n := ft.NumOut(); n > 0 {
+		if n > 1 || ft.Out(0) != _typeOfError {
+			return fmt.Errorf(
+				"optional hook return may only be an error, got %v (%T)",
+				la.Target,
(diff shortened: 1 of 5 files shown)


annotated.go
```diff
@@ -250,23 +250,14 @@ func (la *lifecycleHookAnnotation) apply(ann *annotated) error {
...
-	if !hasOut || !returnsErr {
-		return fmt.Errorf(
-			"hooks must return only an error type, got %v (%T)",
-			la.Target,
-			la.Target,
-		)
+	if n := ft.NumOut(); n > 0 {
+		if n > 1 || ft.Out(0) != _typeOfError {
+			return fmt.Errorf(
+				"optional hook return may only be error, got %v (%T)",
```
sywhang, line 256: We may want to specify that the optional return is `an` error to avoid confusions between a single error vs. multiple errors.

```suggestion
				"optional hook return may only be an error, got %v (%T)",
```

sywhang (approved):
Looks good to me

annotated.go
mway: Done, thanks!

sywhang (approved):

annotated_test.go
```diff
@@ -1450,3 +1455,59 @@ func TestHookAnnotationFailures(t *testing.T) {
...
+
+	for _, tt := range table {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			called.Store(false)
+
+			opts := fx.Options(
+				fx.Provide(tt.annotation),
+				fx.Invoke(func(A) {}),
+			)
```
abhinav, line 1507: optional:
we can avoid sharing `called` between the tests by putting `called` into the container,
and making that the parameter the OnStart functions are accepting.

```
var called atomic.Bool
opts := fx.Options(
  fx.Supply(&called)
  // ...
)
```

And in the test,

```
fx.OnStart(func(called *atomic.Bool) {
  called.Store(true)
})
```

abhinav (approved):

annotated_test.go
mway: Ah good call, thanks!

PR #958, opened 2022-10-11
Add docs/ folder and an intro tutorial

abhinav:
This adds a new docs/ folder to the repository.
We'll put non-API-reference documentation in this folder.

To begin with, this folder contains the following documentation:

- docs/get-started/README holds an introduction to Fx tutorial
- docs/CONTRIBUTING holds guidance for adding new documentation
  (please read before adding documentation)

In the interest of keeping the documentation up-to-date and valid,
we make use of [mdox](https://github.com/bwplotka/mdox/) to:

- reformat Markdown to the same format
- verify that code samples are up-to-date
- all links in the documentation are valid

Further, to ensure that all the code samples are runnable,
this includes tests for all example code.
Testing the example code is a bit complex because we
need to treat it like an application "from outside"
so all tests rely on TestMain-style hijacking to run the programs.

---

Rendered documentation:

- [README](https://github.com/uber-go/fx/blob/abg/tutorial/README.md)
- [docs/CONTRIBUTING](https://github.com/uber-go/fx/blob/abg/tutorial/docs/CONTRIBUTING.md)
- [docs/get-started](https://github.com/uber-go/fx/blob/abg/tutorial/docs/get-started/README.md)


abhinav:
Looks like we can't have the doc tests run on Windows.
It doesn't support sending SIGINT from within the process.
I'll adjust the jobs to run docs test only on Linux.

abhinav:
Note that the only failure here is that parts of the code added in the patch are not covered.
Those lines that are missing are difficult to get coverage data out of because they run a subprocess.
Other than that, this change is good to go.

sywhang (approved):

PR #959, opened 2022-10-11
Remove 1.17 from test matrix

sywhang:
We only support the latest two versions of go versions. Dropping the test matrix for 1.17.


sywhang:
closing in favor of #960

PR #962, opened 2022-10-12
docs: Add GitHub Pages

abhinav:
This migrates docs added in #958 into a GitHub Pages website
powered by Vuepress (same as https://cadenceworkflow.io/).

Besides the boilerplate, this carries the content over almost as-is.

One change we've made here is to split the many sections of the getting started
tutorial into different pages.
This helps keep the length of the document manageable,
and makes the transition between each step more obvious.


abhinav:
This is mostly an RFC.
If we don't want to host the website, we can stick with GitHub's UI.

sywhang (approved):

PR #964, opened 2022-10-12
ci/docs: Fix setup-node action

abhinav:
The setup-node action expects the yarn.lock to be present
in the root of the repository,
or if it's in a subdirectory, for a path to be specified.

https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-data

Further, yarn adds links to the currently used yarn registry to the
yarn.lock.
In our case, it was pointing to the internal registry,
so I've added project-local configuration to use the global yarn
registry.

Lastly, we were running yarn install in the root of the repository,
not the docs subdirectory.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -27,9 +27,11 @@ jobs:
         with:
           node-version: '16'
           cache: 'yarn'
+          cache-dependency-path: docs/yarn.lock
 
       - name: Install dependencies
         run: yarn install
+        working-directory: docs
 
       - name: Build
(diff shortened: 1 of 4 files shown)


abhinav:
The build works but deployment is temporarily broken because we don't have GitHub Pages enabled.
I've messaged @jonathanbaker7 about it.

tchung1118 (approved):
LGTM 👍

jonathanbaker7:
@abhinav is now the admin for this repository.

PR #967, opened 2022-10-12
docs: Application lifecycle

abhinav:
Add a page under a Concepts heading documenting the Fx application lifecycle.
This does not go into too much detail,
nor does it include examples of how to use it.
We can add that in the future.

This makes use of mermaidjs to include the diagram.
GitHub also supports mermaidjs in Markdown files,
so our documentation is still GitHub-compatible.

---

Rendered on GitHub: https://github.com/uber-go/fx/blob/docs-lifecycle/docs/lifecycle.md

Screenshot:
image


sywhang (approved):
:shipit:

PR #982, opened 2022-10-27
Allow setting Zap logger level

### LINKED ISSUE #979: Allow setting level of Zap Logger used for logging
The Zap logger that gets passed to fx via `WithLogger` gets logs at `Info` level for all messages except when things error out, which are logged at `Error` level.

We can consider exposing a way to set the level of these logs.

sywhang:
This extends the `fxevent.ZapLogger` struct to add two new methods: `UseLogLevel` and `UseErrorLevel`.

`UseLogLevel` sets the level of all non-error messages emitted by Fx.

`UseErrorLevel` sets the level of all error messages emitted by Fx. 

When these are not set, the log levels are kept identical to what they are now (Info for non-error logs, Error for error logs).

Fix #979.


abhinav:
>Setting ErrorLevel sets the error logs to be emitted at the specified level, unless the specified level is lower than zapcore.ErrorLevel; if that is the case, the logs will be emitted at zapcore.ErrorLevel.

This seems like a weird behavior.
Was this choice made to work around the fact that we can't determine that the user set the ErrorLevel field?
Could we try something else here to determine whether the field is set?

One option is to make the field a private pointer, and add a `UseErrorLevel(zapcore.Level)` method that sets is value, or we default to Error.

```go
eventLog := fxevent.ZapLogger{Log: log}
eventLog.UseErrorLevel(zap.InfoLevel)
```

If you'd like to keep things consistent, the non-error level field can also get a setter.

sywhang:
@abhinav That's a good suggestion. Let me modify the PR to apply those suggestions. Thanks!

abhinav (commented):
will look again tomorrow but:

fxevent/zap.go
```diff
@@ -24,145 +24,175 @@ import (
...
 
 	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 )
 
 // ZapLogger is an Fx event logger that logs events to Zap.
 type ZapLogger struct {
 	Logger *zap.Logger
+
+	logLevel   *zapcore.Level
```
abhinav, line 34: since info is the zero value, i would drop the pointer on this one.
```suggestion
	logLevel   zapcore.Level  // default: info
```

fxevent/zap.go
```diff
@@ -24,145 +24,175 @@ import (
...
+func (l *ZapLogger) UseLogLevel(level zapcore.Level) {
+	l.logLevel = &level
+}
+
+func (l *ZapLogger) logEvent(msg string, fields ...zap.Field) {
+	if l.logLevel == nil {
+		l.Logger.Log(zapcore.InfoLevel, msg, fields...)
+	} else {
+		l.Logger.Log(*l.logLevel, msg, fields...)
+	}
```
abhinav, line 55: For non-error, with logLevel being a non-ptr, this can just be:
```suggestion
	l.Logger.Log(l.logLevel, msg, fields...)
```

fxevent/zap.go
```diff
@@ -24,145 +24,175 @@ import (
...
+}
+
+func (l *ZapLogger) logError(msg string, fields ...zap.Field) {
+	if l.errorLevel == nil {
+		l.Logger.Log(zapcore.ErrorLevel, msg, fields...)
+	} else {
+		l.Logger.Log(*l.errorLevel, msg, fields...)
+	}
```
abhinav, line 63: Consider something DRYer here:

```
lvl := zapcore.ErrorLevel
if l.errorLevel != nil { lvl = *l.errorLevel }
l.Logger.Log(lvl, msg, fields...)
```

abhinav (approved):

Issue #936, opened 2022-09-15
OnStart/OnStop annotations: Suport non-standard signatures


abhinav:
**Is your feature request related to a problem? Please describe.**

fx.OnStart and fx.OnStop annotations currently require that hooks accept a context.Context as their first argument, and return an error as their last result. This requires adding an unnecessary anonymous function for cases where there's no context or error support.

**Describe the solution you'd like**

It should be possible to do this:

```go
type Foo struct{ ... }
func New() *Foo { .. }
func (*Foo) Start() { .. }

fx.Annotate(New, fx.OnStart((*Foo).Start))

// Note: (*Foo).Start is sugar for func(f *Foo) { f.Start() }
```

Currently, if you feed this into fx.New, this will fail because Foo.Start does not accept a context or return an error.

**Describe alternatives you've considered**

Write anonymous functions:

```go
fx.Annotate(New, fx.OnStart(func(f *Foo) { f.Start() }))
```

**Is this a breaking change?**
No

jasonmills:
We should probably support all the variations since we want to also support the lifecyle patten.

```lang=go
 // *Foo must be provided by New
func (*Foo) Start() {} 
fx.Annotate(New, fx.OnStart( (*Foo).Start ))

// honor error from start function, but not context, and New must be provide *Foo
func (*Foo) Start() error {}
fx.Annotate(New, fx.OnStart( (*Foo).Start )) 
```

abhinav:
Yeah that sounds right. Make context and error independently optional.

jkanywhere:
I think this issue is fixed by https://github.com/uber-go/fx/pull/953 which adds support for functions with and without context parameter and error return.

PR #983, opened 2022-10-28
Rewrite annotated to apply annotations in certain order

tchung1118:
Currently all annotations that are passed to fx.Annotate are applied at once when annotated.Build() is called. 
This has led to an issue where lifecycle hook annotations (fx.OnStart and fx.OnStop) do not work properly when 
used with other parameter/result annotations. (uber-go#937)

This PR splits up existing code that applies annotations and makes annotated.Build() apply each annotation 
one by one, mostly in the order that they're passed in. In the process of applying individual annotations in order, 
it puts lifecycle hook annotations aside and applies them after all the other annotations are applied. 

Also, lifecycle hook annotations have been changed so that it can now take zero or more of the following as their arguments: 
- parameters of the annotated function
- results of the annotated function
- a Lifecycle object
- a context.Context object

In other words, lifecycle hook annotations can no longer pull in extra dependencies outside of things on which 
the annotated constructor is dependent, results that the annotated constructor provides, context.Context 
object which is injected by Lifecycle, and the Lifecycle object itself.

### DIFF
--- a/annotated.go
+++ b/annotated.go
@@ -520,7 +520,7 @@ func (la *lifecycleHookAnnotation) apply(ann *annotated) error {
 
 	if ft.IsVariadic() {
 		return fmt.Errorf(
-			"hooks must not accept variatic parameters, got %v (%T)",
+			"hooks must not accept variadic parameters, got %v (%T)",
 			la.Target,
 			la.Target,
 		)
@@ -530,7 +530,7 @@ func (la *lifecycleHookAnnotation) apply(ann *annotated) error {
 	return nil
 }
 
-// build implements Annotation
+// build builds and returns a constructor after applying a lifecycle hook annotation.
 func (la *lifecycleHookAnnotation) build(ann *annotated) (interface{}, error) {
 	resultTypes, hasError := ann.currentResultTypes()
 	if !hasError {
@@ -690,6 +690,10 @@ var (
 	_groupTag = "group"
 )
 
+// makeHookScopeCtor makes a constructor that provides all possible parameters
+// that the lifecycle hook being appended can depend on. It also deduplicates
+// duplicate param and result types, which is possible when using fx.Decorate,
+// and uses values from results for providing the deduplicated types.
 func makeHookScopeCtor(paramTypes []reflect.Type, resultTypes []reflect.Type, args []reflect.Value) interface{} {
 	type key struct {
 		t     reflect.Type
@@ -869,11 +873,13 @@ func (la *lifecycleHookAnnotation) buildHook(fn func(context.Context) error) (ho
 // Lifecycle OnStart (see Lifecycle type documentation) hooks without building a
 // function that takes a dependency on the Lifecycle type.
 //
-//	fx.Annotate(
-//	  NewServer,
-//	  fx.OnStart(func(ctx context.Context, server Server) error {
-//	      return server.Listen(ctx)
-//	  }),
+//	fx.Provide(
+//		fx.Annotate(
+//			NewServer,
+//			fx.OnStart(func(ctx context.Context, server Server) error {
+//				return server.Listen(ctx)
+//			}),
+//		)
 //	)
 //
 // Which is functionally the same as:
@@ -886,12 +892,40 @@ func (la *lifecycleHookAnnotation) buildHook(fn func(context.Context) error) (ho
 //			    return server.Listen(ctx)
 //		      },
 //	     })
+//		 return server
 //	   }
 //	 )
 //
+// It is also possible to use OnStart annotation with other parameter and result
+// annotations, provided that the parameter of the function passed to OnStart
+// matches annotated parameters and results
+//
+// For example, the following is possible:
+//
+//	fx.Provide(
+//		fx.Annotate(
+//			func (a A) B {...},
+//			fx.ParamTags(`name:"A"`),
+//			fx.ResultTags(`name:"B"`),
+//			fx.OnStart(func (p ParamStruct) {...}),
+//		),
+//	)
+//
+// As long as ParamStruct looks like the following and has no other dependencies
+// besides Context or Lifecycle:
+//
+//	type ParamStruct struct {
+//		fx.In
+//		FieldA A `name:"A"`
+//		FieldB B `name:"B"`
+//	}
+//
 // Only one OnStart annotation may be applied to a given function at a time,
 // however functions may be annotated with other types of lifecylce Hooks, such
-// as OnStop.
+// as OnStop. The hook function passed into OnStart cannot take any arguments
+// outside of the annotated constructor's existing dependencies or results, except
+// for a context.Context, which is injected by Lifecycle, or Lifecycle itself, which
+// is automatically added by the annotation to append the hook function.
 func OnStart(onStart interface{}) Annotation {
 	return &lifecycleHookAnnotation{
 		Type:   _onStartHookType,
@@ -904,11 +938,13 @@ func OnStart(onStart interface{}) Annotation {
 // Lifecycle OnStop (see Lifecycle type documentation) hooks without building a
 // function that takes a dependency on the Lifecycle type.
 //
-//	fx.Annotate(
-//	  NewServer,
-//	  fx.OnStop(func(ctx context.Context, server Server) error {
-//	      return server.Shutdown(ctx)
-//	  }),
+//	fx.Provide(
+//		fx.Annotate(
+//			NewServer,
+//			fx.OnStop(func(ctx context.Context, server Server) error {
+//				return server.Shutdown(ctx)
+//			}),
+//		)
 //	)
 //
 // Which is functionally the same as:
@@ -917,16 +953,44 @@ func OnStart(onStart interface{}) Annotation {
 //	   func(lifecycle fx.Lifecycle, p Params) Server {
 //	     server := NewServer(p)
 //	     lifecycle.Append(fx.Hook{
-//		      OnStart: func(ctx context.Context) error {
+//		      OnStop: func(ctx context.Context) error {
 //			    return server.Shutdown(ctx)
 //		      },
 //	     })
+//		 return server
 //	   }
 //	 )
 //
+// It is also possible to use OnStop annotation with other parameter and result
+// annotations, provided that the parameter of the function passed to OnStop
+// matches annotated parameters and results
+//
+// For example, the following is possible:
+//
+//	fx.Provide(
+//		fx.Annotate(
+//			func (a A) B {...},
+//			fx.ParamTags(`name:"A"`),
+//			fx.ResultTags(`name:"B"`),
+//			fx.OnStop(func (p ParamStruct) {...}),
+//		),
+//	)
+//
+// As long as ParamStruct looks like the following and has no other dependencies
+// besides Context or Lifecycle:
+//
+//	type ParamStruct struct {
+//		fx.In
+//		FieldA A `name:"A"`
+//		FieldB B `name:"B"`
+//	}
+//
 // Only one OnStop annotation may be applied to a given function at a time,
 // however functions may be annotated with other types of lifecylce Hooks, such
-// as OnStart.
+// as OnStart. The hook function passed into OnStart cannot take any arguments
+// outside of the annotated constructor's existing dependencies or results, except
+// for a context.Context, which is injected by Lifecycle, or Lifecycle itself, which
+// is automatically added by the annotation to append the hook function.
 func OnStop(onStop interface{}) Annotation {
 	return &lifecycleHookAnnotation{
 		Type:   _onStopHookType,
--- a/annotated.go
+++ b/annotated.go
@@ -94,18 +94,20 @@ func (a Annotated) String() string {
 	return fmt.Sprintf("fx.Annotated{%v}", strings.Join(fields, ", "))
 }
 
-// field used for embedding fx.Out type in generated struct.
-var _inAnnotationField = reflect.StructField{
-	Name:      "In",
-	Type:      reflect.TypeOf(In{}),
-	Anonymous: true,
-}
-
-var _outAnnotationField = reflect.StructField{
-	Name:      "Out",
-	Type:      reflect.TypeOf(Out{}),
-	Anonymous: true,
-}
+var (
+	// field used for embedding fx.In type in generated struct.
+	_inAnnotationField = reflect.StructField{
+		Name:      "In",
+		Type:      reflect.TypeOf(In{}),
+		Anonymous: true,
+	}
+	// field used for embedding fx.Out type in generated struct.
+	_outAnnotationField = reflect.StructField{
+		Name:      "Out",
+		Type:      reflect.TypeOf(Out{}),
+		Anonymous: true,
+	}
+)
 
 // Annotation can be passed to Annotate(f interface{}, anns ...Annotation)
 // for annotating the parameter and result types of a function.
@@ -200,9 +202,7 @@ func (pt paramTagsAnnotation) parameters(ann *annotated) (
 	inFields := []reflect.StructField{_inAnnotationField}
 
 	// there was a variadic argument, so it was pre-transformed
-	if len(types) > 0 && types[0].Kind() == reflect.Struct &&
-		dig.IsIn(reflect.New(types[0]).Elem().Interface()) {
-		offset := 1
+	if len(types) > 0 && isIn(types[0]) {
 		paramType := types[0]
 
 		for i := 1; i < paramType.NumField(); i++ {
@@ -212,10 +212,8 @@ func (pt paramTagsAnnotation) parameters(ann *annotated) (
 				Type: origField.Type,
 				Tag:  origField.Tag,
 			}
-			if field.Type == _typeOfLifecycle {
-				offset += 1
-			} else if i-offset < len(pt.tags) {
-				field.Tag = reflect.StructTag(pt.tags[i-offset])
+			if i-1 < len(pt.tags) {
+				field.Tag = reflect.StructTag(pt.tags[i-1])
 			}
 
 			inFields = append(inFields, field)
@@ -232,16 +230,13 @@ func (pt paramTagsAnnotation) parameters(ann *annotated) (
 		}
 	}
 
-	offset := 0
 	for i, t := range types {
 		field := reflect.StructField{
 			Name: fmt.Sprintf("Field%d", i),
 			Type: t,
 		}
-		if t == _typeOfLifecycle {
-			offset += 1
-		} else if i-offset < len(pt.tags) {
-			field.Tag = reflect.StructTag(pt.tags[i-offset])
+		if i < len(pt.tags) {
+			field.Tag = reflect.StructTag(pt.tags[i])
 		}
 
 		inFields = append(inFields, field)
@@ -297,10 +292,7 @@ func (rt resultTagsAnnotation) apply(ann *annotated) error {
 func (rt resultTagsAnnotation) build(ann *annotated) (interface{}, error) {
 	paramTypes := ann.currentParamTypes()
 
-	resultTypes, remapResults, err := rt.results(ann)
-	if err != nil {
-		return nil, err
-	}
+	resultTypes, remapResults := rt.results(ann)
 
 	origFn := reflect.ValueOf(ann.Target)
 	newFnType := reflect.FuncOf(paramTypes, resultTypes, false)
@@ -311,10 +303,12 @@ func (rt resultTagsAnnotation) build(ann *annotated) (interface{}, error) {
 	return newFn.Interface(), nil
 }
 
+// results returns the types of the results of the annotated function,
+// and a function that maps the results of the target function,
+// into a result compatible with the annotated function.
 func (rt resultTagsAnnotation) results(ann *annotated) (
 	types []reflect.Type,
 	remap func([]reflect.Value) []reflect.Value,
-	err error,
 ) {
 	types, hasError := ann.currentResultTypes()
 
@@ -327,7 +321,7 @@ func (rt resultTagsAnnotation) results(ann *annotated) (
 	if len(rt.tags) == 0 {
 		return types, func(results []reflect.Value) []reflect.Value {
 			return results
-		}, nil
+		}
 	}
 
 	// if there's no Out struct among the return types, there was no As annotation applied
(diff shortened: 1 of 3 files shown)


annotated.go
```diff
@@ -272,210 +530,356 @@ func (la *lifecycleHookAnnotation) apply(ann *annotated) error {
...
+		case _onStopHookType:
+			la.scope = ann.container.Scope("onStopHookScope")
+		}
+		ctor := makeHookScopeCtor(paramTypes, resultTypes, args)
+		if err := la.scope.Provide(ctor); err != nil {
+			results[0] = reflect.ValueOf(err)
+			return
+		}
+		if err := la.scope.Invoke(invokeFn.Interface()); err != nil {
+			results[0] = reflect.ValueOf(err)
```
tchung1118, line 680: Wrap this error for nicer format

annotated.go
```diff
@@ -519,16 +953,44 @@ func OnStart(onStart interface{}) Annotation {
...
+// matches annotated parameters and results
+//
+// For example, the following is possible:
+//
+//	fx.Provide(
+//		fx.Annotate(
+//			func (a A) B {...},
+//			fx.ParamTags(`name:"A"`),
+//			fx.ResultTags(`name:"B"`),
+//			fx.OnStop(func (p ParamStruct) {...}),
```
sywhang, line 975: nit: in the code sample, let's rename ParamStruct to OnStopParams

annotated.go
```diff
@@ -519,16 +953,44 @@ func OnStart(onStart interface{}) Annotation {
...
+//		FieldB B `name:"B"`
+//	}
+//
 // Only one OnStop annotation may be applied to a given function at a time,
 // however functions may be annotated with other types of lifecylce Hooks, such
-// as OnStart.
+// as OnStart. The hook function passed into OnStart cannot take any arguments
+// outside of the annotated constructor's existing dependencies or results, except
+// for a context.Context, which is injected by Lifecycle, or Lifecycle itself, which
+// is automatically added by the annotation to append the hook function.
```
sywhang, line 993: same modification here as OnStart

annotated.go
```diff
@@ -538,9 +1000,24 @@ func OnStop(onStop interface{}) Annotation {
...
+		dig.IsOut(reflect.New(t).Elem().Interface()))
 }
 
-var _ Annotation = asAnnotation{}
+func isIn(t reflect.Type) bool {
+	return (t.Kind() == reflect.Struct &&
+		dig.IsIn(reflect.New(t).Elem().Interface()))
+}
+
+func checkIfResultTagged(results []reflect.Type) (reflect.Type, bool) {
```
sywhang, line 1016: technically this is just checking whether the result is an fx.Out struct, so if the function is producing an fx.Out struct but is not tagged with ResultTags, it'll still return true from this func, right?

annotated.go
```diff
@@ -538,9 +1000,24 @@ func OnStop(onStop interface{}) Annotation {
...
+	return results[0], isOut(results[0])
```
sywhang, line 1017: also this could panic if `results` is nil.

annotated.go
```diff
@@ -582,31 +1059,193 @@ var _ Annotation = asAnnotation{}
...
-	FuncPtr    uintptr
-	Hooks      []*lifecycleHookAnnotation
+	Target      interface{}
+	Annotations []Annotation
+	ParamTags   []string
+	ResultTags  []string
+	As          [][]reflect.Type
+	FuncPtr     uintptr
+	Hooks       []*lifecycleHookAnnotation
+	container   *dig.Container
```
sywhang, line 1248: add a comment here explaining how this is used, or document it in the implementation part where it's using this container, and add a comment here that refers to that section.

sywhang (commented):
Took a first pass

annotated.go
```diff
@@ -189,13 +288,170 @@ func (rt resultTagsAnnotation) apply(ann *annotated) error {
...
+		field := reflect.StructField{
+			Name: fmt.Sprintf("Field%d", i),
+			Type: t,
+		}
+		if i < len(rt.tags) {
+			field.Tag = reflect.StructTag(rt.tags[i])
+		}
+		newOut.Offsets = append(newOut.Offsets, len(newOut.Fields))
+		newOut.Fields = append(newOut.Fields, field)
+		continue
```
JacobOaks, line 377: Is this continue necessary? It's at the end of the loop.

annotated.go
```diff
@@ -189,13 +288,170 @@ func (rt resultTagsAnnotation) apply(ann *annotated) error {
...
+					taggedFields[j+1] = reflect.StructField{
+						Name: field.Name,
+						Type: field.Type,
+						Tag:  reflect.StructTag(tag),
+					}
+				}
+			}
+			existingOuts = append(existingOuts, reflect.StructOf(taggedFields))
+			continue
+		}
```
JacobOaks, line 365: This has 20 lines vs. the 10 lines outside of the if condition, consider flipping condition to reduce nesting

annotated.go
```diff
@@ -272,210 +529,357 @@ func (la *lifecycleHookAnnotation) apply(ann *annotated) error {
...
+					ctxStructPos = i
+					ctxPos = j
+					continue
+				}
+				fields = append(fields, field)
+			}
+			invokeParamTypes = append(invokeParamTypes, reflect.StructOf(fields))
+			continue
 		}
+		invokeParamTypes = append(invokeParamTypes, origHookFnT.In(i))
```
JacobOaks, line 610: Is there any reason why this isn't just an if / else if / else?

JacobOaks: Or, what might be better is:

```
if t == _typeOfContext && ctxPos < 0 {
    ctxPos = i
    continue
if !isIn(t) {
    invokeParamTypes = append(invokeParamTypes, origHookFnT.in(i))
    continue
}
fields := []reflect.StructField{_inAnnotationField}
<...>
```

annotated.go
```diff
@@ -272,210 +529,357 @@ func (la *lifecycleHookAnnotation) apply(ann *annotated) error {
...
+			scope = ann.container.Scope("onStopHookScope")
+		}
+		ctor := makeHookScopeCtor(paramTypes, resultTypes, args)
+		if err := scope.Provide(ctor); err != nil {
+			results[0] = reflect.ValueOf(fmt.Errorf("Error providing possible parameters for hook installer: %w", err))
+			return
+		}
+		if err := scope.Invoke(invokeFn.Interface()); err != nil {
+			results[0] = reflect.ValueOf(fmt.Errorf("Error invoking hook installer: %w", err))
+			return
```
JacobOaks, line 681: this return isn't technically necessary

JacobOaks, line 681: This return isn't technically necessary. Also, I don't think this reads as well as how you have it currently, but if you combine the ifs into an if / else if, neither return is necessary, can just rely on the one under these two ifs. Not really sure what the preference for this kind of thing is.

annotated.go
```diff
@@ -272,210 +529,357 @@ func (la *lifecycleHookAnnotation) apply(ann *annotated) error {
...
+		var scope *dig.Scope
+		switch la.Type {
+		case _onStartHookType:
+			scope = ann.container.Scope("onStartHookScope")
+		case _onStopHookType:
+			scope = ann.container.Scope("onStopHookScope")
+		}
+		ctor := makeHookScopeCtor(paramTypes, resultTypes, args)
+		if err := scope.Provide(ctor); err != nil {
+			results[0] = reflect.ValueOf(fmt.Errorf("Error providing possible parameters for hook installer: %w", err))
```
sywhang, line 676: Error strings should be lowercase. 
https://github.com/golang/go/wiki/CodeReviewComments#error-strings

annotated.go
sywhang: You can't combine these ifs. While they look like conditionals, they're not conditionals. 

These two things are in plain English:

1. If you try providing the constructor to the scope, and it fails, set the result to an error and return
2. Then try Invoking this function, and if it fails, set the result to another error and return.

2 is dependent on 1 succeeding so it's not an if-else clause.

JacobOaks: I understand how conceptually you'd want to keep it how it is. But isn't this technically the same behavior:

```
if err := scope.Provide(ctor); err != nil {
    results[0] = reflect.ValueOf(fmt.Errorf("Error providing possible parameters for hook installer: %w", err))
} else if err := scope.Invoke(invokeFn.Interface()); err != nil {
    results[0] = reflect.ValueOf(fmt.Errorf("Error invoking hook installer: %w", err))
}
return
```

Currently:
1. If providing the constructor to the scope fails, we set result to an error and return.
2. Otherwise, we try to invoke the function. If this fails, we set the result to an error and return.
3. Finally, if both of those succeed, we return.

With an if / else if clause:
1. If providing the constructor to the scope fails, we set the result to an error. We do not interact with the else if, and jump straight over it to the return statement and return
2. Otherwise, we try invoking the function. If that fails, we set the result to another error. We exit the else if, and return
3. If both succeed, we return.

Aren't these the same?

sywhang: ah, I see what you mean. You're right, it is possible to remove the returns by changing this to else if. 

That being said, I don't think this necessarily helps the readability of the code. 

Here are some thoughts behind that:

1. if-else statements are usually mutually exclusive statements. i.e.
```
if A { 
  // something 
} else if B {
  // another thing
} else {
  // something else 
}
```
In the code above, we take "A/something", "B/another thing", and "C/something else" are all mutually exclusive statements that don't seem correlated. So if we grouped this together, it makes it _seem_ like Provide and Invoke are mutually exclusive.

2. That being said, you could argue that the thing that corresponds to A/B/C above in this code is _failure cases_ , which technically are mutually exclusive - you can't fail in both cases. 

3. With these in mind, the code here is doing a sequence of actions, while checking for failures in each action. It's more natural to not group them under an if-else. 

You can see examples of code like this in the standard library as well where they check for errors and return early if an error is encountered using series of if statements instead of an if-else statement. Example: https://github.com/golang/go/blob/master/src/internal/poll/fd_unix.go#L369.

annotated.go
```diff
@@ -272,210 +529,357 @@ func (la *lifecycleHookAnnotation) apply(ann *annotated) error {
...
+// that the lifecycle hook being appended can depend on. It also deduplicates
+// duplicate param and result types, which is possible when using fx.Decorate,
+// and uses values from results for providing the deduplicated types.
+func makeHookScopeCtor(paramTypes []reflect.Type, resultTypes []reflect.Type, args []reflect.Value) interface{} {
+	type key struct {
+		t     reflect.Type
+		name  string
+		group string
+	}
+	seen := map[key]bool{}
```
sywhang, line 703: It seems like `seen[key]` is never set to false for any `key` which makes it kind of like a hashset. 

In Go, it is more optimal for you to use an empty struct as the value type in such a case (`map[key]struct{}`). Using this allocates 0 byte for the value as opposed to using bool which allocates 1 byte per value.

annotated.go
```diff
@@ -223,6 +479,7 @@ func (la *lifecycleHookAnnotation) String() string {
 }
 
 func (la *lifecycleHookAnnotation) apply(ann *annotated) error {
+	// TODO: maybe allow annotated function to be passed in for la.Target for usability
```
sywhang, line 482: remove this please 😄

annotated.go
JacobOaks: Ah okay. Yeah, I agree it's more readable how it is now. Cool, thanks for the standard library example.

annotated.go
```diff
@@ -713,237 +1402,40 @@ func (ann *annotated) typeCheckOrigFn() error {
...
-			} else {
-				outResults = append(outResults, _nilError)
-			}
-		}
+// parameters returns the type for the parameters of the annotated function,
+// and a function that maps the arguments of the annotated function
+// back to the arguments of the target function and a function that maps
+// values to any lifecycle hook annotations. It accepts a variactic set
+// of reflect.Type which allows for omitting any resulting constructor types
+// from required parameters for annotation hooks.
```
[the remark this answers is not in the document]
tchung1118: This is actually a left-over comment from the previous implementation of `annotated.parameters()` Will either just delete or find the new parameters function on ParamTags and apply the relevant changes in the comment

annotated.go
tchung1118: That's correct. However, since the original function being annotated  cannot have any Out struct in its return (checked by `typeCheckOrigFn()`), it's safe to assume that an `fx.ResultTags` was applied if the first return type is an `fx.Out`. Obviously this might change if we start allowing the original function being annotated to have an `fx.Out` struct in its return values, but that's for later.

annotated.go
```diff
@@ -637,62 +1278,110 @@ func (ann *annotated) Build() (interface{}, error) {
...
 
-	hookFns := make([]reflect.Value, len(ann.Hooks))
-	for i, builder := range ann.Hooks {
-		if hookFn := builder.Build(resultTypes...); !hookFn.IsZero() {
-			hookFns[i] = hookFn
+	ann.postbuild()
+
+	for _, la := range lcHookAnns {
+		if ann.Target, err = la.build(ann); err != nil {
+			return nil, err
```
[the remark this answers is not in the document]
tchung1118: So I checked the implementation of `*lifecycleHookAnnotation.build()`, and apparently it's impossible to return non-nil error. This is because I do the dependency check for the hook function by invoking a function in a private scope when the annotated function is called. So if there's a problem with dependencies of the hook function, it fails while the app/module is running provides, instead of failing while building the annotated function. There's no other failure case because all other sanity checks on hook functions when `Annotation.apply()` is called.

Since I do not want the signature of build() method of Annotation interface to change because other types of annotation may return error when building, should I just have an unhandled error here?

Also a similar case of uncovered code is the call to `scope.Provide` in the hookInstaller. I know this call to Provide cannot fail because I have the full control over what's passed to that Provide, and I have codes to make sure we don't provide an invalid constructor to that scope. (by removing context.Context and deduplicating types)

annotated.go
tchung1118: That being said, just realized this is used only in one place. I'm just going to remove this function and inline the logic where it's used.

annotated.go
```diff
@@ -189,13 +288,170 @@ func (rt resultTagsAnnotation) apply(ann *annotated) error {
 	return nil
 }
 
+// build builds and returns a constructor after applying a ResultTags annotation
+func (rt resultTagsAnnotation) build(ann *annotated) (interface{}, error) {
+	paramTypes := ann.currentParamTypes()
+
+	resultTypes, remapResults := rt.results(ann)
+
```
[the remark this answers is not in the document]
tchung1118: Deleted

annotated.go
```diff
@@ -582,31 +1059,193 @@ var _ Annotation = asAnnotation{}
...
+	fields := []reflect.StructField{
+		{
+			Name:      "Out",
+			Type:      reflect.TypeOf(Out{}),
+			Anonymous: true,
+		},
+	}
+
+	// if already went through a ResultTags annotation,
```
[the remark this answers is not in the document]
tchung1118: Extracted the first part to another function and unified second and third

annotated.go
```diff
@@ -637,62 +1278,110 @@ func (ann *annotated) Build() (interface{}, error) {
...
 	}
 
-	resultTypes, remapResults, err := ann.results()
-	if err != nil {
-		return nil, err
+	// app
+	ann.prebuild()
+
+	var err error
+	var lcHookAnns []*lifecycleHookAnnotation
```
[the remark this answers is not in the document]
tchung1118: Done

annotated.go
```diff
@@ -637,62 +1278,110 @@ func (ann *annotated) Build() (interface{}, error) {
 		return nil, fmt.Errorf("invalid annotation function %T: %w", ann.Target, err)
 	}
 
-	resultTypes, remapResults, err := ann.results()
-	if err != nil {
-		return nil, err
+	// app
```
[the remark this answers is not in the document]
tchung1118: Removed

annotated.go
```diff
@@ -637,62 +1278,110 @@ func (ann *annotated) Build() (interface{}, error) {
...
+	var lcHookAnns []*lifecycleHookAnnotation
+	for _, annotation := range ann.Annotations {
+		switch annotation := annotation.(type) {
+		case *lifecycleHookAnnotation:
+			lcHookAnns = append(lcHookAnns, annotation)
+		default:
+			if ann.Target, err = annotation.build(ann); err != nil {
+				return nil, err
+			}
+		}
```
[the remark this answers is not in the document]
tchung1118: Changed to an `if`

annotated.go
```diff
@@ -637,62 +1278,110 @@ func (ann *annotated) Build() (interface{}, error) {
...
+	ann.postbuild()
+
+	for _, la := range lcHookAnns {
+		if ann.Target, err = la.build(ann); err != nil {
+			return nil, err
 		}
 	}
+	// before returning the transformed function, do a final check to see if an As annotation was applied
+	// if there are more than one Out struct returning, at least one As annotation was applied
+	// if it was, wrap the function one more time to remove the first out type
```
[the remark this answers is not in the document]
tchung1118: Removed

annotated.go
```diff
@@ -637,62 +1278,110 @@ func (ann *annotated) Build() (interface{}, error) {
 		return nil, fmt.Errorf("invalid annotation function %T: %w", ann.Target, err)
 	}
 
-	resultTypes, remapResults, err := ann.results()
-	if err != nil {
-		return nil, err
+	// app
+	ann.prebuild()
```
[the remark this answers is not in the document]
tchung1118: Renamed to `prebuild` to `applyOptionalTag` and `postbuild` to `cleanUpAsResults`

annotated.go
```diff
@@ -628,6 +1267,8 @@ func (ann annotated) String() string {
 // Build builds and returns a constructor based on fx.In/fx.Out params and
 // results wrapping the original constructor passed to fx.Annotate.
 func (ann *annotated) Build() (interface{}, error) {
+	// return ann.build()
```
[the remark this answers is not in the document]
tchung1118: Lol removed

annotated.go
```diff
@@ -488,12 +892,40 @@ func (la *lifecycleHookAnnotation) Build(results ...reflect.Type) reflect.Value
...
+//		FieldB B `name:"B"`
+//	}
+//
 // Only one OnStart annotation may be applied to a given function at a time,
 // however functions may be annotated with other types of lifecylce Hooks, such
-// as OnStop.
+// as OnStop. The hook function passed into OnStart cannot take any arguments
+// outside of the annotated constructor's existing dependencies or results, except
+// for a context.Context, which is injected by Lifecycle, or Lifecycle itself, which
+// is automatically added by the annotation to append the hook function.
```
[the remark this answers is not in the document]
tchung1118: Applied suggested change

annotated.go
```diff
@@ -488,12 +892,40 @@ func (la *lifecycleHookAnnotation) Build(results ...reflect.Type) reflect.Value
 //			    return server.Listen(ctx)
 //		      },
 //	     })
+//		 return server
 //	   }
 //	 )
 //
+// It is also possible to use OnStart annotation with other parameter and result
+// annotations, provided that the parameter of the function passed to OnStart
+// matches annotated parameters and results
```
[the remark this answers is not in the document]
tchung1118: Applied suggested change

annotated.go
tchung1118: Applied suggested change

annotated.go
tchung1118: Done

annotated.go
```diff
@@ -637,62 +1278,110 @@ func (ann *annotated) Build() (interface{}, error) {
...
+		newResultTypes := resultTypes[len(resultTypes)-numRes:]
+		origFn := reflect.ValueOf(ann.Target)
+		newFnType := reflect.FuncOf(paramTypes, newResultTypes, false)
+		newFn := reflect.MakeFunc(newFnType, func(args []reflect.Value) (results []reflect.Value) {
+			results = origFn.Call(args)
+			results = results[len(results)-numRes:]
+			return
+		})
+		ann.Target = newFn.Interface()
+	}
```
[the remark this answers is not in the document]
tchung1118: Good call. Done.

annotated.go
tchung1118: Removed

annotated.go
tchung1118: Done

annotated.go
tchung1118: Done

annotated.go
tchung1118: Fixed

annotated.go
tchung1118: This is cool. Changed `seen` to `map[key]struct{}`

annotated.go
tchung1118: Removed xD

sywhang (approved):
LGTM other than some outlying things re: testing and some updated docstring. We can do some more post cleanups in smaller chunks.

annotated_test.go
```diff
@@ -330,6 +396,24 @@ func TestAnnotatedAsFailures(t *testing.T) {
 			invoke:        func() {},
 			errorContains: "does not implement",
 		},
+		{
+			desc:          "provide when an illegal type As with result tag",
+			provide:       fx.Provide(fx.Annotate(newAsStringer, fx.ResultTags(`name:"stringer"`), fx.As(new(io.Writer)))),
+			invoke:        func() {},
+			errorContains: "does not implement",
```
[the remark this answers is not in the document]
tchung1118: Could you elaborate more on this? I'm assuming this doesn't mean "change this to a longer substring of the error message."

tchung1118:
Also I just tested, and I think this PR also implemented suggestion mentioned in #936.

annotated.go
```diff
@@ -519,16 +942,43 @@ func OnStart(onStart interface{}) Annotation {
...
+//	type OnStopParams struct {
+//		fx.In
+//		FieldA A `name:"A"`
+//		FieldB B `name:"B"`
+//	}
+//
 // Only one OnStop annotation may be applied to a given function at a time,
 // however functions may be annotated with other types of lifecylce Hooks, such
-// as OnStart.
+// as OnStop. The hook function passed into OnStop cannot take any arguments
```
jkanywhere, line 979: Misspelled `lifecylce`.

Also this diff replaces `OnStart` with `OnStop` so `OnStop` is used twice in the sentence. The words "other types" makes me think it should be `OnStop` followed by `OnStart` as before this diff.

"Only one **OnStop** annotation ... however functions may be annotated with _other types_ of lifecylce Hooks, such as **OnStop**"

PR #984, opened 2022-11-01
Refactor Done() to use a broadcasted Signal type for future features. 

jasonmills:
This creates a new public struct type `ShutdownSignal` and moves broadcast of operating system signals to a standalone type in it's own source file. This allows for the cleaner expansion of signaling features.

### DIFF
--- a/signal.go
+++ b/signal.go
@@ -32,7 +32,7 @@ type ShutdownSignal struct {
 	Signal os.Signal
 }
 
-// String will render a Signal type as a string suitable for printing.
+// String will render a ShutdownSignal type as a string suitable for printing.
 func (sig ShutdownSignal) String() string {
 	return fmt.Sprintf("%v", sig.Signal)
 }
--- a/signal.go
+++ b/signal.go
@@ -37,13 +37,18 @@ func (sig ShutdownSignal) String() string {
 	return fmt.Sprintf("%v", sig.Signal)
 }
 
-type signalReceivers struct {
-	m    sync.Mutex
-	last *ShutdownSignal
-	done []chan os.Signal
+func newSignalReceivers() signalReceivers {
+	return signalReceivers{notify: signal.Notify}
 }
 
-func (recv *signalReceivers) Done() chan os.Signal {
+type signalReceivers struct {
+	m      sync.Mutex
+	last   *ShutdownSignal
+	done   []chan os.Signal
+	notify func(c chan<- os.Signal, sig ...os.Signal)
+}
+
+func (recv *signalReceivers) Done() <-chan os.Signal {
 	recv.m.Lock()
 	defer recv.m.Unlock()
 
@@ -57,7 +62,7 @@ func (recv *signalReceivers) Done() chan os.Signal {
 		ch <- recv.last.Signal
 	}
 
-	signal.Notify(ch, os.Interrupt, _sigINT, _sigTERM)
+	recv.notify(ch, os.Interrupt, _sigINT, _sigTERM)
 	recv.done = append(recv.done, ch)
 	return ch
 }
--- a/signal.go
+++ b/signal.go
@@ -40,7 +40,7 @@ func (sig ShutdownSignal) String() string {
 type signalReceivers struct {
 	m     sync.Mutex
 	last  *ShutdownSignal
-	dones []chan os.Signal
+	done []chan os.Signal
 }
 
 func (recv *signalReceivers) Done() chan os.Signal {
@@ -56,7 +56,7 @@ func (recv *signalReceivers) Done() chan os.Signal {
 	}
 
 	signal.Notify(ch, os.Interrupt, _sigINT, _sigTERM)
-	recv.dones = append(recv.dones, ch)
+	recv.done = append(recv.done, ch)
 	return ch
 }
 
@@ -80,11 +80,11 @@ func (recv *signalReceivers) Broadcast(signal ShutdownSignal) error {
 
 func (recv *signalReceivers) broadcastDone(signal ShutdownSignal) (int, int) {
 	var (
-		receivers int = len(recv.dones)
+		receivers int = len(recv.done)
 		unsent    int
 	)
 
-	for _, reader := range recv.dones {
+	for _, reader := range recv.done {
 		select {
 		case reader <- signal.OS:
 		default:
--- a/signal.go
+++ b/signal.go
@@ -86,10 +86,7 @@ func (recv *signalReceivers) Broadcast(signal ShutdownSignal) error {
 }
 
 func (recv *signalReceivers) broadcastDone(signal ShutdownSignal) (int, int) {
-	var (
-		receivers int = len(recv.done)
-		unsent    int
-	)
+	var unsent int
 
 	for _, reader := range recv.done {
 		select {
@@ -99,7 +96,7 @@ func (recv *signalReceivers) broadcastDone(signal ShutdownSignal) (int, int) {
 		}
 	}
 
-	return receivers, unsent
+	return len(recv.done), unsent
 }
 
 type unsentSignalError struct {
--- a/signal.go
+++ b/signal.go
@@ -49,8 +49,10 @@ func (recv *signalReceivers) Done() chan os.Signal {
 
 	ch := make(chan os.Signal, 1)
 
-	// if we had received a signal prior to the call of done, send it's
+	// If we had received a signal prior to the call of done, send it's
 	// os.Signal to the new channel.
+	// However we still want to have the operating system notify signals to this
+	// channel should the application receive another.
 	if recv.last != nil {
 		ch <- recv.last.Signal
 	}
--- a/signal.go
+++ b/signal.go
@@ -78,8 +78,11 @@ func (recv *signalReceivers) Broadcast(signal ShutdownSignal) error {
 	return nil
 }
 
-func (recv *signalReceivers) broadcastDone(signal ShutdownSignal) (receivers, unsent int) {
-	receivers = len(recv.dones)
+func (recv *signalReceivers) broadcastDone(signal ShutdownSignal) (int, int) {
+	var (
+		receivers int = len(recv.dones)
+		unsent    int
+	)
 
 	for _, reader := range recv.dones {
 		select {
@@ -89,7 +92,7 @@ func (recv *signalReceivers) broadcastDone(signal ShutdownSignal) (receivers, un
 		}
 	}
 
-	return
+	return receivers, unsent
 }
 
 type unsentSignalError struct {
--- a/signal.go
+++ b/signal.go
@@ -38,17 +38,14 @@ func (sig Signal) String() string {
 }
 
 type signalReceivers struct {
-	last     *Signal
-	lastLock sync.RWMutex
-	dones    []chan os.Signal
-	doneLock sync.RWMutex
+	m     sync.Mutex
+	last  *Signal
+	dones []chan os.Signal
 }
 
 func (recv *signalReceivers) done() chan os.Signal {
-	recv.doneLock.Lock()
-	defer recv.doneLock.Unlock()
-	recv.lastLock.RLock()
-	defer recv.lastLock.RUnlock()
+	recv.m.Lock()
+	defer recv.m.Unlock()
 
 	ch := make(chan os.Signal, 1)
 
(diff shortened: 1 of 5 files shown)


signal.go
```diff
@@ -0,0 +1,114 @@
...
+	recv.lastLock.RLock()
+	defer recv.lastLock.RUnlock()
+
+	ch := make(chan os.Signal, 1)
+
+	// if we had received a signal prior to the call of done, send it's
+	// os.Signal to the new channel.
+	if recv.last != nil {
+		ch <- recv.last.OS
+	}
```
abhinav, line 59: If we already received a signal, we don't need to modify dones or register a signal.Notify, right?
This can be something like this:

```
ch := make(..)

lastLock.RLock()
defer lastLock.RUnlock()
if recv.last != nil {
  ch <- recv.last.OS
  return ch
}

doneLock.Lock()
defer doneLock.Unlock()
// ...
```

Although I still don't love having to reason about the race conditions with double-locking.
Is a single RWLock out of the question here?

signal.go
```diff
@@ -0,0 +1,114 @@
...
+
+	receivers = len(recv.dones)
+
+	for _, reader := range recv.dones {
+		select {
+		case reader <- signal.OS:
+		default:
+			unsent++
+		}
+	}
```
abhinav, line 78: Do we need to nil out recv.dones after going through them?

signal.go
```diff
@@ -0,0 +1,114 @@
...
+
+	return
```
abhinav, line 80: nit: prefer `return receivers, unset` over naked return here.
Otherwise the reader has to go up and around to find out what's getting returned.

see also https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters,

> Naked returns are okay if the function is a handful of lines. Once it's a medium sized function, be explicit with your return values.

I think this might be on the boundary, but I lean "explicit return" here.

signal.go
```diff
@@ -0,0 +1,114 @@
...
+		err.Channels,
+	)
+}
+
+func (recv *signalReceivers) broadcast(signal Signal) (err error) {
+	recv.lastLock.Lock()
+	recv.last = &signal
+	recv.lastLock.Unlock()
+
+	channels, unsent := recv.broadcastDone(signal)
```
abhinav, line 103: Okay, yeah, looking at this, looks like a single lock should do it?
Because we almost always acquire both locks:

- broadcastDone is not called directly, only by broadcast, so we have both locks
- done takes both locks

signal.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+type signalReceivers struct {
+	last     *Signal
+	lastLock sync.RWMutex
+	dones    []chan os.Signal
+	doneLock sync.RWMutex
+}
+
+func (recv *signalReceivers) done() chan os.Signal {
```
abhinav, line 47: minor habitual thing:
for private/unexported types like signalReceivers,
consider still exporting its methods (as long as the type is private)
to differentiate between the type's public vs private API.

Like, with signalReceivers, 'done' and 'broadcast' are meant to be called by app.go, 'broadcastSignal' is not, but someone reading this code wouldn't know that.
If done and broadcast were exported, that would be clearer.

signal.go
```diff
@@ -0,0 +1,114 @@
...
+
+// Signal represents a operating system process signal.
+type Signal struct {
+	OS os.Signal
+}
+
+// String will render a Signal type as a string suitable for printing.
+func (sig Signal) String() string {
+	return fmt.Sprintf("%v", sig.OS)
+}
```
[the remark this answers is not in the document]
jasonmills: yeah, this is for the follow up to this to support the exit code option for the shutdowner interface.

signal.go
jasonmills: The current behavior does not nil out the dones, we create those channels and return them but do not close them, if we changed the behavior to clean up the channels we've created (by closing them and then flushing the set of receivers) we could introduce cases in which users are reading from a done channel that has been closed which would cause panics. 

This does bring up the point however that in the existing shutdowner (1.18.2 at the time of this comment) we do not guard against writing to a closed done channel, so the user could close this channel and panic the app https://github.com/uber-go/fx/blob/v1.18.2/shutdown.go#L32

signal.go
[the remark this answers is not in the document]
abhinav: Ack. Per offline discussion, please call this ShutdownSignal.

signal.go
jasonmills: I had a side discussion with @sywhang about this the other day, and private seemed like the way to go, but I see your point about it helping a reader understanding that "although this type is private, it's API is these methods", when using public methods. I'll update it.

signal.go
```diff
@@ -0,0 +1,108 @@
...
+		err.Signal,
+		err.Unsent,
+		err.Channels,
+	)
+}
+
+func (recv *signalReceivers) broadcast(signal ShutdownSignal) error {
+	recv.m.Lock()
+	defer recv.m.Unlock()
+	recv.last = &signal
```
abhinav, line 95: Okay after thinking about this a second, I realize now why I don't like the `recv.last`.
It's leaky—it only recognizes `Shutdown()` calls and not the `signal.Notify` call.
That said, this is fine because:

- that's what the old behavior was
- once you add the signal.Notify adapter you're planning on adding, it can be `recv.last` aware in a future change

signal.go
```diff
@@ -0,0 +1,108 @@
...
+
+	channels, unsent := recv.broadcastDone(signal)
+
+	if unsent != 0 {
```
abhinav, line 99: optional: broadcast, and broadcastSignal are pretty small and don't get used independently.
Consider inlining implementation of broadcastSignal here.

Also, minor ordering thing: consider moving broadcast above the error declaration.
On my first read-through of this code, I skipped over broadcast because I thought it was a method on unsentSignalError.

signal_test.go
```diff
@@ -0,0 +1,73 @@
...
+	expected *unsentSignalError,
+) {
+	t.Helper()
+
+	actual := new(unsentSignalError)
+
+	assert.Error(t, err)
+	assert.ErrorContains(t, err, "channels are blocked")
+	assert.ErrorAs(t, err, &actual, "is unsentSignalError")
+	assert.Equal(t, expected, actual)
```
abhinav, line 41: minor: assert.Equal is meaningless if ErrorAs fails. Maybe put it behind a condition.
```suggestion
	if assert.ErrorAs(t, err, &actual, "is unsentSignalError") {
		assert.Equal(t, expected, actual)
	}
```

signal_test.go
```diff
@@ -0,0 +1,73 @@
...
+	t *testing.T,
+	err error,
+	expected *unsentSignalError,
+) {
+	t.Helper()
+
+	actual := new(unsentSignalError)
+
+	assert.Error(t, err)
+	assert.ErrorContains(t, err, "channels are blocked")
```
abhinav, line 39: If you're doing ErrorContains, assert.Error is not needed because it'll check non-nil first.
```suggestion
	assert.ErrorContains(t, err, "channels are blocked")
```

signal_test.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func TestSignal(t *testing.T) {
+	t.Parallel()
+	recv := new(signalReceivers)
+	a := recv.done()
+	b := recv.done()
+
+	assert.NotNil(t, a)
+	assert.NotNil(t, b)
```
abhinav, line 51: nit: unnecessary check. assume non-nil if nil/non-nil isn't conditional/part of the contract.
```suggestion
```

signal_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+	expected := ShutdownSignal{
+		OS: syscall.SIGTERM,
+	}
+
+	err := recv.broadcast(expected)
+
+	assert.NoError(t, err, "first broadcast should succeed")
```
abhinav, line 59: require.NoError, probably since we don't expect the rest to work if this is broken.
I might even inline it:

```
	assert.NoError(t, recv.broadcast(expected), "first broadcast should succeed")
```

signal_test.go
```diff
@@ -0,0 +1,73 @@
...
+	assert.ErrorContains(t, err, "channels are blocked")
+	assert.ErrorAs(t, err, &actual, "is unsentSignalError")
+	assert.Equal(t, expected, actual)
+}
+
+func TestSignal(t *testing.T) {
+	t.Parallel()
+	recv := new(signalReceivers)
+	a := recv.done()
+	b := recv.done()
```
abhinav, line 48: we don't use b ever again so let's just throw it away:
```suggestion
	donec := recv.done()
	_ = recv.done()  // we never listen on this
```

signal_test.go
```diff
@@ -0,0 +1,73 @@
...
+	assertUnsentSignalError(t, err, &unsentSignalError{
+		Signal:   expected,
+		Channels: 2,
+		Unsent:   2,
+	})
+
+	actual := <-a
+	assert.Equal(t, expected.OS, actual)
+
+	assert.Equal(t, expected.OS, <-recv.done(), "expect cached signal")
```
abhinav, line 72: minor: inline the `<-a` so it matches the next assert.Equal visually.
```suggestion
	assert.Equal(t, expected.OS, <-a, "expect recorded signal")
	assert.Equal(t, expected.OS, <-recv.done(), "expect cached signal")
```

signal.go
jasonmills: This should be fixed now.

signal.go
```diff
@@ -0,0 +1,113 @@
...
+	return fmt.Sprintf("%v", sig.Signal)
+}
+
+type signalReceivers struct {
+	m    sync.Mutex
+	last *ShutdownSignal
+	done []chan os.Signal
+}
+
+func (recv *signalReceivers) Done() chan os.Signal {
```
abhinav, line 46: minor: we should return a read only channel

```suggestion
func (recv *signalReceivers) Done() <-chan os.Signal {
```

signal.go
```diff
@@ -0,0 +1,113 @@
...
+
+	for _, reader := range recv.done {
+		select {
+		case reader <- signal.Signal:
+		default:
+			unsent++
+		}
+	}
+
+	return receivers, unsent
```
abhinav, line 97: minor: don't need receivers variable at all
```suggestion
	var unsent int
	for _, reader := range recv.done {
		select {
		case reader <- signal.Signal:
		default:
			unsent++
		}
	}
	return len(recv.done), unsent
```

signal.go
```diff
@@ -0,0 +1,113 @@
...
+	recv.m.Lock()
+	defer recv.m.Unlock()
+	recv.last = &signal
+
+	channels, unsent := recv.broadcastDone(signal)
+
+	if unsent != 0 {
+		return &unsentSignalError{
+			Signal:   signal,
+			Channels: channels,
```
abhinav, line 75:
```suggestion
	total, unsent := recv.broadcastDone(signal)

	if unsent != 0 {
		return &unsentSignalError{
			Signal:   signal,
			Channels: total,
```

signal.go
```diff
@@ -0,0 +1,113 @@
...
+		}
+	}
+
+	return receivers, unsent
+}
+
+type unsentSignalError struct {
+	Signal   ShutdownSignal
+	Unsent   int
+	Channels int
```
abhinav, line 103:
```suggestion
	Unsent int
	Total  int
```

signal.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+func (err *unsentSignalError) Error() string {
+	return fmt.Sprintf(
+		"send %v signal: %v/%v channels are blocked",
+		err.Signal,
+		err.Unsent,
+		err.Channels,
```
abhinav, line 111:
```suggestion
		"send %v signal: %v/%v channels are blocked",
		err.Signal,
		err.Unsent,
		err.Total,
```

abhinav (approved):
LGTM with comments.

Please fix those comments and get another stamp.

signal.go
```diff
@@ -0,0 +1,115 @@
...
+	"os/signal"
+	"sync"
+)
+
+// ShutdownSignal is a signal that caused the application to exit.
+type ShutdownSignal struct {
+	Signal os.Signal
+}
+
+// String will render a Signal type as a string suitable for printing.
```
sywhang, line 35:
```suggestion
// String will render a ShutdownSignal type as a string suitable for printing.
```

sywhang (approved):
👍 LGTM modulo a typo in doc. Please address that before merging.

PR #987, opened 2022-11-11
Cleanup context.Context used in unit tests

jasonmills:
A number of unit tests do not correctly clean up context.Context types used in tests that call `app.Start` and `app.Stop`, the expected behavior is that context's should be canceled when they are no longer needed; canceling a context should force the application lifecycle to stop.

### DIFF
--- a/annotated_test.go
+++ b/annotated_test.go
@@ -1038,7 +1038,8 @@ func assertApp(
 	invoked *bool,
 ) {
 	t.Helper()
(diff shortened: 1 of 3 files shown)


sywhang (commented):
Synced offline, might be able to pass child context instead to make this PR unnecessary.

jasonmills:
Closing as it is no-longer required.

PR #989, opened 2022-11-15
Provides ExitCode Shutdowner Option; and Wait method to receive it.

jasonmills:
This PR provides an option for those who take dependencies on the `Shutdowner` interface to call the `Shutdown` method with an `ExitCode` option, in addition it add a `Wait` method to the application to allow for main programs to wait for the application to be shutdown and to exit with a given exit code. 

Please note that this PR refactors the existing signal relay functionality, and alters application lifecycle slightly. Now `Done` will not receive an `os.Signal` on the channel it returns _unless_ a given FX application has been started.


shutdown.go
```diff
@@ -20,6 +20,13 @@
 
 package fx
 
+import (
+	"context"
+	"time"
+)
+
+const _receiverShutdownTimeout = 10 * time.Millisecond
```
sywhang, line 28: While I think this is a _reasonable_ timeout value, I don't think it's safe to assume that it will be for all applications.

How about just sending `context.Background()` to `receivers.Stop`, and in the future if it's necessary, we can expose a ShutdownOption to set the receivers timeout context directly from the user.

signal.go
```diff
@@ -38,17 +45,93 @@ func (sig ShutdownSignal) String() string {
...
+	done []chan os.Signal
+	wait []chan ShutdownSignal
+}
+
+func (recv *signalReceivers) relayer(ctx context.Context) {
+	defer func() {
+		select {
+		case recv.finished <- struct{}{}:
+		default:
+		}
```
sywhang, line 71: I believe this whole thing can just be a simple send?
```suggestion
	recv.finished <- struct{}{}
```

sywhang (commented):
Took a first pass. Looking good!

shutdown.go
jasonmills: I'll add the option, I don't think it introduces much complexity

shutdown_test.go
```diff
@@ -87,6 +91,43 @@ func TestShutdown(t *testing.T) {
...
 		assert.NotNil(t, <-done2, "done channel 2 did not receive signal")
 	})
+
+	t.Run("with exit code", func(t *testing.T) {
+		t.Parallel()
+		var s fx.Shutdowner
+		app := fxtest.New(
+			t,
+			fx.Populate(&s),
+		)
```
JacobOaks, line 101: I realize this PR didn't add this pattern, but I realize all of these tests share this boilerplate. Maybe worth cleaning this up and converting to a test table?

jasonmills: We couldn't generalize the t.Parallel call without abstracting a whole case runner around populate. 

A new fxtest function would be possible, although it doesn't really reduce boilerplate too much, something like: 

`func NewPopulated(t testingT, target interface{}) fx.App { 
   return fxtest.New(t, fx.Populate(target))
}`

signal.go
```diff
@@ -38,17 +45,104 @@ func (sig ShutdownSignal) String() string {
...
+			return ctx.Err()
+		case <-recv.finished:
+			close(recv.shutdown)
+			close(recv.finished)
+			recv.shutdown = nil
+			recv.finished = nil
+			return nil
+		}
+
+	}
```
sywhang, line 142: This doesn't need to be in a infinite loop. 
```suggestion
	select {
	case <-ctx.Done():
		return ctx.Err()
	case <-recv.finished:
		close(recv.shutdown)
		close(recv.finished)
		recv.shutdown = nil
		recv.finished = nil
		return nil
	}
```

signal.go
```diff
@@ -38,17 +45,104 @@ func (sig ShutdownSignal) String() string {
...
+		return
+	case signal := <-recv.signals:
+		recv.Broadcast(ShutdownSignal{
+			Signal: signal,
+		})
+	}
+}
+
+func (recv *signalReceivers) running() bool {
+	return recv.shutdown != nil && recv.finished != nil
```
sywhang, line 100: Consider adding a comment here that says that this method must be invoked under `recv.m.Lock()`'s protection; otherwise it's prone to race.

sywhang (approved):

PR #990, opened 2022-11-16
Add fx.From annotation

kkHAIKE:
open a new PR for https://github.com/uber-go/fx/pull/938,
because long time no progress update. 😭

### DIFF
--- a/annotated.go
+++ b/annotated.go
@@ -1207,8 +1207,8 @@ var _ Annotation = (*fromAnnotation)(nil)
 //	  return NewRunnerWrap(r)
 //	})
 //
-// When multiple values are returned by the annotated function, each type
-// gets mapped to corresponding positional parameter of the annotated function
+// When the annotated function takes in multiple parameters, each type gets
+// mapped to corresponding positional parameter of the annotated function
 //
 // For example,
 //
--- a/annotated.go
+++ b/annotated.go
@@ -1183,7 +1183,7 @@ type fromAnnotation struct {
 
 var _ Annotation = (*fromAnnotation)(nil)
 
-// From is an Annotation that annotates the parameter(s) for a function (i.e. a
+// From is an [Annotation] that annotates the parameter(s) for a function (i.e. a
 // constructor) to be accepted from other provided types. It is analogous to the
 // [As] for parameter types to the constructor.
 //
--- a/annotated.go
+++ b/annotated.go
@@ -1178,9 +1178,10 @@ func extractResultFields(types []reflect.Type) ([]reflect.StructField, func(int,
 
 type fromAnnotation struct {
 	targets []interface{}
+	types   []reflect.Type
 }
 
-var _ Annotation = fromAnnotation{}
+var _ Annotation = (*fromAnnotation)(nil)
 
 // From is an Annotation that annotates some parameter for a function (i.e. a
 // constructor) that accept interface from other provided type. Very much the opposite of fx.As(...).
@@ -1216,14 +1217,14 @@ var _ Annotation = fromAnnotation{}
 //	  ),
 //	)
 func From(interfaces ...interface{}) Annotation {
-	return fromAnnotation{targets: interfaces}
+	return &fromAnnotation{targets: interfaces}
 }
 
-func (fr fromAnnotation) apply(ann *annotated) error {
+func (fr *fromAnnotation) apply(ann *annotated) error {
 	if len(ann.From) > 0 {
-		return fmt.Errorf("fx.From does not support multiple annotations")
+		return errors.New("cannot apply more than one line of From")
 	}
-	types := make([]reflect.Type, len(fr.targets))
+	fr.types = make([]reflect.Type, len(fr.targets))
 	for i, typ := range fr.targets {
 		t := reflect.TypeOf(typ)
 		if t != nil && t.Kind() != reflect.Ptr {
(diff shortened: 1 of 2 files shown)


kkHAIKE:
@sywhang sure

kkHAIKE:
ping @sywhang 😊

annotated_test.go
```diff
@@ -86,6 +94,349 @@ func (a anotherStringer) String() string {
...
+					func(writer io.Writer) fmt.Stringer {
+						return &fromStringer{}
+					},
+					fx.From(new(*fromStringer)),
+				),
+			),
+			invoke: func(stringer fmt.Stringer) {
+				fmt.Println(stringer.String())
+			},
+			errorContains: "does not implement",
```
tchung1118, line 344: It would be good to include some type information in this error message.

annotated_test.go
```diff
@@ -86,6 +94,349 @@ func (a anotherStringer) String() string {
...
+					func(writer io.Writer, x ...int) fmt.Stringer {
+						return &fromStringer{}
+					},
+					fx.From(new(*fromStringer)),
+				),
+			),
+			invoke: func(stringer fmt.Stringer) {
+				fmt.Println(stringer.String())
+			},
+			errorContains: "does not implement",
```
tchung1118, line 359: Same as the above

annotated.go
```diff
@@ -1176,12 +1176,181 @@ func extractResultFields(types []reflect.Type) ([]reflect.StructField, func(int,
...
+
+func (fr *fromAnnotation) apply(ann *annotated) error {
+	if len(ann.From) > 0 {
+		return errors.New("cannot apply more than one line of From")
+	}
+	fr.types = make([]reflect.Type, len(fr.targets))
+	for i, typ := range fr.targets {
+		t := reflect.TypeOf(typ)
+		if t != nil && t.Kind() != reflect.Ptr {
+			return fmt.Errorf("fx.From: argument must either be nil or a pointer to a type implements some interface: got %v", t)
```
tchung1118, line 1231:
```suggestion
			return fmt.Errorf("fx.From: argument must either be nil or a pointer to a type that implements some interface: got %v", t)
```

tchung1118 (commented):
Took a first pass. Looks good, just a few comments about a minor feature of the functionality and test cases.

annotated.go
```diff
@@ -1176,12 +1176,181 @@ func extractResultFields(types []reflect.Type) ([]reflect.StructField, func(int,
...
+//	})
+//
+// If you don't care about certain parameters, you can use nil as a placeholder.
+//
+//	func NewRunnerWrap(o *Object, r Runner) *BarRunner // implements Runner
+//
+//	fx.Provide(
+//	  fx.Annotate(
+//	    NewRunnerWrap,
+//	    fx.From(nil, new(*FooRunner)),
```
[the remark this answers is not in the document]
kkHAIKE: I can consider bringing the parameters that need to be concerned to the front to remove this feature

annotated_test.go
```diff
@@ -86,6 +94,327 @@ func (a anotherStringer) String() string {
...
+					},
+					fx.From(),
+				),
+			),
+			invoke: func(s fmt.Stringer) {
+				assert.Equal(t, s.String(), "a good stringer")
+			},
+		},
+		{
+			desc: "Provide with variadic function",
```
tchung1118, line 258: I'm wondering how From annotation will work when it's applied directly to the variadic parameter of the constructor. So something like this?
```
fx.Provide(
  fx.Annotate(
    func(a A, b ...B) C {...},
    fx.From(new(StructA), new(StructB)),
  ),
)
```
Could you investigate this? I'm guessing it will have some problems (either with an error or a panic) in the current implementation. In that case, could you add a failure test case if it's failing with an error, or add a check for this case if it panics for some reason?

tchung1118: I'm also not sure if it even makes sense to apply From annotation to the variadic parameter. If it actually doesn't make sense, I think we should return an error when fromAnnotation.apply() is called for this case.

tchung1118 (commented):
Added a few more comments.

annotated_test.go
kkHAIKE: i think it doesn't make sense..

tchung1118 (approved):
There's a minor comment on the documentation, but approving because LGTM otherwise.

annotated.go
```diff
@@ -1176,12 +1176,193 @@ func extractResultFields(types []reflect.Type) ([]reflect.StructField, func(int,
...
 }
 
+type fromAnnotation struct {
+	targets []interface{}
+	types   []reflect.Type
+}
+
+var _ Annotation = (*fromAnnotation)(nil)
+
+// From is an Annotation that annotates the parameter(s) for a function (i.e. a
```
sywhang, line 1186:
```suggestion
// From is an [Annotation] that annotates the parameter(s) for a function (i.e. a
```

sywhang (approved):

sywhang:
Thanks for doing this work!

moredure:
@sywhang , is it possible to use fx.From only on some of the arguments/positional to fx.Provide'ed function?

tchung1118:
Yes, just like fx.As, fx.From will only apply to a subset of the parameters if you pass in fewer provided types than the number of parameters. For example, in the following code:

```
fx.Provide(
  func(a A, b, B) C { ... },
  fx.From(StructA),
)
```

fx.From is only applied to the first argument, and the second argument should stay the same. So the resulting constructor's signature will be:

```
func(sa StructA, b B) C
```

moredure:
Yes, in case of we have a control over constructor code we can reorder parameters to pass interfaces first, but in case of third party code for example constructor `func(*StructAsIs, InterfaceArgument) C`, can we do fx.From(new(StructAsIs), new(StructToInterfaceArgument))?

tchung1118:
I think the code in your latest comment should work, although that specific case seems to be missing from our test suite. So something like this:

```
fx.Provide(
  fx.Annotate(
    func(s *StructAsIs, i InterfaceArgument) C { ... },
    fx.From(new(StructAsIs), new(FromAnotherStruct)),
  )
)
```

should be equivalent to:
```
fx.Provide(
    func(s *StructAsIs,  f FromAnotherStruct) C { return origFunc(s, f) },
)
```

moredure:
Gotcha, thanks.

Issue #991, opened 2022-11-21
`index out of range` on (*Lifecycle).Stop()


mxplusb:
**Describe the bug**

When stopping the application with a `SIGINT`, the lifecycle hooks can panic.

```go
panic: runtime error: index out of range [9] with length 5

goroutine 1079 [running]:
go.uber.org/fx/internal/lifecycle.(*Lifecycle).Stop(0x140003aa0b0, {0x101e480d8, 0x1400098e840})
        /Users/sienna/go/pkg/mod/go.uber.org/fx@v1.18.2/internal/lifecycle/lifecycle.go:151 +0x718
go.uber.org/fx.withTimeout.func1()
        /Users/sienna/go/pkg/mod/go.uber.org/fx@v1.18.2/app.go:784 +0x74
created by go.uber.org/fx.withTimeout
        /Users/sienna/go/pkg/mod/go.uber.org/fx@v1.18.2/app.go:772 +0xd8
exit status 2
```

**To Reproduce**

Unclear, I'm not sure what's causing it. I would assume it's something to do with how the lifecycle hooks are registered.

**Expected behavior**

I expect it to error out instead of panicking.

**Additional context**

I am migrating an existing application to fx.

sywhang:
Thanks for reporting this. I'll try to repro this and triage this accordingly.

sywhang:
@JacobOaks to take a look at this issue.

JacobOaks:
Hey,

So I looked at a couple ways this panic can occur, and they all involve using the lifecycle a little differently. Can you clarify a little more about how you're running your app so I can try to narrow down what's happening? Specifically:

- Are you calling `fx.Run` or `fx.Start`/`fx.Stop`?
- If the latter, when/how are you stopping the app?
- Are you running or starting the app multiple times? Concurrently or once at a time?
- Are you specifying custom timeouts for your hooks or are any of them long-running?

Thanks!

mxplusb:
> * Are you calling `fx.Run` or `fx.Start`/`fx.Stop`?

I am calling `fx.New`, then `fx.Start`, then `fx.Stop` when the signal is caught.

> * If the latter, when/how are you stopping the app?

Just `CTRL+C` directly or `SIGINT` indirectly.

> * Are you running or starting the app multiple times? Concurrently or once at a time?

No, one instance of the application.

> * Are you specifying custom timeouts for your hooks or are any of them long-running?

[...]

PR #994, opened 2022-12-06
Stricter Control over Starting/Stopping Lifecycle

JacobOaks:
Currently, there is nothing preventing somebody from making calls to `app.Start` and `app.Stop` when it doesn't make sense, for example calling `app.Start` or `app.Stop` twice, or trying to run the application twice concurrently. Because of this lack of control, it's possible for users to cause race conditions and panics (#991) by putting the lifecycle is strange positions.

This PR places a small state machine in `Lifecycle` so that we can check if calls to `(*Lifecycle).Start` and `(*Lifecycle).Stop` make sense before going through. This will disallow starting an already starting/started app, and stopping an already stopping/stopped app. Tests to verify that are also added.

### DIFF
--- a/internal/lifecycle/lifecycle_test.go
+++ b/internal/lifecycle/lifecycle_test.go
@@ -144,6 +144,18 @@ func TestLifecycleStart(t *testing.T) {
 		// stop hooks.
 		require.NoError(t, l.Stop(ctx))
 	})
+
+	t.Run("StartWhileStartedErrors", func(t *testing.T) {
+		t.Parallel()
+
+		l := New(testLogger(t), fxclock.System)
+		assert.NoError(t, l.Start(context.Background()))
+		err := l.Start(context.Background())
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "attempted to start lifecycle when in state: started")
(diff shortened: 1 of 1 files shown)


internal/lifecycle/lifecycle.go
```diff
@@ -123,10 +123,21 @@ type Hook struct {
 	callerFrame fxreflect.Frame
 }
 
+type appState int
+
+const (
+	notRunning appState = iota
```
sywhang, line 129: nit: consider renaming this to `stopped`.

internal/lifecycle/lifecycle.go
```diff
@@ -221,6 +247,20 @@ func (l *Lifecycle) Stop(ctx context.Context) error {
 		return errors.New("called OnStop with nil context")
 	}
 
+	l.mu.Lock()
+	if l.state != started && l.state != incompleteStart {
+		l.mu.Unlock()
+		return errors.New("attempted to stop lifecycle when not running")
```
sywhang, line 253: this error message might make more sense if you printed the current state of `l.state` rather than just "not running" here since it could be either not running or starting or stopping.

sywhang (approved):
LGTM

Issue #856, opened 2022-03-04
Private Provides


sywhang:
With the addition of `fx.Module`, it is possible to add a `ProvideOption` that scopes a provided constructor to take the scope of a single `fx.Module`. 

Dig already supports this, because by default, providing to a `dig.Scope` limits the scope to a single Scope. Fx is currently providing using the `dig.Export(true)` to provide every dependency to the global `dig.Container` by default, so this option should be fairly easy to add.

hbdf:
Hey, I can take this addition. Here is what I propose it should look like:
```
fx.PrivateProvide(providerFunction)
```
Here are some rules that we should stick with (if I'm not mistaken, this is what's happening right now when we create a provide without exporting it in dig).

1. In essence, private provides should have the same capabilities as current global provides.
2. A module should use the closest provided object in the path to the container.
3. PrivateProvides can coexists (whitout erroring) with global provides and also with other private provides, given that they're not in the same scope/Module.
4. PrivateProvides will be decorated with all decorators from the scope it's used until the container.

In terms of implementation, I expect it to be really similar to Provide, just don't calling it with `dig.Export(true)` (as mentioned in the issue).

What do you think? :smile:

sywhang:
Sync'd offline with @hbdf. 

Provide doesn't take in a functional option today which is why the proposed design above was under an entirely new API. Instead of this I am proposing we use `fx.Annotate` to expose this:

```go
fx.Module("myModule",
  fx.Provide(
    fx.Annotate(
    NewTypeA,
      fx.Private(true), // name TBD
    ),
)
```

makes the output types of NewTypeA only be provided to "myModule".

sywhang:
Actually, the signature of `Provide` is an `interface{}`. 

@hbdf You could add this using a functional option then.

```
fx.Module("myModule",
  fx.Provide(
    NewTypeA,
    fx.Private(true),
  ),
)
```

where:
```go
func Private(isPrivate bool) provideOption {
  return provideOption{
    Private: isPrivate
  }
}
```

and you can change the implementation of `Provide` to check whether the provided ctor is a type of `provideOption` before injecting it to the container. 

If it's a type of `provideOption` you can mark all the ctors in that variadic parameter as private provide.

tchung1118:
Internal Ref: GO-1736

sywhang:
private provides are now available with v1.19.0 or later releases

PR #995, opened 2022-12-07
Private Provides

JacobOaks:
This change adds the ability to tag provides as "private", meaning the provided constructors will be added specifically to that module's scope, rather than the root Dig container scope, restricting usage of them to that module and modules it contains only.

Users can easily do this by adding `fx.Private` in their call to `fx.Provide()`:

```go
app := fx.New(
    fx.Module("SubModule",
        fx.Provide(
            MyPrivateProvideFunc,
            fx.Private // Now only accessible within SubModule
        )
    ),
    fx.Invoke(MyInvokeFunc) // Cannot rely on "MyPrivateProvideFunc"
)
```

Refs: #856


provide.go
```diff
@@ -72,14 +72,45 @@ type provideOption struct {
...
+// corresponding constructors can only be used within the current module
+// or modules the current module contains. Other modules that contain this
+// module won't be able to use the constructor.
+//
+// For example, the following would fail because the app doesn't have access
+// to the inner-module's constructor.
+//
+//	fx.New(
+//		Module("SubModule", Provide(func() int { return 0 }, Private())),
+//		Invoke(func(a int) {}),
```
tchung1118, line 108: Do we need to add `fx.` in front of these function calls?

module.go
```diff
@@ -147,7 +147,7 @@ func (m *module) provide(p provide) {
 	}
 
 	var info dig.ProvideInfo
-	if err := runProvide(m.scope, p, dig.FillProvideInfo(&info), dig.Export(true)); err != nil {
+	if err := runProvide(m.scope, p, dig.FillProvideInfo(&info), dig.Export(!p.Private)); err != nil {
```
tchung1118, line 150: This comment is not related to this line at all, but maybe it's worth adding `Private` field to `fxevent.Provided` to indicate a Provide was private in the event?

tchung1118 (commented):
First pass. Looks mostly good. Just had a couple comments

module.go
tchung1118: Also, if we do this, maybe we should update ZapLogger and ConsoleLogger implementation as well for logging Provided events

JacobOaks: Hmm good idea, let me know if you have thoughts on how I added it to `ConsoleLogger`

provide.go
```diff
@@ -72,14 +72,45 @@ type provideOption struct {
...
+// module won't be able to use the constructor.
+//
+// For example, the following would fail because the app doesn't have access
+// to the inner module's constructor.
+//
+//	fx.New(
+//		fx.Module("SubModule", fx.Provide(func() int { return 0 }, fx.Private())),
+//		fx.Invoke(func(a int) {}),
+//	)
+func Private() privateProvideOption {
```
tchung1118, line 110: This seems like a linter warning? Should we make a public interface for this to resolve it? I'm not sure what's the best route here.

fxevent/zap.go
```diff
@@ -118,6 +118,7 @@ func (l *ZapLogger) LogEvent(event Event) {
 				zap.String("constructor", e.ConstructorName),
 				moduleField(e.ModuleName),
 				zap.String("type", rtype),
+				zap.Bool("private", e.Private),
```
abhinav, line 121: Consider including this only if private is true. You can do that with something like the moduleField function:

```go
func trueBool(name string, b bool) zap.Field {
  if b {
    return zap.Bool(name, true)
  }
  return zap.Skip()
}
```

This will reduce noise in logs since private provides are rare.

provide.go
abhinav: Yeah, I would declare a new interface and seal it with an unexported method. Then make privateProvideOption implement that. Maybe 'ProvideOption'?

```go
type ProvideOption interface{ provideOption() }
func Private() ProvideOption { ... }
```

---

Separately, consider making `func Private(..)` accept a boolean:

```go
fx.Provide(a, b, c, fx.Private(true))
```

----

Another possibility is to make `fx.Private` a top-level variable with an unexported type.

```go
var Private = privateOption{}
```

Usage: `fx.Provide(..., fx.Private)`.

That way, you can mess with the type later like add new methods, make it implement other interfaces (like if Private needs to be a valid `Option` for something) without committing to it being a provide-specific thing.

abhinav (commented):
Also consider how this feature interacts with decoration.
What should happen when I decorate a privately provided type?
(Consider adding a test for that.)

Lastly, before release you may want to add a section under either Features or on the Modules page about private provides and how to use them.

provide.go
JacobOaks: Thx for the ideas. I opted for the top-level var for the reasons you mentioned and to avoid awkward naming conflicts (there already exists a struct called provideOption)

sywhang (commented):
posting on behalf of @r-hang from group code review.

fxevent/zap.go
```diff
@@ -199,3 +200,10 @@ func moduleField(name string) zap.Field {
 	}
 	return zap.String("module", name)
 }
+
+func trueBool(name string, b bool) zap.Field {
```
r-hang, line 204: nit: I think the name should hint at skipping behavior e.g. trueOrSkip or maybeBool. Something that might flag a someone skimming the code even more that there is something special.

r-hang (approved):

docs/modules.md
```diff
@@ -56,6 +56,24 @@ To write an Fx module:
    )
    ```
 
+5. Lastly, if you want to keep a constructor hidden to your
```
sywhang, line 59:
```suggestion
5. Lastly, if you want to keep a constructor's output types contained to your
```

docs/modules.md
```diff
@@ -56,6 +56,24 @@ To write an Fx module:
...
+   	),
+   	fx.Provide(
+   		fx.Private,
+   		parseConfig,
+   	),
+   	fx.Invoke(startServer),
+   	fx.Decorate(wrapLogger),
+   )
+   ```
+
```
sywhang, line 76: Add some more explanation here about the output of `parseConfig` being only available within the `"server"` module.

fxevent/console_test.go
```diff
@@ -160,18 +160,39 @@ func TestConsoleLogger(t *testing.T) {
...
 			want: "[Fx] PROVIDE	*bytes.Buffer <= bytes.NewBuffer()\n",
 		},
+		{
+			name: "Provided privately",
+			give: &Provided{
+				ConstructorName: "bytes.NewBuffer()",
+				OutputTypeNames: []string{"*bytes.Buffer"},
+				Private:         true,
+			},
+			want: "[Fx] PRIVATE PROVIDE	*bytes.Buffer <= bytes.NewBuffer()\n",
```
sywhang, line 174: in terms of readability, something like PROVIDE(PRIVATE) might be easier to read since PROVIDEs are aligned among private/non-private.

fxevent/event.go
```diff
@@ -132,6 +132,10 @@ type Provided struct {
 
 	// Err is non-nil if we failed to provide this constructor.
 	Err error
+
+	// Private denotes whether the provided constructor was flagged provided
+	// alongside a call to [Private].
```
sywhang, line 137:
```suggestion
	// Private denotes whether the provided constructor is a [Private] constructor.
```

provide.go
```diff
@@ -72,14 +74,43 @@ type provideOption struct {
 }
 
 func (o provideOption) apply(mod *module) {
+	var private bool
+
+	var targets []interface{}
```
sywhang, line 79: you can set the cap of the slice to `len(o.Targets)` to avoid allocations.

sywhang (approved):
👍

PR #1002, opened 2022-12-21
Panic Recovery Option

JacobOaks:
Since https://github.com/uber-go/dig/pull/364 was merged into Dig, we can now built upon it to add the ability to recover from panics in user-provided code into Fx, by providing a `fx.RecoverFromPanics()` option into their application. This is disallowed on the module level, only on the application level.

```go
app := fx.New(
    fx.Provide(func() int {
        panic("terrible sorrow")
    }),
    fx.Invoke(func(a int) {}),
    fx.RecoverFromPanics(), // panic will be set to app.Err now
)

err := app.Err()
var pe dig.PanicError
if errors.As(err, &pe) { // panic happened when we invoke
    // handle panic
}
```

### DIFF
--- a/app.go
+++ b/app.go
@@ -165,8 +165,8 @@ func (t stopTimeoutOption) String() string {
 	return fmt.Sprintf("fx.StopTimeout(%v)", time.Duration(t))
 }
 
-// RecoverFromPanics causes panics that occur in provided, decorated, and
-// invoked functions to be recovered from and wrapped in an dig.PanicError.
+// RecoverFromPanics causes panics that occur in functions given to [Provide],
+// [Decorate], and [Invoke] to be recovered from.
 // This error can be retrieved as any other error, by using (*App).Err().
 func RecoverFromPanics() Option {
 	return recoverFromPanicsOption{}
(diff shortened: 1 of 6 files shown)


app_test.go
```diff
@@ -919,6 +919,43 @@ func TestTimeoutOptions(t *testing.T) {
 	assert.True(t, stopped, "app wasn't stopped")
 }
 
+func TestRecoverFromPanicsOption(t *testing.T) {
+	t.Parallel()
+
+	t.Run("CanNotGiveToModule", func(t *testing.T) {
```
sywhang, line 925:
```suggestion
	t.Run("CannotGiveToModule", func(t *testing.T) {
```

app.go
```diff
@@ -165,6 +165,28 @@ func (t stopTimeoutOption) String() string {
 	return fmt.Sprintf("fx.StopTimeout(%v)", time.Duration(t))
 }
 
+// RecoverFromPanics causes panics that occur in provided, decorated, and
+// invoked functions to be recovered from and wrapped in an dig.PanicError.
```
sywhang, line 169:
```suggestion
// RecoverFromPanics causes panics that occur in functions given to [Provide],
// [Decorate], and [Invoke] to be recovered from.
```

sywhang (approved):

Issue #1008, opened 2023-01-01
Module docs mentions features not released yet


pfernandom:
**Describe the bug**
[The official documentation on modules](https://uber-go.github.io/fx/modules.html#writing-modules) mentions the use of `fx.Private`. However, this struct is not exported in the latest released version `1.18.2`. In order to use it, one would have to use the `master` branch in the Github repo.

**To Reproduce**
Use the latest published version `1.18.2` and try to use `fx.Private` in a module, as described in the docs.

The compiler complains with:

```
Private not declared by package fx  compiler[UndeclaredImportedName]
(https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#UndeclaredImportedName)
```

**Expected behavior**
`fx.Private` should be present.

**Additional context**
While it is understandable that some features are not released yet, either:

- The official documents shouldn't mention unreleased features
- The latest published version should include all features mentioned by the published docs

If this feature is not fully ready for production use, maybe tag an alpha version which does include it and mention it in the docs.
Thank you for all the good work!

abhinav:
@sywhang To fix this, we can change the [docs workflow](https://github.com/uber-go/fx/blob/master/.github/workflows/docs.yml) to run only on pushes of tagged releases.

abhinav:
I've opened #1009 in case you decide to go with option 1 above.

sywhang:
@pfernandom sorry you ran into that; Re: Private, we'll be doing a release shortly that has this feature so please keep an eye out for that!

pfernandom:
That was quick! Thank you for getting to this so fast.

Overall, I've been having a great experience using this tool, so I will definitely look out for this release.

Thank you!

PR #1009, opened 2023-01-02
ci/docs: Run on tagged releases only

abhinav:
This changes the ci/docs workflow
to only run on tagged releases by default
instead of every push to main.

This also updates the existing workflow_dispatch trigger to expect an input specifying the ref to publish docs from. With that, you should be able to specify a branch or tag from the Actions UI.
Example of how this looks from [one of my projects](https://github.com/abhinav/restack/blob/89e6749811ee0cccc4cc8884757ca8d291d73272/.github/workflows/release.yml):
image

Going this route has the disadvantage that
all documentation fixes and updates after a release will not be available on the website until another release or a manual publish which can accidentally also publish documentation for unreleased features.

Refs #1008

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -2,11 +2,20 @@ name: GitHub Pages
 
 on:
   push:
-    branches: [master]
+    tags: ['v*']
 
   # This lets us publish the workflow
   # manually from the GitHub Actions UI.
+  #
+  # It expects a single input:
+  # the Git ref we want to build and publish the docs for.
   workflow_dispatch:
+    inputs:
+      head:
+        description: "Git commitish to check out."
+        required: true
+        type: string
+
 
 # Run at most one publish job at a time,
 # cancelling others if a new one starts.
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -19,8 +28,18 @@ jobs:
...
+      # If manually trigerred, we use the provided ref,
+      # otherwise we use the default ref.
+      - name: Checkout (on release)
+        if:  github.event_name != 'workflow_dispatch'
+        uses: actions/checkout@v3
+      - name: Checkout (manual)
+        if:  github.event_name == 'workflow_dispatch'
         uses: actions/checkout@v3
+        with:
+          ref: ${{ inputs.head }}
```
abhinav, line 42: I believe you can collapse these into one and do:

```
ref: ${{ inputs.head || github.ref }}
```

But I'm not certain about it, so I opted for the duplicated version for now.

sywhang (approved):
Thanks @abhinav !

PR #1017, opened 2023-01-09
Fix Start ctx expiration making app unstoppable

### LINKED ISSUE #1015: 1.19 Regression: It's impossible to shutdown from an OS signal after `startCtx` is `Done()`.
**Describe the bug**
Regression: It's impossible to shutdown from an OS signal after `startCtx` is `Done()`.

**To Reproduce**
1) Start an fx app
2) Wait `startTimeout` (default is 15s).
3) Send SIGTERM to the process
4) Nothing will handle the signal and the app will not shut down

**Expected behavior**
The apps shut down correctly like it did in the previous release.

**Additional context**
This line starts the goroutine that dispatches os signals. 
https://github.com/uber-go/fx/blob/213eb869667072f220b8224d16852a81897f6011/signal.go#L118
The goroutine then blocks on 
https://github.com/uber-go/fx/blob/213eb869667072f220b8224d16852a81897f6011/signal.go#L82-L95

After the `startCtx` is `Done()`, this goroutine exits and thus there's nothing to process OS signals to initiate shutdown. 

Replacing the goroutine invocation by `go recv.relayer(context.TODO())` fixed the issue for me. The context passed here should not be `Done()` until the application stops.

sywhang:
This fixes #1015.

When Start() is called, it passes StartTimeout to the relayer goroutine, which is monitoring the Stop signal, as well as the OS signal handler.

It then exits the goroutine if the start context expires, which would naturally happen if the user uses the default setting (15s timeout) or any finite amount of timeout value.

This causes the app to then be not responsive to any OS signals such as SIGTERM or SIGINT.

This was a regression introduced in 1.19 release via #989 .

To fix this, this commit simply removes the relayer goroutine from selecting on the start context being completed.

Verified that all tests are still passing without any goroutine leak, except one test that was triggering a panic to test the panic handler.

To prevent that one test from opting out every single test into goleak.VerifyNone(t) in every sub test, I pulled out that panic test into a separate test package so that we can continue to use goleak.VerifyNone() method in app_test.

### DIFF
--- a/annotated_test.go
+++ b/annotated_test.go
@@ -1500,6 +1500,8 @@ func assertApp(
 		require.NoError(t, app.Stop(ctx))
 		assert.True(t, *stopped)
 	}
+
+	defer app.Stop(ctx)
 }
 
 func TestHookAnnotations(t *testing.T) {
--- a/app_test.go
+++ b/app_test.go
@@ -918,43 +918,6 @@ func TestTimeoutOptions(t *testing.T) {
 	assert.True(t, stopped, "app wasn't stopped")
 }
 
-func TestRecoverFromPanicsOption(t *testing.T) {
-	t.Parallel()
-
-	t.Run("CannotGiveToModule", func(t *testing.T) {
-		t.Parallel()
-		mod := Module("MyModule", RecoverFromPanics())
-		err := New(mod).Err()
-		require.Error(t, err)
-		require.Contains(t, err.Error(),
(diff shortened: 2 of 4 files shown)


sywhang:
Note to reviewers: will kick off internal Go monorepo build pinning it to this branch

tchung1118 (approved):

JacobOaks (approved):

Issue #1020, opened 2023-01-10
Can `fx.Private` be scoped as a top-level `fx.Option` instead of a `var`


rossb83:
**Is your feature request related to a problem? Please describe.**
I'd like to be able to continue the scoping story started with `fx.Private` but make it a top-level `fx.Option` that can be applied outside of `fx.Provide`. The reason for this is the usual pattern is for `fx.Provide` to already be set in legacy codebases and it would be preferable not to go back and change old code by adding an `fx.private` into it, rather than creating a module around this to "privatize" that global module.

**Describe the solution you'd like**
Suppose we have legacy code that looks like this that we wish not to touch
```golang
package etcdmanagerfx

var Module = fx.Provide(
    NewEtcdManager,
)

type EtcdManager struct {
    etcdAddr string
}

func NewEtcdManager(/* ... */) *EtcdManager {
    // ...
}
```

in my main method I'd like to be able to do something like this to have two etcd's (perhaps a global and a regional)
```golang
package main

func main() {
    fx.New(
        fx.Module("global etcd submodule",
            // ... other dependencies to make etcd global
            etcdmanagerfx.Module,
            // ... other dependencies on top of global etcd
            fx.Private, <-- make all dependencies private to other modules outside this scope
        ),
        fx.Module("regional etcd submodule",
            // ... other dependencies to make etcd regional
            etcdmanagerfx.Module,
            // ... other dependencies on top of regional etcd
            fx.Private, <-- make all dependencies private to other modules outside this scope
        ),
    )
}
```

**Describe alternatives you've considered**
None

**Is this a breaking change?**
No

**Additional context**
None

jkanywhere:
How about `fx.Private` is like `fx.Options` but, like all the content can only be used inside the nearest parent `fx.Module`?
```
package examplefx
var Module = fx.Module("examplefx",
    fx.Private( 
      fx.Provide(...),      // only usable inside examplefx.Module
      fx.Supply(...),       // only usable inside examplefx.Module
      etcdmanagerfx.Module, // only usable inside examplefx.Module
    ), // end Private
    fx.Provide(...), // exposed to consumers of examplefx.Module
    fx.Supply(...),  // exposed to consumers of examplefx.Module

    fx.Module("innermodule",
      fx.Provide(...), // exposed to consumers of innermodule and consumers of examplefx.Module
      fx.Private(...), // only usable inside innermodule. Other parts of examplefx.Module can't use it.
    ), // end inner module
)
```

[...]

PR #1030, opened 2023-01-23
Private Modules

JacobOaks:
Some users have been requesting that our private provides functionality provided by #995 be extended by allowing a way to privatize an entire module (Ref: #1020).

This change modifies fx.Private to implement fx.Option, allowing it to be passed to fx.App and fx.Module.
When passed into a module, the entire module will be marked as private, meaning all provides and supplies of that module and its inner modules will be marked as private.

```go
fx.New(
	fx.Module("SubModule",
		fx.Private,
		fx.Supply(5),
		fx.Provide(func() string { return "b"}),
		fx.Invoke(a int, b str) {}, // Runs w/ a = 5, b = "b"
	),
	fx.Invoke(func(a int) {}), // This will fail
	fx.Invoke(func(b str) {}), // This will also fail
)
```

I opted for this API to stay consistent with the style of API we already allow for passing Private to Provide directly.

This change also extracts all private-related definitions and testing out into separate files, `private.go` and `private_test.go`.


JacobOaks:
Closing for now, missed a couple key things

JacobOaks:
> @JacobOaks out of curiosity: do you remember what was missing here?

I _think_ we were unsure about the utility an entirely private module would provide. After all, an entirely private module, by definition, cannot contribute anything to an application's container. However, invokes within the private module would still run and cause the private module's dependencies to build - so that's an exception. 

I think something closer to @jkanywhere's suggestion on #1020 may be better: it allows easily "privatizing" larger collections of options, while still allowing otherwise mostly private modules to expose key type(s) to other modules/the container to be exposed.

abhinav:
> I think we were unsure about the utility an entirely private module would provide.

Ack, yeah. So the use case I was imagining was something like this.
I have an Fx module that supports _one_ of something, e.g. spinning up an HTTP server.

```go
fx.New(httpserver.Module).Run()
```

For reasons, I want to have two separate HTTP servers.
So I would be able to do something like this:

```
fx.New(
  fx.Module("main", fx.Private(httpserver.Module), /* handlers for main */),
  fx.Module("internal", fx.Private(httpserver.Module), /* handlers for internal */),
)
```

I hadn't fully thought through how private modules would "export" values outside the private module.
I'd originally imagined that Private just inverts the Fx default of [exporting from subscopes](https://pkg.go.dev/go.uber.org/dig#Export), but in hindsight, I don't like that much.

> I think something closer to @jkanywhere's suggestion on https://github.com/uber-go/fx/issues/1020 may be better

I like that! It's intuitively similar to internal/ directories in Go: only siblings and direct parent can use outputs of private modules.
There would be no way for a parent to export the private result to the rest of the graph, but IMO that's not an issue—if that's needed, then thing isn't private.

PR #1042, opened 2023-02-18
ci: Test with Go 1.19, 1.20

abhinav:
Test with Go 1.19 and 1.20,
and use setup-go v3 with built-in caching.


abhinav:
Supersedes #1040 which didn't update staticcheck.

sywhang (approved):

Issue #835, opened 2022-02-10
Verify invalid tags passed to fx.Annotate


sywhang:
Some tags passed with fx.Annotate looks a struct tag like this:
```fx.Annotate(f, fx.ParamTags(`group:"myGroup"`))```

Unfortunately I got very little sleep last night so I fatfingered and made a typo like this while testing something:
```fx.Annotate(f, fx.ParamTags(`group:'myGroup"`))```

It took a while to find out what was wrong because Fx (technically, Dig) silently ignored this tag. 

We could add a verification in Fx to see if the tags are "valid" before passing them to Dig and bail out if the tags are not valid tags that Dig understands.

rossb83:
One other thing I noticed about tags, is that if you provide more tags than there are inputs to `f` it will ignore the extra tags, perhaps this should return error?

PR #1051, opened 2023-02-28
[fx-annotate] Verify invalid tags passed to fx.Annotate #835

lverma14:
Tags passed with fx.Annotate follow the struct format
```
fx.Annotate(f, fx.ParamTags(`group:”myGroup”`))
```
Where the tag is of the format `<tag_name>:”tag_value”` and tags are separated by space.

If a user accidentally passes an invalid tag such as mismatched quotes(#835), the issue is not identified and no error is thrown. This is not desired as the tag gets passed to dig, where it’s ignored as it’s not a valid group tag.

This commit adds checks to identify if invalid tags are passed to fx.Annotate. If identified then it throws an error with the right tag format specified. 
The following checks are now enforced -
 - tag key can only be group, optional or name ( dig only accepts these tags as valid keys currently)
 - In cases of multiple tags  they must be separated by space . Eg ``group:"some" optional:"true"``vim 
 - the tag values must be contained within double quotes.

For e.g. -
```
fx.Annotate(f, fx.ParamTags(`group:'myGroup"`))
```
Should be an invalid tag as it starts from a single quote and ends with a double quote (mismatched quotes instead of “...”) We added checks for ParamTag and ResultTag to verify and capture this invalid tag.
Now, when we pass this invalid tag we get the following error :
```
 Encountered error while applying annotation using fx.Annotate to function, 
tag value should start with double quote i.e. key:"value"
```
Components Modified
annotated.go  - add checks to verify invalid tag usage
annotated_test.go - Add test cases to verify functionality of changes
module_test.go - FIx a typo in testcase
Refs : #835

### DIFF
--- a/annotated.go
+++ b/annotated.go
@@ -144,6 +144,75 @@ type paramTagsAnnotation struct {
 }
 
 var _ Annotation = paramTagsAnnotation{}
+var (
+	errTagSyntaxSpace            = errors.New(`multiple tags are not separated by space`)
+	errTagKeySyntax              = errors.New("tag key is invalid, Use group, name or optional as tag keys")
+	errTagValueSyntaxQuote       = errors.New(`tag value should start with double quote. i.e. key:"value" `)
+	errTagValueSyntaxEndingQuote = errors.New(`tag value should end in double quote. i.e. key:"value" `)
+)
+
+// Collections of key value pairs within a tag should be separated by a space.
+// Eg: `group:"some" optional:"true"`.
+func verifyTagsSpaceSeparated(tagIdx int, tag string) error {
+	if tagIdx > 0 && tag != "" && tag[0] != ' ' {
+		return errTagSyntaxSpace
+	}
+	return nil
+}
+
+// verify tag values are delimited with double quotes.
+func verifyValueQuote(value string) (string, error) {
+	// starting quote should be a double quote
+	if value[0] != '"' {
+		return "", errTagValueSyntaxQuote
+	}
+	// validate tag value is within quotes
+	i := 1
+	for i < len(value) && value[i] != '"' {
+		if value[i] == '\\' {
+			i++
+		}
+		i++
+	}
+	if i >= len(value) {
+		return "", errTagValueSyntaxEndingQuote
+	}
+	return value[i+1:], nil
+
+}
+
+// Check whether the tag follows valid struct.
+// format and returns an error if it's invalid. (i.e. not following
+// tag:"value" space-separated list )
+// Currently dig accepts only 'name', 'group', 'optional' as valid tag keys.
+func verifyAnnotateTag(tag string) error {
+	tagIdx := 0
+	validKeys := map[string]struct{}{"group": {}, "optional": {}, "name": {}}
+	for ; tag != ""; tagIdx++ {
+		if err := verifyTagsSpaceSeparated(tagIdx, tag); err != nil {
+			return err
+		}
+		i := 0
+		if strings.TrimSpace(tag) == "" {
+			return nil
+		}
+		// parsing the key i.e. till reaching colon :
+		for i < len(tag) && tag[i] != ':' {
+			i++
+		}
+		key := strings.TrimSpace(tag[:i])
+		if _, ok := validKeys[key]; !ok {
+			return errTagKeySyntax
+		}
+		value, err := verifyValueQuote(tag[i+1:])
+		if err != nil {
+			return err
+		}
+		tag = value
+	}
+	return nil
+
+}
 
 // Given func(T1, T2, T3, ..., TN), this generates a type roughly
 // equivalent to,
@@ -159,10 +228,18 @@ var _ Annotation = paramTagsAnnotation{}
 //
 // If there has already been a ParamTag that was applied, this
 // will return an error.
+//
+// If the tag is invalid and has mismatched quotation for example,
+// (`tag_name:"tag_value') , this will return an error.
 
 func (pt paramTagsAnnotation) apply(ann *annotated) error {
 	if len(ann.ParamTags) > 0 {
 		return errors.New("cannot apply more than one line of ParamTags")
+	}
+	for _, tag := range pt.tags {
+		if err := verifyAnnotateTag(tag); err != nil {
+			return err
+		}
 	}
 	ann.ParamTags = pt.tags
 	return nil
@@ -285,9 +362,17 @@ var _ Annotation = resultTagsAnnotation{}
 //
 // If there has already been a ResultTag that was applied, this
 // will return an error.
+//
+// If the tag is invalid and has mismatched quotation for example,
+// (`tag_name:"tag_value') , this will return an error.
 func (rt resultTagsAnnotation) apply(ann *annotated) error {
 	if len(ann.ResultTags) > 0 {
 		return errors.New("cannot apply more than one line of ResultTags")
+	}
+	for _, tag := range rt.tags {
+		if err := verifyAnnotateTag(tag); err != nil {
+			return err
+		}
 	}
 	ann.ResultTags = rt.tags
 	return nil
--- a/annotated_test.go
+++ b/annotated_test.go
@@ -1474,6 +1474,131 @@ func TestAnnotate(t *testing.T) {
 	})
 }
 
+func TestAnnotateApplyFail(t *testing.T) {
+	type a struct{}
+	type b struct{ a *a }
+	newA := func() *a { return &a{} }
+	newB := func(a *a) *b {
+		return &b{a}
+	}
+
+	var (
+		errTagSyntaxSpace            = `multiple tags are not separated by space`
+		errTagKeySyntax              = "tag key is invalid, Use group, name or optional as tag keys"
+		errTagValueSyntaxQuote       = `tag value should start with double quote. i.e. key:"value" `
+		errTagValueSyntaxEndingQuote = `tag value should end in double quote. i.e. key:"value" `
+	)
+	tests := []struct {
+		give                 string
+		wantErr              string
+		giveAnnotationParam  fx.Annotation
+		giveAnnotationResult fx.Annotation
+	}{
+		{
+			give:                 "Tags value invalid ending quote",
+			wantErr:              errTagValueSyntaxEndingQuote,
+			giveAnnotationParam:  fx.ParamTags(`name:"something'`),
+			giveAnnotationResult: fx.ResultTags(`name:"something'`),
+		},
+		{
+			give:                 "Tags value wrong starting quote",
+			wantErr:              errTagValueSyntaxQuote,
+			giveAnnotationParam:  fx.ParamTags(`name:"something" optional:'true"`),
+			giveAnnotationResult: fx.ResultTags(`name:"something" optional:'true"`),
+		},
+		{
+			give:                 "Tags multiple tags not separated by space",
+			wantErr:              errTagSyntaxSpace,
+			giveAnnotationParam:  fx.ParamTags(`name:"something"group:"something"`),
+			giveAnnotationResult: fx.ResultTags(`name:"something"group:"something"`),
+		},
+		{
+			give:                 "Tags key not equal to group, name or optional",
+			wantErr:              errTagKeySyntax,
+			giveAnnotationParam:  fx.ParamTags(`name1:"something"`),
+			giveAnnotationResult: fx.ResultTags(`name1:"something"`),
+		},
+		{
+			give:                 "Tags key empty",
+			wantErr:              errTagKeySyntax,
+			giveAnnotationParam:  fx.ParamTags(`:"something"`),
+			giveAnnotationResult: fx.ResultTags(`:"something"`),
+		},
+	}
+	for _, tt := range tests {
+		t.Run("Param "+tt.give, func(t *testing.T) {
+			app := NewForTest(t,
+				fx.Provide(
+					fx.Annotate(
+						newA,
+						tt.giveAnnotationParam,
+					),
+				),
+				fx.Invoke(newB),
+			)
+			assert.ErrorContains(t, app.Err(), tt.wantErr)
+		})
+		t.Run("Result "+tt.give, func(t *testing.T) {
+			app := NewForTest(t,
+				fx.Provide(
+					fx.Annotate(
+						newA,
+						tt.giveAnnotationResult,
+					),
+				),
+				fx.Invoke(newB),
+			)
+			assert.ErrorContains(t, app.Err(), tt.wantErr)
+		})
+	}
+}
+
+func TestAnnotateApplySuccess(t *testing.T) {
+	type a struct{}
+	type b struct{ a *a }
+	newA := func() *a { return &a{} }
+	newB := func(a *a) *b {
+		return &b{a}
+	}
+
+	tests := []struct {
+		give                 string
+		giveAnnotationParam  fx.Annotation
+		giveAnnotationResult fx.Annotation
+	}{
+		{
+			give:                 "ParamTags Tag Empty",
+			giveAnnotationParam:  fx.ParamTags(`  `),
+			giveAnnotationResult: fx.ResultTags(`  `),
+		},
+		{
+			give:                 "ParamTags Tag Empty with extra spaces",
+			giveAnnotationParam:  fx.ParamTags(`name:"versionNum"`, `  `),
+			giveAnnotationResult: fx.ResultTags(`   `, `group:"versionNum"`),
+		},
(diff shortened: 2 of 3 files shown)


annotated.go
```diff
@@ -159,11 +159,20 @@ var _ Annotation = paramTagsAnnotation{}
...
 func (pt paramTagsAnnotation) apply(ann *annotated) error {
 	if len(ann.ParamTags) > 0 {
 		return errors.New("cannot apply more than one line of ParamTags")
 	}
+	for _, tag := range pt.tags {
+		if strings.Contains(tag, "'") {
+			correctedTag := strings.ReplaceAll(tag, "'", "\"")
+			return errors.New("the ParamTags tag has invalid quotations, did you mean " + correctedTag + " ?")
+		}
+	}
```
abhinav, line 175: Hi, just doing a quick drive-by.
This seems a bit too specific. What if my tag value has `'` in it?

```
type foo struct {
  Value string `name:"a'bc"`
```

Fx does not have any rules rejecting that, AFAIK.

Y'all might be able to borrow struct tag validation from the structtag lint check in `go vet`. 
https://cs.opensource.google/go/go/+/master:src/cmd/vendor/golang.org/x/tools/go/analysis/passes/structtag/structtag.go;drc=31881f87873b84709a49ca17195bbe5b3f683acf;l=212

(We don't need the xml/json specialization there.)

lverma14: Thanks @abhinav, that's really helpful! I'll look into that

annotated.go
```diff
@@ -285,10 +294,19 @@ var _ Annotation = resultTagsAnnotation{}
...
+// If the tag is invalid and has mismatched quotation for example,
+// (`tag_name:"tag_value') , this will return an error
 func (rt resultTagsAnnotation) apply(ann *annotated) error {
 	if len(ann.ResultTags) > 0 {
 		return errors.New("cannot apply more than one line of ResultTags")
 	}
+	for _, tag := range rt.tags {
+		if strings.Contains(tag, "'") {
+			correctedTag := strings.ReplaceAll(tag, "'", "\"")
+			return errors.New("the ResultTags tag has invalid quotations, did you mean " + correctedTag + " ?")
```
JacobOaks, line 307: Do we want to surround `correctedTag` with quotes or backticks to help demarcate it in the error message?

annotated_test.go
```diff
@@ -1347,6 +1347,78 @@ func TestAnnotate(t *testing.T) {
...
+				fx.Annotate(
+					newA,
+					fx.ParamTags(`name:"something'`),
+				),
+			),
+			fx.Invoke(newB),
+		)
+		err := app.Err()
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), `the ParamTags tag has invalid quotations, did you mean name:"something" ?`)
```
JacobOaks, line 1365: Thoughts on using [ErrorContains](https://pkg.go.dev/github.com/stretchr/testify/assert#ErrorContains) instead? It looks like it can combine all 3 of these lines in one.

annotated_test.go
```diff
@@ -1347,6 +1347,78 @@ func TestAnnotate(t *testing.T) {
 		assert.Contains(t, err.Error(), "encountered error while applying annotation using fx.Annotate to go.uber.org/fx_test.TestAnnotate.func1(): cannot apply more than one line of ResultTags")
 	})
 
+	t.Run("invalid ParamTag quotations", func(t *testing.T) {
```
JacobOaks, line 1350: +1 to @abhinav 's comment about additional error cases that this wouldn't cover.

I also think we can convert these tests to table tests since much of the logic is the same between tests (https://github.com/uber-go/guide/blob/master/style.md#test-tables)

i.e. the test case struct could look something like

```
type test struct {
    name string
    giveTags string
    giveAnnotation fx.Annotation // set this to either fx.ParamTags or fx.ResultTags
    wantErr string
}
```

JacobOaks (commented):
I saw this is still a draft but some initial thoughts :)

annotated.go
lverma14: Oh yeah, good catch thanks!

annotated_test.go
```diff
@@ -1474,6 +1474,159 @@ func TestAnnotate(t *testing.T) {
...
+		errTagValueSpace             = `tag key value separated by extra space. Use format` + " `tag_key:\"value\"`"
+	)
+	tests := []struct {
+		give           string
+		wantErr        string
+		giveAnnotation fx.Annotation
+	}{
+		{
+			give:           "ParamTags wrong ending quote",
+			wantErr:        errTagValueSyntaxEndingQuote + "`name:\"something\"` ?",
```
JacobOaks, line 1502: Considering specifying in the test message whether it is the key or the value that has the incorrect quote

annotated_test.go
```diff
@@ -1474,6 +1474,159 @@ func TestAnnotate(t *testing.T) {
...
+			give:           "ParamTags key not equal to group, name or optional",
+			wantErr:        errTagKeySyntax,
+			giveAnnotation: fx.ParamTags(`name1:"something"`),
+		},
+		{
+			give:           `ParamTags invalid space after key colon`,
+			wantErr:        errTagValueSpace,
+			giveAnnotation: fx.ParamTags(`name: "something"`),
+		},
+		{
```
JacobOaks, line 1535: The set of tests for ResultTags looks similar to for ParamTags, can we de-duplicate these tests?

annotated_test.go
```diff
@@ -1474,6 +1474,159 @@ func TestAnnotate(t *testing.T) {
...
+				fx.Provide(
+					fx.Annotate(
+						newA,
+						tt.giveAnnotation,
+					),
+				),
+				fx.Invoke(newB),
+			)
+			err := app.Err()
+			assert.ErrorContains(t, err, tt.wantErr)
```
JacobOaks, line 1588: We can combine these
`assert.ErrorContains(t, app.Err(), tt.wantErr)`

annotated_test.go
```diff
@@ -1474,6 +1474,159 @@ func TestAnnotate(t *testing.T) {
...
+		})
+	}
+
+	tests1 := []struct {
```
JacobOaks, line 1592: Let's put two separate sub-tests, one for your valid cases and one for your invalid cases

annotated_test.go
```diff
@@ -1474,6 +1474,159 @@ func TestAnnotate(t *testing.T) {
...
+			giveAnnotation: fx.ParamTags(`  `),
+		},
+		{
+			give:           "ParamTags Tag Empty with extra spaces",
+			giveAnnotation: fx.ParamTags(`name:"versionNum"`, `  `),
+		},
+		{
+			give:           "ParamTags Tag Empty with extra spaces",
+			giveAnnotation: fx.ParamTags(`name:"versionNum"`, `  `),
+		},
```
JacobOaks, line 1607: repeated test cases

annotated_test.go
```diff
@@ -1474,6 +1474,159 @@ func TestAnnotate(t *testing.T) {
...
+				fx.Provide(
+					fx.Annotate(
+						newA,
+						tt.giveAnnotation,
+					),
+				),
+				fx.Invoke(newB),
+			)
+			err := app.Err()
+			require.NoError(t, err)
```
JacobOaks, line 1625: Same comment as above, can combine these

JacobOaks (commented):
Submitting on behalf of @r-hang for group code review.

Let's try to refactor this:
* Let's try to separate separate failure cases into different functions to make the code easier to understand
* Let's try without the "did you mean?" suggestions for now, for this case it's probably not worth the added complexity

annotated.go
```diff
@@ -144,6 +144,82 @@ type paramTagsAnnotation struct {
...
+// Currently dig accepts only 'name', 'group', 'optional' as valid tag keys
+func verifyAnnotateTag(tag string) error {
+	numTags := 0
+	for ; tag != ""; numTags++ {
+		validKeys := map[string]bool{"group": true, "optional": true, "name": true}
+		if err := verifyTagsSpaceSeparated(numTags, tag); err != nil {
+			return err
+		}
+		i := 0
+		// return if tag is empty
```
r-hang, line 200: I think this comment can be removed.

annotated.go
```diff
@@ -144,6 +144,82 @@ type paramTagsAnnotation struct {
...
+		// parsing the key i.e. till reaching colon :
+		for i < len(tag) && tag[i] != ':' {
+			i++
+		}
+		key := strings.TrimSpace(tag[:i])
+		// if key is not equal to group name or optional
+		if _, ok := validKeys[key]; !ok {
+			return errTagKeySyntax
+		}
+		value, err := verifyValueQuote(tag[i+1:])
```
r-hang, line 214: This code has an interesting struct where the value passed through a set of filters and runs through the same set with the remaining processed values. Have you experimented to see if the code is simplified if implemented recursively?

annotated.go
```diff
@@ -144,6 +144,82 @@ type paramTagsAnnotation struct {
...
+		if strings.TrimSpace(tag) == "" {
+			return nil
+		}
+
+		// parsing the key i.e. till reaching colon :
+		for i < len(tag) && tag[i] != ':' {
+			i++
+		}
+		key := strings.TrimSpace(tag[:i])
+		// if key is not equal to group name or optional
```
r-hang, line 210: I think the comment can be removed given the good naming of the map lookup

annotated.go
```diff
@@ -144,6 +144,82 @@ type paramTagsAnnotation struct {
...
+		return errTagSyntaxSpace
+	}
+	return nil
+}
+
+// Given a value from a tag verify that it is contained between double quotes
+// Return an error if the invalid quote format is used for tag value.
+// On sucess return the remaining tag with nil error
+func verifyValueQuote(value string) (string, error) {
+	// starting quote should be a double quote
```
r-hang, line 169: Could we just check if the value string is prefixed and suffixed with double quotes. Could the step to validate tag values is within quotes be replaced by counting the total number of double quotes and ensuring that is exactly 2?

annotated.go
```diff
@@ -144,6 +144,82 @@ type paramTagsAnnotation struct {
...
+}
+
+// Given a string tag the function checks if the tag follows valid struct
+// format and returns an error if it's invalid. (i.e. not following
+// tag:"value" space-separated list )
+// Currently dig accepts only 'name', 'group', 'optional' as valid tag keys
+func verifyAnnotateTag(tag string) error {
+	numTags := 0
+	for ; tag != ""; numTags++ {
+		validKeys := map[string]bool{"group": true, "optional": true, "name": true}
```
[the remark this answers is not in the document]
lverma14: Nice! That'd be really handy

annotated.go
lverma14: The total number of double quotes can be more than 2, as it's valid for a struct string to have quotes in it. 
`key:"val/"ue"` would be a valid tag with more than 2 double quotes. Hence, I don't think it'd be possible unless I misunderstood?

r-hang: Ah understood that this approach does not work because value actually contains the rest of the tag strings.

annotated_test.go
```diff
@@ -1474,6 +1474,134 @@ func TestAnnotate(t *testing.T) {
...
+
+	type a struct{}
+	type b struct{ a *a }
+	newA := func() *a { return &a{} }
+	newB := func(a *a) *b {
+		return &b{a}
+	}
+
+	var (
+		errTagSyntaxSpace            = ` multiple tags are not separated by space`
```
r-hang, line 1488: Could we re-use the error declarations in annotated .go, but call it's errors.Error() value in the test cases?

annotated.go
```diff
@@ -144,6 +144,75 @@ type paramTagsAnnotation struct {
...
+}
+
+// Given a string tag the function checks if the tag follows valid struct
+// format and returns an error if it's invalid. (i.e. not following
+// tag:"value" space-separated list )
+// Currently dig accepts only 'name', 'group', 'optional' as valid tag keys
+func verifyAnnotateTag(tag string) error {
+	tagIdx := 0
+	for ; tag != ""; tagIdx++ {
+		validKeys := map[string]struct{}{"group": {}, "optional": {}, "name": {}}
```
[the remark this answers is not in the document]
lverma14: Thanks! Great catch!

annotated_test.go
r-hang: annoted_test is under fx_test, no need to export the error.

annotated_test.go
```diff
@@ -1474,6 +1474,134 @@ func TestAnnotate(t *testing.T) {
...
+		},
+		{
+			give:                 "ParamTags Tag with \\ ",
+			giveAnnotationParam:  fx.ParamTags(`name:"version\\Num"`, `  `),
+			giveAnnotationResult: fx.ResultTags(``, `group:"version\\Num"`),
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.give, func(t *testing.T) {
+			app := NewForTest(t,
```
r-hang, line 1590: I see you're also using t.Parallel() at the top-level. You'll also need to call t.Parallel() within these subtests if you want them to run in parallel.

annotated.go
```diff
@@ -144,6 +144,75 @@ type paramTagsAnnotation struct {
...
+		return "", errTagValueSyntaxEndingQuote
+	}
+	return value[i+1:], nil
+
+}
+
+// Given a string tag the function checks if the tag follows valid struct
+// format and returns an error if it's invalid. (i.e. not following
+// tag:"value" space-separated list )
+// Currently dig accepts only 'name', 'group', 'optional' as valid tag keys
```
r-hang, line 187:
```suggestion
// Currently dig accepts only 'name', 'group', 'optional' as valid tag keys.
```

annotated.go
```diff
@@ -285,10 +362,18 @@ var _ Annotation = resultTagsAnnotation{}
 //
 // If there has already been a ResultTag that was applied, this
 // will return an error.
+//
+// If the tag is invalid and has mismatched quotation for example,
+// (`tag_name:"tag_value') , this will return an error
```
r-hang, line 367:
```suggestion
// (`tag_name:"tag_value') , this will return an error.
```

annotated.go
```diff
@@ -159,11 +228,19 @@ var _ Annotation = paramTagsAnnotation{}
 //
 // If there has already been a ParamTag that was applied, this
 // will return an error.
+//
+// If the tag is invalid and has mismatched quotation for example,
+// (`tag_name:"tag_value') , this will return an error
```
r-hang, line 233:
```suggestion
// If the tag is invalid and has mismatched quotation for example,
// (`tag_name:"tag_value') , this will return an error.
```

annotated.go
```diff
@@ -144,6 +144,75 @@ type paramTagsAnnotation struct {
...
+// Collections of key value pairs within a tag should be separated by a space.
+// Eg: `group:"some" optional:"true"`
+func verifyTagsSpaceSeparated(tagIdx int, tag string) error {
+	if tagIdx > 0 && tag != "" && tag[0] != ' ' {
+		return errTagSyntaxSpace
+	}
+	return nil
+}
+
+// verify tag values are delimited with double quotes
```
r-hang, line 163:
```suggestion
// verify tag values are delimited with double quotes.
```

annotated.go
```diff
@@ -144,6 +144,75 @@ type paramTagsAnnotation struct {
...
 var _ Annotation = paramTagsAnnotation{}
+var (
+	errTagSyntaxSpace            = errors.New(`multiple tags are not separated by space`)
+	errTagKeySyntax              = errors.New("tag key is invalid, Use group, name or optional as tag keys")
+	errTagValueSyntaxQuote       = errors.New(`tag value should start with double quote. i.e. key:"value" `)
+	errTagValueSyntaxEndingQuote = errors.New(`tag value should end in double quote. i.e. key:"value" `)
+)
+
+// Collections of key value pairs within a tag should be separated by a space.
+// Eg: `group:"some" optional:"true"`
```
r-hang, line 155:
```suggestion
// Eg: `group:"some" optional:"true"`.
```

annotated.go
```diff
@@ -144,6 +144,75 @@ type paramTagsAnnotation struct {
...
+		i++
+	}
+	if i >= len(value) {
+		return "", errTagValueSyntaxEndingQuote
+	}
+	return value[i+1:], nil
+
+}
+
+// Given a string tag the function checks if the tag follows valid struct
```
r-hang, line 184:
```suggestion
// Check whether the the tag follows valid struct
```

r-hang (approved):
lgtm modulo some minor fixups!

sywhang (approved):
👍

PR #1053, opened 2023-03-10
Allow fx.Annotate on In/Out structs with ResultTags/ParamTags, respectively

### LINKED ISSUE #980: Annotation: Param/ResultTags with In/Out
Currently, fx.ParamTags and fx.ResultTags error out when they're used with fx.In and fx.Out. 

It makes sense that fx.ParamTags shouldn't be used with fx.In, and fx.ResultTags with fx.Out, but there's no reason for fx.ParamTags to not be able to be used with fx.Out. 

We should allow the usage of fx.ParamTags annotation with fx.Out, and similarly, fx.ResultTags with fx.In.

sywhang:
It is currently not possible for fx.Annotate to be used on a function that takes in or returns fx.In and fx.Out structs.

This limitation, however, is not necessary when the annotation is different from the struct defined annotations, such as using fx.In with fx.ResultTags Annotation or fx.Out with fx.ParamTags Annotation.

This loosens the constraint so that these combinations of Annotations become valid.

Fixes #980.

### DIFF
--- a/annotated.go
+++ b/annotated.go
@@ -261,6 +261,9 @@ func (pt paramTagsAnnotation) parameters(ann *annotated) (
 // ParamTags is an Annotation that annotates the parameter(s) of a function.
 // When multiple tags are specified, each tag is mapped to the corresponding
 // positional parameter.
+//
+// ParamTags cannot be used in a function that takes an fx.In struct as a
+// parameter.
 func ParamTags(tags ...string) Annotation {
 	return paramTagsAnnotation{tags}
 }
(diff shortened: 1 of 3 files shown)


r-hang (approved):

JacobOaks:
Should we update the "pre-requisites" section here: https://uber-go.github.io/fx/annotate.html#annotating-a-function ?

JacobOaks (approved):

sywhang:
> Should we update the "pre-requisites" section here: https://uber-go.github.io/fx/annotate.html#annotating-a-function ?

Good call. updated.

PR #1063, opened 2023-03-28
Fix race test for Start/Stop hooks

sywhang:
The race test that was added in Start/Stop hooks in #1062. This test can fail if the machine is sufficiently slow that the second Stop hook does not finish running before the test finishes.

This removes the check for the second hook from running - this is still adequately testing the code path we wanted to test since the race would occur on the first stop hook with the second start hook.

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -1236,8 +1236,7 @@ func TestAppStart(t *testing.T) {
 			firstStart  bool
 			secondStart bool
 
-			firstStop  bool
-			secondStop bool
+			firstStop bool
 		)
 		app := New(
 			Invoke(func(lc Lifecycle) {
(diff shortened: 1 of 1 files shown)


JacobOaks (approved):

app_test.go
```diff
@@ -1263,7 +1262,6 @@ func TestAppStart(t *testing.T) {
 						return nil
 					},
 					OnStop: func(context.Context) error {
-						secondStop = true
 						time.Sleep(10 * time.Millisecond)
```
tchung1118, line 1265: I thought this hook was never supposed to run. Does this hook run if the machine is too slow? (and the fix is to just not care about this hook any more?)

tchung1118 (approved):
Approved with a comment

app_test.go
sywhang: so Stop() does not have a timeout here. It can run if the test doesn't finish running before the firstStop hook.

tchung1118: I thought stop hooks run in reverse order. So I guess if the second startHook finishes running some time in between Start() timing out and Stop() being called, the second stopHook will run.

PR #1068, opened 2023-04-17
Disallow using In/Out structs with From/As annotations

### LINKED ISSUE #1060: Disallow using In/Out structs with From/As annotations
In #1053, we eased up the requirement for function signature for applying annotations to allow using In/Out structs with ResultTags/ParamTags. This had an unintended change in behavior where users are now able to use In/Out structs with From/As annotations. For example, the below code used to error out:
```
type Foo interface { ... }
type Bar struct { ... } // Bar implements Foo
struct Res {
  fx.Out

  ABar Bar
}
fx.Annotate(
  func() Res { ... },
  fx.As(new(Foo)),
)
```
Currently the code above applies the As annotation to `ABar` field and transforms the target function's signature accordingly. However, this was an unintended change in behavior, and we should have a design discussion about whether we actually want this before cutting a new release. Since it will require a breaking change to disable this after a new release is cut, I think the safe approach is to disable this for now and enable it after having decided that this is indeed what we want.

ankit209:
In https://github.com/uber-go/fx/pull/1053, we eased up the requirement for function signature for applying annotations to allow using In/Out structs with ResultTags/ParamTags. This had an unintended change in behavior where users are now able to use In/Out structs with From/As annotations. For example, the below code used to error out:

> type Foo interface { ... }
> type Bar struct { ... } // Bar implements Foo
> struct Res {
>   fx.Out
> 
>   ABar Bar
> }
> fx.Annotate(
>   func() Res { ... },
>   fx.As(new(Foo)),
> )

Currently the code above applies the As annotation to ABar field and transforms the target function's signature accordingly. However, this was an unintended change in behavior, and we should have a design discussion about whether we actually want this before cutting a new release. Since it will require a breaking change to disable this after a new release is cut, I think the safe approach is to disable this for now and enable it after having decided that this is indeed what we want.

Fixes #1060

### DIFF
--- a/annotated.go
+++ b/annotated.go
@@ -1138,7 +1138,7 @@ var _ Annotation = (*asAnnotation)(nil)
 //	  return w, r
 //	}
 //
-// As annotation cannot be used in a function that returns an fx.Out struct as a return type.
+// As annotation cannot be used in a function that returns an [Out] struct as a return type.
 func As(interfaces ...interface{}) Annotation {
 	return &asAnnotation{targets: interfaces}
 }
@@ -1325,7 +1325,7 @@ var _ Annotation = (*fromAnnotation)(nil)
 //	  return NewRunnerWraps(r1, r2)
 //	})
 //
-// From annotation cannot be used in a function that takes an fx.In struct as a
+// From annotation cannot be used in a function that takes an [In] struct as a
 // parameter.
 func From(interfaces ...interface{}) Annotation {
 	return &fromAnnotation{targets: interfaces}
@@ -1596,7 +1596,7 @@ func (ann *annotated) cleanUpAsResults() {
 }
 
 // checks and returns a non-nil error if the target function:
-// - returns an fx.Out struct as a result and has either of ResultTags or As annotation
+// - returns an fx.Out struct as a result and has either a ResultTags or an As annotation
 // - takes in an fx.In struct as a parameter and has either of ParamTags or From annotation
 // - has an error result not as the last result.
 func (ann *annotated) typeCheckOrigFn() error {
@@ -1617,7 +1617,7 @@ func (ann *annotated) typeCheckOrigFn() error {
 			continue
 		}
 		if len(ann.ResultTags) > 0 || len(ann.As) > 0 {
-			return errors.New("fx.Out structs cannot be annotated with either of fx.ResultTags or fx.As")
+			return errors.New("fx.Out structs cannot be annotated with fx.ResultTags or fx.As")
 		}
 	}
 	for i := 0; i < ft.NumIn(); i++ {
@@ -1629,7 +1629,7 @@ func (ann *annotated) typeCheckOrigFn() error {
 			continue
 		}
 		if len(ann.ParamTags) > 0 || len(ann.From) > 0 {
-			return errors.New("fx.In structs cannot be annotated with either of fx.ParamTags or fx.From")
+			return errors.New("fx.In structs cannot be annotated with fx.ParamTags or fx.From")
 		}
 	}
 	return nil
--- a/annotated.go
+++ b/annotated.go
@@ -1137,9 +1137,8 @@ var _ Annotation = (*asAnnotation)(nil)
 //	  w, r := a()
 //	  return w, r
 //	}
-// 
-// As annotation cannot be used in a function that takes an fx.In struct as a
-// parameter or returns an fx.Out struct as a return type.
+//
+// As annotation cannot be used in a function that returns an fx.Out struct as a return type.
 func As(interfaces ...interface{}) Annotation {
 	return &asAnnotation{targets: interfaces}
 }
@@ -1327,7 +1326,7 @@ var _ Annotation = (*fromAnnotation)(nil)
 //	})
 //
 // From annotation cannot be used in a function that takes an fx.In struct as a
-// parameter or returns an fx.Out struct as a return type.
+// parameter.
 func From(interfaces ...interface{}) Annotation {
 	return &fromAnnotation{targets: interfaces}
 }
@@ -1597,8 +1596,8 @@ func (ann *annotated) cleanUpAsResults() {
 }
 
 // checks and returns a non-nil error if the target function:
-// - returns an fx.Out struct as a result and has either of ResultTags, As or From annotation
-// - takes in an fx.In struct as a parameter and has either of ParamTags, As or From annotation
+// - returns an fx.Out struct as a result and has either of ResultTags or As annotation
+// - takes in an fx.In struct as a parameter and has either of ParamTags or From annotation
 // - has an error result not as the last result.
 func (ann *annotated) typeCheckOrigFn() error {
 	ft := reflect.TypeOf(ann.Target)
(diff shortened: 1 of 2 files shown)


JacobOaks:
I understand disallowing `fx.From` with `fx.In` and `fx.Out` with `fx.As`. But is there any reason why we should be disallowing `fx.From` (applies to inputs only) with a function that returns an `fx.Out` (is an output) and `fx.As` (applies to outputs only) with a function that returns an `fx.In` (is an input)?

annotated.go
```diff
@@ -1608,18 +1614,25 @@ func (ann *annotated) typeCheckOrigFn() error {
 		if ot.Kind() != reflect.Struct {
 			continue
 		}
-		if len(ann.ResultTags) > 0 && dig.IsOut(reflect.New(ft.Out(i)).Elem().Interface()) {
-			return errors.New("fx.Out structs cannot be annotated with fx.ResultTags")
+		if !dig.IsOut(reflect.New(ft.Out(i)).Elem().Interface()) {
+			continue
+		}
+		if len(ann.ResultTags) > 0 || len(ann.As) > 0 || len(ann.From) > 0 {
```
tchung1118, line 1620: This is actually the old behavior we had for As/From annotations with Out/In structs, which is the safest way to go before easing up on this requirement. However, I agree with Jacob that this is a little overly restrictive, considering As annotation does not affect parameters of a constructor, and From annotation does not affect results of a constructor at all. So we could probably let users use As and From annotations with In and Out structs, respectively.

I'll leave it up to you to decide on how you'd like to proceed.

annotated_test.go
```diff
@@ -1448,10 +1448,40 @@ func TestAnnotate(t *testing.T) {
 
 		err := app.Err()
 		require.Error(t, err)
-		assert.Contains(t, err.Error(), "fx.Out structs cannot be annotated")
+		assert.Contains(t, err.Error(), "fx.Out structs cannot be annotated with either of fx.ResultTags, fx.As or fx.From")
 	})
 
-	t.Run("annotate a fx.In", func(t *testing.T) {
+	t.Run("annotate a fx.Out with As", func(t *testing.T) {
```
tchung1118, line 1454: Depending on what you do with my first comment, you might want to add more test cases. Specifically, with the current implementation, you're missing test cases using fx.From with fx.Out and fx.As with fx.In. If you decide to let users be able to use those cases, you don't need to add them to tests.

annotated.go
```diff
@@ -1608,18 +1614,25 @@ func (ann *annotated) typeCheckOrigFn() error {
...
+			return errors.New("fx.Out structs cannot be annotated with either of fx.ResultTags, fx.As or fx.From")
 		}
+
```
tchung1118, line 1623: Extraneous empty line

tchung1118 (commented):
Leaving couple of comments. Also lint seems to have failed? (try running `make lint`)

annotated.go
ankit209: Makes sense. 
Have removed this unintended restriction and to allow fx.Out struct with From annotation and fx.In struct with As annotation.

ankit209:
> I understand disallowing `fx.From` with `fx.In` and `fx.Out` with `fx.As`. But is there any reason why we should be disallowing `fx.From` (applies to inputs only) with a function that returns an `fx.Out` (is an output) and `fx.As` (applies to outputs only) with a function that accepts an `fx.In` (is an input)?

Agreed Jacob. Thanks for calling this out, I have made the necessary edits.

ankit209:
> Leaving couple of comments. Also lint seems to have failed? (try running `make lint`)

Thanks @tchung1118. Have made the required edits and fixed lint errors as well.

annotated_test.go
```diff
@@ -1448,10 +1448,40 @@ func TestAnnotate(t *testing.T) {
...
+		t.Parallel()
+
+		type I interface{}
+
+		type B struct {
+			// implents I
```
JacobOaks, line 1460:
```suggestion
			// implements I
```

JacobOaks (commented):
This LGTM but I'll wait to see if @tchung1118 has any outstanding comments.

tchung1118 (approved):
LGTM

annotated.go
```diff
@@ -1322,6 +1324,9 @@ var _ Annotation = (*fromAnnotation)(nil)
 //	fx.Provide(func(r1 *FooRunner, r2 *BarRunner) *RunnerWraps {
 //	  return NewRunnerWraps(r1, r2)
 //	})
+//
+// From annotation cannot be used in a function that takes an fx.In struct as a
```
sywhang, line 1328:
```suggestion
// From annotation cannot be used in a function that takes an [In] struct as a
```

annotated.go
```diff
@@ -1591,8 +1596,8 @@ func (ann *annotated) cleanUpAsResults() {
 }
 
 // checks and returns a non-nil error if the target function:
-// - returns an fx.Out struct as a result.
-// - takes in an fx.In struct as a parameter.
+// - returns an fx.Out struct as a result and has either of ResultTags or As annotation
```
sywhang, line 1599:
```suggestion
// - returns an fx.Out struct as a result and has either a ResultTags or an As annotation
```

annotated.go
```diff
@@ -1608,18 +1613,23 @@ func (ann *annotated) typeCheckOrigFn() error {
 		if ot.Kind() != reflect.Struct {
 			continue
 		}
-		if len(ann.ResultTags) > 0 && dig.IsOut(reflect.New(ft.Out(i)).Elem().Interface()) {
-			return errors.New("fx.Out structs cannot be annotated with fx.ResultTags")
+		if !dig.IsOut(reflect.New(ft.Out(i)).Elem().Interface()) {
+			continue
+		}
+		if len(ann.ResultTags) > 0 || len(ann.As) > 0 {
+			return errors.New("fx.Out structs cannot be annotated with either of fx.ResultTags or fx.As")
```
sywhang, line 1620:
```suggestion
			return errors.New("fx.Out structs cannot be annotated with fx.ResultTags or fx.As")
```

annotated.go
```diff
@@ -1608,18 +1613,23 @@ func (ann *annotated) typeCheckOrigFn() error {
...
 		if it.Kind() != reflect.Struct {
 			continue
 		}
-		if len(ann.ParamTags) > 0 && dig.IsIn(reflect.New(ft.In(i)).Elem().Interface()) {
-			return errors.New("fx.In structs cannot be annotated with fx.ParamTags")
+		if !dig.IsIn(reflect.New(ft.In(i)).Elem().Interface()) {
+			continue
+		}
+		if len(ann.ParamTags) > 0 || len(ann.From) > 0 {
+			return errors.New("fx.In structs cannot be annotated with either of fx.ParamTags or fx.From")
```
sywhang, line 1632:
```suggestion
			return errors.New("fx.In structs cannot be annotated with fx.ParamTags or fx.From")
```

annotated_test.go
```diff
@@ -1448,10 +1448,40 @@ func TestAnnotate(t *testing.T) {
...
+
+		app := NewForTest(t,
+			fx.Provide(
+				fx.Annotate(f, fx.As(new(I))),
+			),
+		)
+
+		err := app.Err()
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "fx.Out structs cannot be annotated with either of fx.ResultTags or fx.As")
```
sywhang, line 1481:
```suggestion
		assert.Contains(t, err.Error(), "fx.Out structs cannot be annotated with fx.ResultTags or fx.As")
```

annotated_test.go
```diff
@@ -1471,7 +1501,32 @@ func TestAnnotate(t *testing.T) {
...
+
+		app := NewForTest(t,
+			fx.Provide(
+				fx.Annotate(func(p Param) string { return "ok" }, fx.From(new(B))),
+			),
+		)
+		err := app.Err()
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "invalid annotation function func(fx_test.Param) string")
+		assert.Contains(t, err.Error(), "fx.In structs cannot be annotated with either of fx.ParamTags or fx.From")
```
sywhang, line 1529:
```suggestion
		assert.Contains(t, err.Error(), "fx.In structs cannot be annotated with fx.ParamTags or fx.From")
```

sywhang (approved):

ankit209:
Thanks for correcting the documentation and approving the PR @sywhang

PR #1075, opened 2023-04-29
App.Run: Respect Shutdowner ExitCode

abhinav:
We added support for changing the exit code for a Shutdowner with the
ExitCode option in #989, but this was somehow not respected by App.Run.

This changes App.Run to use the same underlying machinery (`Wait()`) to
decide on the exit code to use.

To test this, we add a new internal/e2e submodule that will hold full,
end-to-end integration tests.
These can be full Fx applications that we run tests against.
This is a submodule so that it can have dependencies that are not
desirable as direct dependencies of Fx,
and it's inside the internal/ directory so that it can consume
Fx-internal packages (like testutil).

The included regression test verifies the behavior described in #1074.
An Fx program using App.Run, and shut down with Shutdowner.Shutdown
and an explicit exit code, does not exit with the requested exit code.
Failure before the fix:

```
% go test
--- FAIL: TestShutdownExitCode (0.01s)
    writer.go:40: [Fx] PROVIDE  fx.Lifecycle <= go.uber.org/fx.New.func1()
    writer.go:40: [Fx] PROVIDE  fx.Shutdowner <= go.uber.org/fx.(*App).shutdowner-fm()
    writer.go:40: [Fx] PROVIDE  fx.DotGraph <= go.uber.org/fx.(*App).dotGraph-fm()
    writer.go:40: [Fx] INVOKE           go.uber.org/fx/internal/e2e/shutdowner_run_exitcode.main.func1()
    writer.go:40: [Fx] RUNNING
    writer.go:40: [Fx] TERMINATED
    main_test.go:46:
                Error Trace:    [..]/fx/internal/e2e/shutdowner_run_exitcode/main_test.go:46
                Error:          An error is expected but got nil.
                Test:           TestShutdownExitCode
FAIL
exit status 1
FAIL    go.uber.org/fx/internal/e2e/shutdowner_run_exitcode     0.016s
```

Resolves #1074

---

There's a follow up to this: https://github.com/abhinav/fx/pull/1.
It depends on the e2e test machinery, so I'll make a PR out of it once this is merged.

Issue #1074, opened 2023-04-29
Terminate my application with exit code


abhinav:
Bug report from #1073 

### Discussed in https://github.com/uber-go/fx/discussions/1073

<div type='discussions-op-text'>

<sup>Originally posted by **nemliyartym** April 28, 2023</sup>
Hi all. I would like to know how can I terminate my application with the exit codes I need. I have a simple `smb `module with a run function that calls `shutdowner.Shutdown(fx.ExitCode(20))` with the code I need.

If in main I just call `app.Run()` then my app exits with code 0. Do I test it with `echo$?`. I changed `Run` to call `Start+Done+Stop` to try and get the exit code from `Done`, but that didn't work for me. I also tried to change `Done`to `Wait`, but in this case, the application hangs on the channel of wait.

Please, tell me what am I doing wrong?)

main.go
```go
func main() {
	app := fx.New(
		//fx.NopLogger,
		//config.Module,
		//logger.Module,

		smb.Module,
	)
	//app.Run()

	if err := app.Start(context.TODO()); err != nil {
		fmt.Println(err)
	}

	done := app.Done()
	fmt.Println(done)

	// wait := app.Wait()
	// sig := <-wait
	// fmt.Println(sig.ExitCode)

	if err := app.Stop(context.TODO()); err != nil {
		fmt.Println(err)
	}
}
```
smb.go
```go
type Smb interface {
}

type smb struct {
	logger *zap.Logger
}

func NewSmb(logger *zap.Logger) Smb {
	return &smb{logger: logger}
}

func Run(logger *zap.Logger, shutdowner fx.Shutdowner) error {
	logger.Info("shutdowner")
       //SET EXIT CODE
	shutdowner.Shutdown(fx.ExitCode(20))
	logger.Info("shutdowner2")
	return nil
}

var Module = fx.Module("smb",
	fx.Provide(NewSmb),
	fx.Invoke(Run),
	fx.Decorate(
		func(log *zap.Logger) *zap.Logger {
			return log.Named("smb")
		},
	),
)
```
</div>

PR #1076, opened 2023-04-29
Shutdowner: Support calling from fx.Invoke

abhinav:
Stacked on top of:

- https://github.com/uber-go/fx/pull/1081
- https://github.com/uber-go/fx/pull/1082

However, since I can't push branches directly to this repository,
this PR shows commits from all PRs.

---

App.Start nils out the "last" signal recorded by signalReceivers,
which it otherwise broadcasts to waiters if it was already received.
This is unnecessary especially because there's a discrepancy in behavior
of using App.Start vs App.Run when shutting down from fx.Invoke.

Given a program that calls Shutdown from fx.Invoke,
when we do:

    app := fx.New(...)

The shutdowner has already sent the signal, and signalReceivers has
already recorded it.
At that point, whether we call App.Start or App.Run changes behavior:

- If we call App.Run, that calls App.Done (or App.Wait after #1075),
  which gives it back a channel that already has the signal filled in.
  It then calls App.Start, waits on the channel--which returns
  immediately--and then calls App.Stop.
- If we call App.Start and App.Wait, on the other hand,
  Start will clear the signal recorded in signalReceivers,
  and then App.Wait will build a channel that will block indefinitely
  because Shutdowner.Shutdown will not be called again.

So even though App.Run() and App.Start()+App.Wait() are meant to be
equivalent, this causes a serious discrepancy in behavior.
It makes sense to resolve this by supporting Shutdown from Invoke.

Refs #1074

### DIFF
--- a/signal.go
+++ b/signal.go
@@ -109,7 +109,6 @@ func (recv *signalReceivers) Start(ctx context.Context) {
 		return
 	}
 
-	recv.last = nil
 	recv.finished = make(chan struct{}, 1)
 	recv.shutdown = make(chan struct{}, 1)
 	recv.notify(recv.signals, os.Interrupt, _sigINT, _sigTERM)
@@ -135,6 +134,7 @@ func (recv *signalReceivers) Stop(ctx context.Context) error {
 		close(recv.finished)
 		recv.shutdown = nil
 		recv.finished = nil
+		recv.last = nil
 		return nil
 	}
 }
--- a/signal.go
+++ b/signal.go
@@ -139,7 +139,7 @@ func (recv *signalReceivers) Stop(ctx context.Context) error {
 	}
 }
 
-func (recv *signalReceivers) Done() chan os.Signal {
+func (recv *signalReceivers) Done() <-chan os.Signal {
 	recv.m.Lock()
 	defer recv.m.Unlock()
 
@@ -157,7 +157,7 @@ func (recv *signalReceivers) Done() chan os.Signal {
 	return ch
 }
 
-func (recv *signalReceivers) Wait() chan ShutdownSignal {
+func (recv *signalReceivers) Wait() <-chan ShutdownSignal {
 	recv.m.Lock()
 	defer recv.m.Unlock()
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
-- No changes yet.
+### Deprecated
+- Deprecate `ShutdownTimeout` option.
 
 ## [1.19.1](https://github.com/uber-go/fx/compare/v1.18.0...v1.19.1) - 2023-01-10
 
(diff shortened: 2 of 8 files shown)


abhinav:
Coverage issue will be resolved by https://github.com/uber-go/fx/pull/1078

JacobOaks:
Spent a little bit looking at this and I think it makes sense.

Have we considered switching the order of execution of `Start` vs. `Wait` in the case of calling `Run` to more closely match the `Start` -> `Wait` -> `Stop` flow? For example, `run` can take a channel-generating function rather than a channel itself, and we can pass `Wait`:
```go
func (app *App) Run() {
    if code := app.run(app.Wait); code != 0 {
        // ...
    }
}

func (app *App) run(doneFn func() <-chan ShutdownSignal) (exitCode int) {
// ...
if err := app.Start(); err != nil { /* ... */ }
sig := <-doneFn()
// ...
}
```
This would cause shutdown from invokes to no longer work when using `Run` making it a breaking change by itself, but in combination with this PR it shouldn't be breaking, and it might be a good idea to prevent future changes/features from introducing additional disparities between `app.Run` and `app.Start` -> `app.Wait` -> `app.Stop`. Just a thought.

abhinav:
That's a good idea! I'll make the switch.

signal.go
```diff
@@ -109,7 +109,6 @@ func (recv *signalReceivers) Start(ctx context.Context) {
 		return
 	}
 
-	recv.last = nil
```
JacobOaks, line 112: If I'm understanding this correctly, since this is never cleared, if a shutdown signal is sent to the app, all successive runs of the app will always immediately terminate too. Not sure if that's an issue/how big of one it is, but I do know some folks rely on running an app multiple times. Maybe there's somewhere else we can clear it (in Stop)?

abhinav: Yeah, that's a good call. Made an update: signalReceivers is now re-usable with your suggested change.
Added a test that starts the app and shuts it down N times with different exit codes.

abhinav:
Actually, that's a problem. Looks like receivers.Stop is called from both, App.Stop and Shutdowner.Shutdown. I'll see if that can be fixed.

abhinav:
Okay, I fixed it but then the ShutdownTimeout option is meaningless. I'll dig a little.

abhinav:
Okay, so @JacobOaks @sywhang, Shutdown has a ShutdownTimeout option that doesn't make sense.
Shutdown can't be a blocking operation.
signalReceivers has the operations: Start, Stop, Broadcast, Wait (and its old version, Done).
Start starts a background goroutine to relay information, and Stop stops it.
It makes no sense for Shutdowner.Shutdown to call Stop if Shutdown is just to unblock a `<-Wait()` and a corresponding Stop.

Let me see if I can wrangle this apart a bit.

abhinav:
A good signal that the ShutdownTimeout option makes little sense is this in the documentation:

```
 // [..] As the Shutdown method will block while waiting for a signal receiver relay
 // goroutine to stop.
```

That's an internal implementation detail.

abhinav:
Updated. Check PRs stacked below this for more context.

JacobOaks (approved):
LGTM!

sywhang (approved):
👍

PR #1089, opened 2023-05-30
Make fx.Populate work with fx.Annotate

tchung1118:
This PR makes fx.Populate work with fx.Annotate. Previously, to fx.Populate a
named dependency, users would have to create a struct that embeds fx.In and has
a field with a tag and pass a pointer to that struct. This could be cumbersome
in cases where the user is intersted in only one named value. With this PR,
users will be able to use fx.Annotate to get the same effect:

```
var a A
...
fx.Populate(
  fx.Annotate(
    &a,
    fx.ParamTags(`name:"A"`),
  ),
)
```

### DIFF
--- a/populate_test.go
+++ b/populate_test.go
@@ -27,8 +27,6 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-
-	"go.uber.org/fx"
 	. "go.uber.org/fx"
 	"go.uber.org/fx/fxtest"
 )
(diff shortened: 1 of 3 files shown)


populate_test.go
```diff
@@ -158,6 +158,42 @@ func TestPopulate(t *testing.T) {
 		assert.False(t, targets.V1 == targets.V2, "fields should be different")
 	})
 
+	t.Run("annotated populate", func(t *testing.T) {
+		t.Parallel()
+
+		type result struct {
+			Out
+
+			V1 *t1 `name:"n1"`
```
sywhang, line 167: can you add a test case where the output is provided via fx.Annotate instead of an Out struct?

sywhang (commented):
👍 nice! Added a suggestion for a test case.

sywhang (approved):

PR #1121, opened 2023-09-27
adding support for wasi

maceip:
go 1.21 added wasi preview 1


sywhang (commented):
Thanks for the PR. 

I think it makes more sense for this to be a build tag change in [app_wasm.go](https://github.com/uber-go/fx/blob/master/app_wasm.go) rather than adding a whole file with the exact same content?

sywhang (approved):

PR #1122, opened 2023-10-02
Include Module Traces in Events

JacobOaks:
Currently, the events that the [fxevent package](https://pkg.go.dev/go.uber.org/fx/fxevent) emits for [provides](https://pkg.go.dev/go.uber.org/fx/fxevent#Provided)/[decorates](https://pkg.go.dev/go.uber.org/fx/fxevent#Decorated)/[supplies](https://pkg.go.dev/go.uber.org/fx/fxevent#Supplied)/[replaces](https://pkg.go.dev/go.uber.org/fx/fxevent#Replaced) include stack traces of where the Fx call was made.

These stack traces often look something like this:
```
somepkg.init (/path/to/some/distant/package/module.go:14)
runtime.doInit1 (/opt/homebrew/Cellar/go/1.21.1/libexec/src/runtime/proc.go:6740)
runtime.doInit (/opt/homebrew/Cellar/go/1.21.1/libexec/src/runtime/proc.go:6707)
runtime.main (/opt/homebrew/Cellar/go/1.21.1/libexec/src/runtime/proc.go:249)
```

While these do show exactly where the provide/decorate/supply/replace statement lives, for larger codebases where a single app may include a wide variety of modules, each with their own chain of sub-modules, this stack trace does a bad job of answering: "How did this constructor get included in my app in the first place?"

This PR proposes module traces, an attempt to alleviate this issue by providing a trace through the module chain by which a provide/decorate/supply/replace found its way to the top-level app declaration.

For this example,

```
  9 func getThing() thing {
 10     return thing{}
 11 }
 12
 13 var ThingModule = fx.Module(
 14     "ThingFx",
 15     fx.Provide(getThing),
 16 )
 17
 18 var ServiceModule = fx.Module(
 19     "ServiceFx",
 20     ThingModule,
 21 )
 22
 23 func main() {
 24     app := fx.New(
 25         ServiceModule,
 26         fx.Invoke(useThing),
 27     )
 28 }
```

The provide event for `getThing` will include the following module trace:

```
main.init (/Users/joaks/go/src/scratch/mod_stacks_scratch/main.go:15)
main.init (/Users/joaks/go/src/scratch/mod_stacks_scratch/main.go:13) (ThingFx)
main.init (/Users/joaks/go/src/scratch/mod_stacks_scratch/main.go:18) (ServiceFx)
main.run (/Users/joaks/go/src/scratch/mod_stacks_scratch/main.go:24)
```

Which shows exactly how `getThing` got included in `app` in order:
* getThing is provided at `main.go:15`
* ThingFx is declared at `main.go:13`
* ServiceFx is declared at `main.go:18`
* The app is declared at `main.go:24`

This makes it more clear to the developer how functions are getting included, especially for large codebase where modules may be distributed all over.


module.go
```diff
@@ -131,6 +135,13 @@ func (m *module) build(app *App, root *dig.Container) {
 	}
 }
 
+func (m *module) getTrace() (trace fxreflect.Stack) {
+	if m.parent != nil {
+		trace = m.parent.getTrace()
+	}
+	return append(trace, m.location)
```
tchung1118, line 142: Usually stack traces start from the latest frame and work its way down. As it is, the order of module trace is reverse of stack trace. I think this might be confusing to users. Also, I think it might be helpful if we can somehow include module's name to the module trace.

tchung1118 (commented):
Looks good. I just have one comment.

module.go
JacobOaks: Makes sense - I updated to have the reverse order. This is probably less efficient due to the repeated `append([]string{location}, locations...)`, but I agree with you about it being more correct. I thought about keeping the order and doing a slice reversal at the end, but thought that might be overkill.

tchung1118 (approved):

module.go
JacobOaks: @tchung1118 and I talked offline a little.

I switched the PR to cache traces for modules when creating them, rather than re-chaining the frames for every provide/decorate/supply/replace. I think this should be faster, especially for applications with lots of constructors/decorators.

app.go
```diff
@@ -432,7 +432,8 @@ func New(opts ...Option) *App {
 		// user gave us. For the last case, however, we need to fall
 		// back to what was provided to fx.Logger if fx.WithLogger
 		// fails.
-		log: logger,
+		log:   logger,
+		trace: []string{fxreflect.CallerStack(1, 0)[0].String()},
```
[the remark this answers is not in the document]
JacobOaks: Good point - I updated to pass depth 2, since we skip 1.

sywhang (approved):

PR #1150, opened 2024-02-06
ci: Use golangci-lint

abhinav:
**Commits are individually reviewable**

---

We've been using a hand-rolled lint setup for a while.
It's limiting, and so we've neglected to keep our lints up to date.

This PR switches to using golangci-lint to run linters.
staticcheck is included and enabled by default.
We add revive and a few others.

This also allows the linter to run in parallel with the tests.

See also uber-go/zap#1323 for a similar change to Zap.

Commits after the first one are fixes to the lints found by golangci-lint.
Some of these are:

- Files should be gofumpt-ed.

- Shouldn't add type in `var x T = y` when type of `y` is `T`.

    ```
    annotated.go:120:15: var-declaration: should omit type reflect.Type from declaration of var _typeOfError; it will be inferred from the right-hand side (revive)
    annotated.go:645:19: var-declaration: should omit type reflect.Type from declaration of var _typeOfLifecycle; it will be inferred from the right-hand side (revive)
    annotated.go:646:19: var-declaration: should omit type reflect.Type from declaration of var _typeOfContext; it will be inferred from the right-hand side (revive)
    internal/lifecycle/lifecycle.go:198:18: var-declaration: should omit type appState from declaration of var returnState; it will be inferred from the right-hand side (revive)
    ```

- Should use `errors.As` and `%w` when wrapping and unwrapping errors.

    ```
    app.go:550:14: type assertion on error will fail on wrapped errors. Use errors.As to check for specific errors (errorlint)
    decorate.go:219:81: non-wrapping format verb for fmt.Errorf. Use `%w` to format errors (errorlint)
    ```

- All public packages should have doc comments.

- `x += 1` should be `x++`.

    ```
    module_test.go:237:5: increment-decrement: should replace p.age += 1 with p.age++ (revive)
    module_test.go:242:6: increment-decrement: should replace p.age += 1 with p.age++ (revive)
    ```

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -14,14 +14,12 @@ jobs:
 
   build:
     runs-on: ${{ matrix.os }}
+    name: Build and test
+
     strategy:
       matrix:
         os: ["ubuntu-latest", "windows-latest"]
         go: ["1.20.x", "1.21.x"]
-        include:
-        - go: 1.21.x
-          os: "ubuntu-latest"
-          latest: true
 
     steps:
     - name: Checkout code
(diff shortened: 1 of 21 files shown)


JacobOaks (approved):
This seems much cleaner! The exclusions you configured make sense. LGTM.

r-hang (approved):
Thanks for your contribution!

PR #1154, opened 2024-02-09
test: add a custom mock clock implementation

### LINKED ISSUE #1135: https://github.com/benbjohnson/clock is EOL/archived.
**Describe the bug**
https://github.com/benbjohnson/clock is archived.

**To Reproduce**
N/A, we can code scanning tools in-house that complain about archived repo (considered EOL)

**Expected behavior**
N/A, code scanning shouldn't complain about `go.mod`.

**Additional context**
based on https://github.com/benbjohnson/clock/network maybe https://github.com/itbasis/go-clock would be a good in-place replacement.

```
╰─ go mod graph | grep github.com/benbjohnson/clock
go.uber.org/fx@v1.20.0 github.com/benbjohnson/clock@v1.3.0
go.uber.org/zap@v1.24.0 github.com/benbjohnson/clock@v1.1.0
go.uber.org/zap@v1.18.1 github.com/benbjohnson/clock@v1.1.0
go.uber.org/zap@v1.19.0 github.com/benbjohnson/clock@v1.1.0
```

abhinav:
benbjohnson/clock has been archived and no longer maintained.
Removing that dependency is desirable.

This change adds a mock clock adapted from Zap's mock clock
(uber-go/zap#1349), but with operations specific to Fx's needs.

There are two areas that require more scrutiny
because of divergence from Zap's mock clock:

WithTimeout requires us to implement a custom `context.Context`
so that it can report context.DeadlineExceeded when it's time.
We cannot just use `context.WithCancelCause` here because the cause
for a context failure is considered different from `ctx.Err`,
so `ctx.Err` would still report `context.Canceled` for a timeout.

When testing that a sleep behaves as expected, there's a data race
between the `Sleep` and the `Add` that progresses time.
To resolve this, this change adds an `AwaitScheduled` method
that blocks until there are operations scheduled for the future.
This is done by using a `sync.Cond`.
This gives us a way to wait until the sleep is scheduled
before we progress time.

This change also had to update Zap to pick up the release with the
custom clock to drop the benbjohnson/clock dependency from the go.mod
completely.

Refs uber-go/zap#1349
Resolves #1135


internal/fxclock/clock_test.go
```diff
@@ -22,44 +22,150 @@ package fxclock
...
 	"testing"
 	"time"
 
-	"github.com/benbjohnson/clock"
 	"github.com/stretchr/testify/assert"
 )
 
-var _ Clock = clock.Clock(nil)
-
-// Just a basic sanity check that everything is in order.
```
abhinav, line 34: Instead of just the system clock sanity test, the test is now used for both, system clock and mock clock to verify basic outer expectations of the clocks.

rabbbit:
Looks like you don't intend to put the clock in an externally addressable
location? :l

I should probably copy paste this into uber/ratelimit too. Not only is it
on bobjohnson but also on an ancient version of it - upgrading breaks the
tests :l

abhinav:
> Looks like you don't intend to put the clock in an externally addressable location? :l

We do! @prashantv has a nicer version of the clock in progress for re-use between projects.

abhinav:
image

unclear what's going on here; i don't have visibility into it

internal/fxclock/clock_test.go
```diff
@@ -22,44 +22,150 @@ package fxclock
...
+		go func() {
+			defer wg.Done()
+
+			clock.Sleep(1 * time.Millisecond)
+		}()
+	}
+
+	clock.AwaitScheduled(N)
+	clock.Add(1 * time.Millisecond)
+	wg.Wait()
```
JacobOaks, line 170: Probably overkill, but should we instead use a channel and expect N messages? Then we can abort after a certain timeout and avoid deadlock if, for whatever reason, a regression causes Sleep to not return.
```
ch := make(chan struct{}, N)
for i := 0; i < N; i++ {
	go func() {
		clock.Sleep(1 * time.Millisecond)
		ch <- struct{}{}
	}()
}
clock.AwaitScheduled(N)
var got int
timeout := time.After(10 * time.Millisecond)
for got < N {
	select {
	case <-ch:
		got++
	case <-timeout:
		t.Fatal("expected all Sleeps to unblock")
	}
}
```

internal/fxclock/clock_test.go
```diff
@@ -22,44 +22,150 @@ package fxclock
...
+	assert.False(t, now.IsZero())
 
-		before := clock.Now()
-		assert.GreaterOrEqual(t, clock.Since(before), time.Duration(0))
+	t.Run("Since", func(t *testing.T) {
+		advance(1 * time.Millisecond)
+		assert.NotZero(t, clock.Since(now), "time must have advanced")
 	})
 
 	t.Run("Sleep", func(t *testing.T) {
```
JacobOaks, line 51: Won't this run into the race you mentioned in the PR summary? I.e., on the off chance advance runs before sleep blocks, wouldn't the Sleep block forever. For `Mock`, isn't this functionality covered by `TestMock_Sleep`?

JacobOaks (approved):
LGTM I think, just a couple questions/comments on your tests, and probably should wait for another maintainer to take another look too.

internal/fxclock/clock.go
```diff
@@ -56,3 +56,199 @@ func (systemClock) Sleep(d time.Duration) {
...
+
+func (c *deadlineCtx) Deadline() (deadline time.Time, ok bool) { return c.deadline, true }
+func (c *deadlineCtx) Done() <-chan struct{}                   { return c.done }
+func (c *deadlineCtx) Value(key any) any                       { return c.inner.Value(key) }
+
+func (c *deadlineCtx) Err() error {
+	c.mu.Lock()
+	err := c.err
+	c.mu.Unlock()
+	return err
```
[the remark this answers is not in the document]
abhinav: Yeah, good call

internal/fxclock/clock_test.go
abhinav: A channel for that would work.

Although I think if the main concern is a deadlock, might make sense to stick with the wait group and just use a channel to enforce a wait specifically on that. I'll make that change.

internal/fxclock/clock_test.go
abhinav: Yeah, you're right! Great catch. Let me fix this.

abhinav: Fixed!

sywhang:
quick first pass, will come back to this later tonight

internal/fxclock/clock.go
```diff
@@ -56,3 +56,198 @@ func (systemClock) Sleep(d time.Duration) {
...
+
+	c.waiters = append(c.waiters, waiter{until: t, fn: fn})
+	c.waiterAdded.Broadcast()
+}
+
+// AwaitScheduled blocks until there are are at least N
+// operations scheduled for the future.
+func (c *Mock) AwaitScheduled(n int) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
```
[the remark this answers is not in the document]
abhinav: The docs also link to [this talk](https://drive.google.com/file/d/1nPdvhB0PutEJzdCq5ms6UI58dp50fcAN/view) by Bryan Mills. See page 37.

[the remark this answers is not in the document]
sywhang: But that's a different lock, right? i.e. the one on the `Cond` struct: 
```go
type Cond struct {

	// L is held while observing or changing the condition
	L [Locker](https://pkg.go.dev/sync#Locker)
	// contains filtered or unexported fields
}
```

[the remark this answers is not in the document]
abhinav: It's the same lock.

https://github.com/uber-go/fx/pull/1154/files#diff-3a247134bac12eee506246c998243ff5546c388572f2a51ed6c53f3205aa4682R85-R86

```go
func NewMock() *Mock {
 	m := &Mock{now: time.Now()}
 	m.waiterAdded = sync.NewCond(&m.mu)
 	return m
```

Let me add a comment here.

sywhang (approved):

PR #1156, opened 2024-02-09
docs: Define Container, fx.Provide, fx.Supply, fx.Invoke

abhinav:
Adds documentation defining the concept of a container
and a couple of the core intrinsic operations that a container suppotrs.

In the future, this should include `fx.Decorate` as an operation,
and can link to more thorough documentation about decoration separately.

### DIFF
--- a/docs/container.md
+++ b/docs/container.md
@@ -54,7 +54,7 @@ Fx provides two ways to provide values to the container:
   This says that Fx should use the provided `*Config` as-is.
 
   **Important**: `fx.Supply` is only for non-interface values.
-  See [When to use fx.Supply](#when-to-use-fx-supply) for more details.
+  See *When to use fx.Supply* for more details.
 
 Values provided to the container are available to all other constructors.
 In the example above, the `*Config` would become available to the `*Logger` constructor,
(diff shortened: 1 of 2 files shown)


abhinav:
Screenshots:
image

image

docs/container.md
```diff
@@ -0,0 +1,156 @@
...
+package fx
+
+type App
+  func New(opts ...Option) *App
+  func (app *App) Run()
+
+type Option
+  func Provide(constructors ...interface{}) Option
+  func Invoke(funcs ...interface{}) Option
+```
```
JacobOaks, line 23: I think your intention is to show some relevant functions for this section and their relevent types, but this throws me off a little because its formatted as a code block and yet isn't really valid Go syntax. I think it's fine, just something to be aware of. Is this a common format for trying to portray this kind of information?

abhinav:
> error: 		container.md:57: link #when-to-use-fx-supply, normalized to: link /home/runner/work/fx/fx/docs/container.md#when-to-use-fx-supply, existing ids: [container providing-values when-to-use-fxsupply using-values when-to-use-fxinvoke]: file exists, but does not have such id

Hm, mdox and vuepress disagree here. I'll just use an explicit anchor.

abhinav:
Never mind, it appears that mdox is just broken for this case. 
It doesn't recognize explicit anchors added with `<a id="foo"></a>`, and it incorrectly drops the '.' when guessing the id.

I'm going to work around this for now, but we should consider dropping its link validation completely in that case.

JacobOaks (approved):
Nice! LGTM - just left one question.

docs/container.md
abhinav: This formatting was borrowed from how godoc shows things:

image

I'm not tied to it, though. We can change it if you'd like.

JacobOaks: Ah okay! Makes sense. I think this is good then.

Issue #1152, opened 2024-02-07
Startup order isn't preserved when using modules


Philio:
**Describe the bug**
I've bundled some commonly used dependencies in modules for convenience, for example db initialisation and web server start up. In one of my web services I implemented them as follows (with the intention that the data migration function runs before the web server starts):

```go
	fx.New(
		...
		service.DatastoreModule,  // Configures and connects to Google cloud datastore
		fx.Invoke(runMigrations), // Runs data migrations
		service.ServerModule,     // Configures routing and starts web server
	).Run()
```

But on startup the web server always starts first and the web server's OnStart hook is called first:

```
[Fx] INVOKE             ***/service.startServer() from module "server"
...
[Fx] INVOKE             main.runMigrations()
...
[Fx] HOOK OnStart               ***/service.startServer.func1() executing (caller: ***/service.startServer)
[Fx] HOOK OnStart               ***/service.startServer.func1() called by ***/service.startServer ran successfully in 3.48µs
[Fx] HOOK OnStart               ***/service.newDatastoreClient.func1() executing (caller: ***/service.newDatastoreClient)
[country] 2024/02/07 02:36:50.404627 Starting HTTP server on 0.0.0.0:8080
[Fx] HOOK OnStart               ***/service.newDatastoreClient.func1() called by ***/service.newDatastoreClient ran successfully in 435.478µs
[Fx] HOOK OnStart               main.runMigrations.func1() executing (caller: main.runMigrations)
[Fx] HOOK OnStart               main.runMigrations.func1() called by main.runMigrations ran successfully in 1.45µs
```

However if I wrap the migrations in a module as well:

```go
	fx.New(
		...
		service.DatastoreModule,
		migrationsModule,
		service.ServerModule,
	).Run()
```

Now it works as expected:

```
[Fx] INVOKE             main.runMigrations() from module "migrations"
...
[Fx] INVOKE             github.com/boxes-ltd/service-common/service.startServer() from module "server"
...
[Fx] HOOK OnStart               ***/service.newDatastoreClient.func1() executing (caller: ***/service.newDatastoreClient)
[Fx] HOOK OnStart               ***/service.newDatastoreClient.func1() called by ***/service.newDatastoreClient ran successfully in 406.037µs
[Fx] HOOK OnStart               main.runMigrations.func1() executing (caller: main.runMigrations)
[Fx] HOOK OnStart               main.runMigrations.func1() called by main.runMigrations ran successfully in 1.42µs
[...]

PR #1160, opened 2024-02-13
doc: Add FAQ page

abhinav:
Adds an FAQ page to the Fx documentation,
aiming to address frequently asked questions.

This includes two entries that came up at work recently.

Includes a test for the `fx.Supply` entry
because I couldn't find the test for it right away.
(I found it in annotated_test.go later but you won't hear me complain
about additional tests.)


docs/faq.md
```diff
@@ -0,0 +1,47 @@
+# Frequently Asked Questions
+
+This page contains answers to common questions and issues with using Fx.
+
+## Does the order of `fx.Option`s matter?
+
+No, the order in which you provide various Fx options
+to `fx.Options`, `fx.New`, `fx.Module`, and others does not matter.
```
JacobOaks, line 8: I don't think this is true for determining the order in which `fx.Invoke`d functions get run, which was a recent point of confusion (ref: https://github.com/uber-go/fx/issues/1152). Consider exempting Invoke from this statement.

JacobOaks (approved):
Awesome, thanks! One small comment.

docs/faq.md
abhinav: Fair. There's a weird situation here.
On paper, there's no guarantee about the order in which invokes run.
But you're right, their order matters.

abhinav: @JacobOaks I realized that this was actually too simplified and lost some information.
Ordering of options relative to fx.New is irrelevant, but ordering of operations relative to each other can matter.
I've rewritten this section to better clarify this.

LMK what you think

abhinav: Actually, looks like we do document the fx.Invoke invocation order.
I'll update this entry.

JacobOaks (approved):
👍 thanks!

PR #1163, opened 2024-02-17
Move advanced features to package documentation

abhinav:
The documentation for parameter structs, result structs,
named values, optional values, value groups, etc.
is currently scattered between In and Out.

First, on the logistics end:
this has made it impossible to link to some of these docs because they
share the same hadings:
https://pkg.go.dev/go.uber.org/fx#hdr-Named_Values goes to the named
values section of fx.In. It's not possible to link to the named values
section of the fx.Out struct (which is also called Named Values).

More importantly, from the doc readability point of view, jumping
between In and Out is annoying and doesn't make for good reading.

This PR moves these to the package-level docs and improves the flow of
how this information is presented:

First, we introduce parameter and result structs.
Then we weave through the two introducing various topics:

    Producing named values        Result structs
    Consuming named values        Parameter structs
    Optional dependencies         Parameter structs
    Producing value groups        Result structs
    Consuming value groups        Parameter structs
    Soft value groups             Parameter structs
    Value group flattening        Result structs
    Unexported fields             Parameter structs

This makes the story of these topics a bit easier to follow.

In the future, we can introduce annotation right
after parameter and result structs,
or between value groups and soft value groups.

Note that this does not touch on the core operations
(provide, invoke, decorate).
It's just moving existing documentation around.


JacobOaks (approved):
👍

doc.go
```diff
@@ -27,13 +27,335 @@
...
+//		Posts    *PostGateway
+//		Votes    *VoteGateway
+//		AuthZ    *AuthZGateway
+//	}
+//
+//	func NewHandler(p HandlerParams) *Handler {
+//		// ...
+//	}
+//
+// Though it's rarely a good idea, constructors can receive any combination of
```
sywhang, line 76: Should we maybe add a bit more detail here on why this is not a good idea?

sywhang (approved):

doc.go
```diff
@@ -27,13 +27,335 @@
...
+//
+// A soft value group can be thought of as a best-attempt at populating the
+// group with values from constructors that have already run. In other words,
+// if a constructor's output type is only consumed by a soft value group,
+// it will not be run.
+//
+// Note that Fx does not guarantee precise execution order of constructors
+// or invokers, which means that the change in code that affects execution
+// ordering of other constructors or functions will affect the values
+// populated in this group.
```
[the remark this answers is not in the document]
abhinav: Fair. I'm going to nix this paragraph in favor of saying that Fx randomizes the order of values in the value group. There's no guarantee about the expected ordering relative to constructor invocation and dependencies.

doc.go
abhinav: I guess it's less of a "good idea" thing and more of a "unnecessary."?
I'll change the wording here.

PR #1164, opened 2024-02-18
docs/sidebar: Move concepts before features

abhinav:
In reading order, concepts like container and lifecycle
should come before we talk about parameter objects, value groups, etc.
(It's also okay to merge concepts and features, but I'm not doing that
here.)

Separately, make the lifecycle page appear as "Lifecycle" in the sidebar
while still having the title "Application Lifecycle".


abhinav:
image

JacobOaks (approved):

PR #1188, opened 2024-04-14
drop io/utils package.

zhanluxianshen:
/kind cleanup

Drop ioutil package.
As it is deprecation , and can be replaced well.

```
Deprecated: As of Go 1.16, the same functionality is now 
provided by package [io](https://pkg.go.dev/io) 
or package [os](https://pkg.go.dev/os), 
and those implementations should be preferred in new code
```

### DIFF
--- a/app_windows_test.go
+++ b/app_windows_test.go
@@ -27,7 +27,6 @@ import (
 	"context"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


r-hang (approved):

zhanluxianshen:
thanks

PR #1191, opened 2024-04-18
Provide Fx types before user types

JacobOaks:
Consider the following example:
```go
func opts() fx.Option {
        return fx.Options(
                fx.WithLogger(func(fx.Lifecycle) fxevent.Logger {
                        return &fxevent.ConsoleLogger{ W: os.Stdout }
                }),
                fx.Provide(func() string { return "" }),
                fx.Provide(func() string { return "" }),
        )
}

func main() {
        fx.New(opts()).Run()
}
```

The relevant issue to surface to the user is that they are double providing the same type. However, the actual error message is:
```
[Fx] ERROR              Failed to start: the following errors occurred:
 -  fx.Provide(main.opts.func3()) from:
    main.opts
        /home/user/go/src/scratch/fx_provide_order/main.go:17
    main.main
        /home/user/go/src/scratch/fx_provide_order/main.go:22
    runtime.main
        /opt/go/root/src/runtime/proc.go:271
    Failed: cannot provide function "main".opts.func3
(/home/user/go/src/scratch/fx_provide_order/main.go:17): cannot provide
string from [0]: already provided by "main".opts.func2
(/home/user/go/src/scratch/fx_provide_order/main.go:16)
 -  could not build arguments for function
"go.uber.org/fx".(*module).constructCustomLogger.func2
        /home/user/go-repos/pkg/mod/go.uber.org/fx@v1.21.0/module.go:292:
    failed to build fxevent.Logger:
    missing dependencies for function "main".opts.func1
        /home/user/go/src/scratch/fx_provide_order/main.go:11:
    missing type:
        - fx.Lifecycle (did you mean to Provide it?)
```
Which contains an additional error related to how the custom logger could not be built.

This is because Fx will try to continue to build the custom logger in the face of DI failure, theoretically to report issues through the right channels. But after an error occurs when providing anything, [Fx refuses to provide any more types](https://github.com/uber-go/fx/blob/master/module.go#L184) - leading to a subsequent error when trying to build this custom logger that depends on the `fx.Lifecycle` type.

This is a common issue that can be misleading for new engineers debugging their fx apps.

I couldn't find any particular reason why user-provided provides are registered before these Fx types, so this PR switches this ordering so that custom loggers can still be built if they rely on the Fx types, which cleans up the error message:
```
[Fx] ERROR              Failed to start: fx.Provide(main.opts.func3())
from:
main.opts
        /home/user/go/src/scratch/fx_provide_order/main.go:17
main.main
        /home/user/go/src/scratch/fx_provide_order/main.go:22
runtime.main
        /opt/go/root/src/runtime/proc.go:271
Failed: cannot provide function "main".opts.func3
(/home/user/go/src/scratch/fx_provide_order/main.go:17): cannot provide
string from [0]: already provided by "main".opts.func2
(/home/user/go/src/scratch/fx_provide_order/main.go:16)
```

### DIFF
--- a/app.go
+++ b/app.go
@@ -480,7 +480,7 @@ func New(opts ...Option) *App {
 		m.build(app, app.container)
 	}
 
-	// Provide Fx types first to increase the change a custom logger
+	// Provide Fx types first to increase the chance a custom logger
 	// can be successfully built in the face of unrelated DI failure.
 	// E.g., for a custom logger that relies on the Lifecycle type.
 	frames := fxreflect.CallerStack(0, 0) // include New in the stack for default Provides
--- a/app.go
+++ b/app.go
@@ -480,10 +480,9 @@ func New(opts ...Option) *App {
 		m.build(app, app.container)
 	}
 
-	for _, m := range app.modules {
-		m.provideAll()
-	}
-
+	// Provide Fx types first to increase the change a custom logger
+	// can be successfully built in the face of unrelated DI failure.
+	// E.g., for a custom logger that relies on the Lifecycle type.
 	frames := fxreflect.CallerStack(0, 0) // include New in the stack for default Provides
 	app.root.provide(provide{
 		Target: func() Lifecycle { return app.lifecycle },
@@ -491,6 +490,10 @@ func New(opts ...Option) *App {
 	})
 	app.root.provide(provide{Target: app.shutdowner, Stack: frames})
 	app.root.provide(provide{Target: app.dotGraph, Stack: frames})
+
+	for _, m := range app.modules {
+		m.provideAll()
+	}
 
 	// Run decorators before executing any Invokes -- including the one
 	// inside constructCustomLogger.
--- a/app_test.go
+++ b/app_test.go
@@ -113,7 +113,7 @@ func TestNewApp(t *testing.T) {
 			[]string{"Provided", "Provided", "Provided", "Provided", "LoggerInitialized", "Started"},
 			spy.EventTypes())
 
-		// Fx types get provided first to increase change of
+		// Fx types get provided first to increase chance of
 		// successful custom logger build.
 		assert.Contains(t, spy.Events()[0].(*fxevent.Provided).OutputTypeNames, "fx.Lifecycle")
 		assert.Contains(t, spy.Events()[1].(*fxevent.Provided).OutputTypeNames, "fx.Shutdowner")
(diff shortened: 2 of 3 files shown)


app.go
```diff
@@ -480,10 +480,6 @@ func New(opts ...Option) *App {
 		m.build(app, app.container)
 	}
 
-	for _, m := range app.modules {
-		m.provideAll()
-	}
-
 	frames := fxreflect.CallerStack(0, 0) // include New in the stack for default Provides
```
sywhang, line 483: Let's add a comment here describing why the Fx-provided dependencies are provided to the scope prior to the user provided types.

app_test.go
```diff
@@ -113,7 +113,30 @@ func TestNewApp(t *testing.T) {
 			[]string{"Provided", "Provided", "Provided", "Provided", "LoggerInitialized", "Started"},
 			spy.EventTypes())
 
-		assert.Contains(t, spy.Events()[0].(*fxevent.Provided).OutputTypeNames, "struct {}")
+		// Lifecycle, Shutdowner, and DotGraph are indices 0, 1, and 2.
```
sywhang, line 116: Want to assert this too? I get that this feels a bit change-detector ish but it already is that way as this test is structured. We may want to consider refactoring these tests in the future to reduce its change-detector ishness.

app_test.go
```diff
@@ -113,7 +113,30 @@ func TestNewApp(t *testing.T) {
...
+
+		spy := new(fxlog.Spy)
+		app := fxtest.New(t, Provide(func() struct{} { return struct{}{} }),
+			WithLogger(func() fxevent.Logger { return spy }))
+		defer app.RequireStart().RequireStop()
+		require.Equal(t,
+			[]string{"Provided", "Provided", "Provided", "Provided", "LoggerInitialized", "Started"},
+			spy.EventTypes())
+
+		assert.Contains(t, spy.Events()[0].(*fxevent.Provided).OutputTypeNames, "fx.Lifecycle")
```
sywhang, line 137: Ah, never mind. You separated it out to its own test.

Consider just putting this in the first test w/o separating it out to its own test.

sywhang (commented):
Did we validate this in internal CI? I don't see a reason it should affect anything but in case.

sywhang (approved):

JacobOaks:
> Did we validate this in internal CI? I don't see a reason it should affect anything but in case.

Yes, this has been validated in internal CI.

sywhang:
@JacobOaks do you want to cut a release?

PR #1195, opened 2024-04-29
CI: test against go1.22

JacobOaks:
This PR modifies CI to test against Go versions 1.21 and 1.22. To avoid forcing users to upgrade to 1.21, I kept `go.mod` version at `1.20.`

Some tests had to be updated to pass in 1.22.

I can remove `stack_120_test.go` if there's a good argument for it, but my thoughts are that we should have a complete & passing test suite for every version >= `go.mod`'s version.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       matrix:
         os: ["ubuntu-latest", "windows-latest"]
-        go: ["1.20.x", "1.21.x"]
+        go: ["1.21.x", "1.22.x"]
 
     steps:
     - name: Checkout code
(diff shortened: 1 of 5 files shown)


tchung1118 (approved):
LGTM

sywhang (approved):

.github/workflows/go.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     strategy:
       matrix:
         os: ["ubuntu-latest", "windows-latest"]
-        go: ["1.20.x", "1.21.x"]
+        go: ["1.21.x", "1.22.x"]
```
rabbbit, line 22: I recently learned a trick via https://github.com/uber-go/tally/pull/258#discussion_r1563278428

PR #1198, opened 2024-05-08
Call signal.Stop when signalReceivers is stopped

### LINKED ISSUE #1197: signalReceivers should call `Signal.Stop(recv.signals)` when stopped
**Describe the bug**

signalReceivers should call `Signal.Stop(recv.signals)` when they are `Stop()`ed.

Because if not, then when a user does:
```
app.Start(...)
app.Close(...)

// block on anything else
select {} // for example
```

the program will fail to exit when the user presses ctrl-c. Which is contrary to what one would expect after starting and stopping the fx app.

**To Reproduce**

Example
```go
func main() {
	app := fx.New()
	app.Start(context.Background())
	app.Stop(context.Background())

	select {}
}
```

ctrl-c should exit the program

**Expected behavior**
the signal handler reverts back to Go's default signal handler

**Additional context**
should be a ~1 line change :)

MarcoPolo:
fixes #1197


signal_test.go
```diff
@@ -100,6 +101,10 @@ func TestSignal(t *testing.T) {
 						}
 					}()
 				}
+				var stopCalledTimes atomic.Uint32
```
JacobOaks, line 104: Do we need this to be atomic? From my read, `Stop` shouldn't return until `stopNotify` completes, so I don't think there should be any concurrent access.

JacobOaks (commented):
Thanks for the PR! I think this makes sense to me. Just one small question.

signal_test.go
MarcoPolo: You're right. Does not need to be atomic.

JacobOaks (approved):

PR #1201, opened 2024-05-14
fx.Self: a parameter to fx.As for providing a type as itself

JacobOaks:
We frequently see requests for folks who want to use `fx.As` to provide a type as another type, while also providing it as itself. To name a few:
* https://github.com/uber-go/fx/discussions/1196
* https://github.com/uber-go/fx/discussions/1148
* https://github.com/uber-go/fx/discussions/1079

This is currently not possible via strictly using `fx.As` + a single constructor, since `fx.As` causes a constructor to no longer provide its original type. The workaround we often give is for folks to do something like this, which involves adding a second "constructor":
```go
fx.Provide(
    newConcreteType,
    func(ct *concreteType) Interface {
        return ct
    }
)
```
which is admittedly not very ergonomic. A somewhat common pattern mistaken to be a workaround is providing the constructor twice instead:
```go
fx.Provide(
    newConcreteType,
    fx.Annotate(newConcreteType, fx.As(new(Interface))),
)
```

This PR adds `fx.Self()`, which returns a special value to indicate an `fx.As` should retain that original return type:
```go
fx.Provide(
    newConcreteType,
    fx.As(fx.Self()),
    fx.As(new(Interface)),
)
```
As an alternative, I considered a new annotation altogether, named something like `fx.AlsoAs`, but adding a special type that can be passed as an argument to `fx.As` directly allows for more fine-tuned control over individual positional return values.

For example, this function's return types can be easily expressed as `*asStringer` and `io.Writer` using `fx.Self()`:
```go
fx.Provide(
	fx.Annotate(
		func() (*asStringer, *bytes.Buffer) {/* ... */ },
		fx.As(fx.Self(), new(io.Writer)), // return values will be: *asStringer, io.Writer
	),
),
```
Whereas something like `fx.AlsoAs` wouldn't provide the ability to skip over the first positional return value entirely.


annotated.go
```diff
@@ -1475,7 +1526,8 @@ type annotated struct {
 	Annotations []Annotation
 	ParamTags   []string
 	ResultTags  []string
-	As          [][]reflect.Type
+	As          [][]asType
+	AsSelf      bool
```
tchung1118, line 1530: Is this field needed? I don't see this used anywhere in this PR.

annotated_test.go
```diff
@@ -699,6 +705,78 @@ func TestAnnotatedAs(t *testing.T) {
...
+		{
+			desc: "self as one As target",
+			provide: fx.Provide(
+				fx.Annotate(
+					func() (*asStringer, *bytes.Buffer) {
+						s := &asStringer{name: "stringer"}
+						b := &bytes.Buffer{}
+						return s, b
+					},
+					fx.As(fx.Self(), new(io.Writer)),
```
tchung1118, line 733: I'm sure this still works, and there's already a test case that covers this, but could we add another case with exactly this and the following annotation added:
```
fx.As(fmt.Stringer)
```

This should provide four types in total:
* *asStringer
* io.Writer
* fmt.Stringer
* *bytes.Buffer

tchung1118 (commented):
Changes look great and pretty isolated. Added a couple of comments.

annotated.go
JacobOaks: Thanks for catching that!

JacobOaks:
> Changes look great and pretty isolated. Added a couple of comments.

Addressed!

tchung1118 (approved):
LGTM

Issue #1180, opened 2024-03-20
fxtest.Lifecycle: Enforce timeout


abhinav:
**Is your feature request related to a problem? Please describe.**

fx.App.Start and fx.App.Stop enforce the context timeout on start/stop hooks by spawning a goroutine:
https://github.com/uber-go/fx/blob/v1.21.0/app.go#L726-L731

fxtest.LIfecycle is supposed to be a lifecycle implementation meant for use in tests.
However, if one runs `lifecycle.Start` or `lifecycle.Stop` on it with an indefinitely blocking operation, the lifecycle will never resolve.

```go
lc := fxtest.NewLifecycle()
lc.Append(fx.StopHook(func() {
  select {}
}))

ctx, cancel := context.WithTimeout(context.Background(), time.Second)
err := lc.Stop(ctx) // blocks forever
```

This does not match the behavior of `fx.App`, where the lifecycle will abort early and return an error.

**Describe the solution you'd like**

fxtest.Lifecycle should similarly enforce the context timeout on the lifecycle operations.

**Describe alternatives you've considered**

Spawn a goroutine from Start/Stop operation to manually enforce the timeout.
This feels unnecessary/incorrect because you don't need to do it for hooks when they're called from App.Start/App.Stop, but do need to do it for Lifecycle.Start/Stop.

**Is this a breaking change?**
This is not a breaking change to the API.
However, there's a possibility that someone was relying on the lifecycle to block previously, and their tests could begin failing.

One argument is that if their tests were violating the timeout anyway, that's likely a bug.
However, if we don't feel comfortable making this change in that way, we can instead use an option or a different constructor to turn this behavior on/off.

PR #1203, opened 2024-05-21
fxtest: Enforceable Timeouts on Lifecycle

JacobOaks:
In Fx, if starting/stopping the application takes longer than the context's timeout, the fx.App will bail early, regardless of the status of the currently running hooks.

This prevents stalling an application when hooks (accidentally) block forever.

In order to test hook behavior, Fx provides fxtest.Lifecycle to interact with. This Lifecycle is a simple wrapper around the actual fx Lifecycle type, meaning it does not check for timeout and bail early like fx.App does. This is an issue because:
* It allows for long-running hooks (which should be considered bugs) to pass tests.
* It allows for tests to completely stall for hooks that block forever.

See #1180 for more details.

This PR adds an option that can be passed to `fxtest.NewLifecycle` to cause it to immediately fail when context expires, similar to fx.App.

```go
lc := fxtest.NewLifecycle(fxtest.EnforceTimeout(true))
lc.Append(fx.StartHook(func() { for {} }))
ctx, _ := context.WithTimeout(context.Background(), time.Second)
err := lc.Start(ctx) // Will return deadline exceeded after a second
```

This PR doesn't provide a way to test timeouts using `RequireStart` and `RequireStop`. However, since those don't take contexts anyways, my assumption is that usage of those APIs represents an intentional decision to not care about timeouts by the test writer.

However, if people feel differently, we can instead do something like expose two new APIs `RequireStartWithContext(ctx)` and `RequireStopWithContext(ctx)` or something (names obviously up for discussion).

### DIFF
--- a/fxtest/lifecycle.go
+++ b/fxtest/lifecycle.go
@@ -127,7 +127,12 @@ func (l *Lifecycle) withTimeout(ctx context.Context, fn func(context.Context) er
 		return fn(ctx)
 	}
 
-	c := make(chan error, 1)
+	// Cancel on timeout in case function only respects
+	// cancellation and not deadline exceeded.
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	c := make(chan error, 1) // buffered to avoid goroutine leak
 	go func() {
 		c <- fn(ctx)
 	}()
--- a/fxtest/lifecycle.go
+++ b/fxtest/lifecycle.go
@@ -66,18 +66,45 @@ func (t *panicT) FailNow() {
 	panic("test lifecycle failed")
 }
 
+// LifecycleOption modifies the behavior of the [Lifecycle]
+// when passed to [NewLifecycle].
+type LifecycleOption interface {
+	apply(*Lifecycle)
+}
+
+// EnforceTimeout will cause the [Lifecycle]'s Start and Stop methods
+// to return an error as soon as context expires,
+// regardless of whether specific hooks respect the timeout.
+func EnforceTimeout(enforce bool) LifecycleOption {
+	return &enforceTimeout{
+		enforce: enforce,
+	}
+}
+
+type enforceTimeout struct {
+	enforce bool
+}
+
+func (e *enforceTimeout) apply(lc *Lifecycle) {
+	lc.enforceTimeout = e.enforce
+}
+
+var _ LifecycleOption = (*enforceTimeout)(nil)
+
 // Lifecycle is a testing spy for fx.Lifecycle. It exposes Start and Stop
 // methods (and some test-specific helpers) so that unit tests can exercise
 // hooks.
 type Lifecycle struct {
 	t  TB
 	lc *lifecycle.Lifecycle
+
+	enforceTimeout bool
 }
 
 var _ fx.Lifecycle = (*Lifecycle)(nil)
 
 // NewLifecycle creates a new test lifecycle.
-func NewLifecycle(t TB) *Lifecycle {
+func NewLifecycle(t TB, opts ...LifecycleOption) *Lifecycle {
 	var w io.Writer
 	if t != nil {
 		w = testutil.WriteSyncer{T: t}
@@ -85,15 +112,40 @@ func NewLifecycle(t TB) *Lifecycle {
 		w = os.Stderr
 		t = &panicT{W: os.Stderr}
 	}
-	return &Lifecycle{
+	lc := &Lifecycle{
 		lc: lifecycle.New(fxlog.DefaultLogger(w), fxclock.System),
 		t:  t,
 	}
+	for _, opt := range opts {
+		opt.apply(lc)
+	}
+	return lc
+}
+
+func (l *Lifecycle) withTimeout(ctx context.Context, fn func(context.Context) error) error {
+	if !l.enforceTimeout {
+		return fn(ctx)
+	}
+
+	c := make(chan error, 1)
+	go func() {
+		c <- fn(ctx)
+	}()
+
+	var err error
+	select {
+	case err = <-c:
+	case <-ctx.Done():
+		err = ctx.Err()
+	}
+	return err
 }
 
 // Start executes all registered OnStart hooks in order, halting at the first
 // hook that doesn't succeed.
-func (l *Lifecycle) Start(ctx context.Context) error { return l.lc.Start(ctx) }
+func (l *Lifecycle) Start(ctx context.Context) error {
+	return l.withTimeout(ctx, l.lc.Start)
+}
 
 // RequireStart calls Start with context.Background(), failing the test if an
 // error is encountered.
@@ -114,7 +166,9 @@ func (l *Lifecycle) RequireStart() *Lifecycle {
 // If any hook returns an error, execution continues for a best-effort
 // cleanup. Any errors encountered are collected into a single error and
 // returned.
-func (l *Lifecycle) Stop(ctx context.Context) error { return l.lc.Stop(ctx) }
+func (l *Lifecycle) Stop(ctx context.Context) error {
+	return l.withTimeout(ctx, l.lc.Stop)
+}
 
 // RequireStop calls Stop with context.Background(), failing the test if an error
 // is encountered.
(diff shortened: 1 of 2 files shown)


fxtest/lifecycle.go
```diff
@@ -66,34 +66,86 @@ func (t *panicT) FailNow() {
...
 	lc *lifecycle.Lifecycle
+
+	enforceTimeout bool
 }
 
 var _ fx.Lifecycle = (*Lifecycle)(nil)
 
 // NewLifecycle creates a new test lifecycle.
-func NewLifecycle(t TB) *Lifecycle {
+func NewLifecycle(t TB, opts ...LifecycleOption) *Lifecycle {
```
abhinav, line 107: Strictly speaking, this is considered a breaking change per Go compatibility guarantees.
Adding variadic options to a function that didn't previously accept them still changes the signature.
It breaks cases like passing a reference to the function, for example.

I'll defer to y'all about whether you'd like to take the risk, since passing `fastest.NewLifecycle` would be a pretty niche use.

Alternatively, we can add one of `NewLifecycleWith`, `NewLifecycleWithOptions`, or `NewStrictLifecycle` instead. (The last defaulting to enforceTimeout = true. In that case, we could drop the EnforceTimeout option completely, and instead just add `Option`s for future expansion.)

fxtest/lifecycle_test.go
```diff
@@ -117,6 +118,108 @@ func TestLifecycle(t *testing.T) {
...
+func TestEnforceTimeout(t *testing.T) {
+	// These tests directly call Start and Stop
+	// rather than RequireStart and RequireStop
+	// because EnforceTimeout does not apply to those.
+
+	t.Run("StartHookTimeout", func(t *testing.T) {
+		t.Parallel()
+
+		wait := make(chan struct{}, 1)
+		defer close(wait)
```
abhinav, line 130: This doesn't need to be buffered. It never gets any values.
Also worth adding a comment that this exists to ensure that OnStart blocks until the end of this test.
```suggestion
		wait := make(chan struct{})
		defer close(wait) // block OnStart until end of test
```

fxtest/lifecycle_test.go
```diff
@@ -117,6 +118,108 @@ func TestLifecycle(t *testing.T) {
...
+		spy := newTB()
+		lc := NewLifecycle(spy, EnforceTimeout(true))
+		lc.Append(fx.Hook{
+			OnStart: func(context.Context) error {
+				<-wait
+				return nil
+			},
+			OnStop: func(context.Context) error {
+				return nil
+			},
```
abhinav, line 141: Can omit for cases where it's a no-op unless you're specifically testing for non-nil function.
```suggestion
```

fxtest/lifecycle_test.go
```diff
@@ -117,6 +118,108 @@ func TestLifecycle(t *testing.T) {
...
+		})
+
+		ctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
```
abhinav, line 144: nit: maybe even smaller timeouts to reduce test sleeps
```suggestion
		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Millisecond)
```

fxtest/lifecycle_test.go
```diff
@@ -117,6 +118,108 @@ func TestLifecycle(t *testing.T) {
...
+				return nil
+			},
+			OnStop: func(context.Context) error {
+				<-wait
+				return nil
+			},
+		})
+
+		require.NoError(t, lc.Start(context.Background()))
+		ctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
```
abhinav, line 168:
```suggestion
		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Millisecond)
```

fxtest/lifecycle_test.go
```diff
@@ -117,6 +118,108 @@ func TestLifecycle(t *testing.T) {
...
+	// rather than RequireStart and RequireStop
+	// because EnforceTimeout does not apply to those.
+
+	t.Run("StartHookTimeout", func(t *testing.T) {
+		t.Parallel()
+
+		wait := make(chan struct{}, 1)
+		defer close(wait)
+
+		spy := newTB()
```
abhinav, line 132: Should we verify that the test was not marked as failed?

```suggestion
		spy := newTB()
		defer func() { assert.Zero(t, spy.failures) }()
```

abhinav (commented):
Agree that it's a bit unexpected that RequireStart and RequireStop don't enforce the timeout, but I also understand, and agree with why—where would the timeout come from? We can expand those APIs in a backwards compatible manner in the future if necessary (e.g. with an option to control their timeouts, or the WithContext method variants which may be more obvious, if a bit verbose).

fxtest/lifecycle.go
```diff
@@ -66,34 +66,86 @@ func (t *panicT) FailNow() {
...
+	}
+	return lc
+}
+
+func (l *Lifecycle) withTimeout(ctx context.Context, fn func(context.Context) error) error {
+	if !l.enforceTimeout {
+		return fn(ctx)
+	}
+
+	c := make(chan error, 1)
```
abhinav, line 130: Worth leaving a quick comment here that this is buffered so as not to leak the goroutine if the test does time out.
```suggestion
	c := make(chan error, 1) // buffered to avoid a goroutine leak
```

fxtest/lifecycle.go
```diff
@@ -66,34 +66,86 @@ func (t *panicT) FailNow() {
...
+	go func() {
+		c <- fn(ctx)
+	}()
```
abhinav, line 133: One last thing: Just for the sake of easier/faster cleanup, we should wrap the context with a WithCancel here so that when this function returns because of a timeout, the context that `fn(ctx)` is operating on is also closed, and the function returns early (if possible).
```suggestion
	ctx, cancel := context.WithCancel(ctx)
	defer cancel()
	go func() {
		c <- fn(ctx)
	}()
```

abhinav (approved):
lgtm with one minor behavioral improvement suggested

fxtest/lifecycle.go
JacobOaks: Hmm, shouldn't the context that `fn` is operating on already be closed if its deadline is exceeded? Or are you referring to a case where `fn` is only returning early when [canceled](https://pkg.go.dev/context#Canceled) and not [exceeded deadline](https://pkg.go.dev/context#DeadlineExceeded)? I can add this to cover this case.

JacobOaks:
> lgtm with one minor behavioral improvement suggested

Thank you for all your helpful review @abhinav!

fxtest/lifecycle.go
abhinav: I don't feel strongly about this; will defer to other maintainers.
I think this is fine as-is. Passing `fxtest.NewLifecycle` as a function parameter is a niche enough use case  that I expect zero breakage.

This isn't authoritative, but just searching public code on GitHub, there are no references to `fxtest.NewLifecycle` not followed by a `(`, so they're always invocations, and won't be broken.
https://github.com/search?q=%2Ffxtest%5C.NewLifecycle%5B%5E%5C%28%5D%2F+path%3A%2F.go%24%2F&type=code

fxtest/lifecycle.go
abhinav: 🤔 You're right. My mistake.
The only way this function returns early is when the context has expired.

fxtest/lifecycle.go
JacobOaks: 👍 @sywhang or @tchung1118 - do you have strong opinions on this? Otherwise I will merge this as-is.

tchung1118: I think this is fine as-is, too.

sywhang: I don't have a strong opinion against this either but I do prefer making a slightly breaking change in this API rather than overloading the API surface with a variant of `NewLifecycle` given how niche the set of use cases that break from this.

sywhang (approved):
👍 :shipit:

Issue #1206, opened 2024-05-27
Private supplies


saturn4er:
**Is your feature request related to a problem? Please describe.**
fx.Private doesn't work with fx.Supply. It just supplies value of option to DI container. It would be nice to be able to use fx.Private in fx.Supply function as it now works with fx.Provide function

**Is this a breaking change?**
No, it's not

JacobOaks:
Hey @saturn4er! This should be supported now with #1207 being merged. Will try to do a release today.

PR #1207, opened 2024-05-29
Support fx.Private w/ fx.Supply

JacobOaks:
`fx.Supply` is essentially an API that allows for conveniently `fx.Provide`ing an exact value, rather than a function that will return that value. For example, `fx.Provide(func() int { return 5 })` is equivalent to `fx.Supply(5)`.

`fx.Private` allows for usage of a provided constructor's results to be restricted to the current module and its child modules.
```go
fx.Module(
	"parent",
	fx.Invoke(func(int) { /* this will error out! */ }),
	fx.Module(
		"child",
		fx.Provide(func() int { return 5 }, fx.Private),
	),
),
```

This PR allows for using `fx.Private` with `fx.Supply` as well, so that folks can enjoy the convenience of `fx.Supply` when they also wish to restrict the usage of the supplied value.
```go
fx.Module(
	"parent"
	fx.Invoke(func(int) { /* this will error out! */ }),
	fx.Module(
		"child",
		fx.Supply(5, fx.Private),
	),
),
```

Ref #1206

Since the behavior between Supply + Private and Provide + Private should be identical, I opted to generalize the existing `fx.Private` tests to run for both Provide and Supply. This keeps the tests a little more DRY but does complicate them/hurt readability. I feel like this is OK since there are a lot of tests, but I also am the one who wrote the tests, so I am biased regarding its readability. Thus, I am happy to break out Supply + Private into its own tests if folks feel strongly that these tests are hard to read.

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -391,134 +391,227 @@ func TestNewApp(t *testing.T) {
 	})
 }
 
-func TestPrivateProvide(t *testing.T) {
+// TestPrivate tests Private when used with both fx.Provide and fx.Supply.
+func TestPrivate(t *testing.T) {
 	t.Parallel()
 
-	t.Run("CanUsePrivateAndNonPrivateFromOuterModule", func(t *testing.T) {
-		t.Parallel()
+	testCases := []struct {
+		desc string
 
-		app := fxtest.New(t,
-			Module("SubModule", Invoke(func(a int, b string) {})),
-			Provide(func() int { return 0 }, Private),
-			Provide(func() string { return "" }),
-		)
-		app.RequireStart().RequireStop()
-	})
+		// provide is either a Supply or Provide wrapper around the given int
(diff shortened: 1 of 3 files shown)


app_test.go
```diff
@@ -391,134 +391,227 @@ func TestNewApp(t *testing.T) {
...
+				assert.Contains(t, err.Error(), "missing type: int")
+			})
+
+			t.Run("MultipleModulesSameType", func(t *testing.T) {
+				t.Parallel()
+
+				var invoked int
+				app := fxtest.New(t,
+					Module("SubModuleA",
+						tt.provide(1, true),
```
sywhang, line 466: naked parameters are hurting the readability. Let's add some comments. 
https://github.com/uber-go/guide/blob/master/style.md#avoid-naked-parameters

```suggestion
						tt.provide(1, true /* private */),
```

sywhang (approved):

PR #1215, opened 2024-06-17
fix: Only register signal handlers if user intends to use them

### LINKED ISSUE #1212: Opt-out of signal handling

A user can't make use of the default behavior of signals in Go programs because we always register [signal handlers](https://github.com/uber-go/fx/blob/master/signal.go#L116).

I would like to use the app lifecycle without it registering a signal handler.

- Add an Option to App that opts-out of registering the signal receiver.

This should be an easy change:
1. Add an option that sets `signalReceivers.{notify,stopNotify}` to no-ops.
2. Add docs around option.

Happy to add once their's consensus.

I use fx as part of my library go-libp2p. An alternative is to expose the `app.Done` in my constructed object and let the user handle the signal explicitly this way. The problem with this approach is that it changes the standard way signal handling works in Go. User's should just concern themselves with `signal.Notify` rather than having to know that `app.Done` has registered a signal handler for them. And if the user hasn't registered with `signal.Notify` they would expect the default behavior from Go.

We do not accept breaking changes to the existing API. Please consider if your proposed solution is backwards compatible. If not, we can help you make it backwards compatible, but this must be considered when we consider new features.

No. It can be provided as an option. The default can remain the same.

**Additional context**

See related issues in go-libp2p:

- https://github.com/libp2p/go-libp2p/issues/2785
- One proposed PR, but I'd favor the ability to opt-out in fx: https://github.com/libp2p/go-libp2p/pull/2829#issuecomment-2161199081

MarcoPolo:
closes #1212, and fixes a regression from #989. Previously we would only register signal handlers if the user intended to use them. #989 changed this behavior [here](https://github.com/uber-go/fx/pull/989/files#diff-6c4b6ed7dc8834cef100f50dae61c30ffe7775a3f3f6f5a557517cb740c44a2dR649).

This regression meant that if you only used app.Start()/app.Stop(), fx would register signal handlers for no reason as the user didn't use app.Done/app.Wait.

### DIFF
--- a/app.go
+++ b/app.go
@@ -704,7 +704,6 @@ func (app *App) start(ctx context.Context) error {
 		if err := app.lifecycle.Start(ctx); err != nil {
 			return err
 		}
-		app.receivers.Start(ctx)
 		return nil
 	})
 }
@@ -742,6 +741,7 @@ func (app *App) Stop(ctx context.Context) (err error) {
 // Alternatively, a signal can be broadcast to all done channels manually by
 // using the Shutdown functionality (see the [Shutdowner] documentation for details).
 func (app *App) Done() <-chan os.Signal {
+	app.receivers.Start() // No-op if running
 	return app.receivers.Done()
 }
 
@@ -752,6 +752,7 @@ func (app *App) Done() <-chan os.Signal {
 // in the [ShutdownSignal] struct.
 // Otherwise, the signal that was received will be set.
 func (app *App) Wait() <-chan ShutdownSignal {
+	app.receivers.Start() // No-op if running
 	return app.receivers.Wait()
 }
 
--- a/app_internal_test.go
+++ b/app_internal_test.go
@@ -136,5 +136,5 @@ func TestStartDoesNotRegisterSignals(t *testing.T) {
 	assert.False(t, calledNotify, "notify should not be called when app starts")
 
 	_ = app.Wait() // User signals intent have fx listen for signals. This should call notify
-	assert.True(t, calledNotify, "notify should not be called when app starts")
+	assert.True(t, calledNotify, "notify should be called after Wait")
 }
--- a/app_internal_test.go
+++ b/app_internal_test.go
@@ -21,8 +21,10 @@
 package fx
 
 import (
+	"context"
 	"errors"
 	"fmt"
+	"os"
 	"sync"
 	"testing"
 
@@ -114,4 +116,25 @@ func TestAnnotationError(t *testing.T) {
 	require.Error(t, err)
 	assert.ErrorIs(t, err, wantErr)
 	assert.Contains(t, err.Error(), wantErr.Error())
+}
+
+// TestStartDoesNotRegisterSignals verifies that signal.Notify is not called
+// when a user starts an app. signal.Notify should only be called when the
+// .Wait/.Done are called. Note that app.Run calls .Wait() implicitly.
+func TestStartDoesNotRegisterSignals(t *testing.T) {
+	app := New()
+	calledNotify := false
+
+	// Mock notify function to spy when this is called.
+	app.receivers.notify = func(c chan<- os.Signal, sig ...os.Signal) {
+		calledNotify = true
+	}
+	app.receivers.stopNotify = func(c chan<- os.Signal) {}
+
+	app.Start(context.Background())
+	defer app.Stop(context.Background())
+	assert.False(t, calledNotify, "notify should not be called when app starts")
+
+	_ = app.Wait() // User signals intent have fx listen for signals. This should call notify
+	assert.True(t, calledNotify, "notify should not be called when app starts")
 }
--- a/app_test.go
+++ b/app_test.go
@@ -2331,13 +2331,37 @@ func TestHookConstructors(t *testing.T) {
 func TestDone(t *testing.T) {
 	t.Parallel()
 
-	done := fxtest.New(t).Done()
+	app := fxtest.New(t)
+	defer app.RequireStop()
+	done := app.Done()
 	require.NotNil(t, done, "Got a nil channel.")
 	select {
 	case sig := <-done:
 		t.Fatalf("Got unexpected signal %v from application's Done channel.", sig)
 	default:
 	}
+}
+
+// TestShutdownThenWait tests that if we call .Shutdown before waiting, the wait
+// will still return the last shutdown signal.
+func TestShutdownThenWait(t *testing.T) {
+	t.Parallel()
+
+	var (
+		s Shutdowner
+	)
+	app := fxtest.New(
+		t,
+		Populate(&s),
+	)
+	defer app.RequireStop()
+	assert.NotNil(t, s)
+
+	err := s.Shutdown(ExitCode(1337))
+	require.NoError(t, err)
+
+	shutdownSig := <-app.Wait()
+	require.Equal(t, 1337, shutdownSig.ExitCode)
 }
 
 func TestReplaceLogger(t *testing.T) {
--- a/app_test.go
+++ b/app_test.go
@@ -2348,20 +2348,30 @@ func TestShutdownThenWait(t *testing.T) {
 	t.Parallel()
 
 	var (
-		s Shutdowner
+		s       Shutdowner
+		stopped bool
 	)
 	app := fxtest.New(
 		t,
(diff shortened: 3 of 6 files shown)


app_internal_test.go
```diff
@@ -115,3 +117,37 @@ func TestAnnotationError(t *testing.T) {
...
+
+			if disableSignalHandler {
+				opt := DisableSignalHandler()
+				opt.apply(app.root)
+			}
+
+			app.Start(context.Background())
+			app.Stop(context.Background())
+
+			if disableSignalHandler {
```
ccoVeille, line 145: assert.Equal(t, disableSignalHandler, !calledNotify)

No if/else needed, no ?

app.go
```diff
@@ -256,6 +256,23 @@ func (l loggerOption) String() string {
 	return fmt.Sprintf("fx.Logger(%v)", l.p)
 }
 
+func DisableSignalHandler() Option {
```
ccoVeille, line 259: If it would be used only for test purpose, I would expect it to have a *testing.T as argument.

This would prevent anyone to use it for non test purpose

app_internal_test.go
```diff
@@ -115,3 +117,37 @@ func TestAnnotationError(t *testing.T) {
 	assert.ErrorIs(t, err, wantErr)
 	assert.Contains(t, err.Error(), wantErr.Error())
 }
+
+// TestDisableSignalHandler tests disabling the default signal handler.
+// It's an internal test to allow us replace the app.receivers.notify function.
+func TestDisableSignalHandler(t *testing.T) {
+	for _, disableSignalHandler := range []bool{false, true} {
```
ccoVeille, line 124: Use a map[string]bool

So you can set `testname` here in the `for` loop, then no if needed

app_internal_test.go
```diff
@@ -115,3 +117,37 @@ func TestAnnotationError(t *testing.T) {
...
+		testName := "Not disabling signal handler"
+		if disableSignalHandler {
+			testName = "Disabling signal handler"
+		}
+		t.Run(testName, func(t *testing.T) {
+			app := New()
+			calledNotify := false
+
+			app.receivers.notify = func(c chan<- os.Signal, sig ...os.Signal) {
```
ccoVeille, line 133: You could have passed a func to func DisableSignalHandler() Option {

So it would be something like that

```go
func DisableSignalHandler(callback func()) Option {
```

So here you would use

```go
var notifyCalled bool
DisableSignalHandler(func () {notifiyCalled = true}
```

Because right now, you have not testing the method you added.

ccoVeille (changes requested):

app_internal_test.go
MarcoPolo: yes, I thought this was clearer. I have no real preference here. Happy to change if maintainers would prefer.

app.go
MarcoPolo: It's generally useful. I want to use it for normal operation (not tests). See attached issue.

app_internal_test.go
MarcoPolo: No preference. Happy to change if maintainers prefer.

app_internal_test.go
MarcoPolo: Maybe there's a misunderstanding? `DisableSignalHandler()` disables fx's default use of `signal.Notify` (via `app.receivers.notify`). This tests that `DisableSignalHandler` effectively prevents fx from registering with `signal.Notify`.

A callback is not required because I'm disabling the notify registration. If I wanted to provide my own notify registration method as an option I would do as you say. However, I don't have a use case for a general purpose option. I think the specific option is a nicer api.

ccoVeille: OK then 👍

app_test.go
```diff
@@ -2331,7 +2331,9 @@ func TestHookConstructors(t *testing.T) {
 func TestDone(t *testing.T) {
 	t.Parallel()
 
-	done := fxtest.New(t).Done()
+	app := fxtest.New(t)
+	defer app.RequireStop()
```
ccoVeille, line 2335: In test t.Cleanup is preferred over defer

signal.go
```diff
@@ -102,7 +102,7 @@ func (recv *signalReceivers) running() bool {
 	return recv.shutdown != nil && recv.finished != nil
 }
 
-func (recv *signalReceivers) Start(ctx context.Context) {
+func (recv *signalReceivers) Start() {
```
ccoVeille, line 105: So the context was not used ?

app_test.go
```diff
@@ -2340,6 +2342,28 @@ func TestDone(t *testing.T) {
...
+		t,
+		Populate(&s),
+	)
+	defer app.RequireStop()
+	assert.NotNil(t, s)
+
+	err := s.Shutdown(ExitCode(1337))
+	require.NoError(t, err)
+
+	shutdownSig := <-app.Wait()
```
ccoVeille, line 2363: Because of GitHub workflow, I cannot comment the Wait method, so I'm commenting here.

I would expect to use the context as argument

This way `Wait` could stop when ctx is done.
The same way  `Stop` does.

It's useless in the unit tests, but meaningful in the real app

ccoVeille: I'm unsure but I would expect the same from `app.Done`

ccoVeille: Last remark, I'm surprised that code in Stop are not closing the channels when leaving on ctx.Done()

ccoVeille (commented):
I found a few things that surprised me. I might be wrong. But I prefer to raise the points

app_internal_test.go
```diff
@@ -115,3 +117,24 @@ func TestAnnotationError(t *testing.T) {
...
+	calledNotify := false
+
+	// Mock notify function to spy when this is called.
+	app.receivers.notify = func(c chan<- os.Signal, sig ...os.Signal) {
+		calledNotify = true
+	}
+	app.receivers.stopNotify = func(c chan<- os.Signal) {}
+
+	app.Start(context.Background())
+	defer app.Stop(context.Background())
```
ccoVeille, line 135: You are not testing it then I think.

When defer is called it's too late, no?

app_internal_test.go
```diff
@@ -115,3 +117,24 @@ func TestAnnotationError(t *testing.T) {
...
+// .Wait/.Done are called. Note that app.Run calls .Wait() implicitly.
+func TestStartDoesNotRegisterSignals(t *testing.T) {
+	app := New()
+	calledNotify := false
+
+	// Mock notify function to spy when this is called.
+	app.receivers.notify = func(c chan<- os.Signal, sig ...os.Signal) {
+		calledNotify = true
+	}
+	app.receivers.stopNotify = func(c chan<- os.Signal) {}
```
ccoVeille, line 132: You should define a variable `stopCalled bool` and set it to true when called.

And  then test its value after calling `Stop`

app_internal_test.go
ccoVeille: The Stop should be after the wait.

Or you use a t.Cleanup(func() {
app.Stop()
require.True(t, stopCalled)
})

JacobOaks:
Thanks for updating the PR @MarcoPolo. Mostly LGTM, just a couple small comments.

ccoVeille:
@MarcoPolo or @JacobOaks tell me if the review I made, makes no sense :sweat_smile:

app_internal_test.go
MarcoPolo: Correct. The only reason I'm setting `.stopNotify` is because I consider it bad form to call the default `signal.Stop` func with a channel that was never registered in the first place.

app_test.go
```diff
@@ -2340,6 +2342,28 @@ func TestDone(t *testing.T) {
...
 }
 
+// TestShutdownThenWait tests that if we call .Shutdown before waiting, the wait
+// will still return the last shutdown signal.
+func TestShutdownThenWait(t *testing.T) {
+	t.Parallel()
+
+	var (
+		s Shutdowner
+	)
```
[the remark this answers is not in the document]
MarcoPolo: Lazy copy/pasting on my part. Fixed.

JacobOaks (approved):

app_internal_test.go
ccoVeille: OK noted

Issue #1219, opened 2024-06-26
Deadlock in `signalReceivers.Stop`


ogaca-dd:
**Describe the bug**
There is a deadlock in [signalReceivers.Stop](https://github.com/uber-go/fx/blob/v1.22.0/signal.go#L120):
- `syscall.SIGINT` is sent to the process
- `Shutdowner.Shutdown()` is called to stop the application 
- [signalReceivers.Stop](https://github.com/uber-go/fx/blob/v1.22.0/signal.go#L120) is called and takes the lock [m](https://github.com/uber-go/fx/blob/v1.22.0/signal.go#L121)
- [signalReceivers.Stop](https://github.com/uber-go/fx/blob/v1.22.0/signal.go#L120) sends a shutdown request ([recv.shutdown <- struct{}{}](https://github.com/uber-go/fx/blob/v1.22.0/signal.go#L130)) and blocks on the [select](https://github.com/uber-go/fx/blob/v1.22.0/signal.go#L132)
- In [signalReceivers.relayer](https://github.com/uber-go/fx/blob/v1.22.0/signal.go#L84), the branch [case signal := <-recv.signals:](https://github.com/uber-go/fx/blob/v1.22.0/signal.go#L92) is executed before [case <-recv.shutdown](https://github.com/uber-go/fx/blob/v1.22.0/signal.go#L90)
- The branch [case signal := <-recv.signals:](https://github.com/uber-go/fx/blob/v1.22.0/signal.go#L92) calls [signalReceivers.Broadcast](https://github.com/uber-go/fx/blob/v1.22.0/signal.go#L177)
- [signalReceivers.Broadcast](https://github.com/uber-go/fx/blob/v1.22.0/signal.go#L177) takes the same lock [`m`](https://github.com/uber-go/fx/blob/v1.22.0/signal.go#L178) as [signalReceivers.Stop](https://github.com/uber-go/fx/blob/v1.22.0/signal.go#L121)

There is a deadlock.

**To Reproduce**
Run the following code:
```go
func main() {
	fx.New(fx.Invoke(func(shutdowner fx.Shutdowner) {
		go func() {
			_ = syscall.Kill(syscall.Getpid(), syscall.SIGINT)
			_ = shutdowner.Shutdown()
		}()
	})).Run()
}
```

**Expected behavior**
No deadlock

JacobOaks:
Hey @ogaca-dd thanks for the report, let me look into this.

[...]

PR #1220, opened 2024-06-26
Fix deadlock caused by race while signal receivers are stopping

JacobOaks:
A user reported a possible deadlock within the signal receivers (#1219).

This happens by:
* `(*signalReceivers).Stop()` is called, by Shutdowner for instance.
* `(*signalReceivers).Stop()` [acquires the lock](https://github.com/uber-go/fx/blob/master/signal.go#L121).
* Separately, an OS signal is sent to the program.
* There is a chance that `relayer()` is still running at this point if `(*signalReceivers).Stop()` has not yet sent along the `shutdown` channel.
* The relayer [attempts to broadcast the signal](https://github.com/uber-go/fx/blob/master/signal.go#L93) received via the `signals` channel.
* `Broadcast()` blocks on [trying to acquire the lock](https://github.com/uber-go/fx/blob/master/signal.go#L178).
* `(*signalReceivers).Stop()` blocks on [waiting for the `relayer()` to finish](https://github.com/uber-go/fx/blob/master/signal.go#L132) by blocking on the `finished` channel.
* Deadlock.

Luckily, this is not a hard deadlock, as `Stop` will return if the context times out, but we should still fix it.

This PR fixes this deadlock. The idea behind how it does it is based on the observation that the broadcasting logic does not necessarily seem to need to share a mutex with the rest of `signalReceivers`. Specifically, it seems like we can separate protection around the registered `wait` and `done` channels, `last`, and the rest of the fields, since the references to those fields are easily isolated. To avoid overcomplicating `signalReceivers` with multiple locks for different uses, this PR creates a separate `broadcaster` type in charge of keeping track of and broadcasting to `Wait` and `Done` channels. Most of the implementation of `broadcaster` is simply moved over from `signalReceivers`.

Having a separate broadcaster type seems actually quite natural, so I opted for this to fix the deadlock. Absolutely open to feedback or taking other routes if folks have thoughts.

Since broadcasting is protected separately, this deadlock no longer happens since `relayer()` is free to finish its broadcast and then exit.

In addition to running the example provided in the original post to verify, I added a test and ran it before/after this change.

Before:
```
$ go test -v -count=10 -run "TestSignal/stop_deadlock" .
=== RUN   TestSignal/stop_deadlock
    signal_test.go:141:
                Error Trace:
/home/user/go/src/github.com/uber-go/fx/signal_test.go:141
                Error:          Received unexpected error:
                                context deadline exceeded
                Test:           TestSignal/stop_deadlock
```
(the failure appeared roughly 1/3 of the time)

After:
```
$ go test -v -count=100 -run "TestSignal/stop_deadlock" .
--- PASS: TestSignal (0.00s)
    --- PASS: TestSignal/stop_deadlock (0.00s)
```
(no failures appeared)

### DIFF
--- a/broadcast.go
+++ b/broadcast.go
@@ -29,7 +29,11 @@ import (
 // broadcaster broadcasts signals to registered signal listeners.
 // All methods on the broadcaster are concurrency-safe.
 type broadcaster struct {
-	// this protects reads/writes to any of the fields in the struct
+	// This lock is used to protect all fields of broadcaster.
+	// 
+	// Methods on broadcaster should protect all concurrent access
+	// by taking this lock when accessing its fields.
+	// Conversely, this lock should NOT be taken outside of broadcaster.
 	m sync.Mutex
 
 	// last will contain a pointer to the last ShutdownSignal received, or
--- a/broadcast.go
+++ b/broadcast.go
@@ -30,7 +30,6 @@ import (
 // All methods on the broadcaster are concurrency-safe.
 type broadcaster struct {
 	// This lock is used to protect all fields of broadcaster.
-	// 
 	// Methods on broadcaster should protect all concurrent access
 	// by taking this lock when accessing its fields.
 	// Conversely, this lock should NOT be taken outside of broadcaster.
--- a/broadcast.go
+++ b/broadcast.go
@@ -51,7 +51,7 @@ func (b *broadcaster) reset() {
 	b.last = nil
 }
 
-// Wait creates a new channel that will receive signals being broadcast
+// Done creates a new channel that will receive signals being broadcast
 // via the broadcaster.
 //
 // If a signal has been received prior to the call of Done,
--- a/broadcast.go
+++ b/broadcast.go
@@ -1,0 +1,176 @@
+// Copyright (c) 2024 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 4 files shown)


broadcast.go
```diff
@@ -0,0 +1,176 @@
...
+	wait []chan ShutdownSignal
+}
+
+func (b *broadcaster) reset() {
+	b.m.Lock()
+	defer b.m.Unlock()
+	b.last = nil
+}
+
+// Wait creates a new channel that will receive signals being broadcast
```
JacobOaks, line 54:
```suggestion
// Done creates a new channel that will receive signals being broadcast
```

broadcast.go
```diff
@@ -0,0 +1,176 @@
...
+	"fmt"
+	"os"
+	"sync"
+)
+
+// broadcaster broadcasts signals to registered signal listeners.
+// All methods on the broadcaster are concurrency-safe.
+type broadcaster struct {
+	// this protects reads/writes to any of the fields in the struct
+	m sync.Mutex
```
lverma14, line 33: Could you leave a comment that documents under what expectations it is safe to hold this lock.

lverma14:
On behalf of group code review by @sywhang  & @r-hang

broadcast.go
JacobOaks: @sywhang I added some elaboration here about the lock. Let me know if it's not what you're looking for.

sywhang (approved):

r-hang (approved):

PR #1221, opened 2024-06-27
Clarify example "Export boundary functions"

JamesSlocumIH:
The following example is confusing. 
- If `Config` is supposed to be a parameter object, it should be named `Params` since the constructor is called `New(...)` and needs defined `fx.In` within. 
- If it is not meant to be an parameter object, I think passing `Params` into the constructor without defining that in the example is confusing.
  - Maybe we should be passing a `Config` instead? i.e. `New(c Config)`

I changed the example to use a parameter object, because I think using the parameter object makes sense here. I think if its not supposed to use a parameter object, updating the names of `Params` or `Config` to align could help make the example more clear.

### DIFF
--- a/docs/modules.md
+++ b/docs/modules.md
@@ -133,7 +133,9 @@ var Module = fx.Module("server",
   ),
 )
 
-type Config struct {
+type Params struct {
+  fx.In
+
   Addr string `yaml:"addr"`
 }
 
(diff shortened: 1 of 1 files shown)


docs/modules.md
```diff
@@ -133,7 +133,9 @@ var Module = fx.Module("server",
   ),
 )
 
-type Config struct {
+type Params struct {
+  fx.In
+
   Addr string `yaml:"addr"`
```
abhinav, line 139: Config isn't intended to be a parameter object. It is being passed into New through a Params struct.
Although I agree that this example is a tad confusing: Params and parseConfig aren't visible, and we just assume that a reader knows what they are.

The text in this example is generated automatically from ex/modules/module.go; it may require some adjustment in the snippets there to update this.

PR #1225, opened 2024-07-10
Support multiple `ResultTags` annotations

### LINKED ISSUE #1210: Support multiple ResultTags annotations
As multiple  `As` annotations are supported, it would make sense to support multiple ResultTag annotations. 

Currently this code fails with `cannot apply more than one line of ResultTags`

```go
func main() {
	app := fx.New(
		fx.Provide(
			fx.Annotate(
				func() *bytes.Buffer {
					return bytes.NewBuffer([]byte("Hello!"))
				},
				fx.ResultTags(`name:"a"`),
				fx.ResultTags(`name:"b"`),
			),
		),
		fx.Invoke(
			fx.Annotate(
				func(a, b *bytes.Buffer) {
					fmt.Println(a.String())
					fmt.Println(b.String())
				},
				fx.ParamTags(
					`name:"a"`,
					`name:"b"`,
				),
			),
		),
	)
	app.Run()
}
```

A work-around is possible by providing multiple copies of the same instance of a type, however it is less intuitive. 

```go
fx.Provide(
    fx.Annotate(
        func() (*bytes.Buffer, *bytes.Buffer) {
            a := bytes.NewBuffer([]byte("Hello!"))
            b := a
            return a, b
        },
        fx.ResultTags(`name:"a"`, `name:"b"`),
    ),
)
```

It should not be a breaking change as nothing (sans tests) should rely on app failure for multiple result tag annotations.

miyamo2:
Closes #1210

This commit makes it possible to specify multiple `fx.ResultTags` annotations in a single `fx.Annotate` call.

```go
fx.Provide(
    fx.Annotate(
        func() *bytes.Buffer {
            return bytes.NewBuffer([]byte("Hello!"))
        },
        fx.ResultTags(`name:"a"`),
        fx.ResultTags(`name:"b"`),
        fx.ResultTags(`name:"c"`),
    ),
)
```

### DIFF
--- a/annotated.go
+++ b/annotated.go
@@ -377,15 +377,12 @@ var _ Annotation = resultTagsAnnotation{}
 // If the tag is invalid and has mismatched quotation for example,
 // (`tag_name:"tag_value') , this will return an error.
 func (rt resultTagsAnnotation) apply(ann *annotated) error {
-	if len(ann.ResultTags) > 0 {
-		return errors.New("cannot apply more than one line of ResultTags")
-	}
 	for _, tag := range rt.tags {
 		if err := verifyAnnotateTag(tag); err != nil {
 			return err
 		}
 	}
-	ann.ResultTags = rt.tags
+	ann.ResultTags = append(ann.ResultTags, rt.tags)
 	return nil
 }
 
@@ -426,12 +423,16 @@ func (rt resultTagsAnnotation) results(ann *annotated) (
 	// if there's no Out struct among the return types, there was no As annotation applied
 	// just replace original result types with an Out struct and apply tags
 	var (
-		newOut       outStructInfo
-		existingOuts []reflect.Type
+		newOut              outStructInfo
+		existingOuts        []reflect.Type
+		existingOutsMapping = make(map[int][]reflect.Type)
 	)
 
 	newOut.Fields = []reflect.StructField{_outAnnotationField}
 	newOut.Offsets = []int{}
+	// to prevent duplicate applying of the same tag to the same type, it is kept in the following format
+	// {"foo": {"fmt.Stringer": struct{}{}, "myStringer": struct{}{}}, "bar": {"fmt.Stringer": struct{}{}}}
+	retaggedTypeMap := map[string]map[string]struct{}{}
 
 	for i, t := range types {
 		if !isOut(t) {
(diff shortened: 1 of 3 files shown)


annotated_test.go
```diff
@@ -1440,12 +1440,62 @@ func TestAnnotate(t *testing.T) {
...
 		)
 
 		err := app.Err()
-		require.Error(t, err)
-		assert.Contains(t, err.Error(), "encountered error while applying annotation using fx.Annotate to go.uber.org/fx_test.TestAnnotate.func1(): cannot apply more than one line of ResultTags")
+		require.NoError(t, err)
+		defer app.RequireStart().RequireStop()
+	})
+
+	t.Run("specify two ResultTags containing multiple tags", func(t *testing.T) {
```
tchung1118, line 1454: We should add a test case for using multiple As annotations and ResultTags annotations. For example, something like the following:
```go
fx.Provide(
    fx.Annotate(
        func() *bytes.Buffer { return bytes.NewBuffer() },
        fx.As(new(io.Reader)),
        fx.As(new(io.Writer)),
        fx.ResultTags(`name:"foo"`),
        fx.ResultTags(`name:"bar"`),
    )
)
```
In fact, what's the expected behavior in this case?
I think the current implementation will produce 4 results from the above code, right? But does that make sense? Does "using multiple ResultTags annotation" only make sense when there is 0 or 1 As annotation? If so, we should probably add additional logic to validate that to apply() method of resultTagsAnnotation.

annotated.go
```diff
@@ -426,12 +423,16 @@ func (rt resultTagsAnnotation) results(ann *annotated) (
...
+		newOut              outStructInfo
+		existingOuts        []reflect.Type
+		existingOutsMapping = make(map[int][]reflect.Type)
 	)
 
 	newOut.Fields = []reflect.StructField{_outAnnotationField}
 	newOut.Offsets = []int{}
+	// to prevent duplicate applying of the same tag to the same type, it is kept in the following format
+	// {"foo": {"fmt.Stringer": struct{}{}, "myStringer": struct{}{}}, "bar": {"fmt.Stringer": struct{}{}}}
+	retaggedTypeMap := map[string]map[string]struct{}{}
```
tchung1118, line 435: Can we validate whether or not a given set of ResultTags annotations will provide conflicting types to the container before we try to build out an actual constructor? (in apply() method) If we validate it beforehand, we won't have to worry about this at all here.

tchung1118 (commented):
Thanks for opening up a PR for this. I added a couple of comments to start with.

PR #1229, opened 2024-08-07
Back to development

JacobOaks:
Return changelog and version to development mode.
Do not merge until a release is tagged with #1228.

PR #1233, opened 2024-08-26
Fix changelog duplicate entry

sywhang:
For some reason #1229 added duplicate entry for 1.22.2


JacobOaks:
Can you fix `version.go` too?

sywhang:
fixed

JacobOaks (approved):
Thanks for this, not sure how that happened.

PR #1235, opened 2024-08-27
Upgrade to Go 1.22, 1.23

abhinav:
Go 1.23 is out, so we can change supported versions to 1.22 and 1.23.
Switch CI to test against 1.22 and 1.23 only,
and merge test files that were previously split on Go versions
back together.


JacobOaks (approved):

.github/workflows/go.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     strategy:
       matrix:
         os: ["ubuntu-latest", "windows-latest"]
-        go: ["1.21.x", "1.22.x"]
+        go: ["1.22.x", "1.23.x"]
```
rabbbit, line 22:
```
    strategy:
      matrix:
        go: ["stable", "oldstable"]
```

and then

```
 go-version: ${{ matrix.go }}
```
+
```
if: matrix.latest
```
below

JacobOaks: Hmm... I don't have a big preference. Sometimes the latest version breaks some tests, and sometimes upgrading involves removing some old build-tagged tests (e.g., this PR). So in that vein it makes sense to make upgrading an explicit action/PR. On the other hand, this would reduce manual ops work.

abhinav: stable oldstable is also a valid option here. I'll defer to you on whether you want that or not.

PR #1240, opened 2024-10-09
feat: add runtime to fxevent.Run events

### LINKED ISSUE #1213: Add constructor runtime to fxevent.Run events
Recently I had to investigate a service not starting up properly due to dependency construction taking a long time in certain zone. While we log fxevent.Run events when a constructor runs and returns results and/or error, it lacks the information on how long a certain constructor took to run, which barred us from being able to further investigate and find out which part of the DI graph could be slowing down and come up with some hypotheses on what could be going on in that zone.

I propose that we include the runtime of each constructor in fxevent.Run events.

Collecting cpu profiles at application start-up, including provides, runs, invokes, and OnStart hooks, would also have enabled us to identify which constructors are running slower than others. However, in this particular situation, the docker container that's starting the service up was being terminated in the middle of DI graph construction, so optionally emitting CPU profiles at start-up wouldn't have helped us.

This is just adding more information to an fxevent. I don't think this is a breaking change.

tchung1118:
This PR adds runtime info of constructors and decorators to Run events,
and the runtime of every constructor and decorator (including ones created
by fx.Supply and fx.Replace) will be logged when logging Run events. This
will help us identify which part of the dependency graph takes long to construct.

Closes #1213

### DIFF
--- a/fxevent/console.go
+++ b/fxevent/console.go
@@ -109,7 +109,7 @@ func (l *ConsoleLogger) LogEvent(event Event) {
 		if e.ModuleName != "" {
 			moduleStr = fmt.Sprintf(" from module %q", e.ModuleName)
 		}
-		l.logf("RUN\t%v: %v%v", e.Kind, e.Name, moduleStr)
+		l.logf("RUN\t%v: %v in %s%v", e.Kind, e.Name, e.Runtime, moduleStr)
 		if e.Err != nil {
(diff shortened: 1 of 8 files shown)


JacobOaks (approved):
LGTM

r-hang (approved):
lgtm! Nice testing for ns/ms/s

sywhang (approved):
nice!

PR #1249, opened 2024-11-29
Add support for `func`-types to `fx.As()`

pelmennoteam:


annotated.go
```diff
@@ -1265,8 +1279,11 @@ func (at *asAnnotation) results(ann *annotated) (
 			continue
 		}
 
-		if !t.Implements(at.types[i].typ) {
-			return nil, nil, fmt.Errorf("invalid fx.As: %v does not implement %v", t, at.types[i])
+		if !((at.types[i].typ.Kind() == reflect.Interface && t.Implements(at.types[i].typ)) ||
+			t.ConvertibleTo(at.types[i].typ)) {
+			return nil,
+				nil,
+				fmt.Errorf("invalid fx.As: %v does not implement or is not convertible to %v", t, at.types[i])
```
JacobOaks, line 1286: Should we make these two separate checks? This will keep the code cleaner and give better error messages.

```
if at.types[i].typ.Kind() == reflect.Interface {
    if !t.Implements(at.types[i].typ) {
        return nil, nil, fmt.Errorf("invalid fx.As: %v does not implement %v", t, at.types[i])
    }
} else if !t.ConvertibleTo(at.types[i].typ) {
        return nil, nil, fmt.Errorf("invalid fx.As: %v cannot be converted to %v", t, at.types[i])
}
```

annotated.go
```diff
@@ -1300,7 +1317,11 @@ func (at *asAnnotation) results(ann *annotated) (
 
 		newOutResult := reflect.New(resType).Elem()
 		for i := 1; i < resType.NumField(); i++ {
-			newOutResult.Field(i).Set(getResult(i, results))
+			if newOutResult.Field(i).Kind() == reflect.Func {
+				newOutResult.Field(i).Set(getResult(i, results).Convert(newOutResult.Field(i).Type()))
+			} else {
+				newOutResult.Field(i).Set(getResult(i, results))
+			}
```
JacobOaks, line 1324: nit: should we flip these conditions to make the control flow more/appear consistent with the check above?
```
if newOutResult.Field(i).Kind() == reflect.Interface {
    newOutResult.Field(i).Set(getResult(i, results))
} else {
    newOutResult.Field(i).Set(getResult(i, results).Convert(newOutResult.Field(i).Type()))
}
```

annotated.go
```diff
@@ -1300,7 +1317,11 @@ func (at *asAnnotation) results(ann *annotated) (
 
 		newOutResult := reflect.New(resType).Elem()
 		for i := 1; i < resType.NumField(); i++ {
```
JacobOaks, line 1319: Is there any reason we would need to check `Value.CanConvert` first since technically `Value.Convert` can panic even if `Type.ConvertibleTo` returns true? I can't think of any reason why a func value wouldn't be convertible if its type is though.

annotated_test.go
```diff
@@ -477,6 +480,32 @@ func TestAnnotatedAs(t *testing.T) {
 				assert.Equal(t, s.String(), "another stringer")
 			},
 		},
+		{
+			desc: "value type convertible to target type",
```
JacobOaks, line 484: nit: consider downscoping the names of these cases since convertible types are locked down to just functions right now
```suggestion
			desc: "function value convertible to target type",
```

annotated_test.go
```diff
@@ -477,6 +480,32 @@ func TestAnnotatedAs(t *testing.T) {
...
+						return "provide func example"
+					}
+				}, fx.As(new(myInvokeFunc))),
+			),
+			invoke: func(h myInvokeFunc) {
+				assert.Equal(t, "provide func example", h())
+			},
+		},
+		{
+			desc: "anonymous value type convertible to target type",
```
JacobOaks, line 497: should we add a test case for a function vaoue that is not convertible to the target type?

annotated.go
```diff
@@ -1145,6 +1146,19 @@ var _ Annotation = (*asAnnotation)(nil)
 // constructor does NOT provide both bytes.Buffer and io.Writer type; it just
 // provides io.Writer type.
 //
+// Example for function-types:
```
JacobOaks, line 1149: Let's update line 1129 as well:
```diff
- // constructor) to be provided as another interface.
+ // constructor) to be provided as another type.
```

JacobOaks (commented):
Thanks for the PR! I have a couple comments on the implementation.

At I high level, I can't think of any particular reason why this shouldn't be allowed, but can you maybe provide some insight into why this is desired over, say, just providing a struct with a call-able method defined on it and using `fx.As(new(someInterface))`? Is it just cleaner in some scenarios?

pelmennoteam:
> Thanks for the PR! I have a couple comments on the implementation.
> 
> At a high level, I can't think of any particular reason why this _shouldn't_ be allowed, but can you maybe provide some insight into why this is desired over, say, just providing a struct with a call-able method defined on it and using `fx.As(new(someInterface))`? Is it just cleaner in some scenarios?

gRPC-server implementation:
```
// interfaces/handlers/grpc/server.go

type Handler[R, V] func(ctx context.Context, request *R) (*V, error) 

type Server struct {
  proto.UnimplementedServer

  listHandler Handler[proto.Request, proto.Response]
}

func (s *Server) List(ctx context.Context, request *proto.Request) (*proto.Response, error) {
  return s.listHandler(ctx, request)
}

func ServerRegistrar(listHandler Handler[proto.Request, proto.Response]) *Server {
  return &Server{listHandler: listHandler}
}
```

list handler:
```
// usecases/grpc/list/handler.go

func Handler(deps...) func(ctx context.Context, request *proto.Request) (*proto.Response, error) {
  return func(ctx context.Context, request *proto.Request) (*proto.Response, error) {
    // do some stuff ...
  }
}
```

I need to do this to register the handler:
```
// app/main.go

fx.Provide(fx.Annotate(func(deps...) grpc.Handler[proto.Request, proto.Response] {
  return list.Handler(deps...)
}))
```

What would I like to do:
```
// app/main.go

fx.Provide(fx.Annotate(list.Handler, fx.As(new(grpc.Handler[proto.Request,proto.Response])))),
```

annotated.go
pelmennoteam: Thx, appropriate changes have been made.

annotated_test.go
pelmennoteam: Thx, I added new test case.

annotated.go
pelmennoteam: fixed

annotated.go
pelmennoteam: Thx, I split this condition

pelmennoteam:
@JacobOaks, could you please review changes again?

PR #1254, opened 2025-01-03
doc(App.Run): Clarify exit code

abhinav:
A user expressed confusion about the exit code of an Fx application
that was interrupted with Ctrl-C.

Clarify in the documentation that the exit code will be non-zero
only if there was a failure, or if Shutdown specifically requested
a non-zero exit code.


r-hang (approved):

abhinav:
lint failure unrelated

abhinav:
Fixing lint issue here: https://github.com/uber-go/fx/pull/1255

sywhang (approved):

PR #1258, opened 2025-01-09
refac: DRY up New, reuse *All methods, delete modules slice

abhinav:
The pattern we appear to have for Fx operations (Provide, Invoke, etc.)
take the form:

    func (*module) thing(t thing)

    func (m *module) thingAll() {
        // perform module-local version of operation
        for _, t := range m.things {
            m.thing(t)
        }

        // recurse into children
        for _, m := range m.modules {
            m.${operation}All()
        }
    }

This means that the following two are equivalent:

    // 1
    for _, m := range app.modules {
        m.thingAll()
    }

    // 2
    app.root.thingAll()

Except (2) is DRYer.

This cleans up New by relying on root-level `*All` methods
instead of manually iterating over modules.

Making this change also highlighted that 'app.modules'
only ever has one entry: the root module.
So we can delete that field from App as well.

Finally, this also renames:

    constructAllCustomLoggers -> installAllEventLoggers
    constructCustomLogger     -> installEventLogger
    executeInvokes            -> invokeAll
    executeInvoke             -> invoke

### DIFF
--- a/app.go
+++ b/app.go
@@ -304,7 +304,6 @@ type App struct {
 
 	container *dig.Container
 	root      *module
-	modules   []*module
 
 	// Timeouts used
 	startTimeout time.Duration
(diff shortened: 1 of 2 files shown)


abhinav:
This change is part of the following stack:

- #1258 ◀

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

sywhang (approved):

PR #1259, opened 2025-01-09
WIP: [PoC] fx.Evaluate

### LINKED ISSUE #1132: fx.Evaluate: Support dynamic graphs by returning fx.Option from constructors
# Feature request

Suppose a program has a `Store` interface with two implementations: a memory-backed version and a Redis-backed version.

```go
type Store interface {
	Get(k string) (v string, error)
	Set(k, v string) error
}

type MemoryStore struct{ m map[string]string }

func NewMemoryStore() *MemoryStore

type RedisStore struct{ /* ... */ }

func NewRedisStore(*redis.Client) *RedisStore
```

The memory store has no dependencies, while the Redis store depends on a Redis client.

```mermaid
flowchart LR
  MemoryStore
  RedisStore -->|NewRedisStore| RedisClient[*redis.Client]
```

The application wants to pick between the two implementations based on a configuration parameter.
Imagine:

```go
func NewStore(cfg Config, memStore *MemoryStore, rediStore *RedisStore) Store {
	if cfg.Development {
		return memStore
	}
	return redisStore
}
```

However, as a result of this, `NewStore` has a hard transitive dependency on the Redis client, even if the Redis store won't be used.

```mermaid
flowchart LR
  MemoryStore
  RedisStore -->|NewRedisStore| RedisClient[*redis.Client]
  Store -->|NewStore| Config & MemoryStore & RedisStore
  Store -.-> RedisClient
```

Ideally, `NewStore` should be able to specify which dependency set it wants to use after looking at the configuration.

I'd like to propose adding a new intrinsic operation to the Fx model to accompany the existing intrinsics: Provide, Invoke, and Decorate.

```go
package fx

func Evaluate(...any) fx.Option
```

The new operation, tentatively named `fx.Evaluate`, will represent an unevaluated part of the graph.
image

Functions passed to `fx.Evaluate` will always be evaluated. They will feed new information to the graph in the form of `fx.Option` return values.
These new operations may form connections that previously did not exist.
image

With this operation, the scenario described above would be solved like so:

```go
func NewStore(cfg Config) fx.Option {
    if cfg.Development {
        return fx.Provide(func(s *MemoryStore) Store { return s })
    }
    return fx.Provide(func(s *RedisStore) Store { return s })
}

// Or alternatively:

func NewStore(cfg Config) fx.Option {
    if cfg.Development {
        return fx.Provide(
            fx.Annotate(NewMemoryStore, fx.As(new(Store))),
        )
    }
    return fx.Provide(
        fx.Annotate(NewRedisStore, fx.As(new(Store))),
    )
}

// In main:

fx.New(
    // ...
    fx.Evaluate(NewStore),
)
```

`fx.Replace` was considered but is insufficient for this functionality.
It cannot conditionally severe the dependency between `NewStore` and `*redis.Client`.
See [this comment](https://github.com/uber-go/fx/discussions/1131#discussioncomment-7470019).

No, this is not a breaking change.

**Additional context**
This functionality has been brought up before and has been referred to with names such as: monadic graph, dynamic provides, late provides.

There was a prior attempt to implement this in [#699](https://github.com/uber-go/fx/pull/699) (tracked in [#698](https://github.com/uber-go/fx/issues/698)). This attempt was rejected because of the following reasons [per this comment](https://github.com/uber-go/fx/issues/698#issuecomment-961294249).

- There was risk of UX confusion by overloading `fx.Provide`. This is not an issue if we introduce a new intrinsic.
- The internals of the library were actively being reworked for `fx.Decorate`. This is done now.
- There were concerns about how "late provides" would interact with lifecycles. Again, this is not an issue with a new intrinsic because new expectations can be established: Evaluate always runs, so lifecycle events there always run.

Fx startup roughly takes the following shape:

```
provides, decorates, invokes = [...], [...], [...]

provideAll(provides)
decorateAll(decorates)
invokeAll(invokes)

start()
```

With the new intrinsic, this could effectively become:

```
provides, decorates, invokes, evaluates = [...], [...], [...], [...]

provideAll(provides)
decorateAll(decorates)

while len(evaluates) > 0:
  e = pop(evaluates)
  ps, ds, is, es = evaluate(e)
  provideAll(ps)
  decorateAll(ds)
  invokes.append(is...)
  evaluates.append(es...)

invokeAll(invokes)
```

abhinav:
**This is just a PoC at this time and not yet ready for review.**

Resolves #1132

## Details

This change introduces the `fx.Evalute` option,
which allows for dynamic generation of provides, invokes, decorates,
and other evaluates based on the state of existing dependencies.

It addresses one of the more significant remaining pain points
with usage of Fx in large applications and ecosystems:
the ability to programmatically generate parts of the dependency graph.

Concretely, this makes the following possible:

```
fx.Evaluate(func(cfg *Config) fx.Option {
    if cfg.Environment == "production" {
        return fx.Provide(func(*sql.DB) Repository {
            return &sqlRepository{db: db}
        }),
    } else {
        return fx.Provide(func() Repository {
            return &memoryRepository{}
        })
    }
}),
fx.Provide(func(...) *sql.DB { ... }),
```

With fx.Evaluate, the dependency on `*sql.DB` is present in the graph
only in production environments.
In development environments, the dependency connection is absent,
and therefore the database connection is never established.

(Today, attempting to write an Fx module that switches on the backend
like this will result in the SQL connection being established even in
development environments.)

## TODO

- [ ] Decide whether nilling out the slice is the right approach
      for tracking what's been already run
- [ ] fx.Module handling
- [ ] fx.WithLogger handling
- [ ] Testing with every other feature
- [ ] Documentation

### DIFF
--- a/app.go
+++ b/app.go
@@ -504,6 +504,23 @@ func New(opts ...Option) *App {
 		return app
 	}
 
+	// At this point, we can run the fx.Evaluates (if any).
+	// As long as there's at least one evaluate per iteration,
+	// we'll have to keep unwinding.
+	//
+	// Keep evaluating until there are no more evaluates to run.
+	for app.root.evaluateAll() > 0 {
+		// TODO: is communicating the number of evalutes the best way?
+		if app.err != nil {
+			return app
+		}
+
+		// TODO: fx.Module inside evaluates needs to build subscopes.
+		app.root.provideAll()
+		app.err = multierr.Append(app.err, app.root.decorateAll())
+		// TODO: fx.WithLogger allowed inside an evaluate?
+	}
+
 	if err := app.root.invokeAll(); err != nil {
 		app.err = err
 
--- a/evaluate.go
+++ b/evaluate.go
@@ -1,0 +1,153 @@
+// Copyright (c) 2024 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package fx
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+
+	"go.uber.org/fx/internal/fxreflect"
+)
+
+// Evaluate specifies one or more evaluation functions.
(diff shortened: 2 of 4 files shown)


abhinav:
This change is part of the following stack:

- #1258
    - #1259 ◀

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

HelloGrayson:
Great idea.

vmihailenco:
Looking forward to this :+1:

ZhengweiHou:
Super looking forward to this! Any updates on the progress?

Issue #1090, opened 2023-06-01
Error on invalid lifecycle hooks


sywhang:
Lifecycle hooks cannot take a dependency on things that the constructor does not have a direct dependency on or produces as a result type. 

When a user mistakenly specifies a hook that does this, it will fail with dependency not satisfied error which can be confusing. We should detect this at Provide time and return an error.

PR #1268, opened 2025-03-26
feat: add a clearer error message during building annotation (#1090)

90Byte:
## Overview
- Add dependencies verification process during the PROVIDE period, building life cycle hooks.
- Add a clearer error message that indicates the user tried to use hooks with insufficient dependencies.

## Changes
- Implement a utility function to validate that all dependencies are from the parameters or results of the annotated constructor.
- Update the unit tests that already serve a similar purpose to this change.
  - "cannot pull in any extra dependency other than params or results of the annotated function"
  - "with unprovided dependency"

## Linked Issue
- #1090

## Notes
I noticed some redundancy with the function `annotated.go:makeHookScopeCtor`; however, I believe refactoring that function is beyond the scope of this issue.


tchung1118 (commented):
Thank you for your contribution. I took an initial pass and left some comments. Let's add some more test cases testing the new type-checking logic for doing this at Provide time.

annotated_test.go
```diff
@@ -2301,7 +2301,7 @@ func TestHookAnnotationFailures(t *testing.T) {
 	}{
 		{
 			name:        "with unprovided dependency",
-			errContains: "error invoking hook installer",
+			errContains: "hook function cannot take any arguments outside",
```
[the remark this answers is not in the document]
90Byte: Okay, I'm working on it. Thank you for your review!

annotated.go
```diff
@@ -687,12 +757,19 @@ func (la *lifecycleHookAnnotation) buildHookInstaller(ann *annotated) (
 	invokeParamTypes := []reflect.Type{
 		_typeOfLifecycle,
 	}
+	if valid := validateHookDeps(origHookFnT, paramTypes, resultTypes); !valid {
+		return reflect.Value{},
+			nil,
+			nil,
+			fmt.Errorf("the %s hook function cannot take any arguments outside of the annotated constructor's parameters's existing dependencies or results, except a context.Context", la.String())
```
tchung1118, line 764: I think it would be useful to users if we can show them which parameter/type is the culprit for this error. This will require updating the signature of `validateHookDeps` to the following:
```
func validateHookDeps(hookFnT reflect.Type, paramTypes []reflect.Type, resultTypes []reflect.Type) (error)
```

And then include necessary information in the error. Something along the lines of "the <OnStart/OnStop> hook function takes in a parameter of "{some type}", but the annotated function does not have parameters or results of that type"

annotated.go
```diff
@@ -662,12 +665,79 @@ var (
...
+				return
+			}
+			continue
+		}
+		for j := 1; j < t.NumField(); j++ {
+			field := t.Field(j)
+			if field.Type == _typeOfContext {
+				continue
+			}
+			k := key{t: field.Type, name: field.Tag.Get("name"), group: field.Tag.Get("group")}
```
tchung1118, line 724: nit
```suggestion
			k := key{
				t: field.Type,
				name: field.Tag.Get("name"),
				group: field.Tag.Get("group"),
			}
```

annotated.go
90Byte: That seems fair, thank you.

Plus, do you think it's important to show all the problematic parameters at once? Personally, I think users usually tackle issues one at a time, since they'll likely come across these problems during development.

Making this change isn't complicated, but it requires some adjustments to the existing code I made.

Could you suggest a preferred format for this project in case all problematic parameters need to be shown at once?

tchung1118: Hmm, let's start with returning one problem at a time. If there's feedback about including all problems in the error, we can make that update later.

90Byte: I just made the changes based on your feedback and pushed the commit. Please take a look.

annotated.go
```diff
@@ -687,12 +761,19 @@ func (la *lifecycleHookAnnotation) buildHookInstaller(ann *annotated) (
 	invokeParamTypes := []reflect.Type{
 		_typeOfLifecycle,
 	}
+	if err := la.validateHookDeps(origHookFnT, paramTypes, resultTypes); err != nil {
```
JacobOaks, line 764: I notice in `validateHookDeps` we are checking for and skipping context.Context. The for loop directly after this call is already going through the params of the hook fn and removing context from them. Should we call `validateHookDeps` after the loop below and pass `invokeParamTypes` in instead of `origHookFnT` to not duplicate this work?

Inserting this call here also confuses the comment above about removing context, since that's not the purpose of this call.

annotated.go
```diff
@@ -662,12 +665,83 @@ var (
...
+			if field.Type == _typeOfContext {
+				continue
+			}
+			k := key{
+				t:     field.Type,
+				name:  field.Tag.Get("name"),
+				group: field.Tag.Get("group"),
+			}
+			if _, ok := seen[k]; !ok {
+				err = fmt.Errorf("the %s hook function takes in a parameter of \"%s\", but the annotated function does not have parameters or results of that type", la.String(), field.Type.String())
```
JacobOaks, line 730: Can we incorporate `name` and `group` tags into this error message to be more precise about what's missing? We can probably either add a helper function for printing `key`, or just add a `String()` method on key.

JacobOaks (commented):
Thanks for your contribution! This look great and should hopefully cause much less confusion. Left a couple tiny comments, and will test this in our internal codebase before approving.

annotated.go
90Byte: I wrote a helper function to format the `key`, so the log displays the tags correctly. I also made this function to format the error message itself since it was a bit redundant, after all.
Thank you for your feedback.

JacobOaks (approved):
LGTM! Did not cause any issues in Uber's codebase either. Thanks for your contribution!

tchung1118 (approved):

PR #1274, opened 2025-09-11
ci: Test against Go 1.24 and 1.25

abhinav:
With the release of Go 1.25, bump CI to run against 1.24 and 1.25,
and bump the minimum required Go version to 1.24
as this is required to use features introduced in Go 1.24.

(We should bump this to at least 1.23 to start using iterators.)

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       matrix:
         os: ["ubuntu-latest", "windows-latest"]
-        go: ["1.23.x", "1.24.x"]
+        go: ["1.24.x", "1.25.x"]
 
     steps:
     - name: Checkout code
@@ -52,7 +52,7 @@ jobs:
     - name: Setup Go
       uses: actions/setup-go@v5
       with:
-        go-version: 1.24.x
+        go-version: 1.25.x
 
     - name: Test
       run: make cover COVER_MODULES=./docs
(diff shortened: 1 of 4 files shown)


rabbbit (approved):
Would be nice to say which features from 1.24 we need perhaps?

abhinav:
> Would be nice to say which features from 1.24 we need perhaps?

We need nothing from 1.24 right now but without this we can't use them either.
Directive was 1.22 earlier which means even iterative (1.23) were unusable and I think Fx could definitely lose some slice allocs with iterators.

rabbbit:
> > Would be nice to say which features from 1.24 we need perhaps?
> 
> We need nothing from 1.24 right now but without this we can't use them either. Directive was 1.22 earlier which means even iterative (1.23) were unusable and I think Fx could definitely lose some slice allocs with iterators.

Interesting - in all cases I've tried to use iterators so far they've turned out to be slower, but I can see how my use cases were "wrong".

abhinav:
No, you're probably not doing anything wrong here.
I haven't benchmarked them extensively but I'm sure they're slower than iterating through a slice.
I've found them useful in certain cases for making code more readable and writable.
E.g. in Fx somewhere we have a walk function for visiting all modules in a module tree; that would be better expressed with an iterator—easier to read and write.

r-hang (approved):

rabbbit:
> No, you're probably not doing anything wrong here. I haven't benchmarked them extensively but I'm sure they're slower than iterating through a slice. I've found them useful in certain cases for making code more readable and writable. E.g. in Fx somewhere we have a walk function for visiting all modules in a module tree; that would be better expressed with an iterator—easier to read and write.

I didn't say I'm doing anything wrong, I said my use cases were wrong;)

PR #1276, opened 2025-09-11
all: modernize

abhinav:
Run the modernize linter across the codebase.
Change generated by running:

```
go run golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize@latest -fix -test ./...
```

Ref: https://pkg.go.dev/golang.org/x/tools/gopls/internal/analysis/modernize

### DIFF
--- a/annotated.go
+++ b/annotated.go
@@ -79,7 +79,7 @@ type Annotated struct {
 	Group string
 
 	// Target is the constructor or value being annotated with fx.Annotated.
-	Target interface{}
+	Target any
 }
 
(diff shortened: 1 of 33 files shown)


abhinav:
Thanks!
The stack has to be landed bottom-up, so after the bottom-most is approved, I will rebase the rest and then we can merge the new bottom-most PR, and so on.
I think I still have privileges to press merge once the PR is approved.

r-hang (approved):

PR #1280, opened 2025-11-06
feat(fx): add Transient provider option for non-cached constructors

rezamokaram:
# Motivation

Currently, fx.Provide() always registers constructors as singletons, meaning their results are reused for the lifetime of the app.
In some cases, it’s desirable to have fresh instances, for example:

- Creating a new per-request context object
- Generating unique IDs or temporary objects
- Managing short-lived resources that shouldn’t be shared globally

Introduces a new fx.Transient() helper that allows constructors to be invoked each time their dependency is requested, rather than once at application start.

This enables transient or per-request dependency lifetimes similar to scoped services in other DI systems. The Transient wrapper registers a factory function that produces a new instance on each call.

Includes unit tests verifying:
- Constructor runs multiple times for transient providers.
- Singleton behavior remains unchanged for regular fx.Provide.

## Backward compatibility

No breaking changes — existing Fx applications continue to behave identically.
The new feature is purely additive.

**Documentation and usage examples will be added in a follow-up PR after this change is approved.**

### DIFF
--- a/provide.go
+++ b/provide.go
@@ -119,6 +119,23 @@ func (o provideOption) String() string {
 	return fmt.Sprintf("fx.Provide(%s)", strings.Join(items, ", "))
 }
 
+// Transient marks a constructor so that fx will provide a factory function
+// that creates a new instance each time it is called. For example,
+//
+//	fx.Provide(fx.Transient(NewA))
+//
+// where `NewA` is `func(...) *A` will register a provider for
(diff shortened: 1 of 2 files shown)


rezamokaram:
Hi @sywhang, just a friendly ping on my PR (#1280) — whenever you have time to review it would be great. Thanks a lot!

Issue #1283, opened 2026-01-14
Delay app.exit() to allow buffered log to flush


yusufthedragon:
**Is your feature request related to a problem? Please describe.**
In my program, I use a log library that utilizes `bufio.Writer`, which copies data into an internal buffer before printing the log to reduce memory usage. When `fx.App` returns exit code 1, it calls `os.Exit()`, which stops the program immediately while the log is still in the internal buffer. This behavior causes the log library to fail to print the log. It would be very helpful if the library had a "buffer" or waiting time before calling `app.exit()` when exit code 1 is returned.

**Describe the solution you'd like**
Add another `fx.Option` to set the buffer time before calling `app.exit()` when exit code 1 is returned. It would be as simple as implementing `time.Sleep()` [here](https://github.com/uber-go/fx/blob/master/app.go#L601).

**Describe alternatives you've considered**
None.

**Is this a breaking change?**
No. We will treat it the same as `fx.StartTimeout()` and `fx.StopTimeout()`, which implement the `fx.Option` interface. To maintain backward compatibility, we will only wait if the value of `fx.Option` is set.

**Additional context**
None.

rabbbit:
Not a fx expert here, but wouldn't doing this via a shutdown hook be both easier and more appropriate? (No hardcoded deadlines, etc)

yusufthedragon:
You are right. Unfortunately the error in my service occurred during start (boot-up), so it won't trigger any hook.

CTrando:
If startup fails, isn't `app.start` wrapped in the `withRollback` function meaning that stop hooks are still run? Implying that if you register a stop hook early enough, that you can execute its shutdown hook on a failed start?

yusufthedragon:
Wow you are absolutely right @CTrando  thank you for your answer, it fix my problem 👍 it's better to close the PR then

PR #1284, opened 2026-01-25
feat(fx): add ExitDelay option to delay os.Exit

yusufthedragon:
Introduce ExitDelay Option and App.ExitDelay to configure a delay before calling os.Exit on non-zero exit. Run will sleep using the
App's clock when a non-zero delay is set. Added unit test to verify behavior of ExitDelay option.


yusufthedragon:
This PR is related to this feature request: https://github.com/uber-go/fx/issues/1283

yusufthedragon:
Closed PR, my problem fixed

